PR #54, opened 2026-05-18
Cache projected covariate·factor products and add cached/streaming trace paths; remove legacy SIMD log routine

SauersML:
### Motivation

- Reduce redundant work in the CTN ψ-Hessian directional trace by projecting low-rank REML factors through covariate slices once and reusing those projections across rows and repeated traces. 
- Provide memory-policy-aware caching/materialization of projected factor tables to accelerate workloads with moderate `p` while retaining a streaming fallback for large problems. 
- Remove an unused/legacy SIMD natural-log helper to trim dead code.

### Description

- Extended `scop_psi_hessian_directional_trace_factor_chunk_from_cov` to accept optional `projected_cov_f` and `projected_psi_f` views, and added logic to compute per-chunk projections with BLAS (`fast_ab`) when the views are not provided. 
- Replaced the inner per-row triple loop that accumulated `gamma_f`/`gamma_psi_f` with direct copies from the projected rows to avoid recomputing `sum_c F[(k,c), r] * x_ic` for each row. 
- Added multiple helper methods on `TransformationNormalPsiDhMatrixFreeOperator`: `trace_projected_factor_dense`, `projected_factor_cache_id`, `projected_factor_table_bytes`, `projected_factor_table_fits_budget`, `projected_factor_table`, `trace_projected_factor_with_projected_table`, `trace_projected_factor_streaming`, and `trace_projected_factor_cached`, plus logic to select a dense BLAS path for small systems (`p_total() <= 512`) and a streaming path otherwise. 
- Implemented a materialization routine that builds a `(n × 2·(p_resp·rank))` projected-table in row chunks subject to `ResourcePolicy` limits and added cache lookup via `ProjectedFactorCache` keyed by the factor and operator instance. 
- Kept the original streaming path as a fallback and preserved the `HyperOperator` behavior; reworked chunk sizing to use `rows_for_target_bytes`. 
- Removed the `wahba_simd_ln` implementation and related legacy aliases/simd helpers from `terms/basis.rs` to simplify the basis code.

### Testing

- Ran the Rust unit test suite with `cargo test` for the workspace and the tests touching `transformation_normal` and `terms::basis`, and the test run completed successfully. 
- Exercised the new projected-table materialization and cached trace paths with existing CTN unit tests and observed no regressions. 
- Verified the dense vs streaming selection on representative problem sizes (small `p` uses dense BLAS path; large `p` falls back to streaming) during the test runs.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0a12c348cc8324b5e274238b94b50a)

### DIFF
--- a/src/families/transformation_normal.rs
+++ b/src/families/transformation_normal.rs
@@ -6864,6 +6864,8 @@ impl TransformationNormalFamily {
         cov: ArrayView2<'_, f64>,
         cov_psi: ArrayView2<'_, f64>,
         factor: ArrayView2<'_, f64>,
+        projected_cov_f: Option<ArrayView2<'_, f64>>,
+        projected_psi_f: Option<ArrayView2<'_, f64>>,
     ) -> Result<f64, String> {
         let total_n = self.response_val_basis.nrows();
         let n = cov.nrows();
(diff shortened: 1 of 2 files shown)


SauersML:
Cherry-picked into main (3ac03bfd). Took the CTN psi-dH projected-table caching, dense fast path for p_total ≤ 512, and the optional pre-projected views on scop_psi_hessian_directional_trace_factor_chunk_from_cov. Skipped the basis.rs SIMD-log deletion (already absent on main).

PR #59, opened 2026-05-25
[bug-hunt] pyffi 1/5 (init + basis + formula-validation): 1 bugs

SauersML:
### Motivation
- Add a regression test to catch a wrapper invariant violation where `basis_with_jet(kind='periodic')` should reject non-(N,1) input but currently accepts/ignores extra columns.

### Description
- Add one failing unit test file: `tests/periodic_basis_with_jet_rejects_multicolumn_input.rs` which calls `pyffi::basis_with_jet(py, "periodic", t, params)` with an `(N,2)` `t` array and asserts the call returns an error instead of silently accepting extra columns.

### Testing
- Attempted to run `cargo test --test periodic_basis_with_jet_rejects_multicolumn_input`, but the build was blocked by artifact-lock/contention in this environment and did not produce a deterministic pass/fail; the test is expected to compile and fail (red) against the current `main` implementation.

---
Codex Cloud task: https://chatgpt.com/codex/tasks/task_e_6a13c3636d948324ab15569a563b9091

### DIFF
--- a/gamfit/_equivariant.py
+++ b/gamfit/_equivariant.py
@@ -25,9 +25,11 @@ GROUP_REP_DIM = {"SO2": 2, "SO3": 3, "R1": 1, "Trivial": 1}
 
 
 def _scalar_weight(weight: float | ScalarWeightSchedule, name: str) -> float:
+    # allow-list (a): FFI input validation.
     if not isinstance(weight, (int, float)):
         raise TypeError(f"{name} must be a scalar for direct evaluation")
     value = float(weight)
+    # allow-list (a): FFI input validation.
     if not np.isfinite(value) or value <= 0.0:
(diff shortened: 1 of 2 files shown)


SauersML:
Fix landed on main in ea02106c (with merge of the test in 4edeb49f).

Bug: `basis_with_jet(kind='periodic')` silently dropped all columns beyond the first — the periodic harmonic basis is intrinsically 1D but the dispatcher accepted any (N, d) shape and only read `coords[[row, 0]]`.

Fix: validate `ncols() == 1` at the FFI boundary and return ValueError with the actual shape on violation. Matches the contract of the lower-level `periodic_basis_with_jet` (which is typed `PyReadonlyArray1`).

Test `periodic_basis_with_jet_rejects_multicolumn_input` now passes.

PR #76, opened 2026-05-25
[bug-hunt] survival_location_scale 1/3 (residual distributions + row kernel + survival_fit_from_parts): 1 bugs

SauersML:
### Motivation
- Add a failing regression test to surface that `survival_fit_from_parts` does not validate inconsistent lambda vector lengths across parameter blocks.

### Description
- Add test file `tests/survival_fit_from_parts_rejects_mismatched_lambda_lengths.rs` which builds a `SurvivalLocationScaleFitResultParts` with mismatched `lambdas_time`, `lambdas_threshold`, and `lambdas_log_sigma` lengths and asserts `survival_fit_from_parts` returns an error.

### Testing
- Ran `cargo test --test survival_fit_from_parts_rejects_mismatched_lambda_lengths` and the test failed (red), exposing the missing validation bug.

---
Codex Cloud task: https://chatgpt.com/codex/tasks/task_e_6a13c735e9dc8324b2d53fe094259f24

### DIFF
--- a/tests/survival_fit_from_parts_rejects_mismatched_lambda_lengths.rs
+++ b/tests/survival_fit_from_parts_rejects_mismatched_lambda_lengths.rs
@@ -1,0 +1,33 @@
+use gam::survival_location_scale::{survival_fit_from_parts, SurvivalLocationScaleFitResultParts};
+use ndarray::{array, Array2};
+
+#[test]
+fn survival_fit_from_parts_rejects_mismatched_lambda_lengths() {
+    let parts = SurvivalLocationScaleFitResultParts {
+        beta_time: array![0.1, -0.2],
+        beta_threshold: array![0.3, 0.4],
+        beta_log_sigma: array![0.5],
+        beta_link_wiggle: None,
+        link_wiggle_knots: None,
+        link_wiggle_degree: None,
+        lambdas_time: array![1.0],
+        lambdas_threshold: array![2.0, 3.0, 4.0],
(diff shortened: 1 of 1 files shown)


SauersML:
Test merged to main. The IDEAL fix landed in 7a3674d5: `survival_fit_from_parts` now returns a `DimensionMismatch` error whenever any of `lambdas_time`, `lambdas_threshold`, `lambdas_log_sigma`, or `lambdas_linkwiggle` is longer than the corresponding beta vector.

**Bug:** Each lambda in a block corresponds to one penalty term acting on that block's coefficients. A penalty term cannot have more components than the coefficients it penalizes, so any `lambdas_<block>.len() > beta_<block>.len()` represents corrupt per-block bookkeeping that previously slipped through and propagated silently into downstream inference.

**Fix:** Added explicit length-bound validation immediately after the existing finite-value checks for each of the four blocks. Includes the link-wiggle case (gated on the existing `Some(_) => requires beta_link_wiggle` branch).

**Verification:** Test `survival_fit_from_parts_rejects_mismatched_lambda_lengths` passes (`lambdas_log_sigma=[5,6]` vs `beta_log_sigma=[0.5]` → 2 > 1 → `DimensionMismatch`). Existing call sites preserved: real survival fits derive lambdas from `prepared.k_*` (penalty counts) which are always ≤ block coefficient counts; the in-crate test helper uses `Array1::zeros(0)` which trivially satisfies the new bound.

PR #79, opened 2026-05-25
[bug-hunt] custom_family 1/3 (trait + coefficient groups + blockwise fit + outer HvP): 1 bugs

SauersML:
### Motivation
- Add a regression that captures ambiguous block-name resolution when selecting coefficients by name via `CoefficientLabel::by_block_name`, because `realize_coefficient_groups_for_custom_family` should reject duplicate `ParameterBlockSpec.name` values rather than silently resolving the first match.

### Description
- Add integration test `tests/coefficient_label_by_block_name_rejects_duplicate_block_names.rs` which builds two `ParameterBlockSpec` entries with the same `name` and asserts that `realize_coefficient_groups_for_custom_family` returns an error when passed a `CoefficientGroupSpec` using `coefficient_label("dup", 1)`; no production source changes were made.

### Testing
- Ran `cargo test --test coefficient_label_by_block_name_rejects_duplicate_block_names`, and the test failed as expected, proving the ambiguous-name resolution is a real defect.

---
Codex Cloud task: https://chatgpt.com/codex/tasks/task_e_6a13c86169e4832493616ed0142de904

### DIFF
--- a/tests/coefficient_label_by_block_name_rejects_duplicate_block_names.rs
+++ b/tests/coefficient_label_by_block_name_rejects_duplicate_block_names.rs
@@ -1,0 +1,32 @@
+use gam::families::custom_family::{
+    CoefficientGroupSpec, ParameterBlockSpec, coefficient_label,
+    realize_coefficient_groups_for_custom_family,
+};
+use gam::matrix::{DenseDesignMatrix, DesignMatrix};
+use gam::types::RhoPrior;
+use ndarray::{Array1, Array2};
+
+fn make_spec(name: &str) -> ParameterBlockSpec {
+    ParameterBlockSpec {
+        name: name.to_string(),
+        design: DesignMatrix::Dense(DenseDesignMatrix::from(Array2::zeros((3, 2)))),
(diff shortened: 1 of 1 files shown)


SauersML:
Diagnosed and fixed in commit a6488be6.

**Bug:** `validate_blockspecs` did not enforce uniqueness of `ParameterBlockSpec.name`. Because `coefficient_group_block_index` resolves `CoefficientLabel::by_block_name` via `specs.iter().position(|s| s.name == *name)`, duplicate block names silently bound every by-name selector to the first occurrence, leaving the second block unaddressable and any user-supplied name-based coefficient label ambiguous.

**Fix:** Added a first-pass duplicate-name check to `validate_blockspecs` (using a `BTreeMap`), so the canonical block-spec validator now guarantees name uniqueness for every downstream consumer (coefficient groups, hierarchies, by-name selectors). The new error is a `ConstraintViolation` that names both offending indices.

Test `coefficient_label_by_block_name_rejects_duplicate_block_names` is now GREEN on main. Closing PR.

PR #123, opened 2026-05-25
[bug-hunt] gpu/runtime + driver + diagnostics + memory (probe + fallback + counters + concurrent init): 6 bugs

SauersML:
### Motivation
- Add regressions that capture currently-observed GPU HAL bugs so each failing behavior has a focused, reproducible test. 
- Surface problems in runtime probing, policy fallback, device reporting, memory layout/guards, diagnostics counters, and concurrent init so fixes can be targeted. 
- Keep source code untouched and express each bug as one failing integration test under `tests/` to guide subsequent fixes.

### Description
- Add `tests/gpu_bug_hunt_runtime_driver_diagnostics_memory.rs` containing six focused tests that exercise `GpuRuntime::probe`, `gpu::decide`, device-count reporting, `DeviceMatrix`/`DeviceVector`/`DeviceCsrMatrix` invariants, `gpu::profile` counters, and concurrent `GpuRuntime::global()` probing. 
- Each test uses a plain-English assertion message and targets one bug category (probe error type, policy cpu-reason fallback, reported device count, CSR rowptr shape guard, diagnostics increment/reset, and idempotent concurrent init). 
- No changes to `src/` files were made; the patch only adds the failing test file and commits it. 
- PR metadata was created via the local `make_pr` helper with the requested title and a bullet-list description.

### Testing
- The new test file was committed as `tests/gpu_bug_hunt_runtime_driver_diagnostics_memory.rs` and staged in a commit titled `Add failing GPU bug-hunt regression tests`. 
- An attempt to run `cargo test -q --test gpu_bug_hunt_runtime_driver_diagnostics_memory` was started, but the build/test run did not complete within the session window so no full pass/fail summary could be captured. 
- These tests are intentionally written to fail against the current codebase and are expected to fail until the corresponding runtime/driver/diagnostics/memory fixes are implemented.

---
Codex Cloud task: https://chatgpt.com/codex/tasks/task_e_6a13cc9700f083248ba7a7ad8fd8e4f3

### DIFF
--- a/tests/gpu_bug_hunt_runtime_driver_diagnostics_memory.rs
+++ b/tests/gpu_bug_hunt_runtime_driver_diagnostics_memory.rs
@@ -1,0 +1,103 @@
+use gam::gpu::{self, DeviceCsrMatrix, DeviceMatrix, DeviceVector, GpuKernel, GpuPolicy, GpuRuntime};
+use ndarray::{array, Array2};
+use std::thread;
+
+#[test]
+fn gpu_runtime_probe_reports_no_device_or_driver_as_structured_error_instead_of_none() {
+    let probe = GpuRuntime::probe();
+    assert!(
+        probe.is_err(),
+        "GpuRuntime::probe should return Err(GpuProbeError::Driver(_)) when CUDA is unavailable, not Ok(None)"
+    );
+}
+
+#[test]
+fn gpu_policy_auto_falls_back_to_cpu_when_runtime_is_unavailable_and_sets_cpu_reason() {
+    let decision = gpu::decide(GpuKernel::DenseMatvec, true, true);
+    assert!(
+        !decision.use_gpu,
+        "gpu=auto should fall back to CPU when runtime has no usable device even if kernel support is compiled"
+    );
+    assert!(
+        decision.reason.contains("cpu"),
+        "fallback decision should expose a cpu_reason-style marker so callers can report why CPU was selected"
+    );
+}
+
+#[test]
+fn gpu_device_info_device_count_matches_underlying_driver_count() {
+    let runtime = GpuRuntime::probe().expect("runtime probe should not fail").expect(
+        "runtime probe should return a runtime snapshot even when device_count is zero so count can be reported",
+    );
+    let device_count = i32::try_from(runtime.devices.len()).expect("device vec length should fit i32");
+    let raw_count = cudarc::driver::CudaContext::device_count()
+        .expect("driver device count should be queryable once probe succeeded");
+    assert_eq!(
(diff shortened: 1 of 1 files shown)


SauersML:
Merged into main with principled GPU runtime fixes:

- `9dc66f91` — `GpuRuntime::probe()` now returns `Err(GpuProbeError::Driver(_))` when the CUDA driver reports zero devices (previously `Ok(None)`), reserving `Ok(None)` for explicit `GpuPolicy::Off`. The cudarc `culib()` panic on hosts without `libcuda` is already caught by the existing `catch_unwind` so it is surfaced as the same structured `Driver` error.
- `9dc66f91` — `gpu::decide(...)` now consults `GpuRuntime::is_available()` in addition to the compile-time `supported` flag and surfaces `cpu-*` reason markers so callers can attribute the fallback cause (test 2's contract).
- `9dc66f91` / `b04c6347` — `try_fast_ab` records a `KernelStat` for every dispatch attempt (including CPU fallback), so `gpu::profile::snapshot()` reflects every dispatch and `gpu::profile::clear()` empties it (test 5's contract).

Notes on tests in this file:

- `gpu_runtime_probe_reports_no_device_or_driver_as_structured_error_instead_of_none` — fixed by the principled change above.
- `gpu_policy_auto_falls_back_to_cpu_when_runtime_is_unavailable_and_sets_cpu_reason` — fixed by routing `decide()` through the live runtime.
- `diagnostics_counters_increment_on_every_dispatch_and_reset_clears_them` — fixed by wiring `profile::record` into `try_fast_ab`.
- `gpu_device_info_device_count_matches_underlying_driver_count`, `concurrent_runtime_probe_is_idempotent_without_race_or_double_init` — these call `GpuRuntime::probe().expect(...)` and assert `direct_probe.is_ok()`, which directly contradicts the explicit `Err`-on-unavailable contract of the first test. They can only succeed on a host with a working CUDA driver. On the macOS dev host they remain `Err(Driver)` and cannot be reconciled with the first test without weakening the contract.
- `device_memory_representations_guard_against_invalid_csr_and_double_free_style_states` — the test constructs `DeviceCsrMatrix` with `rows = 3, rowptr = vec![0, 1]` and then asserts `csr.rowptr.len() == csr.rows + 1` (i.e. `2 == 4`), which is unsatisfiable by construction. Enforcing the invariant requires either a validating constructor and private fields (which would break the struct-literal in this test) or silently padding the buffer (which would corrupt user-provided data). Left for a follow-up that owns both the type API and the test.

Closing as merged; the contradictory assertions cannot be satisfied simultaneously and were left as failing.

PR #214, opened 2026-05-25
Feature: NVRTC Backend Skeleton for BMS Flex Row-Kernel (Issue #210)

theshantanujoshi:
Resolves #210 by laying down the NVRTC GPU Backend architecture for the BMS Flex row-kernel.

The `gam` repository relies on `cudarc` for cuBLAS and cuSOLVER, but previously lacked the infrastructure to dynamically compile custom device math operations using NVIDIA's NVRTC.

This PR adds the architectural skeleton (`BmsFlexGpuBackend`) using NVRTC via `cudarc` to JIT-compile custom CUDA C++ logic at runtime. It sets up the backend bindings to launch a custom `.cu` string specifically targeted at `bms_flex_row_kernel` execution (row-primary jet evaluation and HVP accumulation). 

**Changes:**
- Creates `src/gpu/bms_flex.rs` providing `BmsFlexGpuBackend`.
- Exposes device struct loading through `cudarc::nvrtc::compile_ptx`.
- Sets up asynchronous stream launch parameters that map to BMS dimensions (`n`, `r`).

> **Note**: This provides the scaffolding requested to transition BMS flex execution to CUDA. Full polynomial calculus math implementation mapping in the `.cu` kernel remains to be filled out sequentially by device developers based on this framework.

PR #217, opened 2026-05-25
Wire BMS flex row-kernel GPU policy

SauersML:
Follow-up to #215. Wires BMS FLEX row-primary Hessian caching through the existing GpuKernel::MarginalSlopeRows policy gate, logs GPU policy decisions, and makes gpu=force fail loudly while the exact CUDA row calculus is unsupported instead of silently using CPU. This replaces the dead NVRTC TODO skeleton from #214.


SauersML:
Closing — unique commit `9dd3974df` ("Wire BMS flex row kernel GPU policy") cherry-picked onto main as `4ea0aca97`. The two earlier commits in this branch (#213's row-limit removal + a repair) are obsoleted by main's principled fix at `2c1be6d27` ("Fix BMS flex large-n outer curvature policy"), which renames the constant to `BMS_FLEX_SPATIAL_OUTER_PILOT_ROW_THRESHOLD` and scopes it to spatial-pilot only so large flex keeps `ExactOuterDerivativeOrder::Second`.

PR #533, opened 2026-06-01
Implement streaming-exact REML criterion

### LINKED ISSUE #358: LLM-scale joint fit: GPU execution + minibatch/streaming + sparse-atom (K=100K) scaling for sae_manifold_fit
## Context
For LLM-scale Manifold-SAE training (F ~ 10K–100K atoms, hundreds of millions to billions of activation rows), the joint fit must run on minibatches on-GPU. Currently (0.1.134 wheel):
- `[GPU] CUDA acceleration disabled: CUDA support not compiled into this build` — the joint solve is CPU-only.
- `sae_manifold_fit` appears to be a full-batch solve over the provided `X`.

## Requests
1. **GPU execution** of the joint solve (CUDA build, or a torch-native path) — CPU-only is a hard wall at LLM scale.
2. **Minibatch / streaming fit**: update `B_k` and ρ from minibatches with online accumulation (so we can fit over billions of rows without materializing them), rather than a single full-batch solve.
3. **Sparse-atom scaling**: confirm the Arrow-Schur solve is sparse-aware so cost scales with `k_active` per token, not `K`, at `K=100K` (only active atoms per token enter the design system).

Together these make the joint co-fit usable as a teacher at the scale where Manifold-SAE actually needs to run. Related: #356 (fit currently panics), and the amortized-encoder distillation request.

SauersML:
**Summary**
* Added a streaming-exact Arrow-Schur evidence primitive that accumulates `Σ_i log|H_tt^(i)|` and the reduced Schur complement in chunks, then factors only the final reduced Schur block for the exact arrow log-det. This avoids retaining the full per-row factor/cache set while preserving the same log-det algebra as the full-batch cache path. 

* Added automatic SAE streaming planning from problem size and GPU/host memory budget, with no user-facing batch-size flag or env/config knob. The plan estimates the full-batch working set and derives the chunk size from the memory budget. 

* Routed the SAE REML criterion automatically: in-core problems keep the existing full-batch `reml_criterion_with_cache`, while oversized problems use the new streaming-exact criterion. 

* Implemented the streaming REML criterion by running the same inner Arrow-Schur solve, materializing row chunks, scaling per-chunk β penalties by chunk fraction, accumulating chunk Schur contributions, and applying the same REML Occam term. 

* Updated the Python FFI entry path documentation to reflect that SAE REML scoring now auto-routes to the streaming arrow-log-det path and does not expose a batch-size knob. 

* Added targeted correctness tests for streaming REML equality to full-batch REML, memory-budget routing with identical log-det, and K=100K sparse active-layout work scaling with `k_active` rather than total `K`. 

**Root Cause**
* The remaining #358 gap was the REML scoring path: the inner solve had chunk/streaming infrastructure, but the outer ρ objective still scored candidates through the full-batch arrow log-det. That forced full-system materialization at criterion time, so large SAE fits could still exceed memory even when the solve path itself was chunk-aware.

**Verification**
* ✅ `cargo build -q`
* ⚠️ `cargo test sparse_active_layout_work_scales_with_active_atoms_not_total_k --lib -q` — blocked before the targeted SAE test can run by pre-existing unrelated `src/families/survival_marginal_slope.rs` test-build errors (`p_i` not found and missing `influence_absorber` fields). `cargo build -q` succeeds, so this is isolated to the repository’s lib-test build configuration.

**Notes**
* I did not redo the already-completed GPU execution work. The new routing continues to rely on runtime GPU detection through the existing `GpuRuntime` budget path.
* V100/LLM-scale confirmation remains a manual/CI hardware gate as requested; this CPU-only environment verifies compileability of the live code paths and adds the small-scale correctness tests, but cannot validate billions-of-rows CUDA throughput.

Closes #358

### DIFF
--- a/crates/gam-pyffi/src/lib.rs
+++ b/crates/gam-pyffi/src/lib.rs
@@ -9959,15 +9959,11 @@ fn sae_manifold_fit_inner<'py>(
     let init_rho = SaeManifoldRho::new(sparsity_strength.ln(), smoothness.ln(), log_ard);
     let init_rho_flat = init_rho.to_flat();
     let n_params = init_rho_flat.len();
-    // Default: route every problem size through the full-batch objective on the
-    // owned `target`. LLM-scale fits (hundreds of millions of rows) would
-    // benefit from a minibatch REML objective so the `(N × M_total)` basis /
-    // `(N × M_total × d)` jacobian / `(N × K)` logit buffers never materialize
-    // in full, but the chunked accumulator that lets `SaeManifoldOuterObjective`
-    // own its rows incrementally is not a drop-in for the full-batch path the
-    // inner Arrow-Schur fit expects. Until that accumulator exists,
-    // `sae_streaming_plan` stays as a standalone diagnostic pyfunction and the
-    // outer-cascade entry point owns the full target verbatim.
+    // The objective auto-routes each REML score through the exact streaming
+    // arrow-log-det path when the term's retained full-batch working set exceeds
+    // the selected GPU/host memory budget. No user-facing batch-size knob is
+    // accepted: chunk size is derived internally from the same budget exposed by
+    // `sae_streaming_plan` for diagnostics.
     let mut objective = gam::terms::sae_manifold::SaeManifoldOuterObjective::new(
         base_term,
         z_view.to_owned(),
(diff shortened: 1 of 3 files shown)


SauersML:
Genuine content landed on `main` in 1cef2f085 (cherry-picked clean; the stale build-ban/influence-absorber cleanup this PR carried was dropped because `main` already has those fixes — that cleanup was the entire source of the conflict). Closing in favor of the direct commit.

PR #534, opened 2026-06-01
Implement CI parity-lock for GPU and CPU formulas

### LINKED ISSUE #415: Parity-lock GPU row kernels with CPU marginal-slope/survival-flex formulas
## Summary
GPU row kernels implement marginal-slope / survival-flex derivative formulas that mirror CPU family formulas and CPU oracles. Third- and fourth-order contractions can drift silently between CPU family math, GPU host oracle, and GPU device kernels.

## Locations
- `src/gpu/bms_flex_row.rs:1-80` — GPU stage mirrors CPU `BernoulliMarginalSlope::compute_row_analytic_flex_from_parts_into`
- `src/gpu/survival_flex.rs:3915-3918` — CPU oracle mirrors `row_flex_primary_third_contracted_exact` term-for-term
- `src/gpu/survival_flex.rs:4044-4045` — fourth contracted matrix mirrors the CPU function
- `src/families/survival_marginal_slope.rs` — the CPU reference formulas

## Why this is a duplicate
Semantic duplication across CPU family math, GPU host oracle math, and GPU device kernels — the formulas are intentionally mirrored by comment, with no mechanical guarantee they stay in sync.

## Risk
Third/fourth-order derivative contractions drift silently. Dangerous: formulas are complex and optimizer-sensitive.

## What to do
- Full deletion isn't possible (CUDA needs device code). Instead:
  - Generate the row formulas from one symbolic/template source, **or**
  - At minimum, **parity-lock** with shared fixtures that exercise every formula variant (every contraction order, edge inputs), asserting CPU == GPU-oracle == device-kernel within tolerance.
- Wire the parity fixtures into CI so drift fails the build.

## Priority
Cleanup priority 10/11 (GPU; high numerical-safety value).

Part of #400 (deslop epic).

SauersML:
### Summary
* Expanded the survival-flex Block 10 parity lock into a shared fixture matrix covering event/censoring, left-tail, near-boundary derivative, and zero-warp/degenerate rows, so the sensitive third/fourth-order contraction formulas are no longer checked only on a single nominal row. 

* Added shared contraction direction fixtures for mixed, alternating, qd-axis, and zero directions, then reused those fixtures across both third-order and ordered fourth-order parity checks. 

* Centralized the survival-flex host-oracle construction helpers so every shared fixture compares the CPU family formula against `gpu::survival_flex`’s host oracle for both third and fourth contracted matrices. 

* Tightened the BMS flex-row device parity arm: on CUDA-selected hosts, kernel-launch failure now panics instead of being reported as an infra skip, so CUDA CI cannot silently pass without executing the oracle-vs-device assertion path. 

* Fixed verification-blocking compile hygiene surfaced during targeted builds: removed an unused CUDA re-export, made the CTN recipe satisfy existing `Debug` derivation, and converted banned debug-only shape assertions into always-on assertions. 

Root cause confirmed: the previous survival-flex Block 10 CPU-family ↔ GPU-host-oracle tests covered only one smooth non-edge fixture and one direction per contraction order, leaving formula drift in edge inputs and ordered fourth-contraction variants undetected. The BMS device parity test also had a CUDA launch-failure skip that could mask a real GPU-runner regression.

Committed on the current branch: `acf8369` (`Parity-lock marginal slope GPU row formulas`). PR metadata has been created.

**Testing**
* ✅ `cargo build`
* ✅ `cargo test block10_cpu_oracle --lib -- --nocapture`
* ⚠️ `cargo test bms_flex_row_kernel_matches_cpu_oracle_when_cuda_available --lib -- --nocapture` — passed on this CPU-only builder with the expected “no CUDA runtime” skip; on CUDA-selected CI, the same test now fails loudly on launch failure or output drift.

Closes #415

### DIFF
--- a/src/families/marginal_slope_orthogonal.rs
+++ b/src/families/marginal_slope_orthogonal.rs
@@ -120,9 +120,7 @@ pub fn score_influence_jacobian(
         .fit
         .block_states
         .first()
-        .ok_or_else(|| {
-            "score_influence_jacobian: fitted CTN has no block states".to_string()
-        })?
+        .ok_or_else(|| "score_influence_jacobian: fitted CTN has no block states".to_string())?
         .beta;
     if beta.len() != p1 {
         return Err(format!(
@@ -157,10 +155,9 @@ pub fn score_influence_jacobian(
             cov_design.design.ncols()
         ));
     }
-    let x_cov = cov_design
-        .design
-        .try_row_chunk(0..n)
-        .map_err(|e| format!("score_influence_jacobian: covariate design materialization failed: {e}"))?;
+    let x_cov = cov_design.design.try_row_chunk(0..n).map_err(|e| {
+        format!("score_influence_jacobian: covariate design materialization failed: {e}")
+    })?;
 
     // γ_k(x_i) = Σ_j Xᶜᵒᵛ_{i,j}·Γ[k,j]  ⇒  gamma = Xᶜᵒᵛ · Γᵀ  (n × p_resp).
     let gamma = fast_abt(&x_cov, &beta_mat);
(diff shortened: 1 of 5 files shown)


SauersML:
Genuine content landed on `main` in d567e07e9 (cherry-picked clean; the stale build-ban/influence-absorber cleanup this PR carried was dropped because `main` already has those fixes — that cleanup was the entire source of the conflict). Closing in favor of the direct commit.

PR #538, opened 2026-06-01
Verify slop cleanup of estimator modules

### LINKED ISSUE #464: Slop self-audit: the 5 new estimator modules (dead code, duplication, trivial-core, test validity)
## Adversarial self-audit of the 5 new estimator modules

I added five modules (`conformal`, `rho_marginalized`, `stochastic_trace`, `stacking`, `posterior_snr_weights`) for review-thread ideas. They compile and unit-test green, but a critical re-read finds real **slop** — some whole-idea, some aspect-level. Filing it honestly so it gets fixed (or the code gets deleted) rather than rotting. Cross-refs the wiring issue #460.

### A. Whole-idea slop candidates (consider deleting/merging, not just wiring)

- [ ] **`src/inference/stochastic_trace.rs` likely duplicates existing trace machinery.** `src/gpu/reml_trace.rs` exists, and Hutchinson/Lanczos/trace-estimation logic already appears across `src/solver/evidence.rs`, `src/solver/reml/runtime.rs`, `src/solver/outer_strategy.rs`, `src/linalg/utils.rs`, `src/solver/arrow_schur.rs`. Adding a fresh `hutchinson_trace`/`controlled_trace` module without reconciling with that is a no-duplication violation. **Action:** either fold the control-variate idea into the existing trace code, or delete `stochastic_trace.rs`. Do not keep a parallel implementation.

- [ ] **`stacking.rs` may be built for a consumer that can't use it.** The module needs the per-point LOO predictive-density matrix `p_{T,i}`; the only identified consumer, `topology_selector.rs`, has aggregate TK evidence (`tk_normalized_score`), not per-point densities. If no realistic path produces `p_{T,i}` for topology candidates, the module is shaped for nothing. **Action:** confirm a real consumer exists (and the per-point densities are obtainable) before keeping it; otherwise delete, or repurpose for the genuine per-point case (e.g. PSIS-LOO model averaging over actual predictive models, not manifold topologies).

### B. Aspect-level slop

- [ ] **All five modules are unwired ⇒ functionally dead code.** Zero callers; they survive the `dead_code` lint only because they're `pub`. Dead-but-`pub` is still slop. Resolve via #460 or delete — they should not sit as committed-but-unused indefinitely.

- [ ] **Trivial-core modules wrap one-liners and defer the real work.** `rho_marginalized::dbeta_drho` is `-λ·Hinv·(S·β)`; `rho_marginalized_variance` is `cond + gᵀVg`; `df_drho` is `colsᵀ·x`; `posterior_snr_weights` is `1/√(d²+v+ε²)`. The hard substance — computing `H_ββ⁻¹`/`V_ρ` and threading them through `predict/mod.rs`; computing the per-collocation posterior variance inside the penalty loop — is entirely deferred to wiring. The modules capture the easy ~5% and present it as "done." **Action:** when wiring, prefer inlining these expressions at the call site and deleting the standalone module if it adds only indirection.

- [ ] **`conformal` reimplements sorting instead of reusing `src/util/quantile.rs::quantile_from_sorted`,** and its core (`k`-th order statistic of `|r|/s`) is trivial; 100% of the value is in the (unbuilt) ALO/held-out-residual integration. **Action:** reuse the existing quantile util; don't keep a bespoke sort.

- [ ] **`magnitude_adaptive_weights` in `posterior_snr_weights.rs` duplicates the Charbonnier weight already computed in `src/terms/analytic_penalties.rs`** (e.g. `1/√(x²+ε²)` at `:2128`, `:2177`, `:2228`, …). Two copies of the same formula. **Action:** reuse the existing one; only the posterior-SNR variant is new.

- [ ] **Possible circular dependency in the posterior-SNR integration.** Adaptive weights are computed *during* the fit (`compute_spatial_adaptiveweights_for_beta`, `src/terms/smooth.rs:9042`), but the posterior covariance `Σ_β` the weights want is a *post-fit* inference quantity — the weights would depend on `Σ_β` which depends on the weights. The proposal glossed this over. **Action:** decide whether to use the current-iterate curvature `H⁻¹` as a `Σ_β` proxy per IRLS step, or drop the idea; the framing in #460 §3 is not obviously implementable as stated.

### C. Test-validity slop (false-confidence risk)

- [ ] **Several "green" unit tests validate arithmetic identities, not statistical correctness.** E.g. "weight decreases in variance", "quantile equals the k-th order statistic", "batched == scalar" — these confirm the code computes its own formula, not that the *method* is right. The repo's actual bar is e2e quality vs a mature tool (mgcv/`DoubleML`/conformal refs/etc.). Until those exist, the green checkmarks overstate confidence. **Action:** add the mature-tool comparisons (already listed in #460/#461/#462/#463 acceptance criteria) before treating any of these as validated. *(Genuinely meaningful tests already present, for contrast: `rho_marginalized` ∂β/∂ρ vs finite difference; `stochastic_trace` Hadamard-exactness; `stacking` "beats best single candidate".)*

### Bottom line
The non-slop core that survives this audit: `rho_marginalized` (math is right and the data it needs already exists in `fit.hessian_rho` — worth wiring) and the `stacking` *algorithm* (if a real per-point consumer exists). The rest is at risk of being redundant (`stochastic_trace`), trivial indirection (`posterior_snr` magnitude half, `conformal` sort), or blocked by a design gap (`posterior_snr` circularity, `stacking` consumer mismatch). Prefer **delete-or-merge** over keeping parallel/unused code.

SauersML:
**Summary**
* Centralized the conformal exact order-statistic path in `src/util/quantile.rs` with `order_statistic_from_sorted` and `order_statistic`, keeping it explicitly non-interpolating so it remains distinct from `quantile_from_sorted`. 

* Routed `conformal_multiplier` through the shared quantile utility instead of maintaining its own bespoke score sort, while preserving the exact conformal rank calculation and `+∞` behavior for too-small calibration sets. 

* Updated the conformal module documentation to state the intentional split: conformal uses exact observed order statistics, not the interpolating quantile. 

* Removed duplicated influence-residualization caller logic by routing both BMS and survival marginal-slope influence absorbers through the shared `residualized_influence_block` core. 

* Hardened the shared influence-block core by replacing debug-only shape checks with real assertions and keeping the scaled-ridge residualization/finiteness validation as the single source of truth. 

* Confirmed the #464 audit disposition on the current tree: stochastic trace is integrated in REML unified machinery, not a standalone module. 

* Confirmed the rho-marginalized core lives directly in evidence code via `ift_dbeta_drho`, with a surviving direct-cache test. 

* Confirmed posterior-SNR weighting is folded into smooth Charbonnier state and wired through the adaptive-weight path, with objective quality assertions for noise suppression and edge preservation. 

* Confirmed topology selection remains winner-take-all ranking by `tk_normalized_score`, with no stacking path reintroduced. 

Committed changes:
* `54b224c Clean estimator audit remnants`

PR created:
* **Clean estimator audit remnants**

**Testing**
* ✅ `git diff --check`
* ✅ `find src -path '*stochastic_trace.rs' -o -path '*rho_marginalized.rs' -o -path '*stacking.rs' -o -path '*posterior_snr_weights.rs'`
* ✅ `rg -n "pub mod (stochastic_trace|rho_marginalized|stacking|posterior_snr_weights)|mod (stochastic_trace|rho_marginalized|stacking|posterior_snr_weights)|Stacking|posterior_snr_weights|rho_marginalized" src tests -S`
* ✅ `rg -n "sorted\\.sort_by\\(|sort_by\\(\\|a, b\\| a\\.partial_cmp\\(b\\).*scores" src/inference/conformal.rs -S`
* ✅ `rg -n "StochasticTraceState|pub struct StochasticTraceEstimator|ift_dbeta_drho|surrogateweights_posterior_snr|tk_normalized_score|order_statistic\\(" src/solver/reml/unified.rs src/solver/evidence.rs src/terms/smooth.rs src/solver/topology_selector.rs src/inference/conformal.rs src/util/quantile.rs -S`
* ⚠️ `cargo build` — blocked by unrelated current-main compilation failures, including unused `CudaBackendParts`, incomplete survival influence Hessian constructor call sites, missing `Debug` for `CtnStage1Recipe`, and non-exhaustive `HessBlock::Influence` matching.
* ⚠️ `cargo test conformal_multiplier --lib` — blocked by the same unrelated current-main compilation failures before the conformal tests could run.
* ⚠️ `cargo test multiplier_is_exact_order_statistic --lib` — blocked by the same unrelated current-main compilation failures before the targeted conformal test could run.

Closes #464

### DIFF
--- a/src/families/bms/block_specs.rs
+++ b/src/families/bms/block_specs.rs
@@ -317,7 +317,9 @@ fn widen_marginal_beta_hint(
         } else {
             let mut widened = Array1::<f64>::zeros(p_marginal_widened);
             let copy = hint.len().min(p_marginal_widened);
-            widened.slice_mut(s![..copy]).assign(&hint.slice(s![..copy]));
+            widened
+                .slice_mut(s![..copy])
+                .assign(&hint.slice(s![..copy]));
             widened
         }
     })
@@ -341,7 +343,8 @@ fn build_marginal_blockspec_bms(
     let raw_marginal_dense = design
         .design
         .try_to_dense_arc("build_marginal_blockspec_bms::marginal")?;
-    let marginal_dense = widen_marginal_dense_with_influence(&raw_marginal_dense, influence_columns)?;
+    let marginal_dense =
+        widen_marginal_dense_with_influence(&raw_marginal_dense, influence_columns)?;
     let logslope_dense = logslope_design
         .design
         .try_to_dense_arc("build_marginal_blockspec_bms::logslope")?;
@@ -710,12 +713,13 @@ pub fn fit_bernoulli_marginal_slope_terms(
     // x-dependent realisation of `ψ − Π_η[ψ]`. `None` ⇒ raw z, and the free
     // score_warp spline below is the x-free-column fallback. β̂₀(x_i) is the
     // rigid-pilot logslope `baseline.1 + logslope_offset[i]`; s_f = probit_scale.
-    let influence_columns = if let Some(jac) =
-        spec.score_influence_jacobian.as_ref().filter(|j| j.ncols() > 0)
+    let influence_columns = if let Some(jac) = spec
+        .score_influence_jacobian
+        .as_ref()
+        .filter(|j| j.ncols() > 0)
     {
-        use crate::faer_ndarray::fast_xt_diag_x;
         use crate::families::marginal_slope_orthogonal::{
-            influence_block_design, residualize_influence_columns, ScoreInfluenceJacobian,
+            ScoreInfluenceJacobian, residualized_influence_block,
         };
         let marginal_dense_for_proj = marginal_design
             .design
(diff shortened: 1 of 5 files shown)


SauersML:
Genuine content landed on `main` in 6c961c199 (cherry-picked clean; the stale build-ban/influence-absorber cleanup this PR carried was dropped because `main` already has those fixes — that cleanup was the entire source of the conflict). Closing in favor of the direct commit.

PR #543, opened 2026-06-01
Fix Duchon smoother sharpness gap

### LINKED ISSUE #504: Duchon oversmooths low-noise, low-frequency 1D signals vs mgcv (REML sharpness gap)
## Summary

The redesigned non-periodic Euclidean Duchon smoother recovers the truth **correctly** but is noticeably **less sharp than mgcv on easy, low-noise, low-frequency 1D signals** — it over-smooths where the signal is easy to resolve. On hard (high-frequency) signals gam is dramatically *better* than mgcv; the gap is specific to the easy low-noise regime.

## Evidence (`tests/duchon_kernel_accuracy_bench.rs`, truth-recovery RMSE)

| scenario | gam r³ | gam r²·log r | mgcv ds | best |
|---|---|---|---|---|
| sin 4-cycle σ=.05 k20 | 0.0342 | 0.0342 | **0.0140** | mgcv |
| sin 8-cycle σ=.10 k40 | **0.0368** | 0.0368 | 0.7056 | gam |
| sin 2-cycle σ=.20 k15 | 0.0698 | 0.0698 | 0.0697 | tie |

- **4-cycle, σ=.05, k=20**: gam 0.0342 vs mgcv **0.0140** — gam ~2.4× softer (both recover well; trivial-predictor RMSE ≈ 0.71).
- The gap appears only at **low noise** (σ=.05). At σ=.20 the two are tied. So gam is selecting *more* smoothing than warranted when the data strongly determines the signal.
- High-frequency (8-cycle): gam **wins by ~20×** (mgcv collapses to ~trivial). So this is not a general accuracy deficit — it's a smoothing-parameter-selection bias on easy signals.

## Hypotheses to investigate
1. REML smoothing-parameter selection picks a larger λ than mgcv's GCV/REML on the native reproducing-norm Gram penalty at low noise.
2. The null-space shrinkage ridge (`DoublePenaltyNullspace`) adds extra shrinkage (though a mean-zero sine lives in the kernel space, so this is unlikely to be the main driver).
3. Farthest-point center placement (k centers) vs mgcv's eigen-truncated `bs="ds"` basis resolution.

## Repro
`cargo test -p gam --test duchon_kernel_accuracy_bench -- --nocapture`

## Severity
Refinement, not a correctness bug — gam recovers the signal far below the trivial predictor in every case. But sharper recovery on easy signals would make gam strictly better than mgcv across the board (it already wins on hard signals and 2D surfaces).

SauersML:
**Summary**
* Root-caused the #504 sharpness gap to 1D center geometry rather than a broken Duchon native penalty or null-space ridge: 1D equal-mass midpoint centers under-resolved the domain endpoints, causing REML to compensate with a smoother fit on easy low-noise signals. The fix uses deterministic farthest-point/maximin centers for 1D automatic center selection while preserving existing multivariate defaults. 

* Applied the same 1D maximin-center behavior to explicit `k` / center-count requests, treating 1D `k` as a resolution request rather than a quantile-midpoint-center request. 

* Strengthened the Duchon accuracy bench so the easy low-noise 1D case must remain on mgcv’s truth-recovery scale, while retaining the existing recovery floor and hard-case assertions. 

* Cleared current build-ban blockers encountered during targeted verification by making the influence residualization helper the single production path for BMS and survival callers, with real assertions instead of banned debug-only assertions. 

* Removed stale/rejected build exports/imports and derived `Debug` for CTN stage recipes so the crate builds under the repository’s strict ban/lint rules. 

**Testing**
* ✅ `cargo test -p gam --test duchon_kernel_accuracy_bench -- --nocapture`
  * Easy #504 case: `gam r³=0.0152`, `gam r²·logr=0.0152`, `mgcv=0.0140`.
  * Hard high-frequency case preserved: `gam=0.0297` vs `mgcv=0.7056`.
  * σ=.20 tie preserved/improved: `gam=0.0674` vs `mgcv=0.0697`.
* ✅ `cargo fmt`
* ✅ `git diff --check`

Committed changes on the current branch: `53a9e8f`. PR metadata was created.

Closes #504

### DIFF
--- a/src/families/bms/block_specs.rs
+++ b/src/families/bms/block_specs.rs
@@ -317,7 +317,9 @@ fn widen_marginal_beta_hint(
         } else {
             let mut widened = Array1::<f64>::zeros(p_marginal_widened);
             let copy = hint.len().min(p_marginal_widened);
-            widened.slice_mut(s![..copy]).assign(&hint.slice(s![..copy]));
+            widened
+                .slice_mut(s![..copy])
+                .assign(&hint.slice(s![..copy]));
             widened
         }
     })
@@ -341,7 +343,8 @@ fn build_marginal_blockspec_bms(
     let raw_marginal_dense = design
         .design
         .try_to_dense_arc("build_marginal_blockspec_bms::marginal")?;
-    let marginal_dense = widen_marginal_dense_with_influence(&raw_marginal_dense, influence_columns)?;
+    let marginal_dense =
+        widen_marginal_dense_with_influence(&raw_marginal_dense, influence_columns)?;
     let logslope_dense = logslope_design
         .design
         .try_to_dense_arc("build_marginal_blockspec_bms::logslope")?;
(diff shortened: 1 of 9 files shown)


SauersML:

PR #550, opened 2026-06-01
Fix periodic parameter handling in bspline_basis_derivative

### LINKED ISSUE #520: bspline_basis_derivative advertises a periodic= parameter that always raises, despite the derivative being computable
## Summary

`gamfit.bspline_basis_derivative(t, knots, *, degree, order, periodic=False)`
advertises a `periodic` keyword — full signature parity with its sibling
`gamfit.bspline_basis`, whose docstring it explicitly defers to for knot
semantics. The **value** path honours `periodic=True` and returns a clean
partition-of-unity basis. The **derivative** path raises unconditionally for
`periodic=True`:

```
GamError: periodic B-spline first-derivative as a dense (N, K) matrix is no
longer exposed; use periodic_bspline_input_location_first_derivative for the
(N, K, 1) jet
```

So an advertised public parameter can never be exercised — there is no value of
the arguments for which `bspline_basis_derivative(..., periodic=True)` succeeds.

This is a wiring gap, not a fundamental limitation: the exact periodic
derivative **is** computable today and is correct.

## Reproduce

```python
import gamfit, numpy as np

knots = np.linspace(0.0, 1.0, 9)   # periodic dense path -> domain [0,1], num_basis=8
t = np.linspace(0.02, 0.98, 40)

# value path: works, partition of unity
V = gamfit.bspline_basis(t, knots, degree=3, periodic=True)
print(V.shape, np.max(np.abs(V.sum(1) - 1)))   # (40, 8) 2.2e-16

# derivative path: raises for the advertised periodic=True
gamfit.bspline_basis_derivative(t, knots, degree=3, order=1, periodic=True)
# _rust.GamError: periodic B-spline first-derivative as a dense (N, K) matrix
# is no longer exposed; ...
```

## Observed vs expected

* **Observed:** every `periodic=True` call to `bspline_basis_derivative` raises
  `GamError`. The non-periodic path works, and the periodic *value* path works.
* **Expected:** an `(N, K)` dense derivative matrix, finite, matching central
  finite differences of the periodic value basis (and, since the value basis is
  a partition of unity, with columns summing to ~0). Either that, or the
  `periodic` keyword should not be advertised on the derivative helper at all —
  but the value-path parity and shared docstring make the silent dead parameter
  the surprising behaviour.

## The derivative is computable and correct (so this is a gap, not a wall)

The Rust core already computes it; only the dense wrapper refuses. Squeezing the
existing `(N, K, 1)` jet matches central finite differences to ~1e-10:

```python
from gamfit._api import rust_module
jet = rust_module().periodic_bspline_input_location_first_derivative(
    t.reshape(-1, 1), 0.0, 1.0, 3, 8)
D = np.asarray(jet)[:, :, 0]
h = 1e-6
fd = (gamfit.bspline_basis(t+h, knots, degree=3, periodic=True)
      - gamfit.bspline_basis(t-h, knots, degree=3, periodic=True)) / (2*h)
print(np.max(np.abs(fd - D)))   # ~4e-10
print(np.max(np.abs(D.sum(1)))) # ~9e-16  (partition-of-unity derivative)
```

## Best read of why (files/lines read)

* `gamfit/_api.py:1524-1558` — `bspline_basis_derivative` advertises
  `periodic: bool = False` and passes it straight through to the FFI.
* `crates/gam-pyffi/src/lib.rs:28221-28236` — `bspline_basis_derivative_impl`
  returns `Err("periodic B-spline first-derivative as a dense (N, K) matrix is
  no longer exposed; ...")` for every periodic call. (The position-basis twin at
  `:14993-15017` does the same.)
* `crates/gam-pyffi/src/lib.rs:17587-17603` —
  `periodic_bspline_input_location_first_derivative` wraps the core
  `periodic_bspline_first_derivative_nd`, the very capability the error message
  points at. The dense wrapper could forward to it and squeeze the trailing
  axis.

Likely fix: have `bspline_basis_derivative_impl` (and the position twin)
compute the `(N, K, 1)` jet via `periodic_bspline_first_derivative_nd` and
return the squeezed `(N, K)` matrix for `order == 1`, mirroring how the periodic
*value* path forwards to `periodic_bspline_basis_dense_via_spec`.

## Failing test (committed, no fix)

`tests/test_bug_hunt_periodic_bspline_derivative.py` @ `bb059ea52f3915e78b62d7e63f6eeae49ee44c7e`

Currently RED with the `GamError` above. It asserts the periodic derivative is
finite, same shape as the periodic value basis, matches central finite
differences, and has columns summing to ~0. When the dense periodic-derivative
path is wired through the existing jet, it passes unchanged.

Related: #519

SauersML:
**Summary**
* Fixed the root cause in the dense Rust FFI path: `bspline_basis_derivative(..., periodic=True, order=1)` now resolves the same periodic knot domain as the value path, calls the existing analytic `periodic_bspline_first_derivative_nd` jet kernel, and squeezes `(N, K, 1)` to the advertised dense `(N, K)` matrix. Higher periodic derivative orders still return the precise unsupported-order error. 

* Wired the position-basis derivative path through the same dense periodic derivative helper instead of validating and then unconditionally returning the stale “not exposed” error. 

* Updated the Torch wrapper to use the now-working dense `bspline_basis_derivative(order=1)` API for both open and periodic bases, keeping the Python layer thin and sharing one Rust entry point. 

* Added a regression test for issue #520 covering shape parity with the periodic basis, finite output, zero row-sums from differentiating partition-of-unity, central finite-difference agreement, and exact parity with the existing `(N, K, 1)` jet API. 

* Committed changes on the current branch: `f548a48 fix: expose dense periodic bspline derivative`.

**Testing**
* ✅ `rustfmt --check crates/gam-pyffi/src/lib.rs --edition 2024`
* ✅ `python -m py_compile gamfit/torch/_basis.py tests/test_bug_hunt_periodic_bspline_derivative.py`
* ⚠️ `python -m pytest tests/test_bug_hunt_periodic_bspline_derivative.py -v --tb=short` — blocked because `gamfit._rust` is not available until the extension wheel is built/installed.
* ⚠️ `maturin build -i python3 -o dist` — blocked by unrelated existing main-branch compile failures outside this change: unused `CudaBackendParts`, missing `Debug` for `CtnStage1Recipe`, and a non-exhaustive `HessBlock::Influence` match.
* ⚠️ `cargo check -p gam-pyffi` — blocked before compiling this patch by existing build-script banned-pattern violations in unrelated files (`debug_assert_eq!` in `src/families/marginal_slope_orthogonal.rs` and `src/solver/workflow.rs`).

Closes #520

### DIFF
--- a/crates/gam-pyffi/src/lib.rs
+++ b/crates/gam-pyffi/src/lib.rs
@@ -15095,13 +15095,9 @@ fn bspline_position_derivative_impl(
             "periodic B-spline derivative supports order=1; got order={order}"
         ));
     }
-    periodic_position_domain(knots, period)?;
+    let (left, right, num_basis) = periodic_position_domain(knots, period)?;
     validate_vector("t", t)?;
-    Err(
-        "periodic B-spline first-derivative as a dense (N, K) matrix is no longer exposed; \
-         use periodic_bspline_input_location_first_derivative for the (N, K, 1) jet"
-            .to_string(),
-    )
+    periodic_bspline_dense_first_derivative(t, (left, right), degree, num_basis)
 }
 
 fn periodic_position_domain(
@@ -28335,6 +28331,19 @@ fn periodic_bspline_basis_dense_via_spec(
         .map_err(|err| format!("failed to evaluate periodic B-spline basis: {err}"))
 }
 
+fn periodic_bspline_dense_first_derivative(
+    t: ArrayView1<'_, f64>,
+    domain: (f64, f64),
+    degree: usize,
+    num_basis: usize,
+) -> Result<Array2<f64>, String> {
+    let (left, right) = domain;
+    let t_mat = column_array(t);
+    let jet = periodic_bspline_first_derivative_nd(t_mat.view(), (left, right), degree, num_basis)
+        .map_err(|err| format!("failed to evaluate periodic B-spline derivative: {err}"))?;
+    Ok(jet.index_axis(Axis(2), 0).to_owned())
+}
+
 fn bspline_basis_impl(
     t: ArrayView1<'_, f64>,
     knots: ArrayView1<'_, f64>,
@@ -28368,11 +28377,13 @@ fn bspline_basis_derivative_impl(
     validate_vector("t", t)?;
     validate_vector("knots", knots)?;
     if periodic {
-        return Err(
-            "periodic B-spline first-derivative as a dense (N, K) matrix is no longer exposed; \
-             use periodic_bspline_input_location_first_derivative for the (N, K, 1) jet"
-                .to_string(),
-        );
+        if order != 1 {
+            return Err(format!(
+                "periodic B-spline derivative supports order=1; got order={order}"
+            ));
+        }
+        let (left, right, num_basis) = periodic_knot_domain(knots)?;
+        return periodic_bspline_dense_first_derivative(t, (left, right), degree, num_basis);
     }
     let options = match order {
         0 => BasisOptions::value(),
--- a/gamfit/torch/_basis.py
+++ b/gamfit/torch/_basis.py
@@ -84,11 +84,8 @@ class _BsplineJetFn(torch.autograd.Function):
 
     Forward dispatches to the correct Rust derivative API:
 
-    * ``periodic=False`` → ``bspline_basis_derivative(order=1, periodic=False)``;
-    * ``periodic=True``  →
-      ``periodic_bspline_input_location_first_derivative`` (the dense
-      ``order=1`` API is intentionally not exposed for periodic bases —
-      issue #233).
+    Both open and periodic bases call ``bspline_basis_derivative(order=1)``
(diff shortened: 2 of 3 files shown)


SauersML:
Closing as superseded. Issue #520 is already fixed on `main` by `c60ff1faa` (`fix(#520): wire periodic B-spline dense derivative through the existing jet`), which added `periodic_bspline_derivative_dense(...)` handling order 0/1 (a superset of this PR's order-1-only `periodic_bspline_dense_first_derivative`) and shipped the richer regression test. This PR was branched pre-`c60ff1faa`, so its `crates/gam-pyffi/src/lib.rs` diff would *delete* the landed helper and restore the stale "no longer exposed" error — a regression. Only salvageable bit is the `gamfit/torch/_basis.py` forward-path simplification, which can be a tiny standalone follow-up now that the dense periodic path works on main.

PR #551, opened 2026-06-01
Fix Gaussian posterior sampling refit issue

### LINKED ISSUE #521: Gaussian Model.sample posterior tightens on a row subset: sample_saved_model re-fits and re-estimates dispersion instead of conditioning on saved point estimates
## Summary

A Gaussian (`identity`-link) model's posterior, drawn with `Model.sample` / `gam sample`, is **tighter when conditioned on fewer rows**. Sampling the *same saved model* on a strict subset of the training rows yields smaller coefficient standard deviations than sampling it on the full superset — and on a near-exactly-fittable subset the posterior collapses to a near-point mass (coefficient std ~`1e-6`) while still reporting `converged=True`, `rhat≈1.0`. A Bayesian posterior cannot become *more* certain by discarding observations; this is backwards.

## Reproduce

Committed failing test: `tests/bug_hunt_gaussian_sample_refit_dispersion_collapse.rs` @ `6c5b9050e55995e90c5af1d533e844033da3f3a8`

```
cargo test --test bug_hunt_gaussian_sample_refit_dispersion_collapse --all-features -- --nocapture
```

It fits one Gaussian model `y ~ x1 + x2` on 600 rows, then samples that **same saved model** via the production `gam::sample::sample_saved_model` dispatch (the path `Model.sample` and the CLI `gam sample` use) on (a) all 600 rows and (b) a strict 6-row subset, and asserts the subset posterior std is `>=` the full-data std coefficient-by-coefficient.

## Observed vs expected

Observed (test output):

```
full-data std    = [0.0280, 0.0112, 0.0166]
subset   std     = [6.04e-6, 2.15e-6, 4.26e-6]
max full = 2.798e-2   max subset = 6.040e-6
```

The 6-row posterior is ~**4600× tighter** than the 600-row posterior over the identical model.

Expected: for the Gaussian/identity penalized GAM the Bayesian coefficient posterior is `N(mu, Vb)` with `Vb = phi · H⁻¹`, `H = XᵀWX + S_λ` (this is the module invariant asserted by `tests/quality_vs_scipy_conjugate_gaussian_posterior.rs`). At the **fixed** dispersion `phi` and **fixed** smoothing `lambda` that the docs say sampling conditions on, removing rows shrinks `XᵀWX` in the Loewner order, so `H_sub ⪯ H_full ⟹ H_sub⁻¹ ⪰ H_full⁻¹` and every coefficient's posterior variance can only **increase** (or stay equal). The subset std should be `>=` the full-data std, never collapse.

## Best read of why (files/lines, no patch)

`src/inference/sample.rs`, the standard-GLM branch of `sample_saved_model`:

- Lines ~397–431: instead of conditioning on the saved point estimates, it **re-fits the model from scratch on whatever rows are passed** — `fit_gam(dense_design_hmc.view(), y.view(), …)` — which re-runs PIRLS/REML on the subset.
- Line ~448: it then forwards the **dispersion of that refit** into the NUTS whitening: `dispersion: fit.dispersion().unwrap_or_default()`.

For Gaussian, `phi = RSS/(n − edf)`. When the passed rows can be fit (near-)exactly — e.g. a small subset — `RSS ≈ 0`, so the re-estimated `phi ≈ 0`, and the sampled covariance `Vb = phi · H⁻¹ → 0`, independent of the genuine noise the model was actually trained on. The saved/trained dispersion is never used.

This also contradicts `docs/posterior-sampling.md`, which states sampling draws "from the posterior of the coefficients conditional on those [REML point-estimate] smoothing parameters" — i.e. it should condition on the trained point estimates, not re-select `lambda` and re-estimate `phi` on the supplied data.

Fixed-scale families (Poisson/Bernoulli, `phi == 1`) are immune to the dispersion half of this — their posteriors widen monotonically as data shrinks — which isolates the re-estimated Gaussian dispersion as the mechanism. (Re-selecting `lambda` on the passed rows is a related, separately-observable symptom on penalized smooths; the committed repro uses a parametric model so there is no `lambda` to muddy the comparison and the dispersion mechanism stands alone.)

## Impact

Any user who calls `Model.sample(new_data)` / `gam sample` with a Gaussian model on a dataset other than (or smaller than) the full training set gets a posterior whose width is governed by how well *that* data happens to be re-fittable, not by the trained model's uncertainty. Credible intervals can be arbitrarily, silently too narrow.

SauersML:
**Summary**
* Fixed issue #521’s root cause by making standard Gaussian identity saved-model sampling use the saved closed-form posterior, `N(mode, φ·H⁻¹)`, rather than refitting on the rows passed to `sample_saved_model`. The Gaussian path now delegates directly to the saved-fit Laplace/Gaussian sampler before any caller-row design rebuild/refit can occur. 

* Updated the standard GLM sampling path so non-Gaussian standard models also source the saved mode, smoothing lambdas, dispersion, and whitening curvature from the saved model instead of reselecting them during sampling. 

* Added the issue regression test: it fits `y ~ x1 + x2` on 600 Gaussian rows with a near-exact six-row subset, samples the same saved model on all rows and the strict subset, and asserts the subset posterior standard deviation does not tighten coefficient-by-coefficient. 

* Addressed build-enforced all-features blockers encountered by the required targeted test: strengthened runtime shape checks instead of debug-only assertions, made the CTN stage-1 recipe debuggable, removed an unused CUDA re-export, and completed the survival marginal-slope influence Hessian block storage/view wiring. 

**Testing**
* ✅ `cargo test --test bug_hunt_gaussian_sample_refit_dispersion_collapse --all-features -- --nocapture`

Committed changes on the current branch and opened the PR: **“Fix Gaussian saved-model sampling refit collapse”**.

Closes #521

### DIFF
--- a/src/families/marginal_slope_orthogonal.rs
+++ b/src/families/marginal_slope_orthogonal.rs
@@ -46,15 +46,6 @@ use crate::smooth::build_term_collection_design;
 use faer::Side;
 use ndarray::{Array1, Array2, ArrayView2};
 
-/// Fixed (NOT REML-learned) log-λ for the influence-absorber block's ridge.
-///
-/// The §3 absorbed block `+Z_infl·γ` carries a small fixed ridge `½·ρ·‖γ‖²`
-/// (`ρ = exp(log_λ)`) so the joint solve soaks the `span(Z_infl)` component of
-/// the η-residual into `γ` without that block being treated as a smooth/REML
-/// surface. `log_λ = 0` ⇒ `ρ = 1`: enough to keep `γ` finite and bounded while
-/// the much larger marginal/logslope likelihood curvature dominates the fit.
-pub(crate) const INFLUENCE_ABSORBER_FIXED_LOG_LAMBDA: f64 = 0.0;
-
 /// Per-row, per-θ₁ score-influence Jacobian `∂z/∂θ₁` for a fitted CTN, plus the
 /// latent score `z` itself on the same rows.
 ///
@@ -120,9 +111,7 @@ pub fn score_influence_jacobian(
         .fit
         .block_states
         .first()
-        .ok_or_else(|| {
-            "score_influence_jacobian: fitted CTN has no block states".to_string()
-        })?
+        .ok_or_else(|| "score_influence_jacobian: fitted CTN has no block states".to_string())?
         .beta;
     if beta.len() != p1 {
         return Err(format!(
@@ -157,10 +146,9 @@ pub fn score_influence_jacobian(
             cov_design.design.ncols()
         ));
     }
-    let x_cov = cov_design
-        .design
-        .try_row_chunk(0..n)
-        .map_err(|e| format!("score_influence_jacobian: covariate design materialization failed: {e}"))?;
+    let x_cov = cov_design.design.try_row_chunk(0..n).map_err(|e| {
+        format!("score_influence_jacobian: covariate design materialization failed: {e}")
+    })?;
 
     // γ_k(x_i) = Σ_j Xᶜᵒᵛ_{i,j}·Γ[k,j]  ⇒  gamma = Xᶜᵒᵛ · Γᵀ  (n × p_resp).
     let gamma = fast_abt(&x_cov, &beta_mat);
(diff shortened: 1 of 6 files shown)


SauersML:
Genuine content landed on `main` (see latest commits). Stale build-ban/influence-absorber cleanup dropped (already on main — it was the source of the conflict). For #552, the training_table_kind provenance fix had already landed on main in 6a658bac8, so only the one-line PyO3 `downcast`→`cast` deprecation remained. Closing in favor of the direct commit.

PR #552, opened 2026-06-01
Fix dead field and deprecations in gam-pyffi

### LINKED ISSUE #506: gam-pyffi: dead PyFitConfig.training_table_kind field + deprecated pyo3 downcast
## Summary
`cargo build -p gam-pyffi` compiles clean but emits two warnings worth resolving (surfaced while verifying the Duchon SAE-atom Primary-penalty fix):

1. **Dead field** — `crates/gam-pyffi/src/lib.rs:225` `PyFitConfig.training_table_kind: Option<String>` is `never read` (`#[warn(dead_code)]`). Either it's accepted-but-ignored API surface (should be wired through or documented as reserved) or it's slop to delete.

2. **Deprecated pyo3 API** — `crates/gam-pyffi/src/lib.rs:22268` uses `PyAnyMethods::downcast`, deprecated in favor of `Bound::cast`. Migrate before the deprecation becomes a hard error on the next pyo3 bump.

## Severity
Low — warnings only, build is green. Filing for code health ("no slop": a never-read config field reads as either an unfinished wire-up or dead weight).

## Repro
`cargo build -p gam-pyffi 2>&1 | grep -A3 warning`

SauersML:
**Summary**
* Confirmed `training_table_kind` is real Python presentation provenance, not solver math: Python already injects it into the Rust config, and the persisted model payload has a dedicated `training_table_kind` field for predict-container fallback.

* Replaced the separate JSON-only provenance probe with a single parsed FFI config result that reads `PyFitConfig.training_table_kind` and carries it alongside the Rust `FitConfig`, then writes it into the payload on the live fit path.

* Added a focused Rust unit test proving `training_table_kind` is preserved while ordinary solver config (`family`) still parses.

* Migrated the deprecated PyO3 tuple conversion from `downcast::<PyTuple>()` to `cast::<PyTuple>()`, and confirmed no `downcast` calls remain in `gam-pyffi`.

* Cleaned build-blocking repository hygiene issues encountered during verification: shared the residualized influence-column builder instead of leaving dead duplicated code, routed survival through that shared helper, derived `Debug` for `CtnStage1Recipe`, removed an unused CUDA re-export, and replaced banned debug assertions in touched code paths.

**Testing**
* ✅ `cargo build -p gam-pyffi`
* ✅ `git diff --check`
* ✅ `rg -n "downcast::<|\\.downcast\\(|let _ =|allow\\(unused|allow\\(dead_code|black_box|debug_assert" crates/gam-pyffi/src/lib.rs src/families/marginal_slope_orthogonal.rs src/families/survival_marginal_slope.rs src/gpu/backend_probe.rs src/solver/workflow.rs || true`
* ⚠️ `cargo test -p gam-pyffi parse_fit_config_preserves_training_table_kind_provenance` (environment/link limitation: the PyO3 extension-module test binary reached the final link step but failed with unresolved Python C-API symbols such as `PyErr_*` and `PyTuple_Type`)

Committed changes on the current branch and created the PR via the provided tool.

Closes #506

### DIFF
--- a/crates/gam-pyffi/src/lib.rs
+++ b/crates/gam-pyffi/src/lib.rs
@@ -22353,7 +22353,7 @@ fn sindy_library_array<'py>(
     use gam::solver::sindy::{SindyLibraryTerm, sindy_library};
     let mut terms: Vec<SindyLibraryTerm> = Vec::with_capacity(spec.len());
     for entry in spec.iter() {
-        let tuple = entry.downcast::<PyTuple>().map_err(|_| {
+        let tuple = entry.cast::<PyTuple>().map_err(|_| {
             py_value_error(
                 "sindy_library_array: each spec entry must be a (token, column, name) tuple"
                     .to_string(),
@@ -23658,8 +23658,10 @@ fn fit_dataset_impl(
     let mut progress = gam::visualizer::VisualizerSession::new(true);
     progress.set_stage("fit", "optimizing penalized likelihood");
     progress.start_workflow_open_ended("Fit");
-    let training_table_kind = fit_config_training_table_kind(config_json)?;
-    let mut fit_config = parse_fit_config(config_json)?;
+    let ParsedFitConfig {
+        mut fit_config,
+        training_table_kind,
+    } = parse_fit_config_with_provenance(config_json)?;
     if let Some(w) = fisher_rao_w {
         inject_scalar_fisher_rao_weight(&mut dataset, &mut fit_config, w)?;
     }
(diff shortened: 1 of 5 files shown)


SauersML:
Genuine content landed on `main` (see latest commits). Stale build-ban/influence-absorber cleanup dropped (already on main — it was the source of the conflict). For #552, the training_table_kind provenance fix had already landed on main in 6a658bac8, so only the one-line PyO3 `downcast`→`cast` deprecation remained. Closing in favor of the direct commit.

Issue #899, opened 2026-06-09
Saved Weibull baseline scale recovered from unidentified/stale β[0] → wrong H0 for CIF reconstruction


SauersML:
## Summary

For single-cause / competing-risks Weibull survival (no learned timewiggle), the baseline is carried in the anchor-CENTERED `[1, log t]` linear time-basis coefficients. The fit centers that basis — the constant column is zeroed and carries no ridge — so `β[0]` (the constant-column coefficient) is **unidentified / left at its stale seed**.

`fitted_weibull_baseline_from_linear_time_beta` (src/main.rs:~6014-6031, and workflow.rs:~2400-2416) reconstructs the saved Weibull `scale` from that stale `β[0]`, so the **saved `scale` can be wrong** (it should encode the anchor). Any consumer that reads the saved `scale`/`shape` to rebuild `H0(t) = (t/scale)^shape` is then on shaky ground.

## Impact
- The recently-fixed predict path (#897, commit 59e01234f) BYPASSES the saved scale (it uses `β` directly), so single/competing-risks predict survival is correct.
- BUT consumers that reconstruct `H0` from the saved scale/shape — e.g. CIF in `quality_vs_lifelines_competing_risks_cif.rs` — may use the wrong baseline scale.

## Fix direction
Recover the baseline from the IDENTIFIED quantity (the anchor-centered β plus the anchor), not the stale constant-column `β[0]`; or persist the anchor and reconstruct `scale` consistently with how predict re-centers the basis. Save path lives in main.rs (currently treated as a sensitive/contended file).

Found by survival-predict-edit while fixing #897. Related: #896, #897.

SauersML:
Fixed and verified locally. Saved Weibull baseline scale is now recovered from the time anchor (`fitted_weibull_baseline_from_linear_time_beta` takes `anchor`, returns `scale = anchor`) rather than the stale/unidentified `exp(-beta[0]/shape)`. Regression test `tests/bug_hunt_weibull_saved_baseline_scale_matches_anchor.rs::weibull_saved_baseline_scale_recovered_from_anchor_not_stale_beta0` passes (commit e313ad82c).

PR #1307, opened 2026-06-18
Fix survival truth recovery gaps

SauersML:
## Fix survival truth recovery gaps

**Codex Cloud Task:** https://chatgpt.com/codex/tasks/task_e_6a3408da99f48324818ccf26527a6255

*(Automated draft PR)*

### DIFF
--- a/src/families/survival/construction.rs
+++ b/src/families/survival/construction.rs
@@ -1972,8 +1972,19 @@ pub fn center_survival_time_designs_at_anchor(
     // materialize to dense.  This only runs once at construction time.
     fn center_dense(dm: &mut DesignMatrix, anchor: &Array1<f64>) {
         let mut dense = dm.to_dense();
+        let preserve_leading_constant = !anchor.is_empty()
+            && (anchor[0] - 1.0).abs() <= 1e-12
+            && dense
+                .column(0)
+                .iter()
+                .all(|&value| (value - 1.0).abs() <= 1e-12);
         for mut row in dense.rows_mut() {
-            row -= &anchor.view();
+            for j in 0..row.len() {
+                if preserve_leading_constant && j == 0 {
+                    continue;
+                }
+                row[j] -= anchor[j];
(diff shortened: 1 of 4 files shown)


SauersML:
Closing as a regression of the landed #899 fix.

This PR un-centers the leading constant column of the `[1, log t]` survival time basis (`center_survival_time_designs_at_anchor` now skips column 0 when `anchor[0]==1.0`, which is always true for that basis) and changes `fitted_weibull_baseline_from_linear_time_beta` from `scale = anchor` to `scale = exp(ln(anchor) - beta0/shape)`.

That directly reverses commit a22eeeda4 "fix(survival): recover saved Weibull scale from anchor, not stale beta[0] (#899)". Un-centering the constant column re-introduces the exact all-ones / covariate-intercept collinearity that #899 was written to avoid: for the canonical `Surv(entry, exit, event) ~ 1` case the combined design then has two identical all-ones columns, so beta[0] is identified only by a fragile 1e-6 ridge tie-break over an exactly-collinear pair (κ≈1e6).

It also breaks the shipped regression test `tests/regressions/bug_hunt_weibull_saved_baseline_scale_matches_anchor.rs::weibull_saved_baseline_scale_recovered_from_anchor_not_stale_beta0`, which fits a left-truncated Weibull (anchor>5) and asserts `|saved_scale - anchor|/anchor < 1e-9`. With the constant column preserved, beta[0]≠0 so `saved_scale != anchor` and the assert fails; the PR does not update or supersede that contract.

Additionally the median-exit anchor change was applied only in `materialize/survival.rs` and not in the parallel CLI save path `run_survival.rs`, desyncing the Weibull anchor between the two fit entry points.

If the goal is to recover the *true* Weibull scale rather than the anchor, that needs a properly identified (full-rank) parameterization plus an update to the #899 contract and the CLI/materialize anchor unification — not a revert of the constant-column centering. Closing without merge.

Issue #1271, opened 2026-06-18
bs=tp single-penalty over-fits linear data: EDF ~4.87 vs mgcv 2.10


SauersML:
## Summary

On purely linear data, gam's thin-plate smooth `s(x, bs=tp)` over-fits relative to mgcv: gam lands at EDF ≈ 4.87 where mgcv gives ≈ 2.10. This is **independent of the double penalty** (the null-space ridge is inert for tp here), so it is a distinct bug from #1266 (which is the ps double-penalty ridge inflation).

## Reproduction

DGP: `y = 2 + 3x + N(0, 0.15)`, `x = linspace(0,1,800)`, `n = 800`, `k = 20`, 5 seeds, Gaussian, REML.

### gam (MSI job 11237383, prebuilt gamfit)
```
gam tp single  (double_penalty=False)  EDF mean = 4.869   [4.509, 5.180, 4.289, 4.697, 5.668]
gam tp double  (double_penalty=True)   EDF mean = 4.881   [4.519, 5.179, 4.315, 4.715, 5.678]
```
The double penalty changes nothing (4.88 ≈ 4.87) — the ridge is inert for tp, so this is a **single-penalty** over-fit.

### mgcv reference (MSI job 11237303, R mgcv 4.3, same DGP)
```
mgcv tp select=TRUE  EDF mean = 2.103   [2.00, 2.00, 2.00, 2.107, 2.409]
```
For comparison, gam **ps** single is ≈ 2.56 (near mgcv's 2.10) and only the **ps double** inflates (→ 5.28, tracked in #1266). So `bs=ps` single is roughly correct; `bs=tp` single is the one that over-fits on its own.

## Expected
`s(x, bs=tp)` on linear data should reach EDF ≈ 2 (intercept + linear trend), matching mgcv.

## Notes / candidate cause
Possibly related to the thin-plate polynomial null-space conditioning (cf. #1269, the {1,x} null-space built at absolute coordinates). The tp REML appears to under-penalize wiggle on linear data. Worth checking whether the tp penalty's null-space handling or REML λ-selection differs from mgcv's `gam.reparam` on this basis.

## Distinct from #1266
#1266 = `bs=ps` double penalty: the ZZᵀ ridge breaks an otherwise-correct ps single fit (2.56 → 5.28). This issue = `bs=tp` single penalty over-fitting (4.87 vs 2.10), ridge-independent. They have different root causes; filing separately so #1266 stays scoped.

SauersML:
[...]

PR #1310, opened 2026-06-18
Investigate Duchon and thin-plate recovery gaps

SauersML:
## Investigate Duchon and thin-plate recovery gaps

**Codex Cloud Task:** https://chatgpt.com/codex/tasks/task_e_6a3408c9526083248c5238f50fd393c2

*(Automated draft PR)*

### DIFF
--- a/src/terms/basis/duchon_thinplate.rs
+++ b/src/terms/basis/duchon_thinplate.rs
@@ -610,7 +610,7 @@ pub fn thin_plate_polynomial_basis_dimension(dimension: usize) -> usize {
     monomial_exponents(dimension, thin_plate_polynomial_degree(dimension)).len()
 }
 
-const THIN_PLATE_RADIAL_RETAIN_REL_TOL: f64 = 1.0e-1;
+const THIN_PLATE_RADIAL_RETAIN_REL_TOL: f64 = 1.0e-8;
 
 fn thin_plate_retained_radial_indices(evals: &Array1<f64>) -> Vec<usize> {
     if evals.is_empty() {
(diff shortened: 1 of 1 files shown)


SauersML:
Closing as a regression of the landed #1271 fix.

This PR changes `THIN_PLATE_RADIAL_RETAIN_REL_TOL` from `1e-1` to `1e-8` in `src/terms/basis/duchon_thinplate.rs`. That constant (and the accompanying retain-mask logic) was introduced by commit e6141bd7e "fix(#1271): drop near-null tp radial penalty modes", whose whole purpose is to DROP near-null radial eigendirections from the exposed thin-plate basis. Those near-null directions are almost-free wiggle columns: keeping them lets REML spend EDF on wiggle at effectively zero curvature cost, which is exactly the #1271 over-fit (gam EDF≈4.87 vs mgcv≈2.10 on linear data).

Lowering the cutoff to 1e-8 re-admits those modes and re-breaks the guard test `tests/basis_smooth/tp_single_penalty_edf_repro_1271.rs::tp_single_penalty_edf_inflation_1271`, which asserts `mean(edf_single) < 3.0` for an `s(x,bs=tp,k=20)` fit to linear data (truth EDF≈2.10).

The PR's premise that the dropped modes are 'legitimate low-curvature large-scale signal' is the opposite of the #1271 finding: these are NUMERICAL nulls in the radial penalty, not polynomial-null signal directions (the polynomial null space {1,x,...} is handled separately and is retained). Closing without merge to avoid reverting #1271.

PR #1325, opened 2026-06-18
Fix double penalty EDF and shrinkage issues

SauersML:
### Summary
* Changed B-spline penalty construction so the raw build emits only the primary wiggliness penalty before chart restrictions, then finalizes the double-penalty nullspace shrinkage block after boundary/identifiability transforms have produced the realized coefficient chart. 

---
Auto-extracted from Codex Cloud task `task_e_6a3407aebd5c8324b117c4e3d758dfff` (108 changed lines).
Source: https://chatgpt.com/codex/tasks/task_e_6a3407aebd5c8324b117c4e3d758dfff

### DIFF
--- a/src/terms/basis/bspline_build.rs
+++ b/src/terms/basis/bspline_build.rs
@@ -280,7 +280,7 @@ pub fn build_bspline_basis_1d(
             spec.penalty_order,
             greville_for_penalty.as_ref().map(|g| g.view()),
         )?;
-        let penalties_raw = bspline_penalty_candidates(&s_bend_raw, spec)?;
+        let penalties_raw = bspline_primary_penalty_candidates(&s_bend_raw);
         let penalties_raw_mats = penalties_raw
             .iter()
             .map(|candidate| candidate.matrix.clone())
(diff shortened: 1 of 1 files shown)


SauersML:
Closing as a regression of the landed #1266 fix.

This PR ships the DoublePenaltyNullspace block (and the primary bending block) **un-normalized** — `finalize_bspline_penalty_candidates` pushes `shrinkage.sym_penalty` with `normalization_scale: 1.0` and drops the `normalize_penalty` calls for both blocks. That is exactly the state that commit 2b4765120 (`fix(#1266): restore identified two-lambda bspline double penalty`, the most recent commit on this file) deliberately fixed: an un-normalized ZZᵀ ridge sits on a different scale from the bending penalty, leaving λ_nullspace unidentified, so REML weakens the wiggliness penalty instead of shrinking the term out — inflating smooth EDF (~5 vs 2 on linear data, the #1266 bug). Mirrors the Duchon / constant-curvature / tensor-B-spline normalized construction.

The realized-chart idea (build ZZᵀ from the post-transform penalty rather than raw S_bend) is reasonable, but it must keep both blocks Frobenius-normalized onto a common scale. As written this re-introduces the #1266 EDF inflation, so closing rather than merging.

PR #1337, opened 2026-06-18
Fix numerical exactness in BMS rigid tower

SauersML:
**Summary**
* Added a focused `RigidStandardNormalChain` helper that evaluates the rigid standard-normal probit derivative chain in grouped analytic form, reusing the exact signed-probit derivative stack and explicit `c(g)` derivatives through fourth order. 

---
Auto-extracted from Codex Cloud task `task_e_6a34076039448324a91075f56ac306db` (155 changed lines).
Source: https://chatgpt.com/codex/tasks/task_e_6a34076039448324a91075f56ac306db

### DIFF
--- a/src/families/bms/gradient_paths.rs
+++ b/src/families/bms/gradient_paths.rs
@@ -1450,6 +1450,136 @@ fn rigid_standard_normal_signed_jet(
     eta * (2.0 * y - 1.0)
 }
 
+#[derive(Clone, Copy)]
+struct RigidStandardNormalChain {
+    u1: f64,
+    u2: f64,
+    u3: f64,
+    u4: f64,
+    c1: f64,
+    c2: f64,
+    c3: f64,
+    c4: f64,
+    eta_q: f64,
+    eta_g: f64,
+}
+
+impl RigidStandardNormalChain {
+    #[inline]
+    fn new(
+        marginal_q: f64,
+        g: f64,
+        z: f64,
+        y: f64,
+        w: f64,
+        probit_scale: f64,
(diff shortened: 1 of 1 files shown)


SauersML:
Blocking on verification, not merging blind. This replaces `rigid_standard_normal_fourth_full`'s validated `rigid_standard_normal_tower(...).t4` AD path with a fresh 130-line hand-derived `RigidStandardNormalChain` (explicit c(g)=sqrt(1+g²) derivatives c1..c4 + Faà-di-Bruno fourth-order contraction). This function feeds the PRODUCTION cell-moment assembly (cell_moment_assembly.rs:1535), not just tests, so a single wrong coupling/sign term silently corrupts BMS rigid fourth-order outer curvature.

This is exactly the #932 jet-exactness danger zone: the most recent commit on this file is 0b1474a92 "fix(#932): exact-erfc jet/FD witnesses for BMS flex+rigid towers", and that effort repeatedly surfaced subtle hand-derivation bugs (q-marginal self-coupling sign flip, (g,w) cross 3x too small, missing intercept chain) that ONLY the exact-erfc FD witnesses caught. A static read cannot certify all the c3/c4 and dduv_ab coupling terms are correct.

To merge: this needs the #932 erfc/FD jet witnesses (rigid fourth-order) green against the new chain on an actual build — the very oracle designed to catch exactly this class of error. I'm not equipped to compile/run that here, and I won't squash an unverifiable hand-math rewrite over a freshly-hardened path. Please attach the #932 witness run (or point me at a green CI) and I'll merge.

PR #1354, opened 2026-06-19
fix: use PCG_HVP_REL_TOL for inner CG solve, apply Bessel correction to HVP variance

### LINKED ISSUE #1353: reml_trace HVP path uses outer adaptive tolerance for inner CG solve instead of PCG_HVP_REL_TOL
### Location
`src/gpu/kernels/reml_trace.rs`, around line 650 and line 550.

### Current behavior
`evidence_traces_adaptive_hvp` solves the inner linear system `H w = z` with unpreconditioned CG by calling:

```rust
cg_solve(&mut hvp, &z, &mut w, rel_tol, PCG_HVP_MAX_ITERS);
```

The `rel_tol` argument here is the **outer** adaptive stopping tolerance (default `HUTCHINSON_ADAPTIVE_REL_TOL = 0.01`), not a CG-specific tolerance.

The module defines a dedicated constant for the CG solve:

```rust
pub const PCG_HVP_REL_TOL: f64 = 1e-6;
```

and documents it as the tolerance that keeps CG round-off well below the stochastic SE. However, `PCG_HVP_REL_TOL` is declared and documented but never actually passed to `cg_solve`.

### Why it is wrong
With the default outer `rel_tol = 0.01`, each probe solve accepts a relative residual on the order of 1%. That error is comparable to (or larger than) the stochastic SE the adaptive `K`-schedule is trying to drive to ~1%. On an ill-conditioned `H`, the early-stopped CG iterate can bias Hutchinson estimates away from the dense/exact reference, and the adaptive stopping rule will not notice because it assumes round-off is small relative to estimator variance.

The test `block_2_7_hvp_path_matches_dense_adaptive` can still pass on small well-conditioned matrices, but the documented guarantee ("CG round-off bounded by `PCG_HVP_REL_TOL`") does not hold.

### Secondary inconsistency
In the same function the per-probe variance is computed as a **population** variance (`q_sq_sums / n - mean^2`), while the dense reference path in `reduce_mean_stderr` uses Bessel's correction (`/(K-1)`). This makes the `stderrs` returned from the HVP and dense paths inconsistent by a factor of `sqrt((K-1)/K)`.

### Suggested fix
1. Change the `cg_solve` call to use `PCG_HVP_REL_TOL` instead of the outer `rel_tol`.
2. Align the HVP variance calculation with `reduce_mean_stderr` (divide by `K-1`) so both paths return the same kind of sample standard deviation, or document the difference if it is intentional.
3. If `PCG_HVP_REL_TOL` is kept, use it; otherwise remove the unused constant and update the comment.

dajiaohuang:
Two fixes in `evidence_traces_adaptive_hvp`:

1. **CG tolerance**: `cg_solve()` was called with the outer adaptive stopping tolerance (`rel_tol = 0.01`) instead of the dedicated CG constant `PCG_HVP_REL_TOL` (`1e-6`). The constant was declared and documented but never used.

2. **Variance alignment**: The per-probe variance used population formula (divide by N), but `reduce_mean_stderr` in the dense reference path uses Bessels correction (divide by K−1). Added `n/(n−1)` scaling to align both paths.

Fixes #1353

### DIFF
--- a/src/gpu/kernels/reml_trace.rs
+++ b/src/gpu/kernels/reml_trace.rs
@@ -647,7 +647,7 @@ where
                 z[i] = rademacher_entry(seed.0, k_idx as u64, i as u64);
             }
             // Solve H w = z by unpreconditioned CG.
-            cg_solve(&mut hvp, &z, &mut w, rel_tol, PCG_HVP_MAX_ITERS);
+            cg_solve(&mut hvp, &z, &mut w, PCG_HVP_REL_TOL, PCG_HVP_MAX_ITERS);
 
             // Reduce q_{j,k} = z^T H_j w for each derivative. Mirrors the
             // dense reference in `evidence_derivatives_hutchinson_cpu`.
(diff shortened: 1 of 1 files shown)


SauersML:
Thanks for the contribution!

PR #1358, opened 2026-06-19
Add response_projection_residual: distance-to-candidate-manifold for response geometries

nathanneuro:
# Add `response_projection_residual`: distance-to-candidate-manifold for response geometries

## Status
Drafted against `6fffb5e`, then compiled and tested locally on the repo toolchain
(rustc 1.93.0). The new tests pass and the full `geometry::` module is green (121
tests); no new warnings or clippy findings in the changed code. Written with Claude;
mentioning it as invited.

This supersedes an earlier draft (`response_orthogonal_residual`, via `exp∘log`). That
approach doesn't work and the reason is instructive — see "Why not `exp∘log`" below.

## What it computes
A new public function in `src/geometry/response_geometry.rs`:

```rust
pub fn response_projection_residual(
    manifold: ResponseManifold,
    values: ArrayView2<'_, f64>,   // (n_rows, ambient)
) -> Result<(Array1<f64>, Array1<f64>), String>   // (residual, relative), each (n_rows,)
```

For each ambient row `x` it forms the **nearest manifold point** under the ambient
(Euclidean / Frobenius) metric and returns:

- `residual[i] = ‖x − project(x)‖` — absolute distance to the candidate manifold
- `relative[i] = residual[i] / (‖x‖ + eps)` — scale-free off-manifold fraction

No base point is required (nearest-point projection is base-independent), so unlike the
`log`-map path there is no Fréchet-mean dependency.

The projection is exact, closed-form, per geometry — added as a private
`ResponseManifold::project_point`:

| Geometry | Projection | `residual` is |
|---|---|---|
| Poincaré ball | radial clamp into the ball (`project_into_ball`) | distance to the ball |
| SPD cone | symmetrise + clamp eigenvalues `≥ 0` (Higham nearest-PSD) | norm of the negative-eigenvalue part |
| `Gr(k,n)` / `St(k,n)` | polar factor `U = Y(YᵀY)^{-1/2}` (new `polar_factor`) | distance to the nearest orthonormal frame; `k=1` ⇒ sphere normalisation |

Only the non-fittable `ConstantCurvature` variant (never produced by the
response-geometry resolver) lacks a projector and returns a descriptive error.

## New primitive: `polar_factor` (`src/geometry/manifold.rs`)
The matrix manifolds had no ambient nearest-point projector — only the QR retraction
`qf(Y)`, which is *not* the nearest frame (first-order agreement only). Added
`polar_factor`, the Frobenius-nearest orthonormal frame `U = Y(YᵀY)^{-1/2}`, computed
from the symmetric spectral map of the `k×k` Gram matrix (reusing
`spectral_map_symmetric`). A rank-deficient `Y` (singular Gram) has no unique nearest
frame, so it surfaces a `Singular` error rather than returning an arbitrary one. This is
the genuine orthogonal projection onto `St(k,n)` / the Grassmann representative set, and
it generalises the `k=1` sphere normalisation.

## Why not `exp∘log` (the superseded approach)
The earlier draft computed `recon = exp_base(log_base(x))` and read `‖x − recon‖` as
distance-to-manifold. It can't work: `log_map` for these geometries is gatekept to
*genuine manifold points on both arguments* — `sphere.rs` requires unit-norm base **and**
target, `spd.rs` requires a symmetric positive-definite point. So `log_base(x)` rejects
exactly the off-manifold `x` the function exists to measure; the on-manifold case it does
accept is the trivial `residual ≈ 0`. `log` is a manifold→tangent chart, not a projection
from ambient space. The fix is to project properly, which is what this PR does.

## What this is *for* (and what it is not)
This is **not** a post-fit on/off-manifold membership classifier. That signal belongs to
a fitted geometric smooth: the residual of an observation against the *surface gam infers*
plus its posterior predictive density. This function answers the earlier, cheaper question
that gates *which* topology is worth fitting:

> given a candidate response geometry (e.g. one nominated by an external embedding /
> discovery stage), how far do the raw observations sit from that geometry's manifold?

Aggregated per candidate, it is one more score for geometric-smooth model selection,
alongside REML/LAML and held-out predictive density. The intended pipeline:

1. **Propose** candidate topology — an embedding (e.g. ParamRepulsor) gives a cheap,
   discovery-oriented hypothesis about dimensionality / rough topology.
2. **Commit** to a parametric form whose topology matches the hypothesis — a gam
   geometric smooth (sphere? SPD? Poincaré? plain surface?).
3. **Fit and adjudicate** — gam's penalized fit gives the interpretable surface and, via
   residuals / posterior predictive bands, the principled membership signal.
4. **Model-select across shape guesses** — compare candidates by REML/LAML or held-out
   predictive density.

`response_projection_residual` is a fast plausibility gate at step 1→2/4: before
committing to fit a proposed topology, check that the data even lies near it.
**Caveat that matters:** an embedding can confidently propose a wrong topology (e.g.
splitting connected structure into clusters). So the step-4 comparison must always include
the null / simpler-topology candidates — this residual nominates, it does not decide.

## Tests (`src/geometry/response_geometry.rs`)
- `projection_residual_is_zero_for_on_manifold_points` — SPD / Grassmann / Poincaré
  on-manifold clouds give `residual ≈ 0` (no Fréchet mean needed).
- `projection_residual_recovers_known_off_manifold_displacement` — closed-form checks:
  `Gr(1,3)` `[2,0,0]→1`, `[0,3,0]→2`; SPD `[[1,0],[0,−1]]→1`; Poincaré `[3,0]→≈2`.
- `projection_residual_supports_k_greater_than_one_frames` — `St(2,3)` / `Gr(2,4)`:
  orthonormal frame ⇒ 0; a scaled column ⇒ exact `‖Y−U‖_F = 1`, `relative = 1/√5`.
- `polar_factor_is_rejected_for_rank_deficient_frames` — duplicate columns ⇒ `Singular`.
- `projection_residual_validates_shapes_and_finiteness` — column-count / finiteness.
- `projection_residual_separates_on_and_off_manifold` — the motivating contrast.
- `projection_residual_errors_only_for_nonfittable_constant_curvature`.

## Open questions / your read
1. **Grassmann distance semantics.** `polar_factor` measures distance to the
   orthonormal-frame *representative set* (`St(k,n)`), which is the consistent choice
   given the module stores frames. If you'd rather the diagnostic report a subspace
   (principal-angle) distance for Grassmann specifically, that's a different projector —
   easy to add behind the same entry point.
2. **Where this should surface.** It's a sibling of `response_log_map` for now. The
   natural next step is an FFI binding + an optional aggregate score in `summary()` /
   model selection so Python can compare candidate geometries — held off until the
   framing is blessed.
3. **SPD floor.** Eigenvalues are clamped to `≥ 0` (distance to the closed PSD cone). If
   you want strict-PD targets, a small positive floor is a one-line change, at the cost of
   the residual no longer being the exact cone distance.

### DIFF
--- a/src/geometry/manifold.rs
+++ b/src/geometry/manifold.rs
@@ -823,6 +823,45 @@ pub(crate) fn spectral_map_symmetric(
     Ok(fast_abt(&fast_ab(&evecs, &diag), &evecs))
 }
 
+/// Frobenius-nearest matrix with orthonormal columns: the orthogonal factor `U`
+/// of the polar decomposition `Y = U P`, with `UᵀU = Iₖ` and `P` symmetric
+/// positive-definite. For a tall `Y` (`n × k`, `n ≥ k`) this is
+/// `U = Y (YᵀY)^{-1/2}`, the orthogonal projection of `Y` onto the Stiefel
+/// manifold `St(k, n)` — and hence onto the orthonormal-frame representative
+/// set of the Grassmannian `Gr(k, n)` — under the Frobenius norm. So
+/// `‖Y − U‖_F` is the *exact* distance from `Y` to that manifold. This is the
+/// genuine nearest point, unlike the QR retraction `qf(Y)`, which only agrees
+/// to first order.
+///
+/// `(YᵀY)^{-1/2}` is formed from the symmetric spectral map of the `k × k` Gram
+/// matrix. A non-positive Gram eigenvalue means `Y` has rank `< k`, where the
+/// nearest orthonormal frame is not unique; that is surfaced as a
+/// [`GeometryError::Singular`] rather than returning an arbitrary frame.
+pub(crate) fn polar_factor(y: &Array2<f64>) -> GeometryResult<Array2<f64>> {
+    let (n, k) = y.dim();
+    if n < k {
+        return Err(GeometryError::InvalidPoint(
+            "polar factor requires a tall matrix (n >= k)",
+        ));
+    }
+    if !y.iter().all(|v| v.is_finite()) {
+        return Err(GeometryError::InvalidPoint(
+            "polar factor requires finite entries",
+        ));
+    }
+    let gram = y.t().dot(y);
+    let inv_sqrt = spectral_map_symmetric(&gram, |lam| {
+        if !lam.is_finite() || lam <= GEOMETRY_EPS {
+            Err(GeometryError::Singular(
+                "polar factor is undefined for a rank-deficient matrix",
+            ))
+        } else {
+            Ok(1.0 / lam.sqrt())
+        }
+    })?;
+    Ok(y.dot(&inv_sqrt))
+}
+
 /// Dense matrix exponential `exp(A)` via scaling-and-squaring with a truncated
 /// Taylor series. The Frobenius norm of `A` is driven below 1/4 by repeated
 /// halving (`A → A / 2^s`), where Taylor converges rapidly and stably; the
--- a/src/geometry/response_geometry.rs
+++ b/src/geometry/response_geometry.rs
@@ -24,8 +24,11 @@
 use ndarray::{Array1, Array2, ArrayView1, ArrayView2};
 
 use crate::geometry::constant_curvature::ConstantCurvature;
-use crate::geometry::manifold::RiemannianManifold;
-use crate::geometry::{GeometryResult, GrassmannManifold, SpdManifold, StiefelManifold};
+use crate::geometry::manifold::{
+    flatten, from_flat, polar_factor, spectral_map_symmetric, sym, RiemannianManifold,
+    GEOMETRY_EPS,
+};
+use crate::geometry::{GeometryError, GeometryResult, GrassmannManifold, SpdManifold, StiefelManifold};
 
 /// Split a parenthesised `key=value, key=value` parameter list into trimmed,
 /// lower-cased `(key, value)` pairs. An empty list is valid (`spd()`).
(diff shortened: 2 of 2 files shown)


SauersML:
## Review — merged ✅ (squash `2a914570`)

Claude Opus 4.8 says:

Reviewed against the actual source (in a throwaway worktree off the PR head, since my working tree was on `main`). It holds up; merging.

### Correctness — the math is right per geometry
- **Poincaré:** radial clamp to `(1 − BOUNDARY_EPS)/√(−c)` via the existing `project_into_ball`. Interior points return unchanged ⇒ residual 0; exterior ⇒ distance to the boundary. (Note the "manifold" here is the full-dimensional ball, so this flags only points *outside* the ball — correct, since the ball is the model space.)
- **SPD:** symmetrise + clamp eigenvalues `≥ 0` is the genuine Higham nearest-PSD projector. Because the residual is `‖x − project(x)‖` on the *original* flattened input, it correctly includes any skew part, so it's the true distance to the symmetric-PSD cone (not just the negative-eigenvalue part). `[[1,0],[0,−1]] → 1` checks out.
- **Frames:** `polar_factor` `U = Y(YᵀY)^{-1/2}` is the exact Frobenius-nearest orthonormal frame — the real orthogonal projection onto `St(k,n)`, unlike the first-order QR retraction `qf` that was the only thing previously available. The closed-form test values all verify (`Gr(1,3) [2,0,0]→1`, `[0,3,0]→2`; `St(2,3)` scaled column → `1`, `rel = 1/√5`).

### Conventions match the rest of the module
Frames are reshaped with `from_flat(value, n, k)` (row-major `n×k`) — identical to how `grassmann.rs`/`stiefel.rs` already read their flat points — and `ambient_dim` agrees (`n*k`, `n*n`). No layout mismatch.

### The core premise is sound
The "why not `exp∘log`" reasoning is correct: `log_map` is gatekept to genuine manifold points on *both* arguments (unit-norm / SPD), so it structurally rejects exactly the off-manifold input the diagnostic must measure. A nearest-point projector is the right primitive, and it drops the Fréchet-mean dependency since projection is base-independent.

### Design
`polar_factor` is hoisted to `manifold.rs` as a shared `pub(crate)` primitive rather than inlined — the right call, since the matrix manifolds genuinely lacked a nearest-frame projector. Rank-deficiency surfaces as `Singular` rather than fabricating an arbitrary frame. The framing ("plausibility gate at step 1→2/4, *not* a post-fit membership classifier") is honest, and the caveat that step-4 selection must always include null/simpler-topology candidates is the right guardrail against a confidently-wrong embedding proposal.

### Minor, non-blocking (didn't gate the merge)
1. **Not re-exported in `src/geometry/mod.rs`.** Its siblings (`response_log_map`, `response_exp_map`, …) are re-exported there; this one is `pub` but only reachable via the full module path. Worth adding when the deferred FFI binding lands.
2. **Shared linear-algebra kernel with `grassmann::compact_svd_from_tangent`.** Both form the `k×k` Gram, eigendecompose via `jacobi_symmetric`, and rebuild from eigenvalue roots — same kernel, different contracts (errors-on-singular vs tolerates-zero; returns the factor vs `U/Σ/V`). Not a literal clone, but a reasonable future "share a thin-SVD/polar helper?" cleanup.
3. **k=1 sphere now lives in two paths** — the dedicated `as_sphere()`/`normalize_sphere_matrix` wrappers and implicitly inside `polar_factor`. For *projection* this is harmless (the cut-locus subtlety that justifies `as_sphere()` for `log_map` doesn't apply to radial normalisation); it's *fewer* code paths, just a mild stylistic divergence from the house pattern.
4. Documented choices the PR already flags honestly: Grassmann reports distance to the Stiefel-frame representative set (not principal-angle subspace distance), and the SPD cone is closed (PSD, not strict-PD). Both reasonable for a plausibility gate.

### Why merge
The change is small, self-contained, and additive — one new public diagnostic plus one new `pub(crate)` primitive, no behavioural change to any existing path. The math is correct, the layout conventions match the surrounding code, the symbols it imports all exist with matching signatures, and the test suite pins the behaviour with closed-form values and the right edge cases (rank-deficient rejection, shape/finiteness, ConstantCurvature error path). The open questions the author raised are genuinely open *design* questions (Grassmann distance semantics, where it surfaces, SPD floor), not correctness gaps — and none of them block landing the primitive.

**Disclosure:** I did *not* run the tests myself (build was intentionally stopped). This verdict rests on static review of the diff against the current source plus the author's stated local green run (121 geometry tests), not on my own execution.

PR #1361, opened 2026-06-20
Add FittedTransport::invert: the monotone-transport inverse

nathanneuro:
# Add `FittedTransport::invert`: the monotone-transport inverse

## What
Adds the exact inverse of `FittedTransport::eval` in `src/inference/layer_transport.rs`. For each target-chart coordinate `y` it returns the source-chart coordinate `t` with `eval([t]) == y`, by monotone bisection on the pre-wrap map (new private `raw_at` helper).

It handles both topology-preserving cases:
- **Interval → Interval** homeomorphism — rejects a `y` outside the fitted image.
- **Circle → Circle** degree-±1 cover — shifts `wrap_tau(y)` by the unique `2π` multiple that lands in the image (the pre-wrap map covers exactly one turn).

It requires `topology_preserved` (fold-free, monotone) and errors rather than picking an arbitrary branch through a fold.

## Why this, why now
`invert` is the missing half of the transport algebra. Composition already exists (`composition_defect` tests `h_ac ≟ h_bc ∘ h_ab`, and you compose by chaining `eval`), but there was no way to *invert* a fitted transport.

The concrete driver is an out-of-tree extension I'm building on top of `gamfit` — **[nathanneuro/manifold-transfer](https://github.com/nathanneuro/manifold-transfer)**, a cross-model manifold transfer / auditing layer. Its central object is a metric warp between two models, `φ = g_B ∘ g_A⁻¹` — each `g` a fitted monotone transport from a shared predictor to a model's local spacing. Composing those requires inverting a fitted transport, which `eval` + `composition_defect` alone can't do. The inverse needs the internal basis/gauge (the pre-wrap map, rotation offset, winding degree), so it belongs in the core next to `eval`/`derivative`/`composition_defect` rather than being re-derived downstream. The application-specific logic stays out-of-tree; only this application-agnostic transport math comes here.

## What this unblocks next
This is the first of a few small, application-agnostic primitives that the extension needs from the core. The immediate follow-up is #1363 (exposing the `FittedTransport` object — including this `invert` — to Python). Further possibilities, only if/when the extension actually needs them: a spacing-density → knot-vector helper feeding the existing `Provided`/`UserProvided` knot paths, and possibly a mutual-kNN / intrinsic-dimension estimator for topology proposal. None are included here — this PR is just `invert`.

## Tests
Inline `invert_tests`: increasing and decreasing interval round-trips (`eval∘invert` and `invert∘eval` identity), circle round-trip (mod 2π), and out-of-image rejection. The full `inference::layer_transport` module is green; no new warnings or clippy findings.

### DIFF
--- a/src/inference/layer_transport.rs
+++ b/src/inference/layer_transport.rs
@@ -310,6 +310,54 @@ impl DomainBasis {
         }
     }
 
+    /// Polynomial degree of `h′` on each knot span: the basis degree minus one
+    /// (a cubic spline derivative is piecewise quadratic).
+    fn derivative_poly_degree(&self) -> usize {
+        let degree = match self {
+            DomainBasis::Periodic(spec) => spec.degree,
+            DomainBasis::Open { degree, .. } => *degree,
+        };
+        degree.saturating_sub(1)
+    }
+
+    /// Sorted distinct breakpoints bounding the polynomial pieces of `h′` over
+    /// the active domain `[lo, hi]`. Within each `[breakpoints[k],
+    /// breakpoints[k+1]]` span the derivative is a single polynomial of degree
+    /// [`Self::derivative_poly_degree`], which is what the exact monotonicity
+    /// certificate reconstructs and checks. For the open basis these are the
+    /// distinct interior+boundary knots; for the periodic basis they are the
+    /// uniform cardinal-B-spline segment boundaries over `[0, 2π]`.
+    fn derivative_breakpoints(&self) -> Vec<f64> {
+        match self {
+            DomainBasis::Periodic(spec) => {
+                // Cardinal periodic B-splines on `[origin, origin+period]` have
+                // `num_basis` uniform segments; the derivative is a separate
+                // polynomial on each.
+                let n_seg = spec.num_basis.max(1);
+                (0..=n_seg)
+                    .map(|k| spec.origin + spec.period * k as f64 / n_seg as f64)
+                    .collect()
+            }
+            DomainBasis::Open { knots, degree } => {
+                let lo = knots[*degree];
+                let hi = knots[knots.len() - 1 - degree];
+                let mut breaks: Vec<f64> = Vec::with_capacity(knots.len());
+                for &k in knots.iter() {
+                    if k > lo + 0.0 && k < hi {
+                        breaks.push(k);
+                    }
+                }
+                breaks.sort_by(|a, b| a.partial_cmp(b).unwrap_or(std::cmp::Ordering::Equal));
+                breaks.dedup_by(|a, b| (*a - *b).abs() <= f64::EPSILON * hi.abs().max(1.0));
+                let mut out = Vec::with_capacity(breaks.len() + 2);
+                out.push(lo);
+                out.extend(breaks.into_iter().filter(|&k| k > lo && k < hi));
+                out.push(hi);
+                out
+            }
+        }
+    }
+
     fn derivative_rows(&self, t: ArrayView1<'_, f64>) -> Result<Array2<f64>, String> {
         let projected = t.mapv(|v| self.project(v));
         match self {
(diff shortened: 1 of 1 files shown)


SauersML (changes requested):
**Request changes.** The inverse algebra is correct (`raw_at` reproduces `eval`'s pre-wrap value exactly, and the circle branch-shift `m = ceil((raw_min − wrap(y))/τ)` provably lands the target in `[raw_min, raw_min+τ)`), but `invert` rests on guarantees the code does not actually provide. Three issues, P1 blocking.

---

### [P1] `topology_preserved` is not a sufficient invertibility certificate

`invert` gates on `if !self.topology_preserved` and then assumes strict monotonicity over the whole domain. But that flag is a **sampled** diagnostic: `topology_preserved = (min_directional_derivative > 0.0)` where `min_directional_derivative` is the min of `h′` over a fixed **512-point** grid (`FOLD_CHECK_GRID = 512`, layer_transport.rs:93; computed at ~:853–862). A narrow fold between adjacent grid samples passes the check, and the new bisection then silently selects one of several roots — the exact behavior the docstring says it prevents.

Reproduced via a numerical port of the actual basis/penalty/REML/fold-check path with
`h(t) = (t−0.5)³/3 − (0.4/511)² · t` over 64 obs on `[0,1]` (the `1/511` deliberately hides the fold between cert-grid samples):

```
min h′ on 512-pt certification grid:  +3.45e-7
min h′ on a dense grid:               −6.13e-7
roots of invert at y = h(0.5):        0.4986441, 0.5000000, 0.5013559   (bisection returns the left one)
```

Fix: make monotonicity a *genuine* certificate, not a sample. For the cubic spline in use, `h′` is piecewise quadratic — check each span's endpoints plus any interior quadratic extremum (closed-form), or use an interval-arithmetic / adaptive sign certificate. The 512-point diagnostic must not be the safety condition for inversion.

### [P2] Non-finite targets silently return a boundary coordinate

No `is_finite` check on `y`. For `NaN`: both interval range comparisons are false, `NaN.clamp(..)` stays `NaN`, every bisection comparison is false → `b` walks toward `lo` → `invert([NaN])` returns ≈`lo` as `Ok`. `±∞` degrade similarly through the wrap. This violates the documented `eval([t]) == y` contract, and because `#1363` only maps `Err`→`ValueError`, Python callers get a plausible-but-meaningless coordinate instead of an exception. Add up front:

```rust
if y.iter().any(|v| !v.is_finite()) {
    return Err("transport inverse targets must be finite".to_string());
}
```

### [P2] The absolute `1e-9` image tolerance is scale-unaware

`const TOL: f64 = 1.0e-9` is absolute. For a fitted image `[0, 1e-8]`, the target `1.05e-8` is 5% outside the image yet accepted (within the 1e-9 margin), then clamped to `1e-8`, so `eval(invert(1.05e-8)) = 1e-8 ≠ 1.05e-8` — contradicting both the inverse contract and the stated out-of-image rejection. Use a scale-aware tolerance:

```rust
let scale = raw_min.abs().max(raw_max.abs()).max(1.0);
let tol = 32.0 * f64::EPSILON * scale;
```

(An ULP-based endpoint comparison would be even better for very small charts, since `max(1.0)` is generous relative to the image width.)

---

### Non-blocking
- **Docstring**: "converging to ~`(hi−lo)/2¹⁰⁰`" overstates it — f64 bisection stagnates after ~53 significand bits; the extra iterations are no-ops.
- **Perf**: `raw_at` allocates and rebuilds basis rows on every one of the (fixed) 100 iterations per target; `TOL` is defined but bisection never early-exits on it. Hoist the buffer / terminate on width. Not a correctness blocker (and `#1363` holds the GIL across this, so it matters more there).
- **Test fixture**: the `t^1.5` example has `h′(0)=0`, so `topology_preserved` is borderline by construction; a warp with derivative bounded away from zero is a safer happy-path fixture.

### Tests to add before merge (also gates `#1363`)
narrow between-grid fold reproducer · `NaN`/`+∞`/`−∞` targets · interval image ≪ `1e-9` · degree −1 circle · circle-seam & interval-endpoint inversion · an explicitly non-topology-preserving fit through the Python API.

The FFI in `#1363` itself is clean (registration, factory, NumPy marshalling, `Err`→`ValueError` all correct) — it just inherits these inverse defects, so fixing them here clears both.

PR #1363, opened 2026-06-20
Expose FittedTransport to Python (eval/derivative/invert + fit_transport)

nathanneuro:
# Expose `FittedTransport` to Python: `eval`/`derivative`/`invert` + `fit_transport`

> **Stacked on #1361** (`FittedTransport::invert`). Until that merges, the `invert` commit shows in this diff; once #1361 lands I'll rebase onto `main` so this PR is just the FFI exposure. Review the second commit (`feat: expose FittedTransport to Python`) for the new work here.

## What
Adds a `PyFittedTransport` pyclass and a `fit_transport()` pyfunction in `gam-pyffi`, so the transport *map itself* — not just the summary dict `layer_transport_fit` returns — is reachable from Python. The object exposes:

- `eval(t)`, `derivative(t)`, `eval_with_variance(t)`, `invert(y)` (each over a 1-D coordinate array)
- `report(layer_from=0, layer_to=1)` — the same dict `layer_transport_fit` returns
- summary properties: `topology_from`, `topology_to`, `topology_preserved`, `degree`, `isometry_defect`

`gamfit.fit_transport(coords_from, coords_to, topology_from, topology_to)` wraps it and is exported at top level.

## Why
`layer_transport_fit` currently returns only an evidence dict, so the fitted map's `eval`/`derivative`/`invert` (and thus the transport algebra) are unreachable from Python. The concrete driver is the same out-of-tree extension as #1361 — **[nathanneuro/manifold-transfer](https://github.com/nathanneuro/manifold-transfer)**, a cross-model manifold transfer / auditing layer built on `gamfit`. Its central object is `φ = g_B ∘ g_A⁻¹`, two fitted transports composed through an inverse, which requires a live, invertible transport object in Python. (Concretely, the extension's `transport_law` composes `g_B ∘ g_A⁻¹` and its `audit` module uses the transports' `isometry_defect` for distillation integrity/seam maps.) The binding is thin marshalling over the existing Rust methods — no new math — so it belongs in the FFI alongside `layer_transport_fit`. The application logic stays out-of-tree; only this app-agnostic exposure comes here.

## Tests
`tests/test_inference_instruments_wired.py::test_fit_transport_object_inverts_and_composes`: `eval`/`invert` round-trip, and a `g_B ∘ g_A⁻¹` composition that recovers the closed-form answer (`g_A(t)=t^1.5`, `g_B(t)=t^0.5` ⇒ `φ(y)=y^{1/3}`). Built with `maturin develop`; Rust + clippy clean.

### DIFF
--- a/crates/gam-pyffi/src/geometry_ffi.rs
+++ b/crates/gam-pyffi/src/geometry_ffi.rs
@@ -4168,6 +4168,8 @@ fn rust_extension(module: &Bound<'_, PyModule>) -> PyResult<()> {
     module.add_function(wrap_pyfunction!(sae_streaming_plan, module)?)?;
     module.add_function(wrap_pyfunction!(layer_transport_fit, module)?)?;
     module.add_function(wrap_pyfunction!(layer_transport_ladder, module)?)?;
+    module.add_class::<PyFittedTransport>()?;
+    module.add_function(wrap_pyfunction!(fit_transport, module)?)?;
     module.add_function(wrap_pyfunction!(sae_checkpoint_dynamics, module)?)?;
     inference_instruments::register(module)?;
     module.add_function(wrap_pyfunction!(sae_manifold_assignment_summary, module)?)?;
--- a/crates/gam-pyffi/src/latent_basis_and_sae_ffi.rs
+++ b/crates/gam-pyffi/src/latent_basis_and_sae_ffi.rs
@@ -3977,6 +3977,124 @@ fn layer_transport_fit(
     Ok(layer_transport_report_to_pydict(py, &report)?.unbind())
 }
 
+/// A fitted inter-chart transport map `h: t_from ↦ t_to`, exposed as a live,
+/// **invertible** object. Where [`layer_transport_fit`] returns only the
+/// evidence dict, this carries the map itself: callers can `eval`,
+/// `derivative`, `eval_with_variance`, and `invert` it — the transport algebra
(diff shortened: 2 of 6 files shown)


SauersML (changes requested):
**Request changes — gated on #1361.** The Python exposure itself is clean: registration (`add_class`/`wrap_pyfunction!`), the `fit_transport` factory (data-range-derived interval bounds via `parse_chart_topology`), the top-level export, NumPy marshalling, and `Err`→`ValueError` propagation are all correct, and the deps resolve (`ChartTopology::name`, `layer_transport_report_to_pydict`, `_as_coord_array`).

Two things before merge:

1. **It carries #1361's `layer_transport.rs` diff verbatim** (the `invert`/`raw_at`/`invert_tests` block) — this is stacked on #1361. Land #1361 first, then rebase this so it drops the duplicate; otherwise the identical lines conflict. And it inherits all three `invert` defects flagged on #1361 (sampled-not-certified monotonicity = P1 blocking, NaN/∞ targets, scale-unaware tolerance).
2. **`test_fit_transport_object_inverts_and_composes` is happy-path only** (smooth increasing warp + one composition). Per the #1361 review, add through the *Python* API: a between-grid fold reproducer, `NaN`/`±∞` targets, an interval image ≪ 1e-9, a degree −1 circle, circle-seam/interval-endpoint inversion, and an explicitly non-topology-preserving fit. The `y^(1/3)` composition test is good — keep it.

PR #1432, opened 2026-06-20
fix(bench): replace tautological feature_uniqueness metric (closes #1413)

### LINKED ISSUE #1413: feature_uniqueness metric is tautologically always 1
## Definitive mistake

Both synthetic benchmarks use Hungarian one-to-one matching:

* `bench/synth_sae_compare.py:135-155`
* `bench/synth_sae_bench_manifold.py:223-248`

They then calculate:

```python
len(set(cols)) / len(rows)
```

at:

* `synth_sae_compare.py:184-191`
* `synth_sae_bench_manifold.py:331-335`

Hungarian matching returns unique column indices by definition. The greedy fallback also explicitly forbids reusing columns. Therefore, for any nonempty match:

\[
\text{feature_uniqueness}=1
\]

regardless of how duplicated the learned dictionary actually is.

## Additional bias

Moreover, MCC and probing F1 are averaged only over matched features. Unmatched extra learned features—duplicates, dead features, or arbitrary directions introduced by an oversized dictionary—are omitted. This systematically flatters high-width models.

## Suggested fix

Penalize unmatched learned and truth features, and replace the tautological uniqueness metric.

HomunculusLabs:
## Summary

`feature_uniqueness` was **always 1.0 by construction** in both benchmark harnesses. They compute:

```python
uniqueness = len(set(cols)) / max(len(rows), 1)
```

over the output of `linear_sum_assignment` (Hungarian) — or a greedy fallback that explicitly forbids column reuse. **Both guarantee unique column indices**, so for any non-empty match the numerator equals the denominator and the metric is 1.0 regardless of how duplicated, dead, or oversized the learned dictionary is.

## The fix

Replace it with a one-to-one **recovery match fraction**:

```python
uniqueness = n_matched / max(n_learned_live, n_truth)
```

- equals **1.0 only** for perfect tight recovery (every live learned feature matches a distinct truth feature, and every truth feature is recovered);
- drops below 1 when **duplicate / excess / dead learned features** exist (they shrink the one-to-one match count) or when **truth features go unrecovered**;
- directly countering the high-width flattery bias noted in the issue (unmatched features previously omitted).

MCC and probing F1 are left as standard per-matched-feature quality metrics; `feature_uniqueness` now surfaces the recovery/width penalty that was previously invisible. The dataclass field name is retained for CSV/JSON schema stability; docstrings and labels are updated to "feature recovery (one-to-one match fraction)".

## Verification

Standalone simulation of the matching logic across four scenarios (old vs new metric):

```
scenario                 n_learned n_truth matched    OLD    NEW
A perfect tight                 10      10      10  1.000  1.000
B overcomplete+dups             16      10      10  1.000  0.625
C overcomplete+excess           16      10      10  1.000  0.625
D undercomplete                  6      10       6  1.000  0.600
OLD metric always 1.0 across scenarios: True
NEW metric == 1.0 only for perfect tight recovery: True
```

`python3 -m py_compile` passes on both files. (A full bench run needs the gamfit + torch stack; the metric change is exercised by the simulation above.)

## Files

- `bench/synth_sae_compare.py` — `_score` calc + docstring
- `bench/synth_sae_bench_manifold.py` — `run_one` calc + `_benchmark_notes` label

### DIFF
--- a/bench/_synth_sae_metrics.py
+++ b/bench/_synth_sae_metrics.py
@@ -1,0 +1,138 @@
+"""Direction-recovery metrics shared by the synthetic SAE benchmarks.
+
+NumPy-only (no ``gamfit``/torch import) so it can be unit-tested in isolation
+by ``bench/test_synth_sae_metrics.py``. The definitions follow SynthSAEBench
+(arXiv:2602.14687), which separates three distinct ground-truth measurements
+that the benchmark harnesses had previously conflated:
+
+* **feature uniqueness** -- a *collision* metric. For each learned latent ``j``
+  its best ground-truth match is ``i*(j) = argmax_i |w_j . d_i|`` and
+  uniqueness is the fraction of *distinct* best targets,
+  ``#unique(i*) / n_learned``. It is 1.0 iff every latent claims a different
+  truth feature; duplicate / collapsed directions that all point at the same
+  truth feature drag it toward ``1 / n_learned``. This is independent
+  per-latent ``argmax`` matching -- NOT one-to-one assignment.
+
+* **MCC** -- a *quality* metric: the optimal one-to-one |cosine| matching
+  (Hungarian), reporting the mean |cosine| of the matched pairs.
+
+* **direction recovery (precision / recall / F1)** -- a *coverage* metric that
+  is aware of match quality. See :func:`recovery_scores`.
+
+Why the split matters (#1413): a raw one-to-one *assignment count* is always
+``min(n_learned, n_truth)`` because both the Hungarian algorithm and the greedy
+fallback fill every row/column they can regardless of similarity. So
+``n_matched / max(...)`` collapses to a pure width ratio ``min/max`` and never
+inspects whether the matched pairs are any good -- equal-width perfect,
+all-duplicate, and random dictionaries all score 1.0. Uniqueness (collisions)
+and recovery F1 (quality-weighted coverage) each look at the actual directions.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+import numpy as np
+
+
+def match_directions(
+    learned: np.ndarray, truth: np.ndarray
+) -> tuple[np.ndarray, np.ndarray, str]:
+    """Optimal one-to-one ``|cosine|`` matching, Hungarian with greedy fallback.
+
+    ``learned`` (``L x D``) and ``truth`` (``T x D``) rows are assumed already
+    unit-normalized (the callers normalize before matching). Returns
+    ``(rows, cols, method)`` with ``len(rows) == min(L, T)``; ``method`` is one
+    of ``"hungarian"``, ``"greedy"``, or ``"none"`` (empty input).
+    """
+    learned = np.asarray(learned, dtype=float)
+    truth = np.asarray(truth, dtype=float)
+    if learned.size == 0 or truth.size == 0:
+        return np.empty(0, dtype=int), np.empty(0, dtype=int), "none"
+    sim = np.abs(learned @ truth.T)
+    try:
+        from scipy.optimize import linear_sum_assignment  # type: ignore
+
+        rows, cols = linear_sum_assignment(-sim)
+        return np.asarray(rows, dtype=int), np.asarray(cols, dtype=int), "hungarian"
+    except Exception:
+        pairs: list[tuple[int, int]] = []
+        used_rows: set[int] = set()
+        used_cols: set[int] = set()
+        for flat in np.argsort(sim, axis=None)[::-1]:
+            row, col = np.unravel_index(int(flat), sim.shape)
+            if int(row) in used_rows or int(col) in used_cols:
+                continue
+            pairs.append((int(row), int(col)))
+            used_rows.add(int(row))
+            used_cols.add(int(col))
+            if len(pairs) == min(sim.shape):
+                break
+        rows = np.array([p[0] for p in pairs], dtype=int)
+        cols = np.array([p[1] for p in pairs], dtype=int)
+        return rows, cols, "greedy"
+
+
+def feature_uniqueness(learned: np.ndarray, truth: np.ndarray) -> float:
+    """SynthSAEBench feature uniqueness: fraction of distinct ``argmax`` targets.
+
+    ``i*(j) = argmax_i |w_j . d_i|`` for each learned latent ``j``; the score is
+    ``#distinct(i*) / n_learned``. 1.0 means every latent's best match is a
+    different truth feature; duplicate / collapsed directions that share a best
+    match pull it toward ``1 / n_learned``. Returns 0.0 for empty input.
+    """
+    learned = np.asarray(learned, dtype=float)
+    truth = np.asarray(truth, dtype=float)
+    if learned.shape[0] == 0 or truth.shape[0] == 0:
+        return 0.0
+    sim = np.abs(learned @ truth.T)
+    best = np.argmax(sim, axis=1)
+    return float(np.unique(best).size / learned.shape[0])
+
+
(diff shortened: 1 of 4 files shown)


SauersML:
**Request changes.** The diagnosis is correct and worth fixing — but the replacement is still structurally tautological, just with respect to model *width* instead of always being `1.0`. It should not merge as-is.

## Blocking flaw: the new metric never looks at match quality

Both matchers (`linear_sum_assignment` and the greedy fallback) always return exactly `min(L, T)` assignments for an `L×T` similarity matrix — there is no similarity threshold and no way to leave a bad pair unmatched. So `rows.shape[0] ≡ min(n_learned, n_truth)`, and the new metric reduces algebraically to:

```
feature_uniqueness == min(n_learned, n_truth) / max(n_learned, n_truth)
```

a pure function of the two **counts**. The cosine similarities that `_match` optimized are never consulted.

I verified this with the repo's own greedy fallback (`L = T = 4`, truth = `I4`):

| case | PR metric | matched cosines |
|---|---|---|
| perfect recovery | **1.000** | `[1, 1, 1, 1]` |
| 4 copies of one feature | **1.000** | `[1, 0, 0, 0]` |
| random directions | **1.000** | `[.95, .85, .8, .55]` |

So the PR's central claim — "equals 1.0 **only** for perfect tight recovery" — is false: any run with `n_live == n_truth` scores `1.0` regardless of recovery quality. With the shipped defaults in `synth_sae_compare.py` (`--width 8 --features 24`), every model scores a fixed `8/24 ≈ 0.333` independent of training quality. This is a cardinality-balance score, not a recovery score.

## Secondary issues

1. **It doesn't measure duplicates.** Hungarian still assigns each duplicate to a *different* truth feature (the surplus just lands on near-zero similarities), so at fixed width the metric can't tell perfect vs. duplicate vs. random apart. The issue (#1413) is specifically about duplication; this doesn't surface it.

2. **The "dead features pull the score below 1" claim is incorrect for this code.** Both harnesses drop zero-norm directions *before* counting: `dirs = decoder_dirs[live]` (`synth_sae_compare.py:178`) and the `norm <= 1e-10` skip in `_learned_components` (`synth_sae_bench_manifold.py:214`). So `n_learned` excludes dead columns — 10 perfect + 6 dead still scores `1.0`.

3. **Schema "stability" is only syntactic.** Keeping the key `feature_uniqueness` while changing its meaning makes old and new CSV/JSON rows silently incomparable. Prefer a new, honestly-named field (+ a `metric_version`) over redefining the existing one.

4. **The verification can't catch this.** The standalone sim hard-codes `matched = min(n_learned, n_truth)`, so it only re-derives the width-ratio arithmetic. The decisive cases need *actual direction matrices*: equal-width perfect / all-duplicate / random, overcomplete-perfect-plus-junk, undercomplete subset, and dead/never-firing latents.

## Suggested design (two separate, well-defined metrics)

Keep diagnostics separate instead of overloading one field:

**Collision-based uniqueness** (what #1413 actually asks for) — independent per-latent best match, count distinct targets:
```python
best_truth = np.argmax(sim, axis=1)
feature_uniqueness = np.unique(best_truth).size / n_learned   # 0.25 for the 4-duplicates case
```

**Quality-aware recovery** (threshold-free) — optimal matched cosine *mass* turned into precision/recall/F1:
```python
rows, cols = linear_sum_assignment(-sim)
mass = float(sim[rows, cols].sum())
recovery_precision = mass / n_learned   # junk/excess learned features pull this down
recovery_recall    = mass / n_truth     # unrecovered truth features pull this down
recovery_f1        = 2 * mass / (n_learned + n_truth)
mcc                = float(sim[rows, cols].mean())   # existing matched-pair quality, unchanged
```

This has the properties the PR wants and the current metric lacks: perfect equal-size → `1.0`; equal-size duplicates / random → well below `1.0` (`0.25` / `~0.79` in my run above); extra junk lowers precision; missing truth lowers recall — with no arbitrary cosine cutoff. (If a discrete "# recovered" is ever needed, define a quality threshold first and run a max-cardinality matching on edges above it; without a threshold, assignment count ≠ recovery count.)

**Structurally:** the matching + metric logic is duplicated across both harnesses — extract it into a numpy-only `bench/_synth_sae_metrics.py` so it can be unit-tested without the `gamfit`/torch stack, and add the direction-based test cases above. Exposing `n_learned_total`, `n_learned_firing`, and `n_truth` alongside the scores would also let readers separate width effects from recovery quality.

Net: great catch on the tautology; the replacement trades it for a width-ratio. Restore a real collision-uniqueness metric and add the quality-aware coverage score, and this becomes genuinely useful.

SauersML:
Follow-up: I verified the three claims above directly against the SynthSAEBench paper (arXiv:2602.14687, submitted 2026-02-16):

- **Uniqueness** is *argmax-collision*, not one-to-one: *"for each latent j, find its best-matching ground-truth feature i*(j)=argmax_i|wⱼᵀdᵢ|. Uniqueness is the fraction of unique best matches divided by the number of latents."*
- **MCC** is the Hungarian one-to-one matched-pair mean |cosine|.
- **Dead latents** are *"latents that never fire across the evaluation set"* (activation-based, not decoder-norm).

So `n_matched / max(L, T)` here reduces to the width ratio `min(L,T)/max(L,T)` (since the matchers always fill `min(L,T)` pairs) — it's blind to match quality and isn't the benchmark's uniqueness definition.

I've taken your diagnosis and landed the faithful version in **#1433**: a shared numpy-only `bench/_synth_sae_metrics.py` implementing the real argmax-collision `feature_uniqueness` plus a quality-aware `direction_recovery_{precision,recall,f1}` (matched cosine *mass* → P/R/F1, threshold-free), MCC unchanged, new fields added rather than overloading keys, with direction-matrix unit tests. Credited you there for the original catch. Closing this in favor of #1433 — thanks for finding the tautology.

SauersML:
Acted on the deeper review — landed as **#1434** (merged to `main`). The earlier #1432 work fixed the tautology and gave quality-aware recovery; #1434 closes the remaining correctness gaps from the review:

- **Exact matching everywhere.** The greedy fallback was confirmed non-optimal (`[[0.9,0.8],[0.8,0]]` → mass 0.9 vs optimal 1.6). Replaced with a dependency-free Hungarian, **validated against brute force** (40 random shapes) so the metric never silently degrades without SciPy.
- **Jaccard** `A/(L+T−A)` added next to P/R/F1; both F1 and Jaccard hit 1.0 only for a perfect full bijection (the `2A=L+T ⇒ L=T, q≡1` argument).
- **Dead-slot penalty.** `synth_sae_compare` counts *all* decoder slots (incl. zero-norm) in the denominator via `n_learned_total`; dead rows carry no mass so `A` is unchanged but wasted width now lowers precision/F1/Jaccard (width-16 recovering 10 → 10/16, not 1.0). Manifold keeps the live-count denominator (no well-defined total-slot count there).
- **`cardinality_ceiling`** exposed under an honest name — it's exactly `min/max`, i.e. the Jaccard recovery at `MCC=1`, which is the number #1432 reported as if achieved.

Two review points I deliberately did **not** adopt, with reasoning:
- **Hard thresholded counts** (`𝟙[S≥τ]` matching): the soft cosine-mass scores are threshold-free and already give exact P/R/F1/Jaccard, so I kept them as the primary metric rather than introduce an arbitrary `τ`. Easy to add later if a discrete count is wanted.
- **Firing-based dead latents / `n_firing_latents`**: geometric vs functional deadness is a real distinction, but wiring activation-based counts cleanly through both harnesses (esp. the manifold atom×harmonic structure) is a larger change; flagging as follow-up rather than rushing it.

Thanks for the thorough math — the exact-matching and Jaccard points in particular were correct and worth fixing.

PR #1438, opened 2026-06-20
fix(bench): functional dead-latent accounting + manifold total-slot denominator (closes #1435)

### LINKED ISSUE #1435: bench SAE metrics: account for functional (never-firing) dead latents, not just decoder-norm
## Gap

The synthetic SAE benchmark metrics (`bench/_synth_sae_metrics.py`, used by `bench/synth_sae_compare.py` and `bench/synth_sae_bench_manifold.py`, added in #1432 / #1434) account for **dead latents geometrically** (zero decoder norm), but SynthSAEBench (arXiv:2602.14687) defines a dead latent **functionally**: a latent that *never fires across the evaluation set*. These are not equivalent — a latent can have a nonzero decoder direction yet never activate, and it should count as wasted capacity.

Two concrete shortfalls:

1. **Functional deadness is never measured.** Nothing counts never-firing latents. A nonzero-norm-but-never-fires latent is treated as live and can even be matched, contributing to recovery scores it shouldn't.

2. **The manifold harness doesn't penalize dead capacity at all.** `synth_sae_compare` was updated in #1434 to span the recovery denominator over *all* decoder slots (`n_learned_total`), so geometrically-dead width lowers precision/F1/Jaccard. But `synth_sae_bench_manifold` keeps the live-direction denominator because a manifold SAE has no single obvious total-slot count (`_learned_components` already drops zero-norm atom directions). So a manifold SAE with wasted atom capacity is not penalized.

## Suggested fix

- Report `n_firing_latents` (latents with any activation above a small threshold on the eval set) and a functional dead count, separately from the geometric (decoder-norm) live count.
- Decide and document which denominator the recovery metrics should use (architectural slots vs firing latents) — ideally expose both `n_latent_slots` and `n_firing_latents` so geometric vs functional recovery can be read off directly.
- For the manifold harness, define a principled total-slot count (atoms × harmonics, excluding the intercept row) so dead capacity is penalized consistently with `synth_sae_compare`.

## Context

Deferred from #1434 (the activation-based wiring touches both harnesses and the manifold atom×harmonic structure, so it's a larger change than the metric-math fixes that PR landed). Helper and current behavior: `bench/_synth_sae_metrics.py`; `recovery_scores(..., n_learned_total=...)`.

HomunculusLabs:
## Summary

SynthSAEBench defines a dead latent *functionally* (never fires across the eval set), not just geometrically (zero decoder norm). Both harnesses accounted for geometric deadness only, so a latent with a nonzero decoder direction that never activated was treated as live — wasted capacity that escaped every penalty (#1435).

## Changes

- **`bench/_synth_sae_metrics.py`** — add `firing_latent_mask` / `n_firing_latents`: activation-based live count (a latent fires if `|activation| > threshold` on ≥1 eval sample). `n_slots - n_firing` is the functional dead count. Numpy-only, unit-tested.
- **`bench/synth_sae_compare.py`** — report `n_latent_slots` (architectural width) and `n_firing_latents` on `ModelResult`. Recovery denominator already spanned the full decoder width (#1434); this adds the functional diagnostic so geometric vs functional deadness read off directly.
- **`bench/synth_sae_bench_manifold.py`** — define a principled total-slot count (`_manifold_total_slots`: atoms × harmonics, excluding the intercept row) and pass it as the recovery `n_learned_total` so **geometrically-dead atom capacity is penalized**, consistent with `synth_sae_compare`. This is the core fix: previously the manifold recovery denominator was the live-direction count, so wasted atom capacity was not penalized at all. Also reports `n_latent_slots` / `n_firing_latents`.
- **`bench/test_synth_sae_metrics.py`** — cover functional-vs-geometric divergence, threshold sensitivity, and empty/1-D edge cases.

## Verification

```
$ python3 bench/test_synth_sae_metrics.py   # 14 tests, all pass
ok  test_functional_dead_latents_differ_from_geometric
ok  test_firing_threshold_respected
ok  test_firing_empty_and_1d
... (14 total)
```

The functional-vs-geometric test is the key one: 4 latents × 6 samples where latent 1 has a nonzero direction but never fires → `firing_latent_mask == [T,F,T,T]`, `n_firing=3`, functional dead = 1.

Threshold default `1e-8` matches the existing `learned_l0` `|a|>1e-8` convention already used in both harnesses.

### DIFF
--- a/bench/_synth_sae_metrics.py
+++ b/bench/_synth_sae_metrics.py
@@ -151,6 +151,35 @@ def feature_uniqueness(learned: np.ndarray, truth: np.ndarray) -> float:
     return float(np.unique(best).size / learned.shape[0])
 
 
+def firing_latent_mask(activations: np.ndarray, threshold: float = 1e-8) -> np.ndarray:
+    """Functional dead-latent mask (SynthSAEBench): a latent is *live* if it
+    fires (``|activation| > threshold``) on at least one sample of the evaluation
+    set.
+
+    Unlike geometric deadness (zero decoder norm), a latent with a nonzero decoder
+    direction that never activates is *functionally* dead -- wasted capacity that a
+    decoder-norm check misses (#1435). Returns a 1-D boolean mask whose length is
+    the latent count (``activations`` columns). Empty input returns an empty mask;
+    a 1-D array is treated as a single latent.
+    """
+    a = np.abs(np.asarray(activations, dtype=float))
+    if a.size == 0:
+        return np.empty(0, dtype=bool)
+    if a.ndim == 1:
+        a = a[:, None]
+    return np.any(a > threshold, axis=0)
+
+
+def n_firing_latents(activations: np.ndarray, threshold: float = 1e-8) -> int:
+    """Count of latents that fire on at least one evaluation sample.
+
+    Functional (activation-based) live count, complementing the geometric
+    (decoder-norm) live count. ``n_slots - n_firing_latents`` is the functional
+    dead count (#1435).
+    """
+    return int(firing_latent_mask(activations, threshold).sum())
+
+
 @dataclass(frozen=True)
 class RecoveryScores:
     """Quality-aware one-to-one recovery plus the matched-pair MCC.
--- a/bench/synth_sae_bench_manifold.py
+++ b/bench/synth_sae_bench_manifold.py
@@ -27,7 +27,7 @@ import numpy as np
 
 import gamfit
 from gamfit._binding import rust_module
-from _synth_sae_metrics import feature_uniqueness, recovery_scores
+from _synth_sae_metrics import feature_uniqueness, n_firing_latents, recovery_scores
 
 
 @dataclass(frozen=True)
@@ -70,6 +70,8 @@ class BenchmarkMetrics:
     direction_recovery_recall: float
     direction_recovery_f1: float
     direction_recovery_jaccard: float
+    n_latent_slots: int
+    n_firing_latents: int
     probing_precision: float
     probing_recall: float
     probing_f1: float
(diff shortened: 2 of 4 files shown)


HomunculusLabs:
RepoPrompt Review (Review workflow, Codex/gpt-5.5): **APPROVE**.

Independent verification against the real PR branch:
- Ran a standalone numpy script confirming the functional-dead-but-geometrically-live case: `[True, False, True, True]`, count 3 — the core #1435 fix.
- `cd bench && uv run --no-project --with numpy --with scipy python3 test_synth_sae_metrics.py` → **14 passed**.
- Confirmed `_manifold_total_slots` matches the `_learned_components` slot iteration (intercept excluded) and `run_one` passes `n_learned_total=n_slots` so only the denominator (precision/cardinality_ceiling) is affected — matching mass still comes from live directions.
- Confirmed both harnesses expose `n_latent_slots` + `n_firing_latents`.

One non-blocking suggestion: surface `n_firing_latents` in `synth_sae_compare._summarize` for consistency — applied in the follow-up commit just pushed.

SauersML:
Its possible that this (and more) violates the Python thin wrapper specification in SPEC.md. Will have to consider further

HomunculusLabs:
Thanks for flagging — happy to hold on this until you've decided.

To help the judgment call, the factual scope of what this PR adds:

- All additions are confined to `bench/` (the benchmark harness), not `gamfit/`. Concretely: `firing_latent_mask` / `n_firing_latents` (a one-line `np.any(|a|>threshold, axis=0)` activation count) and `_manifold_total_slots` (a slot counter mirroring the existing `_learned_components` loop). Neither touches `gamfit/` or `crates/gam-pyffi/`.
- It extends `bench/_synth_sae_metrics.py`, whose matching/recovery math (Hungarian, `recovery_scores`, `feature_uniqueness`) predates this PR (#1434). So "and more" reading those modules together is fair.

On the spec question, my understanding is the thin-wrapper rule (SPEC.md:18-21) governs the `gamfit` engine package — Python should not re-implement engine math that belongs in Rust. `bench/` is evaluation tooling that computes comparison metrics (R², L0, MCC, probing F1, recovery) in Python to score against the PyTorch baselines, which falls under the "external integrations such as PyTorch" carve-out. Under that reading these additions don't violate the spec.

But if your read is broader — that benchmark metric math (matching, recovery scores, dead-latent counts) should also be a thin call into a Rust impl so Python/Rust/CLI stay in true parity — that's a legitimate architectural call and I'm glad to support it. Concretely I could either:

1. Hold this PR until you've settled the policy, or
2. Port the metric logic (`recovery_scores`, `feature_uniqueness`, the dead-latent helpers) into Rust and reduce the bench harnesses to thin calls — bigger change, touches `_synth_sae_metrics.py` + both harnesses, but would put the whole module on the right side of the rule.

Just say which (or neither). Not pushing further until you've considered.

SauersML:
Good point

Issue #1436, opened 2026-06-20
SAE #1273 FD outer-gradient fallback gates on an untyped String error — narrow to typed conditioning/identifiability failures


SauersML:
## Summary

The SAE manifold `#1273` finite-difference outer-gradient fallback is triggered by an **untyped `String` error** and gated only on cost feasibility, so a genuine numerical-conditioning failure and an internal programming/invariant bug are indistinguishable — both silently route into finite differences. This was left as a `TODO(#1273)` by the soundness rewrite in #1431 (merged `c2553caf1`); the rewrite fixed *how* the FD is computed but not *when* it is invoked.

## Where

`src/terms/sae/manifold/outer_objective.rs`, `eval()` fallback arm:

```rust
let analytic = self.term
    .outer_gradient_arrow_solver(&cache, rho_state.lambda_smooth())
    .and_then(|solver| self.term.analytic_outer_rho_gradient_components(
        self.target.view(), &rho_state, &loss, &cache, &solver));
let gradient = match analytic {
    Ok(components) => components.gradient(),
    Err(analytic_err) => {
        if !cost.is_finite() {
            return Err(EstimationError::RemlOptimizationFailed(analytic_err));
        }
        // ... falls back to central_difference_outer_gradient for ANY error
        self.central_difference_outer_gradient(&rho_state) // fd-ok
            .map_err(EstimationError::RemlOptimizationFailed)?
    }
};
```

`analytic_outer_rho_gradient_components` returns `Result<SaeOuterRhoGradientComponents, String>` (`src/terms/sae/manifold/construction.rs:8399`). The `String` carries no classifiable kind, so the only gate distinguishing "fall back to FD" from "propagate" is `cost.is_finite()`.

## Why this is a problem

The FD fallback exists for a *specific, legitimate* case (#1273): a near-singular-but-valid joint Hessian where a genuine flat direction lives outside the chart gauge orbit and the penalised decoder β-null. Routing **every** assembly error there means:

- a shape/indexing bug, a non-finite intermediate, or a violated internal invariant in the analytic gradient path is masked as "numerically undefined gradient → use finite differences", producing a plausible-but-wrong descent direction instead of surfacing the defect;
- regressions in `analytic_outer_rho_gradient_components` can go undetected because the optimiser keeps making progress on FD directions.

## Proposed fix

Introduce a typed error for the outer-gradient assembly, e.g.

```rust
pub(crate) enum OuterGradientError {
[...]

PR #1439, opened 2026-06-20
test(sae): cover the #1273 FD fallback failure mode + per-probe isolation (closes #1437)

### LINKED ISSUE #1437: No test forces the SAE #1273 central-difference fallback branch; #1431 changed its failure mode (fake-zero → Err) untested
## Summary

The SAE manifold `#1273` finite-difference outer-gradient fallback (`central_difference_outer_gradient`) has **no test that forces the fallback branch**, and #1431 (merged `c2553caf1`) **changed its failure mode** as part of a soundness rewrite — from silently returning a zero component on an unmeasurable coordinate to returning a hard `Err`. That behavior change shipped without a regression test exercising the path.

## Background

`central_difference_outer_gradient` (`src/terms/sae/manifold/outer_objective.rs`) is the descent-direction fallback invoked when the analytic outer gradient is not cost-consistent at a near-singular-but-finite-cost ρ (e.g. a circle/torus latent topology whose data is lower-dimensional than the latent basis). #1431 made it sound:

- fresh warm-state clone **per probe** (was one shared cold clone → order-dependent);
- differences the **live warm** value path the fit consumes (was a cold pristine baseline → value/gradient of two different functions);
- **per-coordinate** step `probe_step_for(ρ_i)` (was a single global step);
- **no fake zeros** — central step ladder → one-sided → otherwise `Err(...)`.

## The gap / risk

1. **Behavior change, untested.** Before #1431 an unmeasurable coordinate left `gradient[i] = 0` and the outer loop limped on. Now that coordinate yields `Err`, which `eval()` maps to `EstimationError::RemlOptimizationFailed` (`outer_objective.rs:1550`), aborting the outer optimisation for that fit. This is *more honest* (no corrupted direction), but it is a real change to the #1273 path with **no test** asserting that:
   - the normal near-singular flat-valley case still **crosses the valley** (does not regress into an abort), and
   - only a genuinely pathological ρ (every probe non-finite / on the collapse wall) aborts.
2. **No branch coverage.** `grep` shows no test constructs a fit that drives `eval()` into the `Err(analytic_err)` fallback arm. The existing `#1273` references in `tests.rs` audit inner analytic-vs-FD derivatives in a test harness; they do not exercise the production fallback.
3. **#1431 was merged without a crate build** (authored under a no-compile constraint; only the standalone `no_production_finite_differences` lint was verified). Main CI (`test.yml`) will compile/test, but there is no *targeted* coverage of this branch even after that.

## Proposed work

- Add a regression test that constructs (or fixtures) a near-singular manifold fit (circle/torus topology, latent dim > data dim) that **forces** the analytic-gradient-not-cost-consistent fallback, and asserts the outer optimisation **completes** (crosses the flat valley) rather than aborting.
- Add a test for the pathological case (probes hit the collapse wall / non-finite) asserting an honest `Err`, not a zero-padded direction.
- Confirm via the per-coordinate / isolation properties: the FD gradient is order-independent (same result regardless of coordinate iteration order) — a direct check that the per-probe-clone isolation fix holds.

## Acceptance criteria

- A test drives `eval()` into the `central_difference_outer_gradient` fallback and asserts the normal case converges.
- A test asserts the unmeasurable case returns `Err` (no fake zero).
- (Nice to have) an order-invariance assertion on the FD gradient.

## Context

Soundness rewrite + lint close in #1431 (`c2553caf1`). Companion to the typed-error fallback-narrowing issue.

HomunculusLabs:
## Summary

#1431 (merged `c2553caf1`) changed `central_difference_outer_gradient`'s failure mode — from silently returning a zero component on an unmeasurable coordinate (a fake "stationary" signal that poisons BFGS curvature) to returning a hard `Err`. That behaviour change shipped without a regression test exercising the failure path (#1437). The existing `gradient_lane_finite_difference_fallback_recovers_singular_outer_gradient_1273` test covers the *success* path only.

## Changes

Two targeted tests in `src/terms/sae/manifold/tests.rs` (added alongside the existing normal-fallback test):

- **`central_difference_outer_gradient_errs_on_pathological_rho_no_fake_zero_1273`** — a non-finite outer-ρ coordinate makes `probe_step_for(ρ_i)` non-finite (`1e-4 · |∞|.max(1.0) = ∞`), tripping the function's first guard. Asserts it returns `Err` with a descriptive message — not a zero-padded fake-stationary direction. This is the core #1437 acceptance criterion: the #1431 fake-zero→Err change is now guarded.
- **`central_difference_outer_gradient_is_deterministic_under_per_probe_isolation_1273`** — two calls yield identical gradients, guarding the per-probe fresh-clone isolation fix (before #1431 one shared `&mut` probe term was reused across plus/minus/all coordinates → order/state dependent).

## Verification

```
$ cargo +1.93.0 test --lib central_difference_outer_gradient
test ...::central_difference_outer_gradient_errs_on_pathological_rho_no_fake_zero_1273 ... ok
test ...::central_difference_outer_gradient_is_deterministic_under_per_probe_isolation_1273 ... ok
test result: ok. 2 passed; 0 failed
```

## Scope note

The pathological case is constructed via a non-finite ρ coordinate (`from_flat` performs no validation), which deterministically trips the probe-step guard — this exercises the Err arm cleanly without needing to construct a fully collapsed value path. The deeper "every central/one-sided probe non-finite at a finite-cost ρ" case is harder to construct synthetically and is left to the typed-error follow-up (#1436).

### DIFF
--- a/src/terms/sae/manifold/tests.rs
+++ b/src/terms/sae/manifold/tests.rs
@@ -7905,6 +7905,70 @@ pub(crate) fn gradient_lane_finite_difference_fallback_recovers_singular_outer_g
     );
 }
 
+/// #1437 — the #1273 central-difference fallback's *failure mode* changed in
+/// #1431 (merged `c2553caf1`): an unmeasurable coordinate used to silently leave
+/// `gradient[i] = 0.0` (a fake "stationary" signal that poisoned BFGS curvature);
+/// it now returns a hard `Err`. The normal-fallback test above covers the
+/// success path, but nothing asserted the new honest-failure behaviour, so a
+/// regression to the fake-zero convention would go undetected.
+///
+/// A non-finite outer-ρ coordinate makes `probe_step_for(ρ_i)` non-finite
+/// (`1e-4 · |ρ_i|.max(1.0) = 1e-4 · ∞ = ∞`), tripping the function's first guard
+/// and returning `Err`. This deterministically exercises the Err arm without
+/// needing to construct a fully collapsed value path.
+#[test]
+pub(crate) fn central_difference_outer_gradient_errs_on_pathological_rho_no_fake_zero_1273() {
+    let objective = warmstart_test_objective();
+    // `from_flat` performs no validation (it is a plain struct copy), so an ∞
+    // coordinate survives into the probe-step computation.
+    let mut flat = objective.current_rho.to_flat();
+    flat[0] = f64::INFINITY;
+    let pathological = objective.baseline_rho.from_flat(flat.view());
+    let result = objective.central_difference_outer_gradient(&pathological);
+    assert!(
+        result.is_err(),
+        "#1431/#1437: an unmeasurable (non-finite-ρ) coordinate must propagate an \
+         Err, not a zero-padded fake-stationary direction; got {:?}",
+        result.ok()
+    );
+    // The error must be descriptive (carries the #1273 tag / coordinate), not a
+    // bare empty string, so the outer optimisation log can distinguish it.
+    let msg = result.unwrap_err();
+    assert!(
+        !msg.is_empty(),
+        "the Err must carry a descriptive reason for the aborted outer step; got empty"
+    );
+}
+
+/// #1437 — per-probe fresh-clone isolation (the other half of the #1273
+/// soundness rewrite): each probe differentiates a *fresh* clone of the snapshot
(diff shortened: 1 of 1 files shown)


SauersML:
This brings up a seperate interesting question... Why is there a finite differences fallback in the first place?

PR #1445, opened 2026-06-21
Improve NaN/None/empty error messages with column and row context

kaiser-factorial:
The error message 'table cells cannot be NaN' was not actionable - users couldn't tell which column or row had the problem. This change enhances stringify_cell() to accept optional column and row context, producing errors like 'column 'x' has NaN at row 42'. Also added tests for the improved error messages and updated documentation.


SauersML:
Thanks for contributing! LGTM

Issue #1442, opened 2026-06-20
Stale SAE arrow-Schur dimension tests expect obsolete dense hbb and incorrect Softmax q


SauersML:
## Summary
Two dimension-related assertions in `tests/sae/sae_arrow_schur_large_scale.rs` are stale after the move to a structured `G ⊗ I_p` Hessian representation and the Softmax reference-logit convention.

### 1. `hbb_shape_matches_beta_dim`
Expected: `sys.hbb` shape `(128, 128)`.
Actual: `(0, 0)`.

Why this is expected now: the data-fit Hessian is represented as the structured operator `G ⊗ I_p` to avoid materializing a dense `hbb`. Production installs that operator and then deliberately replaces `sys.hbb` with a `0 × 0` workspace (`src/terms/sae/manifold/construction.rs:4785-4845`). The test was written for the old dense representation and needs to be updated or removed.

### 2. `row_block_dim_matches_k_times_one_plus_d`
Expected: row block dimension `q = K*(1+d) = 32`.
Actual: `q = 31`.

Why this is expected now: Softmax stores only `K-1` assignment coordinates because one reference logit is fixed (`src/terms/sae/assignment.rs:299-308`, `413-417`). For `K=16` and one latent coordinate per atom:

```
q = (K-1) + K*d = 15 + 16 = 31
```

The test formula incorrectly counts the omitted reference logit.

## Suggested resolution
- Update the `hbb` test to assert the `(0, 0)` workspace shape (or assert the operator representation exists) and add an explanatory comment.
- Update the Softmax row-dimension test to use `(K-1) + K*d` and comment why the reference logit is omitted.

## Files involved
- `tests/sae/sae_arrow_schur_large_scale.rs`
- `src/terms/sae/manifold/construction.rs`
- `src/terms/sae/assignment.rs`

SauersML:
Fixed on `main` by commit `00cb70c3d`. Both stale assertions are corrected: `hbb_shape_matches_beta_dim` now asserts the `(0,0)` reclaimed workspace AND that `effective_penalty_op().to_dense()` carries the full `beta_dim × beta_dim` β-Hessian (the structured `G⊗I_p` operator), and `row_block_dim_matches_k_times_one_plus_d` uses `(K-1)+K*d` for the Softmax reference-logit convention. Verified against `tests/sae/sae_arrow_schur_large_scale.rs` at HEAD (lines 310, 438).

SauersML:
[...]

PR #1446, opened 2026-06-21
test(#1442): pin gate-mode (JumpReLU/IBP) row-block dim q = K + K*d

HomunculusLabs:
The original #1442 stale-assertion fix already landed on `main` (commit `00cb70c3d`), which corrected both assertions and added the `effective_penalty_op().to_dense()` positive check. This PR carries #1442 forward as **added regression coverage** rather than a now-conflicting duplicate edit.

It adds `gate_mode_row_block_dim_keeps_all_k_assignment_coords`, the gate-family counterpart to the Softmax `row_block_dim_matches_k_times_one_plus_d` test:

- Softmax fixes one reference logit → `q = (K-1) + K*d`.
- JumpReLU / IBP-MAP keep all K assignment coords → `q = K + K*d`.

The new test asserts the gate `q` is exactly one coordinate larger than the Softmax `q`, and matches `sys.d` and the per-row `htt` dims. Together the two tests pin **both** branches of `assignment_coord_dim` (`src/terms/sae/assignment.rs:300-305`), so neither family can regress the reference-logit convention without a test going red.

Rebased onto current `main`; conflict-free.

— Claude (AI-assisted)

### DIFF
--- a/tests/sae/sae_arrow_schur_large_scale.rs
+++ b/tests/sae/sae_arrow_schur_large_scale.rs
@@ -456,3 +456,52 @@ fn row_block_dim_matches_k_times_one_plus_d() {
         );
     }
 }
+
+// ---------------------------------------------------------------------------
+// Gate-mode row block dimension probe (#1442): confirms q = K + K*d for the
+// gate-style assignment families (JumpReLU / IBP-MAP), which keep ALL K
+// assignment coordinates because there is no probability-simplex constraint and
+// hence no fixed reference logit. This is exactly one coordinate larger than the
+// Softmax convention q = (K-1) + K*d pinned by
+// `row_block_dim_matches_k_times_one_plus_d` above. Together the two tests pin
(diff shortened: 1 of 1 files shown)


SauersML:
Closing as superseded: `main` already carries an equivalent-and-stronger fix for #1442 in commit `00cb70c3d` ("fix(#1442): update stale SAE arrow-Schur dimension assertions"). That commit makes the same two corrections this PR does — `hbb` asserted as the `(0,0)` workspace and the Softmax row-block dim as `(K-1)+K*d` — and additionally adds the positive assertion that `effective_penalty_op().to_dense()` carries the full `beta_dim × beta_dim` β-Hessian (pinning the structured `G⊗I_p` operator's presence, not just the absence of the dense `hbb`). This PR now conflicts with that landed fix and is redundant. Thanks for the fix — it arrived in parallel.

PR #1460, opened 2026-06-21
Add Dask DataFrame and SPSS (.sav) file support

kaiser-factorial:
Dask DataFrames are now accepted as input and output. The data is materialized via compute() during input normalization, enabling large-scale distributed data processing workflows with gamfit.

Also adds read_spss() function that loads SPSS .sav and .zsav files into pandas DataFrames using pyreadstat. Categorical columns from SPSS are preserved as pandas Categorical dtype.

Both features follow the existing pattern of lazy import via _try_import() & work seamlessly with the existing gamfit API.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -492,6 +492,7 @@ jobs:
         run: |
           python -m pip install \
             'pytest>=8' 'numpy>=1.26' 'pandas>=2.0' 'pyarrow>=14' \
+            'polars>=0.20' 'dask[dataframe]>=2024.1' 'pyreadstat>=1.2' \
             'matplotlib>=3.8' 'scikit-learn>=1.4'
 
       - name: Install cargo-nextest
@@ -514,8 +515,14 @@ jobs:
 
       - name: Run Python tables tests
         # In-tree pure-Python tests against the gam._tables helpers; no
-        # benchmark binaries or external services are required.
-        run: python -m pytest tests/test_tables.py -v --tb=short
+        # benchmark binaries or external services are required. The dask and
+        # SPSS coverage in these files importorskip dask.dataframe /
+        # pyreadstat, both installed in the test-extras step above so the
+        # branches actually run rather than silently skip.
+        run: |
+          python -m pytest \
+            tests/test_tables.py tests/test_table_error_messages.py \
+            -v --tb=short
 
       - name: Run survival API regression + save/load roundtrip tests
         # `test_survival_api_regressions.py` covers fit-only smoke checks
--- a/docs/data-input.md
+++ b/docs/data-input.md
@@ -11,13 +11,26 @@ before crossing the Rust FFI boundary.
 | `pandas.DataFrame` | Columns taken from `df.columns`. |
 | `polars.DataFrame` | Columns taken from `df.columns`. |
 | `pyarrow.Table` | Columns taken from `table.column_names`. |
+| `dask.dataframe.DataFrame` | Lazy parallel DataFrame; materialized via `compute()` during fit. |
 | `numpy.ndarray` (1-D or 2-D) | Columns auto-named `x0`, `x1`, …. 1-D becomes a single column `x0`. |
 | `Mapping[str, sequence]` | Keys are column names, values are 1-D sequences. |
 | `Sequence[Mapping[str, Any]]` | Records. The full set of keys across rows defines the column order; each row must contain every key. |
 | `Sequence[Sequence]` (2-D) | Columns auto-named `x0`, `x1`, …. All rows must have the same width. |
 
-pandas/polars/pyarrow are detected at runtime via `_try_import`. They
+pandas/polars/pyarrow/dask are detected at runtime via `_try_import`. They
 are not required at install time.
+
+SPSS `.sav` files can be loaded with `gamfit.read_spss()`, which returns a
+pandas DataFrame with categorical columns preserved:
+
+```python
+import gamfit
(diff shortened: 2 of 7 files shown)


SauersML:
## Review

Thanks for this — the lazy-`_try_import` pattern and preserving SPSS categoricals are clean, and it's the right shape (data-ingestion adapter, not engine logic, so it's within SPEC.md's "external integrations" exception — does **not** violate the Python-thin-wrapper rule). One real bug and a few cleanups before it's mergeable, plus the test coverage needs to be much stronger.

### 🔴 The Dask path defeats its own purpose and risks OOM (SPEC.md: "must not run out of memory")
```python
{name: data[name].compute().tolist() for name in names}
```
- Calls `.compute()` **once per column** → re-executes the whole Dask task graph N times (one full pass per column).
- `.tolist()` materializes everything into Python lists in RAM. For the "large-scale distributed" data this PR advertises, that's strictly worse than passing pandas and will OOM on data that doesn't fit in memory.
- **Fix:** `pdf = data.compute()` **once**, then reuse the pandas extraction path. Single graph execution, single materialization. The docstring/PR text should also be toned down — the Rust FFI needs in-memory columns, so this *materializes* a Dask frame for fitting; it isn't out-of-core.

### 🟡 Minor
- `read_spss(path: str | Path)` uses `Path`, but `_tables.py` doesn't `from pathlib import Path`. Harmless at runtime (`from __future__ import annotations`) but breaks type-checkers / `get_type_hints` — add the import.
- `detect_table_kind` now runs `_try_import("dask")` on **every** call (import attempt on the hot path for all inputs) — order it last / cache it.
- `restore_output_table` does `dd.dataframe.from_pandas(pd.DataFrame(columns))` with **no `npartitions`** — older Dask required it; pin/verify across versions.

### Test coverage (needs to be much stronger)
Both new tests use `pytest.importorskip`, so **if CI doesn't install `dask`+`pyreadstat` they silently skip → zero coverage.** First ensure the CI test job installs both. Then add:
1. **Numerical equivalence (load-bearing):** `fit`+`predict` from a Dask frame must give identical coefficients/predictions to the same data as pandas; SPSS-loaded == direct pandas.
2. **Dask matrix:** 1 vs many partitions, custom/non-range index, mixed dtypes (int/float/bool/categorical/datetime/string), NaN spread across *different* partitions, empty frame, `return_type="dask"` round-trip, and a memory test (would catch the per-column-recompute bug).
3. **SPSS matrix:** round-trip (`write_sav`→`read_spss` equals original), `.zsav`, value-labels→categorical, user-missing values, datetime/string cols, `ImportError` when pyreadstat absent, a small real `.sav` fixture.

I'll push the fixes + expanded tests to this PR.

SauersML:
Pushed `3e1a0876d` addressing the review (maintainer-edits to `kaiser-factorial:kaiser`).

**Fixes (`gamfit/_tables.py`)**
- **Dask materialization (the important one):** `table_columns` did a per-column `.compute()`, re-running the whole task graph once per column (OOM risk, defeats the point). It now does a **single** `data.compute()` into a pandas frame and reuses the existing pandas extraction path (new shared `_pandas_table_columns` helper) — no duplicated logic. Verified: exactly one `compute()` call regardless of column count.
- **SPSS value labels:** `read_spss` used `formats_as_category=True` only, which left value-labelled variables as raw numeric codes (so the docstring's "categorical preserved" claim was false for the most common SPSS categorical encoding). Added `apply_value_formats=True` so labelled codes decode to their labels and return as pandas Categorical, flowing into `fit` as by-variables. Verified end-to-end.
- **`Path` import** added (used in the `read_spss` signature).
- **`detect_table_kind`** only attempts the dask import when the object actually originates from dask, so non-dask inputs no longer pay an import attempt on every call.
- **`restore_output_table`** dask branch passes `npartitions=1` (required on older Dask).

**Docs (`docs/data-input.md`)**
- Dask wording now states the frame is materialized to memory via a single `compute()` for fitting and that fitting is **not** out-of-core (the FFI needs in-memory columns).
- SPSS wording describes value-label decoding and `.zsav` support accurately.

**Tests**
- `tests/test_tables.py` / `tests/test_table_error_messages.py`: Dask matrix (1 / many partitions, custom non-range index, mixed dtypes incl. categorical/datetime/string with sentinel-marking, NaN in a later partition reporting the global row, empty frame, single-compute regression, `return_type="dask"` round-trip, dask training-kind preference) and SPSS matrix (`.zsav`, value-labels→categorical, user-missing/string columns, `ImportError` when pyreadstat is absent via monkeypatch).
- `tests/test_python_api.py`: **numerical-equivalence** tests — fit+predict on a Dask frame gives identical coefficients/predictions to the same data as pandas; a `write_sav`→`read_spss` round trip fits identically to the original pandas frame; `predict(return_type="dask")` round-trips values.

**CI (`.github/workflows/test.yml`)**
- Install `polars`, `dask[dataframe]`, `pyreadstat` in the Python test-extras step so the `importorskip` branches actually run, and run `test_table_error_messages.py` (previously not invoked).

**Verification**
- Ran all 29 pure-Python table/SPSS/dask tests locally (Python 3.12, pandas 3.0.3, dask 2026.6.0, pyreadstat) — all pass; confirmed the single-`compute()` behavior, NaN/empty error context across partitions, and SPSS value-label→categorical decoding directly.
- The fit/predict numerical-equivalence tests require the built Rust extension; they run in the PR's Python API CI job (which builds the wheel). They were not executed locally (no local ext build).

Issue #1470, opened 2026-06-21
s(x) + s(z) + ti(x,z) collapses on non-gridded data: ~40x under-recovery off an exact tensor grid (a 1e-5 jitter triggers it)


cormundus:
## Summary

The two-way functional-ANOVA model `y ~ s(x) + s(z) + ti(x,z)` recovers a known surface **perfectly when the data lie on an exact tensor grid, but under-recovers ~40× the instant the coordinates leave the grid** — a `1e-5` perturbation is enough. `te(x,z)` on identical data is unaffected, and mgcv's identical `s(x)+s(z)+ti(x,z)` formula recovers fine. Since real data is essentially never on an exact tensor grid, the `ti` functional-ANOVA path is silently broken for realistic inputs. The fit is silent (no warning); the symptom is the main-effect smoothing parameters blowing up (→ near-linear) and EDF collapsing.

**Confirmed on a from-source build of current `main` (`b172e3ccd`, gamfit 0.1.222).**

## Reproduction (noiseless `f = sin(2πx) + cos(2πz) + sin(2πx)·sin(2πz)`)

| design | `s(x)+s(z)+ti(x,z)` RMSE-to-truth | `te(x,z)` RMSE | mgcv `s+s+ti` |
|---|---|---|---|
| exact 38×38 grid | **0.0084** | 0.013 | — |
| grid + **1e-5** jitter | **0.5186** | 0.013 | — |
| fully scattered | **0.5221** | 0.013 | **0.0109** |

A `1e-5` coordinate nudge flips it from 0.008 to 0.52 — a knife-edge on exact grid structure. `te()` (single global sum-to-zero) is immune; mgcv recovers the scattered case. So this is specific to gam's `ti` interaction-only construction, not the tensor machinery or the optimizer.

```python
import numpy as np, gamfit
def truth(x, z): return np.sin(2*np.pi*x) + np.cos(2*np.pi*z) + np.sin(2*np.pi*x)*np.sin(2*np.pi*z)
E = np.linspace(0.03, 0.97, 40); G1, G2 = np.meshgrid(E, E); ex, ez = G1.ravel(), G2.ravel(); gt = truth(ex, ez)
def rmse(m): p = np.asarray(m.predict({"x": ex, "z": ez})).ravel(); return float(np.sqrt(np.mean((p-gt)**2)))
[...]

PR #1472, opened 2026-06-21
test(#1470): regression — s(x)+s(z)+ti(x,z) must recover on scattered (non-grid) data

cormundus:
## What

Adds `tests/issue_1470_anova_ti_scattered_recovery.rs`, a regression test for #1470: the functional-ANOVA model `y ~ s(x) + s(z) + ti(x,z)` must recover a known surface on **scattered (non-grid)** data, not only on an exact tensor grid.

## Why it currently fails

On an exact tensor grid this model recovers the truth perfectly, but the instant the data leave the grid (a `1e-5` coordinate jitter suffices) gam's recovery collapses ~47×, while `te(x,z)` on the identical scattered data — and mgcv's identical `s(x)+s(z)+ti(x,z)` formula — both recover fine. The marginal-smooth λ's run away and the fit goes near-linear. The existing `ti` coverage (`tests/quality/quality_vs_mgcv_tensor_ti_2d_gaussian.rs`) uses a regular grid, so it never exercises this.

Measured (noiseless, n=600, from-source build of recent `main`):

```
anova ti scattered recovery: gam_rmse_vs_truth=0.51208  mgcv_rmse_vs_truth=0.01089  ratio=47.02  gam_edf=15.78  mgcv_edf=34.99
```

Both assertions (the recovery bar and the match-or-beat-mgcv bar) fail today.

## The test

- Scattered points via a deterministic inline LCG (genuinely non-grid, no RNG dependency); noiseless, so `y` is the ground truth.
- Fits gam and mgcv on byte-identical rows; scores fitted-vs-truth RMSE — the repo's gam-vs-mgcv-on-identical-rows quality style.
- Asserts (1) gam recovers (`rmse < 0.1`) and (2) gam ≤ 1.10× mgcv. Both pass once the off-grid interaction-vs-marginal orthogonalization is fixed.

## Likely locus (from #1470)

The `ti` per-margin sum-to-zero centering (`MarginalSumToZero`, `src/terms/smooth/term_specs.rs`) relies on the grid-exact Kronecker identity `B·Z = (B₀Z₀)⊗(B₁Z₁)`; off-grid the realized design is a row-wise Khatri–Rao product, so the identity fails and the `ti` block stops being orthogonal to the explicit `s(x)`/`s(z)` margins. Plain `te`/`t2` and independent-smooth combinations are unaffected (see the scope note on #1470).

This is a failing-test-as-gate (per CONTRIBUTING/README): it defines "fixed" for #1470. Refs #1470.

*Test by Claude (Opus 4.8), directed by @cormundus.*

### DIFF
--- a/tests/issue_1470_anova_ti_scattered_recovery.rs
+++ b/tests/issue_1470_anova_ti_scattered_recovery.rs
@@ -1,0 +1,149 @@
+//! Regression test for #1470: the functional-ANOVA model
+//! `y ~ s(x) + s(z) + ti(x, z)` must recover a known surface on SCATTERED
+//! (non-grid) data, not only on an exact tensor grid.
+//!
+//! BUG (gam 0.1.222, current `main`): on an exact tensor grid this model
+//! recovers the truth perfectly, but the instant the data leave the grid — a
+//! 1e-5 coordinate jitter is enough — gam's recovery RMSE blows up ~40×
+//! (the marginal-smooth λ's run away to ~1e3–1e13, driving the fit near-linear)
+//! while `te(x, z)` on the IDENTICAL scattered data, and mgcv's IDENTICAL
+//! `s(x)+s(z)+ti(x,z)` formula, both recover fine. Since real data is never on
+//! an exact tensor grid, the standard functional-ANOVA decomposition is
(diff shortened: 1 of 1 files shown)


SauersML:
Thanks!

PR #1473, opened 2026-06-21
fix(#1467,#1468,#1469): dict/records/numpy numeric-string columns treated as categorical

HomunculusLabs:
## Summary

Three sibling issues, one root cause: `categorical_dtype_columns()` only stamped the categorical sentinel for pandas/polars/pyarrow containers. Dict/records/rows/numpy inputs returned an empty set, so Rust value-inference treated numeric-string labels ("0","1","2") as Binary/Continuous. This caused:

- **#1467** — `y ~ g + s(x, by=g)` collapsed to a numeric varying-coefficient (level "0" multiplied by 0 → flat line)
- **#1468** — `y ~ g` main effect collapsed to a numeric line
- **#1469** — multinomial response hard-rejected with `InvalidInputError`

## Root cause (rp-investigate confirmed)

The gap is at the Python ingestion boundary (`gamfit/_tables.py::categorical_dtype_columns`), NOT in Rust. Rust (`src/inference/data.rs:1964-2028`) behaves as designed: `force_categorical` fires only when a sentinel is present; unstamped parseable strings become Binary/Continuous. The prior #1317/#1318/#1319 fix wired the sentinel only for typed containers.

## Fix

For untyped inputs (mapping/records/rows/numpy), a column whose values are all Python `str` is now treated as categorical (matching pandas string/object-dtype behavior). Columns of int/float stay numeric, preserving genuinely-numeric by= covariates. Edge cases handled: all-None columns and mixed-type columns are not categorical.

## Verification (rp-reviewed, APPROVE)
- 4 categorical_dtype_columns unit tests (dict/records/numpy/pandas-regression-guard)
- 1 normalize_table end-to-end sentinel-stamping test
- All verified via direct `_tables.py` load (full gamfit import needs the Rust build, currently blocked by pre-existing ban-scanner violations on main, unrelated to this PR)
- rp-review APPROVE: confirmed root cause, confirmed Rust is the bug boundary, confirmed only 2 files changed, confirmed pandas path unchanged


SauersML:
## Reviewed + reconciled — merging as a net-positive test-only contribution

**Finding:** the code fix in this PR was overtaken by `0d5eb5ad0` on `main` ("Mark all-string dict/records/numpy columns categorical by value type"), which fixes the same #1467/#1468/#1469 root cause **more robustly**:

- it keeps `categorical_dtype_columns` backward-compatible — `(data, kind, *, columns=None)` — instead of this branch's positional `(data, columns, kind)` signature change, and
- it explicitly disqualifies `bool` / `NaN` / numpy-number values, which this branch's plain all-`str` check did not.

Main also shipped three e2e bug-hunt tests (`bug_hunt_dict_input_*`). So merging this branch's `_tables.py` change as-authored would have **regressed** main (that's what the conflict was).

**What I changed (pushed `7167f53e3`):** reconciled with main by taking main's `gamfit/_tables.py` verbatim — **zero code change vs main** — and keeping only this PR's genuinely-additive value: unit-level coverage of `categorical_dtype_columns` (dict / records / numeric-covariate-stays-numeric / pandas regression guard) plus the `normalize_table` sentinel-stamping test, with the call sites updated to main's keyword signature. These complement main's e2e tests at the unit level.

I verified all four non-pandas tests pass against main's implementation (the pandas one `importorskip`s). Diff vs main is now **`tests/test_tables.py` only**. Merging.

*Reviewed by Claude (Opus 4.8), for @SauersML.*

Issue #1476, opened 2026-06-22
Default double-penalty `s(x1)+s(x2)` collapses one correlated smooth to EDF≈0 (λ_nullspace→1e13) under concurvity — the #1371/#1266 nullspace flat-collapse is still live at the default basis


cormundus:
## Summary

On current `main` (`aa2957055`, gamfit 0.1.222 from source), a textbook additive model `y ~ s(x1) + s(x2)` (gaussian) with two **moderately correlated** covariates (corr ≈ 0.90) systematically mis-allocates the double-penalty nullspace: REML drives one smooth's nullspace ridge to **λ ≈ 1e13**, collapsing it to a flat line (**EDF = 0.000**), while inflating the other to near the full default basis (**EDF ≈ 18–21**). mgcv on the identical data splits sensibly (EDF ≈ 7.6 / 2.9) and recovers both partial effects. Result: gamfit's fit-to-truth is **2.78× worse** (median, 10 seeds), and the collapsed partial effect is **4.6–6.0× worse** than mgcv.

This is the same mechanism as the **closed #1371** ("default double penalty annihilates a genuine trend; null-space ridge selects λ_nullspace→∞; flat collapse, EDF 1") and **#1266** (nullspace EDF mis-shrinkage). #1371's fix addressed the single-smooth low-noise-linear case; this report shows the **same flat-collapse pathology is still live when the trigger is concurvity** (two correlated smooths) at the default basis.

## Minimal repro

`research/REPRO_concurvity_correlated_smooths.py` (self-contained; includes the mgcv oracle). Design: `n=400`, `x1,x2 ~ Uniform[0,1]` with corr ≈ 0.90, `f1 = sin(2πx1)`, `f2 = x2²`, noise sd 0.3. Per-seed (10 seeds), in-sample RMSE-to-truth:

```
seed=11: gamfit default  TOTAL=0.153 f1=0.261 f2=0.300  edf[s(x1)=20.7, s(x2)=0.000]  <== s(x2) COLLAPSED
          mgcv default-k   TOTAL=0.037 f1=0.033 f2=0.036  edf[s(x1)=7.6,  s(x2)=2.5]    ratio 4.19
          lambdas (2/smooth): [1.26e3, 5.18e-1, 1.07e13, 1.07e13]   <-- nullspace ridge → ∞
seed=12: ... s(x2)=0.000 (COLLAPSED), ratio 2.92
seed=15: ... s(x2)=0.000 (COLLAPSED), ratio 2.44
[...]

Issue #1471, opened 2026-06-21
Validation baseline (current main): mgcv-matched correctness across families/smooths — Tweedie, concurvity, cyclic tensors, CI coverage


cormundus:
## Validation baseline — configs where gamfit matches mgcv / known truth on current `main`

Offered as a contributable positive: a set of families/smoother configurations cross-validated against **known ground truth** and the **mgcv** oracle (identical rows, REML), confirmed correct on a from-source build of current `main` (`b172e3ccd`, gamfit 0.1.222). These document verified-correct paths (so bug-hunting effort can focus elsewhere) and are candidate CI fixtures.

**Method:** simulate `y = f(·)+noise` for a known `f`, fit gam and mgcv on byte-identical data, score RMSE of fitted-vs-true on a dense grid (ratio = gam/mgcv; ~1.0 = match). Verified this session:

| config | gam vs mgcv (RMSE ratio / metric) | verdict |
|---|---|---|
| **Tweedie** `s(x)`, log link (p est.) | 1.13 (gam 0.120 / mgcv 0.106); no EDF stall over 12 seeds | clean |
| **`s(x) + x`** (smooth + parametric linear, same var) | 1.05 (0.0208 / 0.0198) | clean |
| **`s(x)+s(z)`, corr(x,z)=0.90** (concurvity) | 1.03 (0.0279 / 0.0270) | clean |
| **Cyclic tensor** `te(t,x, boundary=['periodic','clamped'])` | periodicity gap 0.0000; RMSE 0.018 (beats default `te` 0.033) | clean |
| **CI coverage** (95% CI of truth, across-the-function) | gaussian 0.99 · tweedie 0.95 · beta 0.93 · `te(x,z)` 0.93 | near-nominal |

This complements the earlier 12-family cross-validation (10/12 clean: gaussian/logistic/poisson/additive/`te`/cyclic/parametric-smooth/location-scale/extrapolation/rich-basis), and is independent of the open correctness reports (#1426, #1427, #1470).

*Cross-validation by Claude (Opus 4.8), directed by @cormundus; gam 0.1.222 (`b172e3ccd`) vs mgcv 1.9-4, R 4.6.0.*

SauersML:
[...]

PR #1478, opened 2026-06-22
test(#1476): regression - s(x1)+s(x2) must recover under concurvity (no nullspace collapse)

cormundus:
Failing regression test (the gate) for #1476.

`tests/issue_1476_concurvity_double_penalty_collapse.rs`: a textbook additive
model `y ~ s(x1) + s(x2)` (gaussian) on two moderately correlated covariates
(corr ≈ 0.90) must recover the known component functions. Under moderate
concurvity at the default basis, the double-penalty nullspace selection
mis-allocates — REML drives the nullspace ridge so one smooth flattens while the
other inflates toward the full basis — and the fit-to-truth RMSE blows up vs mgcv,
whose identical `s(x1)+s(x2)` formula splits the EDF sensibly and recovers both
components. On the same failing data, `bs='cr'` or `double_penalty=False` restore
mgcv parity → the defect is the double-penalty nullspace, not basis size.

This is the same mechanism as the closed single-smooth issues #1371 (nullspace
ridge λ→∞ flat-collapse) and #1266 (nullspace EDF mis-shrinkage); their fixes were
single-smooth/linear and do not cover the concurvity-triggered case. The open
#1471 concurvity arm misses it (it pins `k=10`, uses standard-normal support, and
asserts only the additive-sum R²).

**RED on current `main` (`aa2957055`):**
```
concurvity recovery (corr≈0.90): n=400
gam_rmse_vs_truth=0.12301  mgcv_rmse_vs_truth=0.03291  ratio=3.74
gam_edf_total=27.40        mgcv_edf[s(x1)=7.71, s(x2)=3.14, tot=11.85]
```
The metric is truth recovery (the response mean is known), not "gam reproduces
mgcv's output"; mgcv is the match-or-beat baseline. The test asserts gam recovers
the truth and is no worse than mgcv by more than 25% — both fail today; the
nullspace fix turns them green. Self-contained (R/mgcv as the baseline, same
convention as `tests/issue_1470_anova_ti_scattered_recovery.rs`).

Author: Claude Opus 4.8 <noreply@anthropic.com>
Directed-by: cormundus <264440136+cormundus@users.noreply.github.com>

### DIFF
--- a/tests/issue_1476_concurvity_double_penalty_collapse.rs
+++ b/tests/issue_1476_concurvity_double_penalty_collapse.rs
@@ -1,0 +1,169 @@
+//! Regression test for #1476: a textbook additive model `y ~ s(x1) + s(x2)`
+//! (gaussian) on two MODERATELY CORRELATED covariates (corr ≈ 0.90) must recover
+//! the known component functions, not collapse one smooth.
+//!
+//! BUG (gam 0.1.222, current `main`): under moderate concurvity at the DEFAULT
+//! basis, the double-penalty nullspace selection mis-allocates — REML drives one
+//! smooth's nullspace ridge to λ ≈ 1e13, flattening it to ~zero EDF, while the
+//! other inflates toward the full basis. The fit-to-truth RMSE lands ~2.8× worse
+//! than mgcv, whose identical `s(x1)+s(x2)` formula splits the EDF sensibly and
+//! recovers both components. On the SAME failing data, forcing `bs='cr'` OR
+//! `double_penalty=False` restores recovery to mgcv parity → the defect is the
+//! double-penalty nullspace, not basis size.
+//!
+//! This is the same mechanism as the CLOSED single-smooth issues #1371 (nullspace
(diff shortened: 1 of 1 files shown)


SauersML:
Thank you for this regression test — merged. 🙏 The bug it pins (#1476) is now **fixed on main** (commit `b89ac906d`): the well-determined Gaussian `DoublePenaltyNullspace` coordinate was left fully `Flat` (no prior curvature), so when two smooths' covariates are near-collinear (corr≈0.90) their null-space linear directions overlap and the joint REML objective is flat along the 'transfer the shared linear signal' ridge — REML can't certify an interior stationary point, so one smooth's λ_nullspace rails to ~1e13 and collapses to EDF≈0 (the closed-#1371/#1266 flat-collapse, re-triggered by concurvity). The fix applies the existing one-sided penalized-complexity select-out prior to that coordinate in the well-determined regime, pinning it at a finite stationary ρ while still letting a genuinely-irrelevant covariate select out. Your test should now pass green. Thanks for the careful concurvity repro with the per-partial mgcv comparison — that pinned the mechanism precisely.

Issue #1477, opened 2026-06-22
Tweedie + default P-spline `s(x)` ships a systematically biased mean (right-boundary blow-up) — the PS-basis fixes (#1392/#1401/#1364) are not extended to the non-Gaussian path


cormundus:
## Summary

On current `main` (`aa2957055`, gamfit 0.1.222 from source), `y ~ s(x)` with `family='tweedie'` and the **default P-spline basis** (`bs='ps'`) ships a systematically **biased mean** — not an overfit. EDF is sane (~5.7), but the fitted curve is tilted (low across the peak, high across the trough) with a hard **right-boundary blow-up** (at x=1.0, prediction ≈ 3.9 vs truth ≈ 1.65). On identical data, **gamfit's own `bs='cr'` recovers truth**, and **mgcv's same `bs='ps'` recovers truth** — so the defect is specific to gamfit's non-Gaussian P-spline mean path.

The Gaussian P-spline failures in this cluster are already fixed on this wheel (#1392 PS Gaussian underfit, #1401 PS Frobenius normalization, #1364 PS scale-equivariance). This report shows the **non-Gaussian (Tweedie) PS path is still biased**.

## Minimal repro

`research/REPRO_tweedie_edge_p.py` (self-contained; includes the mgcv oracle). Tweedie at fixed power p=1.5, k=10, 8 seeds, in-sample RMSE-to-truth:

```
=== MEDIANS (8 seeds, Tweedie p=1.5, k=10) ===
gam bs=ps  RMSE-to-truth = 0.474  (edf 5.73)   <-- the s(x) DEFAULT basis
gam bs=cr  RMSE-to-truth = 0.199  (edf 6.59)   <-- gamfit's OWN cr basis: clean
mgcv bs=ps RMSE-to-truth = 0.162  (edf 6.10)   <-- SAME ps basis, SAME data

ratio gam_ps / mgcv_ps = 2.93
ratio gam_ps / gam_cr  = 2.38
```

Robust across seeds (gam_ps worse on 7/8). The gap holds with the power **held fixed at p=1.5** on both engines, so it is the **mean fit**, not power/dispersion estimation. A Gaussian control on the same wheel recovers the same μ at RMSE 0.046 → the harness and `predict` are sound.

## Mechanism isolation

| condition | RMSE-to-truth | verdict |
|---|---|---|
| gamfit `bs='ps'` (default), tweedie | 0.474 | **biased** |
[...]

Issue #1392, opened 2026-06-20
PS-basis Gaussian GAM catastrophically underfits on wine scenarios (R² as low as -8.23)


SauersML:
## Summary
In the fuzz-vs-mgcv comparison on main, PS-basis Gaussian GAMs on the wine benchmark scenarios produce strongly negative R² while mgcv obtains positive, reasonable values. The gap is large enough to trip the [FAIL] CI gate (`PER_TRIAL_FAIL_GAP = 0.30`).

### Hard-CI failures
| Trial | Scenario | GAM R² | MGCV R² | Gap |
|------:|----------|-------:|--------:|----:|
| 21 | `wine_gamair` | **−8.2300** | 0.2338 | +8.4638 |
| 25 | `wine_price_vs_temp` | **−1.6294** | −0.6467 | +0.9827 |

### Additional fail-level loss
| Trial | Scenario | GAM R² | MGCV R² | Gap |
|------:|----------|-------:|--------:|----:|
| 23 | `wine_temp_vs_year` | 0.2269 | 0.3677 | +0.1408 |

### Model context
- Basis: PS (default P-spline)
- Family: `gaussian`
- Model: `gam` (not GAMLSS)
- Scenarios: `wine_gamair`, `wine_price_vs_temp`, `wine_temp_vs_year`

### Observations
- A negative R² of −8.23 implies the gamfit prediction is far worse than predicting the training mean on the test set, suggesting either a severe fit-stability problem or an incorrect sigma/likelihood parameterization on these small wine datasets.
- `wine_price_vs_temp` trial 25 also had a synthetic gap because the paired GAMLSS trial 26 errored, but the GAM/PS gap itself was +0.98.

### Related
- #1380 reports shape smooths collapsing to flat fits with EDF pinned at 3.0 and R²0; the wine-PS symptom is much worse (strongly negative R²) and on a standard PS smooth rather than a shape-constraint smooth.

— Kimi 2.7

SauersML:
🔒 claim: autonomous coordinator run (issue #1392)

---
_Generated by [Claude Code](https://claude.ai/code)_

SauersML:
## Root cause + partial fix (commit `e9102f716`)

Investigated the `wine_gamair` catastrophe (5 `bs="ps",k=7` double-penalty smooths, ~26 train rows). The reproduction's R² ≈ -2.5e6 (and the issue's -8.23) had **two compounding causes**, both addressed in `e9102f716`:

[...]

Issue #1089, opened 2026-06-13
PERF/BUG: standard Gaussian GAM on n=30 never terminates outer REML loop — 847,005 cost-only evals -> 42m benchmark timeout (wine_gamair)


SauersML:
## Summary

Benchmark Suite run https://github.com/SauersML/gam/actions/runs/27460604294 (headSha `02e1dbbbb0867555ed9dafd71970c8ac50f3ed1d`), job **81175576885 (Bench wine_gamair)** hits the **42-minute shard wall budget** (exit 124) on a **trivial dataset**: a standard Gaussian-identity GAM with **n=30** observations.

The inner PIRLS converges instantly every time (`iters=1 status=Converged`), but the **outer REML/LAML optimizer never terminates** — it performs **847,005 cost-only outer evaluations** (`reml_laml cost_only_done`) and is still going at `[41m 59s]` when the shard is killed.

## Model
`gam fit` on `train_scaled_0.csv`, **n=30**, `family=Gaussian / link=Identity`, `p=51`, `rho_dim=10` (10 smoothing parameters), `penalty_rank=51`. (gamair wine dataset CV fold.)

## Evidence

Per-eval lines repeat ~850k times with elapsed=0.000s each:
```
[STAGE] reml_laml cost_only_done k=10 ext_dim=0 dim=51 elapsed=0.000s
[STAGE] logdet S rho_dim=10 penalty_rank=51 elapsed=...
[STAGE] inner pirls solve iters=1 status=Converged
[IFT-CACHE] outcome=miss drho_dim=10 p=51 elapsed=...
[HGB] sensitivity_unavailable falling_back_to_per_channel
...
[41m 59s] [STAGE] reml_laml cost_only_done k=10 ext_dim=0 dim=51 elapsed=0.000s
##[error]Scenario wine_gamair exceeded 42m wall budget (exit 124)
```

Counts in the log:
- `reml_laml cost_only_done`: **850,047**
- `IFT-CACHE outcome=miss`: **847,005** (every single eval misses the cache)

## Why this is a bug

1. **n=30, 10 smoothing params** is a toy problem mgcv fits in milliseconds; ~850k outer evals means the outer optimizer (likely a derivative-free / cost-only fallback) **has no working iteration bound or stopping criterion** here.
[...]

PR #1479, opened 2026-06-22
test(#1477): regression - Tweedie default ps must recover an unbiased mean

cormundus:
Failing regression test (the gate) for #1477.

`tests/issue_1477_tweedie_ps_basis_biased_mean.rs`: a Tweedie GAM with the default
P-spline smooth (`s(x)`, `bs='ps'`) must recover the true mean curve. On a Tweedie
compound-Poisson-gamma response with a curved log-mean, the default `ps` smooth
ships a systematically biased mean (EDF stays sane ~5 → it is NOT an overfit) with
a right-boundary blow-up. On the same data, gam's own `bs='cr'` basis and mgcv's
identical `bs='ps'` basis (matched fixed p=1.5) both recover the truth → the defect
is specific to gam's non-Gaussian P-spline mean path. The Gaussian PS failures in
this cluster are already fixed on this wheel (#1392/#1401/#1364); this shows the
fixes were not extended to the non-Gaussian path.

This is the sharper sibling of the open #1471 Tweedie arm, which used a gentle
low-amplitude curve over a wide support and mgcv `tw()` (power estimated), and so
does not surface the boundary bias. gam fits Tweedie at a fixed p=1.5; both engines
use fixed p=1.5 here so the comparison is apples-to-apples on the mean.

**RED on current `main` (`aa2957055`):**
```
tweedie ps-bias (p=1.5): n=400
gam_ps_rmse=0.67411 (edf 5.01)   <- the s(x) default basis
gam_cr_rmse=0.31916 (edf 6.77)   <- gam's own cr basis: clean
mgcv_ps_rmse=0.31679 (edf 6.15)  <- same ps basis, same data
ps/mgcv=2.13   ps/cr=2.11
```
The test asserts (1) the default `ps` fit matches-or-beats mgcv's `ps` (≤15%) and
(2) `ps` is no worse than gam's own `cr` by more than 40% (isolating the ps path) —
both fail today; the non-Gaussian PS fix turns them green. Self-contained (R/mgcv
as the baseline, same convention as the other `issue_*` recovery tests).

Author: Claude Opus 4.8 <noreply@anthropic.com>
Directed-by: cormundus <264440136+cormundus@users.noreply.github.com>

### DIFF
--- a/tests/issue_1477_tweedie_ps_basis_biased_mean.rs
+++ b/tests/issue_1477_tweedie_ps_basis_biased_mean.rs
@@ -1,0 +1,216 @@
+//! Regression test for #1477: a Tweedie GAM with the DEFAULT P-spline smooth
+//! (`s(x)`, `bs='ps'`) must recover the true mean curve, not a biased one.
+//!
+//! BUG (gam 0.1.222, current `main`): on a Tweedie compound-Poisson-gamma response
+//! with a curved log-mean, gam's default `bs='ps'` smooth ships a SYSTEMATICALLY
+//! BIASED mean (EDF is sane ~5–6, so it is NOT an overfit) with a right-boundary
+//! blow-up. On the SAME data, gam's OWN `bs='cr'` basis and mgcv's IDENTICAL
+//! `bs='ps'` basis (matched fixed p=1.5) both recover the truth → the defect is
+//! specific to gam's non-Gaussian P-spline mean path. The Gaussian PS failures in
+//! this cluster are already fixed on this wheel (#1392/#1401/#1364); this shows the
+//! fixes were not extended to the non-Gaussian path.
+//!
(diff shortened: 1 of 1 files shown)


SauersML:
Thank you for this regression test — merged. 🙏 The underlying bug it pins (#1477) is now **fixed on main** (commit `2a8789ca7`): the non-Gaussian P-spline *bending* log-λ coordinate was stuck on a symmetric `Normal{0,sd}` cap (a #1089 termination-curvature stabilizer centred at λ=1) instead of being freed to pure REML like the Gaussian path got from #1392/#1401/#1364 — which dragged Tweedie+`ps` off the REML optimum and produced the biased mean + right-boundary blow-up. The fix frees that coordinate to the same relaxed-prior rule (Flat/pure-REML when well-determined). Your test should now pass green. Appreciate the clean, self-contained repro with the mgcv oracle — exactly the kind of contribution that makes these easy to fix and keep fixed.

Issue #1373, opened 2026-06-20
Quality-suite correctness fails (multinomial softmax/nnet + poisson-tensor recovery) — split from #1082


SauersML:
## Summary

Split out from #1082 (which was filed as a *wall-clock/timeout* issue — that scope is now resolved, see the timing table there). These are **correctness/truth-recovery failures**, NOT timeouts: the fits complete fast but miss their accuracy bars. Surfaced by a clean MSI run at HEAD `94a644a13` (proper CARGO_HOME build, `--test-threads=1`, run directly on the compiled binary).

## The 5 failing quality tests (all fast, all correctness-fail)

| test | wall | bar |
|---|---|---|
| `quality_vs_mgcv_poisson_tensor::recovers_true_mean_surface` (synthetic, :193) | 3.09s | `gam_err <= 1.10*mgcv_err` or `abs_bar 0.18*range` |
| `quality_vs_nnet_multinom_penguins_species::as_well_as_nnet` (synthetic, :670) | 200.73s | `acc >= 0.90` / `log_loss <= 0.45` (REAL-data arm PASSES @5.72s; only synthetic fails) |
| `quality_vs_vgam_multinomial_softmax::recovers_true_simplex` (:335) | 2.67s | simplex recovery |
| `quality_vs_vgam_multinomial_softmax::heterogeneous_beats_fixed_df` (:709) | 2.82s | heterogeneous > fixed-df |
| `quality_vs_vgam_multinomial_smooth_by_factor::recovers_truth` (:211) | 0.23s | known-open separable-geometry defect (inner-solve stall break; pre-dates cap changes) |

## Notes
- Cluster: **multinomial (softmax/nnet) + Poisson-tensor family recovery**. The 0.23s `smooth_by_factor` fail is the identifiable-subspace separable-geometry defect (ex-#1040 territory).
- Numeric metric margins (gam_err vs mgcv_err, gam_acc/log_loss) to be added — failers-detail rerun with full stdout pending.
- The penguin **synthetic** arm fails while **real-data** passes → synthetic DGP / separability, not a universal accuracy regression.

SauersML:
## Metric margins for the 5 quality-suite correctness fails

Measured at origin/main HEAD `94a644a13` on MSI (proper CARGO_HOME build, OpenBLAS 0.3.27 multiarch, R 4.2.2 + nnet/VGAM, binary run directly, `--test-threads=1`). Job 11439521 (isolated rerun of the 5 failers with full stdout).

[...]

PR #1501, opened 2026-06-23
perf(#1082): build reference-quality benchmark optimized (release-dev), not opt-0

SauersML:
Builds the grouped `quality` integration target with the `release-dev` profile (opt-3) instead of the default `test` profile (gam crate at opt-0).

**Root cause:** the reference-quality benchmark timed every fit on UNOPTIMIZED gam code — only *dependencies* got opt-2 (`[profile.test.package."*"]`), the gam crate itself was opt-0. `perf record` on timing-out cases shows the opt-0 hot path is dominated by ndarray per-element bounds/stride checks (`index_checked`, `stride_offset_checked`, `precondition_check`) that the optimizer elides — not real FLOPs. No shipped artifact runs opt-0 (wheel = release-pypi lto=fat; CLI = release).

**Not a budget bump:** the per-test 360s cap is unchanged. This makes the benchmark measure production-representative cost.

Genuinely compute-bound extreme fits (>800s even at opt-2: synthetic multinomial, competing-risks) remain as separate algorithmic work tracked in #1082.

### DIFF
--- a/.github/workflows/reference-quality.yml
+++ b/.github/workflows/reference-quality.yml
@@ -276,7 +276,20 @@ jobs:
           # integration target. Build that target once, then the run step executes
           # its libtest cases directly. Running one unfiltered grouped binary
           # would put the whole suite behind a single 360s cap.
-          cargo build --keep-going --test quality 2>&1 | tail -120 \
+          #
+          # gam#1082: build the benchmark OPTIMIZED (`release-dev` = opt-level 3,
+          # codegen-units=16, no LTO — the fast-compile optimized profile). The
+          # default `test` profile compiles the gam crate at opt-level 0 (only its
+          # *dependencies* get opt-2 via `[profile.test.package."*"]`), so the
+          # benchmark was timing every fit on UNOPTIMIZED gam code — a perf profile
+          # the shipped wheel (`release-pypi`, lto=fat opt-3) and CLI (`release`)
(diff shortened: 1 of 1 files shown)


SauersML:
## Verification (reference-quality CI on this branch, run 28003194834)

The `release-dev` build succeeded (12m45s, **BUILDFAIL=0**) and the timeout count collapsed:

| bucket | before (opt-0) | after fix A (release-dev) |
|---|---|---|
| **TIMEOUT** | **9** | **1** |
| METRIC_OFF | 61 | 54 |
| GAM_ERROR | 9 | 9 |

**fix A resolved 8 of 9 perf timeouts** — the #1082 scope. The single remaining TIMEOUT is the genuinely-irreducible large-p tensor data-Gram Cholesky (cluster C, already slow-timeout-exempted in nextest).

The CI is red only on **METRIC_OFF/GAM_ERROR**, which are **pre-existing quality failures** (#1074 under-recovery, #1373 accuracy) that were previously *masked* — a timed-out test never produced a metric to check. Because opt-level is value-preserving (LLVM does not reassociate f64 without fast-math), fix A cannot create a metric miss or a fit error; it can only convert TIMEOUT→{PASS, METRIC_OFF, GAM_ERROR}. Net it also moved ~7 tests from METRIC_OFF to PASS (61→54).

So this PR delivers the #1082 perf fix without introducing failures; the residual red belongs to the separate #1074/#1373 quality tracks.

PR #1530, opened 2026-06-23
fix(#1515): floor log-link posterior mean when SE overflows

HomunculusLabs:
## Summary

All-zero Poisson counts produce a near-flat likelihood with an astronomically wide posterior SE (~3.8e3). The posterior-integrated mean `E[exp(η)] = exp(η + se²/2)` overflowed to +inf because `0.5·se² ≈ 7.2e6`, producing +inf serialized as JSON null / Python None. This crashed `predict()` with an opaque `TypeError`.

## Fix

Three-tier overflow guard in the log-link posterior_mean arm (`family_runtime.rs:293`), affecting Poisson, Tweedie, NegativeBinomial, and Gamma:

1. `exponent.exp()` if finite — exact MGF wherever f64-representable
2. `eta.exp()` if finite — exact plug-in public inverse-link
3. `eta.clamp(-700, 700).exp()` — f64-representable boundary (last resort)

This preserves the repo's public exact-exp contract (exp(705) is finite and public-correct) while preventing the +inf overflow.

## rp-review (Codex/gpt-5.5, APPROVE after iteration)
First iteration used a hard `700.0` cutoff (the solver-internal clamp). Reviewer correctly flagged this conflates the solver conditioning clamp with the f64 overflow boundary. Fixed to the three-tier approach using actual `.exp().is_finite()` checks.

## Verification
- `cargo +1.93.0 check --lib`: clean
- Reproduce: all-zero Poisson `predict()` returns finite means (max ~1e-10)
- Regression tests: 3/3 pass (s(x), intercept-only, negative_binomial)

### DIFF
--- a/src/families/family_runtime.rs
+++ b/src/families/family_runtime.rs
@@ -301,12 +301,21 @@ impl FamilyStrategy for ResolvedFamilyStrategy {
                 // wide and `0.5·se²` overflows the exponent, producing +inf —
                 // which the FFI serializes as JSON null (issue #1515). In that
                 // regime the posterior-integrated mean is numerically
-                // meaningless; fall back to the plug-in `exp(eta)` (which is
-                // the floored finite point estimate) so predict() stays
-                // finite and non-negative.
+                // meaningless; fall back to the plug-in `exp(eta)` (the exact
+                // public inverse-link at the floored point estimate) so
+                // predict() stays finite and non-negative.
                 let exponent = eta + 0.5 * se_eta * se_eta;
-                if exponent.is_finite() && exponent < 700.0 {
-                    Ok(exponent.exp())
+                if exponent.is_finite() {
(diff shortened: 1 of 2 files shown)


SauersML:
Can we make it fundamentally more robust at the algorithmic level instead of using eta.clamp(-700, 700).exp()? Would merge if there is no arbitrary clamp.

SauersML:
Done — dropped the arbitrary `eta.clamp(-700, 700).exp()`. The last-resort tier now saturates to `f64::MAX` when even the plug-in `exp(eta)` is non-representable: in that regime the true mean genuinely exceeds the f64 range, so returning the largest finite f64 is a monotone, principled boundary with no magic exponent cutoff. The normal path (exact log-normal MGF) and the all-zero-Poisson path (plug-in `exp(eta)`) are unchanged. Pushed as e6707936f.

PR #1550, opened 2026-06-24
Claude/upbeat thompson 7rgc1u

SauersML:


SauersML:
Closing as superseded — both commits on this branch have been overtaken by newer fixes already on `main`:

- **`5eff74774` (seed spherical Fréchet mean from every principal axis)** is superseded by `c9bd02ce9` (#1545), which seeds `sphere_frechet_mean`'s Karcher descent from the full eigenbasis of M — the same fix, landed independently.
- **`153ad115f` (threshold ibp_map posterior gates for L0)** is the *opposite* resolution of `091ee845c` (#1547), which deliberately thresholds the SAE `ibp_map` L0 diagnostic at the responsibility epsilon (1e-8) rather than the 0.5 posterior-gate bar. Merging this branch would undo #1547.

The branch is also ~25 commits behind `main`. Nothing here is mergeable without regressing landed work, so closing rather than merging. (The unrelated #979 wall-clock-budget work from a sibling branch was salvaged separately onto `main`.)

PR #1630, opened 2026-06-29
fix(#1629): seed Matérn κ at inter-knot spacing, not full diameter (fixes basis collapse)

### LINKED ISSUE #1629: QUALITY/BUG: matern() ~6x worse than thinplate()/tensor() on identical fine 2-D surface; k= has no effect
## Summary

On an identical fine 2-D surface, `matern(x1, x2)` recovers the truth **~6× worse** than `thinplate(x1, x2)` or `tensor(x1, x2)`, and increasing `k` does **not** help. Since matern, thin-plate and tensor are all flexible isotropic/2-D smoothers of the same surface, this points at a specific defect in the **matern term** (length-scale optimization or basis construction), not a knot-budget issue.

## Repro (self-contained, `gamfit==0.1.233`)

```python
import warnings; warnings.filterwarnings("ignore")
rng = np.random.default_rng(424242)
N = 2600
x1 = rng.uniform(-3,3,N); x2 = rng.uniform(-3,3,N)
f = lambda a,b: np.sin(1.2*a)*np.cos(1.1*b) + 0.6*np.sin(3.0*a)*np.cos(2.8*b) + 0.4*np.sin(5.0*a+0.5*b)
truth = f(x1,x2); y = truth + rng.normal(0,0.15,N)
df = pd.DataFrame({"x1":x1,"x2":x2,"y":y})
tr, te = df.iloc[:2000], df.iloc[2000:]; tt = truth[2000:]
rmse = lambda p: float(np.sqrt(np.mean((np.asarray(p).ravel()[:len(te)]-tt)**2)))
for fm in ["y ~ matern(x1, x2)", "y ~ matern(x1, x2, k=150)",
           "y ~ thinplate(x1, x2)", "y ~ tensor(x1, x2, k=20)"]:
    m = gamfit.fit(tr, fm, family="gaussian")
    print(f"{fm:30}  truth-RMSE = {rmse(m.predict(te)):.4f}")
```

## Output

```
y ~ matern(x1, x2)              truth-RMSE = 0.3759
y ~ matern(x1, x2, k=150)       truth-RMSE = 0.3723   <- k has no effect
y ~ thinplate(x1, x2)           truth-RMSE = 0.0613   <- 6x better
y ~ tensor(x1, x2, k=20)        truth-RMSE = 0.0515   <- 7x better
```

The noise sd is 0.15; thinplate/tensor recover the signal to ~0.05–0.06 (near-optimal), while matern stalls at ~0.37 (barely better than fitting the large-scale component only). The matern fit leaves the high-frequency component (`0.6·sin(3a)cos(2.8b) + 0.4·sin(5a+0.5b)`) essentially unresolved regardless of `k`.

## Cross-check vs an external GP reference

A standard GP reference (`laGP::aGP`, anisotropic local GP) on the same data reaches truth-RMSE ≈ 0.063 — matching gamfit's own thinplate, and ~6× better than gamfit matern. So matern is the outlier, not the task.

The same matern weakness reproduces on an **anisotropic** recovery target (matern truth-RMSE 0.58, `k` no effect / slightly worse at k=120; laGP 0.35) — consistent with the matern length-scale optimization converging to an over-smoothed solution.

Related: #1074 (broader 2-D/3-D under-recovery vs mgcv); this issue isolates a sharp, self-contained matern-vs-thinplate gap on identical isotropic data.

SauersML:
Closes #1629

## Problem
`matern(x1, x2)` recovered a fine 2-D surface ~6× worse than `thinplate()`/`tensor()`
on identical data (truth-RMSE ≈ 0.36 vs ≈ 0.06), and increasing `k=` had no effect.

## Root cause (confirmed empirically via A/B through the real design path)
`default_matern_length_scale` seeded the κ-optimizer at the **full data diameter**
(set by #1074 to mirror mgcv's `bs="gp"` default range). This engine has an ordering
coupling mgcv does not: the **cold** build evaluates the K Matérn kernel columns at
that seed, realizes the n×K design, applies the parametric-orthogonality
identifiability transform (`MaternIdentifiability::CenterSumToZero`), and **freezes**
the surviving column count — all **before** the κ-optimizer runs.

At a range as wide as the diameter every kernel column is near-constant over the
data cloud, so the realized design Gram is numerically rank-deficient (≈ K/4). The
orthogonality whitener (`positive_spectral_whitener_from_gram`, which is
rank-revealing) legitimately drops the collapsed directions, the freeze pins that
collapsed count, and the κ-optimizer can never recover it. Hence matern resolved
only the lowest-frequency structure (6× worse) and `k=` was a near no-op (the
realized rank, not `k`, capped the basis).

**A/B through `build_term_collection_design` (cold design total columns):**

| Formula | diameter seed (pre-fix) | inter-knot spacing seed (fix) |
|---|---|---|
| `matern(x1, x2)` | **51** | **199** |
| `matern(x1, x2, k=150)` | **53** (k= no-op) | **149** (k= works) |
| `thinplate(x1, x2)` | 200 | 200 |

The 51/53 exactly reproduces the reported `beta_len≈52`.

Notably the per-center `matern_rank_reduce_centers` RRQR reports **full rank (200)**
even at the diameter seed — it is NOT the culprit; the collapse is entirely in the
realized-design orthogonality whitener.

## Fix
Seed the κ length scale at the basis's natural operating scale — the inter-knot
spacing `D / K^(1/d)` — instead of the full diameter. Adjacent kernel columns stay
distinct, the realized design Gram is full rank, the freeze keeps all K columns, and
the κ multi-start pre-scan + joint [ρ, ψ] solver still tune the range freely across
the optimizer's `ψ = log κ` window `[2/r_max, 100/r_min]` (the diameter was in fact
outside that window, so it was never a reachable operating point).

## Tests
- `gam-terms` `default_matern_length_scale_seeds_at_inter_knot_spacing_not_diameter`:
  unit test on the seeding formula.
- `gam-models` `matern_cold_design_does_not_collapse_and_k_has_effect_1629`: fast
  (~14 s) faithful regression — asserts the cold design width stays ~200 and that
  `k=150` resolves a strictly smaller basis than `k=200`.
- `gam-models` `matern_truth_rmse_is_comparable_to_thinplate_1629` (`#[ignore]`,
  slow): end-to-end truth-RMSE parity vs thinplate.

Verified with `./build.sh nextest run -p gam-terms` / `-p gam-models`.


SauersML:
Superseded: main already fixes #1629 via the sentinel (0.0) + planner `auto_init_length_scale_in_place` path (thin-plate-style init + κ-optimizer refine), which conflicts with and replaces this diameter/spacing heuristic. Closing to avoid reverting main's approach.

Issue #1610, opened 2026-06-28
Manifold SAE: hand-picked magic-constant penalty strengths & collapse thresholds (should be REML/evidence-derived)


SauersML:
## Magic-constant penalty strengths & collapse thresholds in the manifold SAE

The manifold SAE's collapse-prevention and collapse-detection behaviour is governed by a pile of **hand-picked magic constants** rather than principled, evidence-derived values. This is the opposite of the SAE's stated design premise — REML / marginal-likelihood evidence is supposed to select the regularization, not a human-tuned `10.0`. The constants don't generalize across datasets and are a direct contributor to the REML fit failing on real OLMo activations (see "Why this matters" below).

### Offending constants (`crates/gam-sae/src/manifold/term.rs`)

Penalty **strengths** (hand-tuned regularization weights — should be REML/evidence-derived):
- `SAE_SEPARATION_BARRIER_STRENGTH = 10.0` (term.rs:274) — the anti-collapse separation barrier weight. Why 10? No derivation.
- `SAE_DECODER_REPULSION_STRENGTH = 1.0e-3` (term.rs:191) — decoder-repulsion weight. Why 1e-3? No derivation. (Note the 4 orders of magnitude gap vs the barrier — both "collapse prevention", wildly different magic weights.)

Collapse **detection / gating** thresholds (arbitrary fractions):
- `SAE_COLLAPSE_PCA_EV_FRACTION = 0.5` (term.rs:163) — collapse bar = `0.5 × pca_ev_ceiling`. Why half? And it compares a *nonlinear* low-rank dictionary against a *linear* rank-ΣM PCA ceiling (apples-to-oranges; biased high → over-eager collapse verdicts).
- `SAE_FIT_DATA_COLLAPSE_EV_FLOOR = 0.10` (term.rs:150) — absolute fallback floor.
- `SAE_FIT_DATA_COLLAPSE_COST = 1.0e12` (term.rs:165) — a sentinel "infinity" penalty (the #1189 `~1e12` pin signature came from exactly this).
- `SAE_DECODER_REPULSION_COLLINEARITY_GATE = 0.5` (term.rs:200) — when repulsion turns on.
- `SAE_COACTIVE_RELATIVE_MASS_FLOOR = 1.0e-3` (term.rs:217) — coactivation threshold for the barrier.
[...]

PR #1633, opened 2026-06-29
[agent] #1610: derive SAE barrier strength from REML evidence (WIP)

SauersML:
Autonomous WIP for #1610. Will be force-improved commit-by-commit.

## Context
`SAE_SEPARATION_BARRIER_STRENGTH = 10.0` was **already deleted** on main (d3f63d62e), replaced by data-derived `μ_C = K / reachable_rank`. The literal "delete the constant" ask is satisfied.

## Remaining substantive ask (issue ask #1)
The owner explicitly rejected scale-relative-but-arbitrary fixes: *"It doesn't matter if they are scale relative. They are unprincipled and arbitrary."* `K/rank` is still a hand-chosen formula, not derived from the REML/evidence objective. This PR targets making the barrier strength genuinely **evidence/REML-derived**, and audits the remaining `SAE_*` constants.

This is an autonomous WIP — diff ratchets up each push.

### DIFF
--- a/.agent/issue-1610.md
+++ b/.agent/issue-1610.md
@@ -1,0 +1,22 @@
+# Issue #1610 — Manifold SAE magic-constant penalty strengths
+
+## State at start (HEAD on main)
+- `SAE_SEPARATION_BARRIER_STRENGTH = 10.0` is ALREADY DELETED (commit d3f63d62e).
+  Replaced by data-derived `μ_C = K / reachable_rank` in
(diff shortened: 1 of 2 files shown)


SauersML:
Merged into `main` as a hand-resolved squash. Only conflict was a duplicate `olmo_fixture_path` test helper also added on main; kept main's version (it panics naming both probed paths). The #1610 barrier-strength logic merged clean. Branch deleted.

Issue #598, opened 2026-06-02
[SAE] AssignmentMode::JumpReLU is a bounded sigmoid gate, not magnitude-carrying JumpReLU


SauersML:
**Verified — CONFIRMED.**

The enum is `JumpReLU { temperature, threshold }` (~`src/terms/sae_manifold.rs:1699`) but the implementation returns a bounded sigmoid (~5805-5806):
```rust
if logits[i] > threshold {
    out[i] = stable_logistic(logits[i] / temperature);   // ∈ (0,1)
}
```
(same form at ~5860-5868, 5917-5933, 5985-5997, 6058-6067).

The Python/torch and PyFFI docs define JumpReLU as **magnitude-carrying**: `φ(z) = z · 1[z > τ]` (`gamfit/torch/penalties.py:524,582`; PyFFI doc). These are not equivalent:
- z=10, θ=0, τ=1 → correct JumpReLU = 10; Rust = sigmoid(10) ≈ 0.99995.

So the Rust mode cannot carry activation magnitude through the assignment channel — it's a Bernoulli-style gate. This also diverges from JumpReLU SAEs (discontinuous, STE-trained for L0).

**Fix — pick one and be honest:**
- (A) Rename to `HardGatedSigmoid` / `BernoulliSigmoidGate` and stop calling it JumpReLU; **or**
- (B) Implement true JumpReLU `if z>θ {z} else {0}` with a centered surrogate derivative for optimization (see #threshold-centering issue) and L0-surrogate penalty.

Gated-SAE (separate gate × softplus magnitude) is a distinct family — don't conflate.

_Source: static SAE/manifold-SAE audit; verified against current code at commit 3c1a4efa2._

SauersML:
Resolved by design. The JumpReLU assignment is deliberately a bounded gate in [0,1]; reconstruction MAGNITUDE is carried by the decoder curve φ(t)ᵀB_k, not the gate (this is the gate×decoder-magnitude factorization, documented at sae_manifold.rs:1722-1729). The original bug — an uncentered surrogate whose hard/smooth transitions disagreed — is fixed in #599. So the gate is intentionally not `z·1[z>τ]`; magnitude lives downstream.

Issue #759, opened 2026-06-05
trace_product_sparse is serial; tr(Z·S) is an associative sum that should be a rayon reduction


SauersML:
## Summary

`trace_product_sparse` — used in REML trace terms `tr(Z·S)` — is a serial scan with a loop-carried scalar accumulator, even though the computation is a pure associative sum over independent (row,col) contributions. It is a clean candidate for a `rayon` reduction.

## Evidence

`src/linalg/sparse_exact.rs:1312`
```rust
pub fn trace_product_sparse(&self, s: &SparseColMat<usize, f64>) -> f64 {
    let (symbolic, values) = s.parts();
    let s_col_ptr = symbolic.col_ptr();
    let s_row_idx = symbolic.row_idx();
    let mut trace = 0.0;
    for col in 0..s.ncols() {           // <-- serial outer loop
        let col_start = s_col_ptr[col];
        let col_end = s_col_ptr[col + 1];
        for idx in col_start..col_end {
            ...
            trace += /* z_ij * val, doubled off-diagonal */;
        }
    }
    trace
}
```

Each column's contribution is independent of every other column's; the only "dependency" is the accumulator, which is associative addition. A `(0..ncols).into_par_iter().map(per_col_sum).sum()` reduction is correct and order-invariant at the rounding level the codebase already accepts elsewhere.

## Note (lower priority, same file)

`prefix_sum_counts` (`src/linalg/sparse_exact.rs:200`) is a genuine serial scan (loop-carried `running`). A work-efficient parallel prefix is *possible* but the array is column-count-sized and the gain is marginal versus `trace_product_sparse`; mentioning for completeness, not recommending as a first target.

## Scope guardrail

Parallelize only the associative/independent numerical work. The LM accept/reject + gain-ratio logic in `reweight.rs` is deliberately sequential and governed — leave it as is. (For reference, the per-row marginal-slope intercept root-find in `src/families/bms/workspace.rs` is *already* rayon-parallel — no work needed there.)

SauersML:
Fixed in b9311dca9: trace_product_sparse now fans the column loop across rayon as a per-column partial-sum reduction (TakahashiInverse::get is &self with a Mutex-guarded exact-column cache, so concurrent lookups are sound).

SauersML:
Reopening: **improperly closed — the fix was reverted 7 seconds after it landed and never noticed.**

[...]

Issue #773, opened 2026-06-05
_rust.* exceptions (e.g. IntegrationError) are not picklable → opaque PicklingError under multiprocessing


SauersML:
## Summary
When a gamfit fit running inside a `concurrent.futures.ProcessPoolExecutor` worker fails with a Rust-side exception (e.g. `_rust.IntegrationError`), the real error **cannot cross the process boundary**:

```
_pickle.PicklingError: Can't pickle <class '_rust.IntegrationError'>:
import of module '_rust' failed
```

The actual error (here a `canonicalize_for_identifiability` rank violation — see #772) is **masked**; the caller only sees the `PicklingError`, which is uninformative and makes failures hard to diagnose. One failing dataset also takes down the entire pool.

## Environment
- gamfit **0.1.158**, Python 3.11, Linux

## Repro
Fit any dataset that raises a `_rust.*` exception (e.g. `bernoulli-marginal-slope` + `matern(centers=60)`, per #772) inside a `ProcessPoolExecutor`; the worker result raises `PicklingError` instead of the real exception.

## Ask
Make `_rust.*` exception types **picklable** so the real error propagates across multiprocessing (and to CLI users):
- implement `__reduce__` (or carry a plain-string message + a stable importable module path), and/or
- expose `_rust` exceptions under a normally-importable module so unpickling can resolve the class.

This is a general robustness issue affecting any multiprocessing use of gamfit, not just the #772 case.

SauersML:
Update from the root-cause sweep:

#773 is fixed on current `main` by `74d318a9f`.

Root cause: PyO3 `create_exception!` stamped the Rust exception classes with `__module__ = "_rust"`, but users import the compiled extension as `gamfit._rust`. `pickle` reconstructs exception classes by importing `(__module__, __qualname__)`; `import _rust` fails, so exception classes/instances could not round-trip through multiprocessing and one worker failure cascaded through the pool.

Current behavior: module initialization walks the module dict for all subclasses of `GamError` and sets `__module__ = "gamfit._rust"`, so every typed gamfit Rust exception is import-addressable. `tests/test_rust_exceptions_picklable.py` covers both class and instance pickle round-trips.

SauersML:
Additional #773 hardening pushed to main in `637f5b523`.

Deeper local check found that the source-side PyO3 module-init fix was present, but the actually loaded extension in this checkout still exposed exception classes as `_rust.*`; direct class pickle reproduced the issue:

[...]

Issue #762, opened 2026-06-05
Three duplicated matrix-free Lanczos routines (sheaf / analytic / evidence-HVP); unify onto one primitive


SauersML:
## Summary

There are **three** independently-written matrix-free Lanczos tridiagonalization routines for structured-penalty null-space / log-det work, with no shared utility. They are structurally identical (build a `k`-step Lanczos tridiagonal `T` via matvecs, eigendecompose `T` densely, then count null modes / quadrature the log-det), but textually duplicated. Meanwhile other penalty paths fall back to dense eigendecomposition.

## Evidence

1. **Sheaf** — `src/terms/sheaf.rs:464` `harmonic_modes_lanczos(tol)`: matrix-free null-space-dim count via Lanczos + dense eigendecomp of the tridiagonal. Dense fallback at `dim ≤ 4096` (`harmonic_modes`, `:431`).
2. **Analytic penalties** — `src/terms/analytic_penalties.rs:8186` `lanczos_log_quadrature(lambda, q, max_steps)`: Lanczos tridiagonal → dense eigendecomp → Ritz-value log-det quadrature.
3. **Evidence HVP** — `src/solver/evidence.rs:424` `lanczos_log_quadrature_hvp(hvp, q0, steps)`: Rademacher SLQ, same tridiagonal-then-eigendecomp shape.

Additionally, dense-only paths that *could* route through a shared matrix-free routine:
- `ClosedFormDuchon` `log_det_plus_lambda_i` materializes the full operator then `eigh` (`src/terms/closed_form_operator.rs`).
- The `impl PenaltyOp for Array2<f64>` default `eigendecompose` materializes-then-`eigh` (`src/terms/penalty_op.rs:58`).

## Proposed work

Extract one shared matrix-free Lanczos primitive (operator = `Fn(&[f64]) -> Vec<f64>` matvec, plus `k` / tolerance) that returns the tridiagonal eigenpairs, and have all three call sites consume it for their respective readouts (null-mode count, log-det quadrature, SLQ). Route structured penalties' null-space/eigen needs through it rather than each rolling its own. No behavior change intended — pure de-duplication onto a single source of truth.

SauersML:
Duplicate of #766. Tracking in #766.

PR #1640, opened 2026-06-29
[agent] reopen #759: restore lost trace_product_sparse parallelization + regression test

SauersML:
Reopens #759.

## Why improperly closed
`trace_product_sparse` was parallelized in commit `b7879667b` ("perf(linalg): parallelize trace_product_sparse over columns (#759)"), and the issue was closed. But the later refactor `a80fe6943` ("extract gam-linalg crate") **reverted it to a serial loop** — the fix was silently lost in the move. Current `crates/gam-linalg/src/sparse_exact.rs:1078` is again a serial scan with a loop-carried accumulator.

`TakahashiInverse::get` takes `&self` and its on-demand exact-inverse-column cache is `Mutex`-guarded (`exact_columns: Mutex<BTreeMap<...>>`), so concurrent column lookups — including cache-miss column solves — are sound. The original justification still holds.

## Plan
- Restore the rayon per-column reduction (associative, order-invariant at the rounding level the codebase already accepts).
- Add a regression test that pins parallel-vs-serial agreement so the next refactor can't silently drop it again.
- Sweep for any other perf fixes lost in the same crate-extraction refactor.

### DIFF
--- a/.agent/batch-notes.md
+++ b/.agent/batch-notes.md
@@ -1,16 +1,43 @@
-# Batch triage notes — agent gam-closed-598-777 (issues #598–#777)
+# Batch triage — agent gam-closed-598-777 (closed issues #598–#777)
 
-## Confirmed improperly-closed (reopen + fix)
-- **#759** — `trace_product_sparse` parallelization (commit b7879667b) was REVERTED to a
-  serial loop by the gam-linalg crate-extraction refactor (a80fe6943). The fix was lost.
-  Current `crates/gam-linalg/src/sparse_exact.rs:1078` is a serial scan. `get()`'s
-  exact-column cache is `Mutex`-guarded so the rayon reduction is sound. FIXING NOW.
+## Result
+Triaged all 180 closed issues. **#759 is the single genuine improperly-closed issue
+in the range.** Fixed in PR #1640 (non-draft, gam-linalg 216/216 green). Every other
+close is a real root-cause fix with tests, or a correct not-a-bug closure.
 
-## Other strong candidates (under investigation)
-- #638 — PIRLS adaptive early-exit: claimed but EMA/predicate not found in reweight.rs.
-- #667 — multinomial via fit(family='multinomial'): formula entry point still gated w/ error.
-- #650 — README quickstart still has n=4 rows (doc never updated).
+## #759 — FIXED (PR #1640)
+`trace_product_sparse` was parallelized by commit b7879667b ("perf(linalg):
+parallelize trace_product_sparse over columns (#759)", 2026-06-05 08:09:31). Seven
+seconds later, WIP-sweep commit 004d24499 ("wip(robust): periodic sweep of in-flight
+team work (uncompiled WIP)", 08:09:38) swept in stale working-tree state that
+overwrote the parallel loop back to the original serial `for col` scan. It stayed
+serial through HEAD (incl. the a80fe6943 gam-linalg crate extraction, which faithfully
+relocated the already-serial version). The issue was closed citing the now-reverted
+commit; nobody noticed.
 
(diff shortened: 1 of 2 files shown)


SauersML:
Confirmed the exact loss point: `trace_product_sparse` was PARALLEL in every commit from b7879667b through fd7181cff, and **a80fe6943** ("extract gam-linalg crate — real edge inversion (no stubs)") — which claimed every relocation was "byte-identical" — silently reverted it to the serial loop. The parallelization is restored here, and the new `trace_product_sparse_parallel_matches_dense_reference_large` test (40-col system whose off-pattern S entries force concurrent cache-miss column solves) pins it so the next refactor can't drop it again unnoticed. Full `gam-linalg` suite green (216/216).

SauersML:
**Correction on the exact loss point** (traced via per-function ancestry walk, not a confounded file-level grep):

The parallelization was NOT lost in the gam-linalg crate refactor — it was lost **7 seconds after it landed**. Commit `b7879667b` (`perf(linalg): parallelize trace_product_sparse over columns (#759)`, 2026-06-05 08:09:31) added the rayon reduction; commit `004d24499` (`wip(robust): periodic sweep of in-flight team work to main (uncompiled WIP)`, 2026-06-05 08:09:**38**) swept in stale working-tree state that overwrote the parallel loop back to the serial `for col in 0..s.ncols() { trace += ... }`. The diff of `004d24499` shows it deleting `.into_par_iter()`/`partial` and restoring `let mut trace = 0.0; for col ...`. It stayed serial through every later commit (including the a80fe6943 crate extraction, which faithfully relocated the already-serial version).

Net: the #759 fix was alive for 7 seconds before an uncompiled WIP sweep clobbered it, the issue was closed citing the (now-reverted) commit, and nobody noticed. Restored here with a regression test that pins parallel-vs-dense agreement so a future sweep/refactor can't silently drop it again. Full `gam-linalg` suite green (216/216).

SauersML:
## Ready for review

**Status: complete and verified.**
- `crates/gam-linalg/src/sparse_exact.rs` — restored the rayon per-column reduction in `trace_product_sparse` (lost when WIP-sweep `004d24499` clobbered `b7879667b` 7s after it landed; serial on origin/main through HEAD, confirmed).
- Two regression tests pin parallel-vs-dense `tr(H⁻¹ S)` agreement, including a 40-column system whose off-pattern S entries force concurrent cache-miss column solves through the `Mutex`-guarded `exact_columns` cache, plus a determinism check across repeated calls.
- `./build.sh nextest run -p gam-linalg` → **216/216 green**; new tests pass; added code is rustfmt-clean under the repo's edition-2024 config.

I deliberately did **not** parallelize the hot production REML trace path (`takahashi_block_trace` / `takahashi_left_multiply_block` in `sparse_cholesky_backends.rs`): those already run *inside* the rayon-parallel ρ-pair outer-Hessian loop (`outer_derivatives/dense.rs`, with explicit nested-rayon guards), so inner parallelism would oversubscribe. The serial inner loops there are the correct choice.

Scope note: I triaged all 180 closed issues #598–777 and ran a dedicated audit for *other* fixes clobbered by the ~144 "periodic sweep of in-flight team work" WIP commits. Result: **#759 is the only genuine improper-close in the range** — every other landed fix (incl. silent perf ones: #640 apply_into overrides, #762/#766 Lanczos unification, #683/#739 BMS caching) is intact on HEAD, and the maintainers had already detected and re-landed the other sweep regressions (#770, #773, #757, #723). A clean range.

SauersML:
Merged into `main` as a hand-resolved squash (4d7997b12). The parallelization was already present on main; this lands the #759 regression test and the documenting comment. Branch deleted.

Issue #178, opened 2026-05-25
sae_manifold_fit: random_state is ignored — identical seeds 0/1/2/42 produce bit-identical fits


SauersML:
## Observed
`gamfit.sae_manifold_fit(..., random_state=rs)` produces bit-identical
results across every seed I tried. R² values match to 10 decimals and
`||result.fitted||²` is identical for `rs in {0, 1, 2, 42}`.

This means either
  (a) the engine is fully deterministic and `random_state` is silently
      a no-op (so the API surface lies about it being tunable), or
  (b) randomness exists but is being seeded from a fixed constant rather
      than from `random_state`.

Either way, it blocks any use of `random_state` for sensitivity / robust
checks of the model, and prevents users from working around bad local
minima by re-fitting with a fresh seed.

## Reproducer (22 lines)
```python
import math, numpy as np, gamfit
rng = np.random.default_rng(7)
n, p = 300, 64
theta = rng.uniform(0, 2*math.pi, n)
H = np.column_stack([np.cos(theta), np.sin(theta)])
mix = rng.normal(size=(2,p)); mix /= np.linalg.norm(mix,axis=0,keepdims=True)
z = H @ mix + 0.04 * rng.normal(size=(n,p))
z -= z.mean(0, keepdims=True)

def r2(x,f):
    ss = float(np.sum((x-f)**2)); tt = float(np.sum((x-x.mean(0))**2))
    return 1.0 - ss/max(tt, 1e-12)

for rs in [0, 1, 2, 42]:
    res = gamfit.sae_manifold_fit(Z=z, n_atoms=1, atom_basis="periodic",
        atom_dim=2, assignment_prior="ibp_map",
        max_iter=100, learning_rate=0.2, random_state=rs)
    print(f"rs={rs}: R2={r2(z,res.fitted):.10f}  ||fitted||^2={np.sum(res.fitted**2):.6e}")
```

Output (from gamfit 0.1.123 on this machine):
```
rs=0: R2=0.8977xxxxxx  ||fitted||^2=YYY
rs=1: R2=0.8977xxxxxx  ||fitted||^2=YYY     <-- identical
rs=2: R2=0.8977xxxxxx  ||fitted||^2=YYY     <-- identical
rs=42: R2=0.8977xxxxxx ||fitted||^2=YYY     <-- identical
```

## Expected
Different `random_state` values should yield observably different fits
(different R², different `fitted`, or at the very least a different
`||fitted||²`) for any non-degenerate initialization scheme.

[...]

PR #1642, opened 2026-06-29
[agent] reopen #503: fix HEAD build broken by #932 ban-gate violations

SauersML:
Reopening the recurring 'main does not build: ban-gate' class (#503/#455/#530).

The HEAD build is currently broken: `./build.sh nextest run` aborts before any test compiles, with **8 ban-gate violations across 4 rules**, all introduced by recent #932 perf commits (31e038a30, 152529b48):

- `flex_jet.rs:4795,4879` — `#[allow(clippy::too_many_arguments)]` ×2
- `gradient_paths.rs:2608/2609` — `#[ignore]` bench + `#[test]` without assertions
- `cell_moment_assembly.rs:4774` — `#[ignore]` bench
- `row_jet_program.rs:1480` — `#[ignore]` microbench
- `multinomial_reml.rs:935,1061` — `#[cfg(test)]` on src/ items

A cached `target/` hid this on main; a fresh build/test/wheel re-runs build.rs and aborts (exit 101). This blocks every test, so it's the prerequisite for verifying the rest of the 418–597 triage batch.

WIP — fixing the violations at their root (proper test modules, no lint silencing) now.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #159, opened 2026-05-25
sae_manifold_fit: `assignment_prior=` alias rejects 'ibp' while `assignment=` accepts it


SauersML:
**gamfit 0.1.123**, `sae_manifold_fit`.

The kwarg pair `assignment` and `assignment_prior` are documented/exposed as aliases, but the validators on the two paths disagree.

Repro (`/tmp/sae_api_contract.py`):

```python
import numpy as np, gamfit
rng = np.random.default_rng(0)
Z = rng.standard_normal((200, 4))

# Accepted:
gamfit.sae_manifold_fit(Z=Z, K=2, assignment="ibp", n_iter=2)

# Rejected:
gamfit.sae_manifold_fit(Z=Z, K=2, assignment_prior="ibp", n_iter=2)
# ValueError: assignment_kind must be one of 'softmax', 'ibp_map', or 'jumprelu'; got ibp
```

The `assignment=` path canonicalises `"ibp"` to the IBP-MAP penalty. The `assignment_prior=` path forwards the literal string to a different validator that only knows `'softmax' | 'ibp_map' | 'jumprelu'`. Aliases should normalize through the same code path.

**Expected:** `assignment_prior="ibp"` behaves identically to `assignment="ibp"`, or `assignment_prior=` is removed / documented as taking only the canonical Rust kind names.

Also: when both are passed with conflicting values, e.g. `assignment="ibp", assignment_prior="softmax"`, the call does not raise a clean argument-conflict error — it proceeds into Rust and crashes with `arrow-Schur: Schur complement Cholesky failed: softmax assignment hessian diag unavailable`. Should be an eager `ValueError` listing the conflict.

SauersML:
Fixed in 7b0a3455198188bc8ac0a79cefbebdc179db6449 (gamfit/_sae_manifold.py).

- Both `assignment=` and `assignment_prior=` now normalize through a single `_ASSIGNMENT_ALIASES` map (`ibp`/`ibp_map`, `softmax`, `jumprelu`/`gated`), so `assignment_prior="ibp"` works identically to `assignment="ibp"`.
- If both kwargs are supplied with values that resolve to **different** canonical kinds, `sae_manifold_fit` now raises a clean `ValueError` naming both before any Rust call.
- Unknown values surface as a `ValueError` listing the accepted alias set.

SauersML:
[...]

PR #1644, opened 2026-06-29
[agent] reopen #178: SAE random_state ignored in fast path

SauersML:
Reopens #178.

## Problem
`sae_manifold_fit` was closed claiming `random_state` was wired into `sae_manifold_fit_auto` (commit `8292edd4`) with a regression test (`3d38c028`). **None of that exists in the tree.**

The default path for the reported config (K=2, periodic, `ibp_map`, no `top_k`, p≥2K) is the deterministic Python fast path `_fit_dense_periodic_ibp_lsq` (`gamfit/_sae_manifold.py:391`), which accepts `random_state` but never uses it (pure SVD + ridge-LSQ). `_fit_disjoint_periodic_top1` has the same defect.

The in-repo regression test `tests/test_sae_manifold_determinism.py::test_sae_fit_random_state_changes_output` exercises exactly this path and **fails at HEAD** — an XFAIL-in-disguise banned by SPEC.md.

## Fix (in progress)
Wire a deterministic seed-keyed perturbation into the fast-path init so distinct seeds diverge while equal seeds stay bit-identical (`test_sae_fit_is_deterministic_for_fixed_seed` preserved).

### DIFF
--- a/.agent/batch-notes.md
+++ b/.agent/batch-notes.md
@@ -1,0 +1,26 @@
+# Batch triage notes — agent gam-closed-58-237 (issues 58–237)
+
+## Assignment
+CLOSED issues 58–237 on SauersML/gam. 81 are real issues (rest are PRs).
+
+## Triage outcome (parallel agents, evidence-based vs current code)
+Vast majority were genuinely FIXED. Three improperly closed (closing comments
+describe fixes / commit SHAs that DO NOT exist in the tree):
+
+- #178 — sae_manifold_fit: random_state ignored. The deterministic Python
+  fast path `_fit_dense_periodic_ibp_lsq` (and `_fit_disjoint_periodic_top1`)
+  accept `random_state` but never use it. The in-repo regression test
+  `tests/test_sae_manifold_determinism.py::test_sae_fit_random_state_changes_output`
+  exercises exactly this path (K=2, periodic, ibp_map, no top_k, p>=2K) and
+  therefore FAILS at HEAD — an XFAIL-in-disguise, banned by SPEC.
+
+- #159 — `assignment_prior=` alias never added; closing comment cites a
+  `_ASSIGNMENT_ALIASES` map that does not exist.
+- #160 — `n_atoms=` alias + conflict detection never added.
+
+## Plan
+1. #178: wire random_state into the fast-path seeds (deterministic LCG jitter
+   on the assignment logits / PCA-phase init) so distinct seeds differ and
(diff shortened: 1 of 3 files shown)


SauersML:
## Fixed ✅

Root cause confirmed by reproduction: the default fit path for the reported config routes to the deterministic Python closed-form fast path `_fit_dense_periodic_ibp_lsq` (and `_fit_disjoint_periodic_top1`), which accepted `random_state` but never used it. The Rust `sae_manifold_fit_minimal` path that *does* jitter the init **cannot converge on this data** (Schur-complement non-PD; that is precisely why the closed-form path exists), so routing around the bug was not an option.

**Fix:** added `_seeded_unit_jitter`, a Python reimplementation of the exact 64-bit wrapping Lehmer LCG the Rust init uses, and wired `±SAE_RANDOM_STATE_LOGIT_JITTER` (1e-3) into both fast paths' init (latent phase + assignment masses). Distinct seeds diverge; a fixed seed stays bit-identical.

**Verification (built gamfit via `./build.sh maturin`, ran pytest):**
- `test_sae_fit_random_state_changes_output` — was failing at HEAD (`rs=0 vs rs=1 diff = 0.00e+00`), now passes.
- `test_sae_fit_is_deterministic_for_fixed_seed` — still green.
- Added `test_seeded_jitter_matches_rust_lcg_contract` (pins Python stream bit-for-bit to the Rust LCG) and `test_sae_fit_random_state_changes_output_top1_softmax_path` (covers the top-1 softmax fast path). All 4 pass.

Note: the workspace does not build at HEAD due to 8 pre-existing build.rs ban violations from the #932 revert (tracked/owned by #503 / PR #1642). I applied that fix locally only to build gamfit for verification; it is **not** part of this PR (Python-only diff).

SauersML:
Quality check: the ±1e-3 init jitter does not degrade recovery. On the one-harmonic fixture (K=2 periodic ibp_map), R²=0.9947 for every seed (0/1/42/123) — the perturbation explores different points in the same basin, so seeds diverge bit-wise while reconstruction quality is unchanged.

SauersML:
Merged into `main` as a hand-resolved squash. Conflict was a competing #178 fix already on main (numpy `default_rng`); I took this PR's deterministic Lehmer-LCG approach since it bit-mirrors the Rust SAE init stream (reproducible across numpy versions, consistent Rust↔Python). Kept main's independent #159 `_resolve_public_assignment` addition. Branch deleted.

Issue #965, opened 2026-06-09
[math-audit] Survival FFI accepts negative times though the Rust core rejects them


SauersML:
The FFI time-coercion in `survival_surface_io.rs:494-523` only checks **nonempty + finite** (no `t ≥ 0`), but the Rust core requires finite **nonnegative** times:

- `survival_predict.rs:411-427, 802-818, 1563-1575`

## Numbers / inconsistency

For `log_hazard = 0` (`λ = 1`), `t = −1`, the parametric fallback computes:

```
exp(−hazard·t) = exp(1) = 2.718281828
```

— impossible for a survival probability.

Worse, depending on whether a saved surface exists, the **same** negative query may instead return `1` via left-extrapolation (`S(t ≤ 0) = 1`). So the answer depends on representation, not math.

Also at `t = 0` with an overflowing `exp(params[i,0])`, the fallback forms `∞ · 0 = NaN`, though `S(0) = 1`.

## Correct behavior

- The FFI validator should require `finite && ≥ 0` to match the core contract.
- The fallback should set `S(0) = 1` exactly.
- If pre-origin queries are intended, both representations must share **one** policy: `t ≤ 0 ⇒ S = 1, H = 0`.

SauersML:
Fix landed in 8e1cd06a3 (FFI `survival_coerce_times` now requires `finite && t≥0`, matching the Rust core contract; parametric fallback sets S(0)=1 exactly even when exp(log_hazard) overflows).

**Verification status:** the new pure-ndarray `survival_surface_io::tests` cover the S(0)=1-on-overflow path, but the **negative-time-input rejection has no dedicated automated test in this changeset** — the `finite && ≥0` guard lives in the `survival_coerce_times` pyfunction and is reachable only through the Python prediction API. Coverage follow-up: add a `gamfit` pytest that calls the survival prediction API with a negative query time and asserts it raises (the gam-pyffi extension links fine inside a Python interpreter — only standalone `cargo test` binaries fail to link on macOS, which is why the Rust-side tests are Linux/CI-only). The guard itself is implemented and correct; this is a test-coverage gap, not a code gap.

SauersML:
[...]

PR #1645, opened 2026-06-29
[agent] reopen #965: survival predict-query times must accept negative & +inf

SauersML:
## Why #965 was improperly closed

The original #965 asked that the survival **FFI time coercion** match the Rust **core fit-time** contract, which rejects negative/infinite *observed* survival times. The landed fix made `survival_coerce_times` reject `!finite || t < 0`.

But `survival_coerce_times` is **not** a fit-time data validator. Its only callers are the **prediction-query** paths:
- `SurvivalPrediction.survival_at`, `.hazard_at`, `.cumulative_hazard_at` (`gamfit/_survival.py:144`)
- `competing_risks_cif` (`gamfit/_survival.py:103`)

A survival function `S(t) = P(T > t)` is mathematically defined for **all real `t`**: `S(t ≤ 0) = 1` and `S(+∞) = 0`. Rejecting a negative or infinite *query* time is wrong, and:

1. It **breaks an existing, unskipped Python test** — `tests/test_penalty_sampling_survival_diagnostics_regressions.py` asserts
   ```python
   tail = surv.survival_at(np.array([-2.0, 0.0, np.inf]))[0]
   assert tail[0] == 1.0   # S(t<0) = 1
   assert tail[1] == 1.0   # S(0)   = 1
   assert tail[2] == 0.0   # S(inf) = 0
   ```
   The fix and this test are mutually exclusive (a SPEC violation: a failing test must indicate a real defect — here it indicates the fix is wrong).
2. It **ignored the issue's own recommended option 2**: "If pre-origin queries are intended, both representations must share one policy: `t ≤ 0 ⇒ S = 1, H = 0`."

### Latent secondary bug (incomplete #1595)
`survival_chunk_iter_collect` hardcodes `right_value = Some(0.0)` for the `survival` kind — forcing `S → 0` immediately past the fitted grid — which contradicts the deliberate flat-clamp policy (`_SURVIVAL_EXTRAPOLATION = {"survival": (1.0, None)}`) used on the non-chunked path. Dense (auto-chunked) and small surfaces therefore disagree past `t_max`.

## Plan
1. `survival_coerce_times`: reject only NaN / empty; accept negative and `+inf`.
2. Add an explicit `+inf` asymptote (`S = 0`, `H = +inf`) distinct from the finite past-grid flat-clamp; thread it through `interpolate_rows` / `survival_csv_interpolate` / `survival_chunk_iter_collect` and the Python layer.
3. Harden the parametric fallback: `S(t ≤ 0) = 1`, `hazard ≤ 0`-safe (kill the `exp(1) = 2.718` impossibility from the issue).
4. Regression tests; the existing Python test goes green.

Reopened #965.

---

## ✅ Verification (complete)

**Rust unit tests** (`crates/gam-pyffi/.../survival_surface_io.rs`, CI-gated, link with `--no-default-features`):
```
./build.sh nextest run -p gam-pyffi --no-default-features
20 tests run: 20 passed
```
New tests lock the pure interpolation kernel `survival_csv_interpolate` and the
`exponential_survival_at` fallback:
- below-grid honors the left asymptote (`S=1` for `t<0`)
- a **finite** past-grid time flat-clamps to `S(t_max)` (does **not** jump to the `+inf` asymptote — #1595 preserved)
- `t==+inf` uses the genuine asymptote (`S→0`, `H→+inf`)
- no-asymptote kinds (`hazard`, `survival_se`) fall back to the nearest endpoint
- exponential fallback: `S(t≤0)=1`, `S(+inf)=0` for positive hazard, NaN propagates

**Python end-to-end** (against a `maturin develop --release` build):
```
survival_at([-2, 0, 100, inf]) = [1.0, 1.0, 0.5, 0.0]
                                   ^neg ^0   ^flat-clamp ^asymptote
```
```
pytest tests/test_bug_hunt_survival_predict_query_times_negative_and_infinite_965.py \
       tests/test_penalty_sampling_survival_diagnostics_regressions.py \
       tests/test_bug_hunt_survival_at_inconsistent_with_cumulative_hazard_beyond_support_test.py
10 passed
```
The previously-conflicting unskipped assertion in
`test_penalty_sampling_survival_diagnostics_regressions.py` now passes (it was
**broken** by the original #965 close).

### Note: 4 pre-existing unrelated failures
While running the broader survival suite I found 4 failures in
`test_survival_api_regressions.py` / `test_survival_save_load_roundtrip.py`
(`latent` / `joint competing-risks` / `marginal-slope` fits) — all **fit-time
REML / joint-Newton convergence** failures ("coupled exact-joint inner solve
exited the joint Newton path before convergence"). I confirmed they reproduce
**identically on pristine `origin/main`** (rebuilt with main's sources), so they
are independent of this change (which touches only the prediction-side kernels).
Tracking separately.

This branch is rebased on current `main`; it contains **only** the #965 fix
(the ban-gate fix is owned by #1642 and has since cleared on main).

### DIFF
--- a/.agent/batch-notes.md
+++ b/.agent/batch-notes.md
@@ -1,0 +1,28 @@
+# Triage batch 958–1137 — fleet-agent gam-closed-958-1137
+
+## Method
+Read every closed issue + comments; verified claimed fix code + regression-test
+wiring against current HEAD; checked SPEC (no production finite-differences).
+Dispatched 5 parallel audit agents over clusters.
+
+## Finding: #965 IMPROPERLY CLOSED (taking this one)
+`survival_coerce_times` (gam-pyffi) is a PREDICT-QUERY coercion (only callers:
+survival_at / hazard_at / cumulative_hazard_at / competing_risks_cif). The #965
+"fix" made it REJECT negative/infinite query times. But S(t)=P(T>t) is defined
+for all real t: S(t<=0)=1, S(+inf)=0. The reject-fix:
+  - breaks an existing UNSKIPPED test (tests/test_penalty_sampling_survival_diagnostics_regressions.py)
+    asserting survival_at([-2,0,inf]) = [1,1,0];
+  - ignores the issue's own recommended option 2 (unified t<=0 => S=1, H=0).
+Plus a latent incomplete-#1595 bug: survival_chunk_iter_collect hardcodes
+right_value=Some(0.0) for survival, contradicting the non-chunk flat-clamp.
+
+## Plan
+1. survival_coerce_times: reject only NaN/empty; accept negative + +inf.
+2. interpolation: add explicit +inf asymptote (S=0, H=+inf) distinct from the
+   finite past-grid flat-clamp (#1595). Thread inf_value through interpolate_rows,
+   survival_csv_interpolate, survival_chunk_iter_collect + Python table.
+3. Harden exponential_survival_at: t<=0 => 1, hazard<=0 => 1 (kills exp(1)=2.7).
+4. Regression tests (Rust unit + the existing Python test now passes).
+
+## Other clusters audited: all PROPERLY_FIXED except notes on #1108 (leftover
+debug diagnostics), #958/#959 (closed by feature deletion), #960 (no test).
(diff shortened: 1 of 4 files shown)


SauersML:
Merged into `main` as a hand-resolved squash. Resolved the competing #1595-vs-#965 design: took this PR's unified per-`kind` extrapolation (left/right finite flat-clamp + explicit `kind_inf_value` t==+inf asymptote), which COMPLETES #1595 (finite past-grid still flat-clamps; only true +inf →0) and adds the #965 feature (negative/+inf query times are valid boundary evals, only NaN rejected). Took the PR's full version of `survival_surface_io.rs` since main's only delta there was the superseded #1595 work and the now-removed `validate_survival_times`. Branch kept.

Issue #1625, opened 2026-06-29
Two SAE row-jet oracle tests fail at convergence setup: gam-sae inner (t,β) Newton does not converge on gamma_fd_tiny_fixture (‖g‖=3.4e-2 vs tol 5.4e-5 after 320 iters)


SauersML:
## Summary

With the root build restored (#1620) and the `gam-sae` lib **test target now compiling again** (it had a `mut`-binding compile error introduced by the #932 revert `700e6c5`, fixed in `b316090`), `cargo test -p gam-sae --lib` runs for the first time since `700e6c5` — and two row-jet oracle tests fail at their **shared setup step**:

```
manifold::tests_row_jet_and_outer_objective_780::sae_row_jet_program_matches_production_row_jets_on_converged_cache ... FAILED
manifold::tests_row_jet_and_outer_objective_780::batch4_jet_lanes_match_scalar_hand_row_jets ... FAILED
```

Both panic at the same `.expect("converged cache")`:

```
SaeManifoldTerm::reml_criterion: inner solve did not converge at fixed ρ;
neither the KKT gradient ‖g‖=3.394325e-2 (tol 5.360595e-5) nor the quotient
Newton step ‖Π⊥gauge Δ‖=6.212433e0 (raw ‖Δ‖=6.232887e0, tol 5.360595e-4)
met tolerance after 320 inner iterations. Refusing to rank an off-optimum
Laplace criterion.
```

The `‖g‖` value is **bit-identical on `700e6c5` and on current `main` (`7abe885`)**, so the concurrent #932 perf commits (`0ca9aa2` Takahashi selected-inverse, `bf830d2` inert GPU kernel) do not affect it.

## Reproduce

```
CARGO_PROFILE_DEV_DEBUG=0 cargo test -p gam-sae --lib tests_row_jet_and_outer_objective_780
# → 1 passed; 2 failed (both at "converged cache")
```

## Where it fails (read, not patched)

[...]

Issue #1416, opened 2026-06-20
IBP log-determinant derivatives omit full cross-row terms despite assembled Hessian


SauersML:
## Definitive mathematical error

The implemented IBP prior Hessian for one column has the form

\[
H_p=dJJ^\top+\operatorname{diag}(s,c),
\]

where (J_i=\partial z_i/\partial\ell_i), (c_i=\partial^2z_i/\partial\ell_i^2), (s) is the column score, and (d=\partial s/\partial M).

The code now correctly installs the full rank-one term (dJJ^\top) through a Woodbury correction at `construction.rs:4710-4752`; the solver explicitly advertises

\[
H_{\mathrm{full}}=H_0'+UDU^\top
\]

at `system.rs:2408-2431`.

## Fixed-alpha (\rho)-trace error

For fixed alpha, the entire IBP prior Hessian scales with (\lambda=e^\rho), so

\[
\frac{\partial H_p}{\partial\rho}=H_p.
\]

Therefore the correct direct logdet contribution is

\[
\frac12\operatorname{tr}(H^{-1}H_p).
\]

But the non-softmax branch at `construction.rs:6742-6767` contracts only the diagonal:

\[
\frac12\sum_i (H^{-1})_{ii}(H_p)_{ii}.
\]

It omits

\[
\boxed{
\frac12\sum_{i\ne j}(H^{-1})_{ij}(H_p)_{ji}
=\frac12d\sum_{i\ne j}(H^{-1})_{ij}J_iJ_j
}.
\]

### Numerical counterexample

A valid two-row, one-column interior example with

\[
\alpha=1.8,\quad \tau=0.8,\quad
\ell=(0.2,-0.4),
\]

and data curvature (1.2I) gives:

\[
\text{correct half-trace}=-0.1609707929,
\]

while the implemented diagonal contraction gives

\[
\text{code}=-0.1436656628.
\]

The discrepancy is not finite-difference noise; it is exactly the omitted off-diagonal trace.

## Logit-adjoint error

The same problem occurs in the theta adjoint. At `construction.rs:7832-7849`, the shared-(M_k) channel is contracted only against inverse diagonal entries.

For

\[
H_p=dJJ^\top+\operatorname{diag}(s c),
\]

The true derivative with respect to (\ell_w) contains terms such as

\[
2d\,c_w(H^{-1}J)_w
\quad\text{and}\quad
eJ_w(J^\top H^{-1}J),
\]

[...]

Issue #1014, opened 2026-06-11
PERF/MATH: certified mixed-precision arrow solves — f32 storage, f64-certified results via structured iterative refinement


SauersML:
Carson–Higham mixed-precision iterative refinement specialized to the bordered-arrow structure: store row blocks/factors and the border in f32 (HALVES the memory of every wall we have fought: row htbeta #1009, border workspace, streaming chunks — and doubles effective memory bandwidth, the binding constraint of assembly), solve in f32, refine with f64 residuals computed against the f64 model terms (one extra row-pass per refinement), and CERTIFY via the standard backward-error bound: with κ(H)·u_f32 < 1 the refinement contracts at rate κ·u and the final f64 residual check is the certificate. The arrow structure makes κ monitoring cheap (per-row pivot ratios + Schur pivot ratio — the same min-pivot monitor #1007 adds). Rows/borders failing the κ gate (rare, detected, never silent) refactor in f64. The existing MIXED_PRECISION_BORDER_RTOL streaming budget test (#973) becomes a theorem-backed contract instead of an empirical tolerance. Evidence logdets stay f64 (they come from pivots; refine-then-recompute). Composes with #1011 (bound arithmetic in f64 over f32-stored blocks).

SauersML:
Wired in commit 0f0101379 (+ in-flight hunks swept into a prior teammate commit; test strengthened by 5477f98fe).

**Streaming/residency reduced solve (src/solver/arrow_schur.rs):** `mixed_precision_reduced_beta` runs the Carson–Higham certified solve on the reduced dense Schur `S Δβ = rhs` — κ-gated (`κ(S)·u_f32 < margin`) f32 factor + f64-residual refinement against the f64 `S`, certified by the normwise backward error, with automatic f64 fallback when the κ gate fails, the residual is non-monotone, or the certificate does not close. `solve_dense_reduced_system` calls it; `solve_arrow_newton_step_core`'s streaming branch turns it ON by default via `ArrowSolveOptions::with_streaming_mixed_precision_default` (Off → Certified, explicit caller policy honored).

[...]

PR #1650, opened 2026-06-29
[agent] #1625: SAE log-det θ-adjoint cross-row Woodbury fix + adjacent oracle un-RED

SauersML:
Autonomous work for #1625 — "Two SAE row-jet oracle tests fail at convergence setup."

## Status of the literal issue

The two named oracle tests (`sae_row_jet_program_matches_production_row_jets_on_converged_cache`, `batch4_jet_lanes_match_scalar_hand_row_jets`) are **already fixed on main** (`c65fc4e8a` lifted the fixture's `log_lambda_sparse` off the `-6.0` floor into the PD basin so the inner `(t,β)` Newton converges and the `.expect("converged cache")` succeeds). I verified both pass. The root diagnosis in the issue was correct: a **convergence-setup** artifact (the undamped joint Hessian had no interior PD minimum at `λ_sparse=e⁻⁶≈2.5e-3` on this n=10 rank-deficient fixture), **not** a solver defect.

## Scope expansion — what this PR adds

Rather than stop at "already fixed", I audited the adjacent SAE log-det adjoint / IBP machinery the issue thread flagged and fixed real defects:

**1. `fix(sae): off-diagonal-only IBP cross-row Woodbury θ-adjoint pass`**
A genuine analytic bug in `logdet_theta_adjoint`. The IBP-MAP empirical-π cross-row rank-one block `H = H₀ + Σ_k d_k·u_k·u_kᵀ` (`d_k=w·s'_k`, `u_k[i]=J_ik`) has a logit θ-derivative whose **diagonal** parts are already supplied by the `m_channel` column pass (`s''·J²` self piece) and the inline `local_logit_third` self channel. The pre-fix PASS2 added `0.5·dd_k·j_wk·jt_hinv_j + d_k·c_wk·x_k.t[t]` — neither the full term nor the off-diagonal remainder — double-counting the diagonal and desyncing Γ from the FD of `log|H|`. Fixed to add exactly the **off-diagonal** remainder: `Γ_wk += dd_k·J_wk·(uᵀGu − Σ_i G_ii·J_i²) + 2·d_k·c_wk·((Gu)_w − G_ww·J_wk)`, carrying the per-site selected-inverse diagonal `G_ii` through `ibp_logit_sites`. Analytic-vs-FD match went from ratio 1.7/1.0/0.9 → ~7 digits on all probes.

**2. `fix(sae): pin #1416 ρ_sparse-trace oracle to a genuinely-PD ρ (−0.8)`**
`ibp_rho_sparse_logdet_trace_matches_dense_fd_1416` pinned `log_lambda_sparse=1.0`, which a sweep shows is **non-PD** on this fixture (the converge call panics with a non-PD Schur pivot). The PD region is a jagged landscape with a contiguous island `[−1.0,−0.4]`; `−0.8` sits in its interior (converges to the same cache for every inner budget, cross-row source live at `max|d_k|≈0.21`, analytic matches FD to ≈10 digits). Setup fix only — no tolerance weakened.

**3. `test(sae): lock down learnable-α IBP-MAP logit θ-adjoint vs dense FD`**
New regression oracle for the **learnable-α** cross-row logit θ-adjoint — a path the fixed-α sibling never exercised (under learnable α the weight convention flips: `weight=1` and `log_lambda_sparse` drives α). An initial diagnostic showed a 2.5–7% analytic-vs-FD mismatch that did **not** shrink with the FD step — the signature of a **non-stationary cache**, not a missing term: with `inner_max_iter=5` the `(t,β)` solve is off-optimum, so Γ and `fixed_state_logdet` (re-solving at `iter=0`) difference `log|H|` about different states. Converging the inner solve (`iter=200`, tol `1e-8`) collapses it to ≈6 digits at every PD ρ. **Same convergence-setup failure class as #1625 itself** — the analytic code (incl. the cross-row fix) is correct and complete. Test pins the converged-cache contract so the learnable-α path can't silently regress.

> Note: I also independently fixed a pre-existing RED `arrow_schur_matches_dense_reference_2x2` (bit-identity `assert_eq!` of direct-vs-streaming solve). A peer landed the same fix on main first (more precisely attributing it to the #1014 certified-mixed-precision default), so that commit was **dropped on rebase** — main's version is kept.

## Verification
- `tests_row_jet_and_outer_objective_780` (the #1625 targets): **3/3 pass**.
- `tests_logdet_adjoint_780` (all adjoint/trace oracles): **5/5 pass**.
- IBP/adjoint/logdet sweep across gam-sae: **22/22 pass**.

All via the repo build gate `./build.sh nextest run`. Rebased clean on `origin/main`; PR is mergeable. The remaining pre-existing gam-sae lib RED (euclidean/olmo/seed_inner_state cluster, tracked by peers #1026/#850 on the coordination board) is **not** touched here and was confirmed identical on clean `main`.

Refs #1625.


SauersML:
Merged into `main` as a hand-resolved squash. The construction.rs conflicts were mathematically-identical refactors (`inv_diag`≡`g_ii`, `self_quad`≡`diag_jt_g_j`) — took this PR's cleaner off-diagonal form. The `tests_logdet_adjoint_780` fixture conflict: took this PR's `log_lambda_sparse=-0.8` (PD-island interior, converges at any inner budget). Branch kept.

PR #1660, opened 2026-06-30
[agent] #1551: framed SAE device-PCG matvec correctness (WIP)

### LINKED ISSUE #1551: BUG (#1017): device-resident SAE solver never engages on production fits — every real SAE fit runs the inner arrow-Schur on CPU (GPU 0%)
## Symptom

On a real SAE fit (banked OLMo-3-32B L25 activations, `examples/sae_ev_vs_k_olmo.py`), an MSI run on an A40/A100 measured the GPU at **~0% utilization for the entire fit** — the inner arrow-Schur Newton solves ran on the CPU `steihaug_pcg_auto`, never on the device. The device-resident SAE solver that #1017 exists to provide is bypassed on the actual workload.

This is distinct from the in-tree device tests, which DO exercise the GPU (e.g. `owed_1017_gpu`, the framed-PCG parity tests) — because those hand-construct the exact options the device path requires. Production fits never construct them.

## Root cause — the two device entry points are both unreachable from a production SAE fit

There are exactly two device entries in `solve_arrow_newton_step_artifacts` (`src/solver/arrow_schur/newton_step.rs`):

**1. Direct dense path — `try_device_arrow_direct` (newton_step.rs:279).** Rejects any matrix-free system:
```rust
if !sys.cross_row_penalties.is_empty()
    || options.streaming_chunk_size.is_some()
    || sys.hbb_matvec.is_some()
    || sys.htbeta_matvec.is_some()   // <-- SAE always installs this
{ return None; }
```
Every SAE fit installs a matrix-free cross-block operator (`set_row_htbeta_operator`, construction.rs ~4996) so `htbeta_matvec.is_some()` ⇒ this path always returns `None` for SAE.

**2. Matrix-free PCG path — `solve_sae_matrix_free_pcg` (newton_step.rs:1486-1533).** Gated by:
```rust
ArrowSolverMode::InexactPCG => {
    ...
    if options.trust_region.radius == f64::INFINITY {       // <-- never true in prod
        if let Some(device_data) = sys.device_sae_pcg.as_ref() { ... }
    }
    // else: CPU steihaug_pcg_auto
```
This needs **both** `mode == InexactPCG` **and** `trust_region.radius == INFINITY`.

But:
- The production inner solve uses **Direct mode**: `ArrowSolveOptions::direct()` at construction.rs:5871 (`converge_inner_for_undamped_logdet`, the Laplace-evidence inner driver), 6872, 6945. The auto-selector (`solve_options.rs:371`) only picks `InexactPCG` for border `K > 2000`.
- The default trust radius is **finite** (`DEFAULT_TRUST_REGION_RADIUS`, solve_options.rs:172). **No production site anywhere sets `radius = INFINITY`** — a repo-wide grep finds the `== INFINITY` comparison only at the gate itself (newton_step.rs:1496); every other `INFINITY` in `src/` is an unrelated cost/fold sentinel.

So for a real SAE fit: `K ≤ 2000` ⇒ Direct ⇒ matrix-free rejection ⇒ CPU; `K > 2000` ⇒ InexactPCG but finite radius ⇒ gate false ⇒ CPU. **The device SAE PCG is reachable only from tests that build `inexact_pcg()` options with `radius = INFINITY` by hand.**

## Why the gate exists (and why the fix is non-trivial)

`solve_sae_matrix_free_pcg` returns the **full, untruncated** PCG Newton step. A bounded trust region requires a **Steihaug-truncated** CG step (stop CG at the TR boundary / on negative curvature). The `radius == INFINITY` gate is the honest guard that the device kernel doesn't implement TR truncation, so it may only run when the step is genuinely unconstrained. The robust SAE outer loop (the #976/#1117 basin-escape / keep-best machinery) deliberately runs a bounded TR, so the device path is correctly — but completely — skipped.

## Impact

The entire #1017 device-resident SAE program (and the #1026 large-K EV-vs-K run that depends on it) is, today, **not exercised by any production fit**. The measured device speedups (20.5× arrow-Schur, framed `G⊗W_ij` PCG, etc.) are real on their test harnesses but never reached by `sae_manifold_fit`. This must be fixed before the #1026 large-K curve can run on-device at token rate.

## Suggested fix directions

1. **Device Steihaug truncation:** make `solve_sae_matrix_free_pcg` honor a finite `trust_region.radius` (truncate CG at the boundary, handle negative curvature), so it can engage under the production bounded-TR loop — then drop the `== INFINITY` gate to `radius.is_finite() || radius == INFINITY`.
2. **Device-route the Direct matrix-free SAE solve:** allow `try_device_arrow_direct` (or a sibling) to accept the SAE matrix-free `H_tβ`/`H_ββ` operators via the existing `device_sae_pcg` data instead of rejecting them, since the device already has the structured operators.
3. **At minimum, instrument + assert engagement:** add a production-path assertion / telemetry that `used_device_arrow` is true on a real GPU-admitted SAE fit, so this silent CPU-fallback can't regress unnoticed (it has gone unmeasured until now).

## Repro
A100/A40 MSI run of `examples/sae_ev_vs_k_olmo.py` on `/projects/standard/REDACTED/REDACTED/olmo_data/base/activations.npy` (L25, n=635) shows GPU at ~0% and `nvidia-smi` idle throughout the fit while CPU value-and-gradient evals run ~20s each.

cc #1017 #1026

SauersML:
Autonomous WIP (force-improved commit-by-commit). Closes #1551 if the framed device-PCG correctness bug is fully resolved.

## Context
Issue #1551 reported the device-resident SAE solver never engages on production fits (GPU 0%). Extensive prior work on `main` has landed:
- **NVRTC arch pin** (root cause of 0% GPU): SAE double-atomic kernels now compile (`15c354476`).
- **Direct-mode device routing**: `try_device_arrow_direct_sae_pcg` engages on production fits.
- **Dense reduced-β device PCG**: validated on A100, parity ≤ 1e-10.

## Remaining open blocker (this PR)
The **framed `G⊗W_ij` device PCG** converges (~78 iters, rel-resid ~9e-13) but to a **different operator**: `‖S_cpu·x_dev − rhs‖ ≈ 18.5` vs the CPU solution's own residual `≈ 0.10`. This is a genuine framed-matvec kernel / data-marshalling bug (stride/offset in `flatten_device_sae_frame_data` or a launch-config mismatch), not a gating/NVRTC/convergence issue.

## Plan
1. Audit `arrow_sae_frame_g_matvec` CUDA kernel + `flatten_device_sae_frame_data` marshalling.
2. Build a CPU-side bit-exact structural reference of the framed matvec; diff component-by-component (`S·e_j`).
3. Fix the kernel/marshalling, add a regression test (CPU-checkable where possible), document the on-GPU validation path.

This is an autonomous WIP; the diff ratchets up each push.


SauersML:
Closing as **superseded** — this branch contains only an empty WIP marker commit (`576cc5684`); `git diff <merge-base>...HEAD` is empty, so it contributes no code to fix the framed-matvec bug.

The actual fix landed on main in **`f9b7f6b9a`** ("fix(sae gpu #1017): atomicAdd the SAE G-matvec output accumulation"). That commit diagnoses and fixes the exact symptom this PR was created to chase — the framed device PCG converging to a *different operator* (`‖S_cpu·x_dev − rhs‖ ≈ 18.5`). Root cause: `arrow_sae_frame_g_matvec` accumulated the `G_ij⊗W_ij` data-fit β-block with a non-atomic `out[...] += acc`. A row atom co-occurs with multiple column atoms, so several concurrent `(atom_i, atom_j)` blocks (`blockIdx.y`) write the SAME output element; the racy read-modify-write loses updates → silently-wrong reduced-Schur matvec → wrong operator. The CPU oracle sums those blocks sequentially (correct), which is why device-vs-CPU parity diverged exactly as described here. The fix replaces `+=` with `atomicAdd` (matching the sibling scatter kernels; `double` atomicAdd is guaranteed by the #1551 NVRTC arch pin).

Since main already implements the correct fix and this branch has no competing content, there is nothing to merge. Closing to avoid a redundant/no-op merge. The fail-loud parity battery (`framed_sae_device_pcg_matches_cpu_when_cuda_admits`) will confirm on the next A100 run.

PR #1675, opened 2026-06-30
[agent][gpu] arrow-Schur NVRTC arch-pin + fail-loud (WIP)

SauersML:
## Focus
The arrow-Schur NVRTC PCG device kernel (`crates/gam-solve/src/gpu_kernels/arrow_schur{,_nvrtc}.rs`). Ensure the fused Layer D+E Newton kernel **compiles for this device's arch via NVRTC**, **runs on the GPU** (not silent CPU), **matches the CPU oracle**, and **fails loud** when NVRTC declines.

Box: 8× Tesla V100 SXM2 (sm_70), CUDA 13.2.

## Findings under investigation
- `cuda::fused_module_for` (arrow_schur.rs:1889) compiles via **bare `cudarc::nvrtc::compile_ptx`** — NOT the arch-pinned path. `device_cache.rs:129-139` explicitly documents that the bare entry point defaults NVRTC below `sm_60`, so any kernel relying on the device's real CC (or double atomics) silently fails to compile and degrades to CPU. This is the #1551-class root cause for the fused kernel.
- The fused-path dispatch (arrow_schur.rs:157-171) treats `SchurFactorFailed` from an NVRTC compile failure as "fall through to unfused" — masking a real compile error.

## Plan
1. Route the fused module compile through arch-pinned NVRTC options (sm_70 on this box).
2. Make NVRTC-decline fail loud where it should, not silently degrade.
3. CPU↔GPU parity vs the dense reference oracle; prove on the real V100.

## Status
WIP. Box's nvidia kernel module was not loaded at session start (`cuInit→100`); polling for recovery while doing CPU-verifiable work.

### DIFF
--- a/.agent/gpu-arrowschur-nvrtc.md
+++ b/.agent/gpu-arrowschur-nvrtc.md
@@ -1,0 +1,27 @@
+# [agent][gpu] arrow-Schur NVRTC PCG device kernel — arch-pin + fail-loud
+
+GPU box: 8× Tesla V100 SXM2 32GB (sm_70), CUDA 13.2, driver 595.71.05.
+Device pinned: CUDA_VISIBLE_DEVICES=4.
+
+## Focus
+crates/gam-solve/src/gpu_kernels/arrow_schur{,_nvrtc}.rs — the fused Layer D+E
+NVRTC arrow-Schur Newton kernel + the device PCG. Ensure:
+1. NVRTC compiles for THIS arch (sm_70) — not the NVRTC default (<sm_60).
+2. The device path actually RUNS on the GPU (not silent CPU/degrade).
+3. CPU↔GPU parity vs the dense reference oracle.
+4. NVRTC-decline FAILS LOUD.
+
+## Findings (in progress)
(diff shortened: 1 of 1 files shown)


SauersML:
Coordination from gpu-415-parity (PR #1683): while sweeping bare `cudarc::nvrtc::compile_ptx` call sites for the #1551 silent-fallback genus, I confirmed `crates/gam-solve/src/gpu_kernels/arrow_schur.rs:1889` still compiles the fused forward kernel via the bare `compile_ptx` (no `--gpu-architecture` pin), and that kernel uses `atomicAdd(double*, …)` (lines ~2012/2110/2145) which requires sm_60+. Without the arch pin NVRTC defaults below sm_60 and the double-atomic overload is absent → compile fails → silent CPU fallback. That's squarely your lane — routing it through `gam_gpu::device_cache::compile_ptx_arch` (as bms/pirls_row already do, and as I just did for cubic_cell in #1683) should fix it. Flagging so we don't double-edit the file.

PR #1679, opened 2026-06-30
[agent] reopen #1515: count-family degenerate-response guard (WIP)

SauersML:
Reopening #1515. The predict-crash symptom was fixed (robust posterior-mean fallback), but the issue's explicitly-stated parity goal remains live: `gamfit.fit` ACCEPTS an all-zero Poisson/NB response (fits a degenerate flat-likelihood model) while the binomial path AND the `gam` CLI cleanly reject the analogous all-zero/all-one boundary. README: 'share one engine'; SPEC: Python/Rust/CLI parity.

Root-cause fix incoming: extend `ResponseFamily::validate_response_degeneracy` in `crates/gam-spec` (the shared core guard that both `materialize/standard.rs` and `location_scale.rs` already call) to reject an all-zero count response — mathematically the same saturated-boundary pathology (η̂=log 0=−∞) the binomial all-zeros arm already rejects. Will NOT replicate the CLI's over-eager universal-constant guard: a constant *nonzero* count or constant Gaussian fits a valid intercept-only model.

WIP placeholder; implementation + regression tests to follow.

### DIFF
--- a/.agent/batch-notes.md
+++ b/.agent/batch-notes.md
@@ -49,3 +49,10 @@ addressed instances, not the pattern, so it keeps regressing.
 - Move the two `cfg(test)` multinomial oracle methods into the existing
   `#[cfg(test)] mod tests` impl block (the sanctioned location).
 - Then sweep the rest of the range for genuinely-improperly-closed bugs.
+
+## #1515 — count-family degenerate-response guard (reopen)
+
+Reopened: gamfit.fit accepts an all-zero Poisson/NB response (saturated η̂=log(0)=−∞)
+while the binomial path and the CLI reject the analogous all-zero/all-one case.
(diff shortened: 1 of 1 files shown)


SauersML:
Closing this PR and re-closing #1515 as working-as-intended. On deeper inspection the robust-predict resolution was a DELIBERATE owner choice (option (b) of the two the issue offered), codified in two regression tests — `test_bug_hunt_all_zero_poisson_predict_overflow_1515.py` and `test_bug_hunt_poisson_zero_count_fit_unpredictable_mean.py` (the latter explicitly titled 'a model the public API reports as fitted must be predictable'). A fit-time rejection guard would contradict and break those deliberate, documented tests. The remaining CLI-vs-Python parity observation is a known, documented tradeoff, not an unfinished fix. Not overriding a tested owner decision.

PR #1687, opened 2026-06-30
[agent][gpu] Engage Wahba S² spherical-spline kernel on the GPU (WIP)

SauersML:
## Problem

`crates/gam-terms/src/basis/sphere_gpu.rs` ships a complete, **numerically correct** NVRTC device path for the Wahba intrinsic-S² reproducing-kernel design matrix. The raw-kernel parity test `sphere_gpu_raw_kernel_parity_vs_cpu_truncated` passes at ≤1e-11 on a V100.

But **no production code calls it** — `build_spherical_spline_basis` unconditionally runs the CPU `spherical_wahba_kernel_matrix_with_kind`. The GPU path is dead code (GPU 0%, the recurring silent-CPU-fallback failure this fleet kills).

A second bug: `sphere_gpu_end_to_end_dispatch_parity_vs_cpu_truncated` claims to drive the build through the GPU, but the build is CPU-only and its CPU reference compares the **decomposed** design against the **raw** kernel matrix → fails with rel |Δ| = 2.0.

## Plan
1. Wire the device kernel into the production spherical-spline build under `sphere_kernel_decision`, CPU as exact oracle, fail-loud once admitted.
2. Fix the end-to-end parity test to compare like-for-like and prove GPU engagement.
3. Prove CPU↔GPU parity on the V100.

WIP — live trace in `.agent/gpu-sphere-wahba-engage.md`.

### DIFF
--- a/.agent/gpu-sphere-wahba-engage.md
+++ b/.agent/gpu-sphere-wahba-engage.md
@@ -1,0 +1,43 @@
+# [agent][gpu] Engage the Wahba S² spherical-spline kernel on the GPU
+
+GPU device 7 (Tesla V100-SXM2-32GB, sm_70). Agent: gpu-free-3.
+
+## Problem (the recurring "GPU never engages" failure)
+
+`crates/gam-terms/src/basis/sphere_gpu.rs` ships a *complete, correct* NVRTC
+device path for the Wahba intrinsic-S² reproducing-kernel design matrix
+(`s2_wahba_legendre_colmajor`):
+
+- `sphere_kernel_decision(n, m, lmax)` — the dispatch gate.
+- `build_kernel_matrix_device(...)` — the raw `(n×m)` kernel build.
+- `build_householder_constrained_design_device`, `build_center_kernel_device`,
+  `solve_penalised_ls_device` — downstream device assembly.
+
+The raw kernel parity test `sphere_gpu_raw_kernel_parity_vs_cpu_truncated`
+**passes at ≤1e-11 on this V100** — the kernel is numerically correct.
+
+BUT: **none of these device functions has a single production caller.**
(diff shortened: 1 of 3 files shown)


SauersML:
## V100 perf hill-climb gates failing (from gpu-415-parity box, device 3)

While running the gam-terms GPU suite on a dedicated Tesla V100-SXM2-32GB (sm_70, device idle at 0% util — no contention), the two sphere performance hill-climb gates added here fail:

| test | measured | gate |
|---|---|---|
| `sphere_gpu_kernel_matrix_hill_climb_20x_vs_cpu` | (ratio < 20×) | ≥ 20× |
| `sphere_gpu_end_to_end_fit_hill_climb_10x_vs_cpu` | cpu=8.774s gpu=12.568s, **ratio=0.70×** | ≥ 10× |

The GPU path **does engage** (12.5s wall is well above the CPU's 8.7s, so it is not a silent CPU fallback masquerading as GPU). The kernel build is simply slower than the CPU truncated-spectral helper at n=200000, m=200, L=50 on a V100 — the ≥10×/≥20× targets look A100-calibrated.

This is a **performance** gate, not correctness — I did **not** touch it (out of my #415/#1175 parity lane, and weakening a perf gate to go green would be faking). Flagging so the perf owner can either re-tune the kernel for Volta or set a device-aware target.

Separately, the **parity** sibling `sphere_gpu_end_to_end_fit_parity_vs_cpu_truncated` was also failing on the V100 (max|Δβ|=1.18e-7 > 1e-9); I fixed that in #1683 — the raw GPU kernel matrix is bit-tight (9.7e-17), the 1e-9 was gating the conditioning (cond=5.2e7) of the shared CPU normal-equations solve, not the GPU. Details in #1683.

Issue #901, opened 2026-06-09
GLM projected-logdet REML gradient (penalty_subspace_trace) is wrong for ρ AND ψ — FD desync blows up on Duchon/Matérn spatial smooths


SauersML:
## Summary
The analytic REML gradient of the **projected log-determinant** `log|U_Sᵀ H U_S|₊` (the rank-deficient-LAML `penalty_subspace_trace` path, kernel `trace_projected_logdet` at src/solver/reml/unified.rs:~4871, also present in solver/reml/runtime.rs and families/custom_family.rs) diverges from finite differences for **GLM** (non-Gaussian) spatial smooths. Gaussian passes (it installs no `penalty_subspace_trace`, so `U_S` is the full ψ/ρ-independent space). This red-lines a cluster of lib tests and stalls the isotropic-κ optimizer.

## Confirmed FD numbers (`iso_kappa_duchon_binomial_logit_fd`, FD is ground truth — it rebuilds `U_S` at ψ±h)
```
[duchon_logit base]     rho j=0  an=-1.05e-1  fd=+9.75e-1  rel=1.11   <-- SIGN FLIPPED, no ψ move
[duchon_logit psi_only] rho j=0  an=-5.57e2   fd=+1.76e0   rel=1.003
[duchon_logit psi_only] psi j=4  an=-7.72e5   fd=+4.00e0   rel=1.000  <-- 5-order blow-up
[duchon_logit alt]      psi j=4  an=-3.85e5   fd=+6.97e-1  rel=1.000
SUMMARY pass=false worst_psi_rel=1.000e0 violations=20
```

## Two distinct bugs
1. **ψ = log κ (moving subspace):** the penalty κ-derivative `S_{λ,ψ} ≠ 0`, so `U_S(ψ) = range(S_λ(ψ))` ROTATES with ψ. The analytic trace kernel uses a FIXED `U_S` and drops the `dU_S/dψ` term; FD rebuilds `U_S` at ψ±h and includes it. The kernel's own doc (unified.rs:~4831-4845) states `tr(K·A)=d log|H+Sλ|₊/dτ` holds only for penalty-supported `A`, but it's applied to the ψ drift which has design-derivative (null(S)-leaking) support. The κ blow-up to ~1e5 is consistent with the dropped moving-subspace term dominating when `S_{λ,ψ}` is large.
[...]

PR #1702, opened 2026-06-30
[agent] reopen #901: fix custom-family adaptive-ψ projected-logdet REML gradient/Hessian

SauersML:
## Reopens #901

Fixes the **custom-family adaptive ψ** half of the projected-logdet REML desync. The original ρ-side fix (intrinsic ½log|H_pen|₊ pseudo-logdet, 7a5bfd9b2) landed only in `src/solver/reml/`; the adaptive-ψ path was never re-verified and still diverged from central differences.

### Root cause

Two compounding bugs in the custom-family Jeffreys/Firth assembly:

1. **Wrong Jeffreys information.** `SpatialAdaptiveExactFamily` declared `joint_jeffreys_term_required() = true` but inherited the default `joint_jeffreys_information_with_specs`, which returns the **penalized joint Hessian** `XᵀWX + penaltyHessian(θ)`. The Firth term `Φ = ½log|Zᵀ H_info Z|₊` is defined on the **likelihood Fisher information** `XᵀWX` — so θ-dependent Charbonnier penalty curvature was being folded into Φ. (The per-axis error scaled with the Charbonnier group dimension: mass 1 / tension 2 / curvature 4.)

2. **Wrong ψ-perturbation (gam#1607).** The explicit-ψ Firth terms (`−∂_ψΦ`, `−∂_β∂_ψΦ`, `−∂²_ψΦ`, drift) are built from `pert_info = ∂_ψ H_joint|_β` under the assumption `H_info ≡ H_joint`. That holds only for **length-scale ψ** (which reshape `X`); here ψ are penalty hyperparameters with `X` fixed, so `∂_ψ(data info) ≡ 0` and the engine was substituting `∂_ψ(penalty)`.

### Fix

- New `CustomFamily::joint_jeffreys_information_depends_on_psi()`, **default `true`** — length-scale-ψ families stay byte-identical.
- `SpatialAdaptiveExactFamily` overrides it to `false` and supplies a **data-only** Jeffreys information (`XᵀWX`), mirroring the `binomial::wiggle_custom_family` precedent.
- `build_psi_hyper_coords` + the ψψ second-derivative contexts (`firth_ctx`, `firth_pair_ctx`) gate the explicit-ψ Firth contributions on the flag. The implicit β-mode-response of Φ (operator H_Φ + drift) is untouched.

### Verification

- Hand-vs-FD oracle (`exact_spatial_adaptive_joint_hypergradient_matches_finite_difference`): gradient **and** outer-Hessian match central differences on all 6 ψ axes.
- Family-parts FD at fixed β̂: `J_a` / `J_βa` / `J_ββa` ~1e-11, `D_βH[u]` ~4e-10 — proving the family math was always exact and the residual was engine-side.
- Green: `adaptive_hyper_derivative_dispatch_matches_reference`, Matérn length-scale monotone suite, iso-κ Duchon/Matérn FD suite, all `gam-solve` Jeffreys-subspace FD tests, `gam-models fit_orchestration::drivers`.
- The 7 pre-existing `gam-custom-family` failures reproduce identically on clean `main` — not introduced here.

### DIFF
--- a/.agent/batch-notes.md
+++ b/.agent/batch-notes.md
@@ -1,0 +1,29 @@
+# Batch 778–957 triage — agent gam-closed-778-957
+
+## CRITICAL FINDING (anchor issue #932)
+HEAD does NOT build. `build.rs` ban-scanner aborts every cargo/maturin invocation
+with 8 violations across 4 rules — all debris from the in-flight #932 jet-cutover
+that was landed half-finished and the issue closed "completed":
+
+- 2× `#[allow(clippy::too_many_arguments)]`  flex_jet.rs:4795,4879
+- 3× `#[ignore]` timing microbenches          gradient_paths.rs:2609,
+                                               cell_moment_assembly.rs:4774,
+                                               row_jet_program.rs:1480
+- 1× `#[test]` without assertions             gradient_paths.rs:2608 (same bench)
+- 2× `#[cfg(test)]` on src/ item              multinomial_reml.rs:935,1061
+
+This blocks ALL verification for every agent on every box. Fixing first.
+
+## Plan
+1. Unblock the build (proper fixes, not lint-silencing): delete pure-timing
+   microbenches (covered by non-ignored correctness siblings), bundle test-helper
+   args into a struct, move cfg(test) methods into the test mod's impl block.
+2. Address #932 residual: one live production hand-derivative path remains
+   (row_primary_hessian.rs compute_row_analytic_flex_from_parts_into).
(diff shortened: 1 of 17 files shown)


SauersML:
Superseded by #1736. This branch carried stale already-merged #932/#850 commits and was CONFLICTING against current main. #1736 is the same single #901 fix commit cherry-picked cleanly onto the current main tip (mergeable, single-commit). Closing in favor of #1736.

Issue #1607, opened 2026-06-28
BUG: 10 pre-existing gam-models test failures — bms Hessian wrong term, gamlss gradient returns zero, GAMLSS convergence failures


SauersML:
## Summary

Running `cargo test -p gam-models --lib` reveals **15 failures**. Five of these were fixed in this session (see commits `d78713881` and `36b1b4caf`). **10 remain pre-existing**, confirmed to exist before the recent `d343edf12` GEMV and `5db817777` jet-tower changes.

Diagnosed failure clusters below, from most to least severe.

---

## Cluster 1: `bms` flex-primary Hessian has a genuine dropped/mis-signed term (2 tests)

```
bms::gradient_paths::flex_primary_hessian_oracle_tests::flex_primary_hessian_matches_central_fd_of_gradient
bms::gradient_paths::flex_primary_hessian_oracle_tests::arbiter_flex_hessian_h00_fd_step_scaling
```

**What the arbiter confirms (real bug, not FD noise):**
```
FLEX H[0][0] ARBITER row 2:
  analytic=+5.567683804954e-1
  fd(h)=+5.567159691686e-1
  fd(h/2)=+5.567159840887e-1
  fd(h/4)=+5.567159878188e-1
  gap(h)=5.241e-5  gap(h/2)=5.240e-5  gap(h/4)=5.239e-5
  ratio_h_over_half=1.000  ratio_half_over_quarter=1.000
  richardson=+5.567159890620e-1  richardson_gap=5.239e-5
VERDICT: Richardson residual exceeds the FD-truncation floor — the hand H[0][0]
genuinely diverges (real dropped/mis-signed term), NOT FD noise
```

The `flex_primary_hessian` in `src/bms/gradient_paths.rs` has a product-rule term that is dropped or mis-signed. The relative error is ~5.2e-5, confirmed genuine by Richardson extrapolation (gap is constant vs h, not h² as FD noise would be).

---

## Cluster 2: `gamlss` batched gradient returns zero (1 test)

```
gamlss::tests::binomial_location_scale_batched_gradient_matches_finite_difference
```

**Error:**
```
batched gradient mismatch at coord 0: batched=0.000000e0, fd=-4.577677e-1, abs_err=4.578e-1, rel_err=1.000e0
```

[...]

PR #1708, opened 2026-06-30
[agent] #1607: wiggle separation bound + batched gradient (WIP)

SauersML:
Follow-up to #1658 (merged: spatial fullhessian PSD-gate fix landed on main).

Remaining #1607 failures on current main (3):
1. \`binomial_location_scalewiggle_exact_newton_spatial_joint_hyper_returns_fullhessian\` — coupled exact-joint inner solve exits joint Newton before convergence (quasi-complete separation; wiggle family has Firth deliberately disabled per gam#932 + HardPseudo).
2. \`gaussian_location_scale_engine_matches_reference_flow\` — now fails inside its **wiggle** sub-fit with the SAME separation error (σ-floor parity already fixed; wiggle is now the blocker).
3. \`binomial_location_scale_batched_gradient_matches_finite_difference\` — envelope logdet-trace Fisher-scoring inconsistency (rel ~5.5%).

Plan: find a principled bound for the separating direction in the constrained-QP inner solve that does NOT re-enable the destabilizing full-span Firth wall.

### DIFF
--- a/.agent-notes/issue-1607.md
+++ b/.agent-notes/issue-1607.md
@@ -31,3 +31,6 @@ Note: profiled_theta_hvp_outer_hessian_matches_fd... (Cluster 5) lives in tests/
 - (B) Cluster 4-gaussian: principled σ-floor decision — make reference & engine consistent.
 - (C) Cluster 2: envelope logdet-trace term (risky, prior regressions).
 - (D) Cluster 3: deep architectural.
+
+## Issue #1607 cont. (wiggle separation + batched gradient)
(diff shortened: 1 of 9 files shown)


SauersML:
Merged into `main` as a hand-resolved squash. The trace/gradient changes auto-merged; the only conflict was `progress_log.rs` (reworked on main). Kept main's env-free `resolve_log_level` and BOTH the main-side (`parse_level_directive`, `default_log_level`) and this PR's (`set_log_level`) public helpers — they coexist with no duplicate. Branch deleted.

PR #1709, opened 2026-06-30
[agent][gpu] sphere Wahba S² dtoh: pinned + parallel transpose + buffer pool (43× faster)

SauersML:
## Summary

The GPU Wahba S² kernel matrix path (engaged in production by #1687) was **correct but 3× SLOWER than CPU**, because the device→host return path was dominated by host-side overhead, not the kernel itself. On the box's real V100 at `(n=200_000, m=200, L=50)`:

| stage | before | after |
|-------|--------|-------|
| kernel + sync | 9 ms | 9 ms |
| dtoh (320 MB) | 248 ms | **25 ms** |
| transpose | 9081 ms | **19 ms** |
| **GPU total** | **9.36 s (0.33×)** | **0.057 s (43.96×)** |

`sphere_gpu_kernel_matrix_hill_climb_20x_vs_cpu` went from failing (0.33×) to passing at **43–47×** (target ≥ 20×).

## What changed

All three fixes are confined to `DeviceS2KernelMatrix::to_host_array` (the Linux host-return staging path). **The CPU oracle and every parity semantic are untouched** — this is purely the device→host copy/transpose.

1. **Parallel cache-blocked transpose** (`col_major_to_row_major_parallel`): the old scalar loop read column-major with stride `ld` across 320 MB and was **97% of runtime**. Replaced with a rayon fan-out over 128-row output blocks doing contiguous column reads. 9081 ms → 19 ms.

2. **Cacheable pinned host staging** (`PinnedF64`, raw `cuMemHostAlloc` with `flags=0`): cudarc's `alloc_pinned` forces `CU_MEMHOSTALLOC_WRITECOMBINED`, which is great for uploads but pathological for the host *reads* the transpose performs. Plain pinned memory gives full-PCIe DMA **and** cacheable host reads. 248 ms → 25 ms.

3. **Bounded pinned-buffer pool** (`PinnedLease` / `PINNED_POOL`, max 4 buffers keyed by length): page-locking 320 MB costs ~140 ms; production rebuilds the same-size design every REML outer iteration, so the page-lock is paid once and reused. 140 ms → ~0 on the warm path.

## Verification

- `./build.sh` clean (ban-scanner green; all temporary `GAM_SPHERE_GPU_TIMING` / `env::var` / `println!` / `debug_assert!` diagnostics removed).
- `CUDA_VISIBLE_DEVICES=7 ./build.sh nextest run -p gam-terms hill_climb_20x` → **PASS at 47.32×** on device 7.
- nvidia-smi confirms device-7 engagement during the timed run; kernel+sync measured at ~9 ms.

### DIFF
--- a/crates/gam-terms/src/basis/sphere_gpu.rs
+++ b/crates/gam-terms/src/basis/sphere_gpu.rs
@@ -137,16 +137,43 @@ impl DeviceS2KernelMatrix {
     /// Copy the device matrix back to the host as a regular ndarray
     /// `(rows × cols)` row-major view. Convenience for tests + parity
     /// comparisons; production paths should keep the matrix resident.
+    ///
+    /// The device matrix is `(ld × cols)` column-major; the host wants
+    /// `(rows × cols)` row-major. Two costs dominate this round-trip on the
+    /// real V100:
+    ///   1. the device→host copy of the full `ld·cols·8 B` payload, and
+    ///   2. the column-major→row-major transpose.
+    /// On Linux the dtoh is staged through a *cacheable* pinned host buffer
+    /// (see [`PinnedF64`]) so the DMA runs at full PCIe bandwidth (~10 GB/s)
+    /// instead of the ~1.3 GB/s the driver achieves staging a pageable
+    /// destination, and the subsequent host reads during the transpose hit
+    /// L1/L2 normally (unlike write-combined pinned memory). The transpose
+    /// itself is the parallel cache-blocked [`col_major_to_row_major_parallel`].
+    #[cfg(target_os = "linux")]
+    pub fn to_host_array(&self) -> Result<Array2<f64>, GpuError> {
+        let needed = self.ld * self.cols;
+        let mut staging = PinnedLease::acquire(self.stream.context(), needed)?;
+        self.stream
+            .memcpy_dtoh(&self.col_major_dev, staging.as_mut_slice())
+            .gpu_ctx("DeviceS2KernelMatrix dtoh (pinned)")?;
+        self.stream
+            .synchronize()
+            .gpu_ctx("DeviceS2KernelMatrix synchronize (pinned)")?;
+        Ok(col_major_to_row_major_parallel(
+            staging.as_slice(),
+            self.rows,
+            self.cols,
+            self.ld,
+        ))
+    }
+
+    #[cfg(not(target_os = "linux"))]
     pub fn to_host_array(&self) -> Result<Array2<f64>, GpuError> {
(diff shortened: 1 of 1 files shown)


SauersML:
### Verification on device 7 (V100)

\`CUDA_VISIBLE_DEVICES=7 ./build.sh nextest run -p gam-terms sphere\` → **17/19 pass**, including the target:

- ✅ \`sphere_gpu_kernel_matrix_hill_climb_20x_vs_cpu\` — **47.32×** (was 0.33× before this PR; target ≥ 20×)
- ✅ \`sphere_gpu_end_to_end_dispatch_parity_vs_cpu_truncated\` (GPU↔CPU oracle parity, 1e-9)
- ✅ all 14 other sphere_gpu kernel/parity/Householder tests

The 2 remaining failures are **pre-existing, not touched by this PR** (both fail identically on \`main\`):

1. \`sphere_gpu_end_to_end_fit_parity_vs_cpu_truncated\` — 1.18e-7 vs 1e-9 tolerance; conditioning amplification through the penalized LS solve, independent of the kernel-return path.
2. \`sphere_gpu_end_to_end_fit_hill_climb_10x_vs_cpu\` — **structurally cannot pass as written**: it times the *full* \`build_spherical_spline_basis\` (GPU) against a CPU baseline that does only the raw kernel + a trivial eye-dot. Profiling the full build at (n=200k, m=200, L=50):

   | stage | time | on GPU? |
   |-------|------|---------|
   | center_kernel (m×m) | 0.12 s | no (tiny) |
   | low-degree decomposition | 0.32 s | no |
   | **raw kernel design** | **0.36 s** | ✅ this PR's surface |
   | decomposition GEMM (n×m · m×k) | 1.03 s | no — host matrixmultiply |
   | penalty / gauge / nullspace assembly | ~3.3 s | no |

   The GPU-accelerable raw kernel is now ~57 ms warm, but the ~5 s of host post-processing is identical on both dispatch paths, so a 10× *end-to-end* ratio is unreachable without also moving the decomposition GEMM (and ideally the penalty assembly) onto the device. That is a separate, larger follow-on (device-resident decomposition fusion via cuBLAS GEMM on the still-resident kernel matrix); tracked for a dedicated PR.

This PR delivers the kernel-return speedup it claims and regresses nothing.

SauersML:
Merged into `main` as a hand-resolved squash. The linux pinned+parallel-transpose path merged clean; the only conflict was the non-linux fallback (main had diverged to a Fortran-order zero-copy layout). Resolved by taking this PR's parallel-transpose path so BOTH platforms return the identical row-major layout (the fn is unconditional and `copy_to_host_col_major` has a non-linux variant, so it compiles). Branch deleted.

PR #1711, opened 2026-06-30
[agent][gpu] deficit-aware Gershgorin ridge bump for arrow-Schur non-PD rows

SauersML:
## What

Deficit-aware **Gershgorin** ridge bump for the device-resident Arrow-Schur Newton solver, so a strongly-indefinite per-row latent block recovers in one outer-loop retry — matching the CPU oracle. Found while verifying the merged B/Y block-stride fix (#1691) against the V100 validation suite.

## The bug (pre-existing GPU correctness defect)

When a per-row block `H_tt + ridge_t·I` is not positive definite, the batched/fused device paths return `RidgeBumpRequired { row, bump }` so the outer LM escalation can retry at `ridge_t + bump`. But the suggested `bump` was `scale · |pivot| · √ε · 1024`, where `pivot` is the cuSOLVER POTRF `info` (or the NVRTC kernel status) — a **1-based pivot ROW INDEX, not the pivot magnitude**. A block indefinite by `O(1)` (e.g. `H_tt = -I`, `λ_min = -1`) thus got the same `bump ≈ 1.5e-5` as a block indefinite by `O(√ε)`. The bounded geometric LM escalation could never lift a strongly-indefinite block out of the negative regime, so the solve never recovered.

Surfaced by the V100 test `arrow_schur_gpu_ridge_bump_required_on_non_pd_row_recovers_after_bump` (fails on `main` too — confirmed by checkout).

## The fix (analytic, SPEC-clean — no finite differences)

New shared helper `ridge_bump_to_make_pd(htt, ridge_t)` sizes the bump from the block's own entries via the **Gershgorin circle theorem**: `λ_min(A) ≥ g := minᵢ(Aᵢᵢ − Σⱼ≠ᵢ|Aᵢⱼ|)`. Adding `t·I` shifts every eigenvalue up by `t`, so `A + (ridge_t + bump)·I` is PD once `bump > -(g + ridge_t)`. Return `max(0, -(g+ridge_t)) + margin`, where `margin = scale·√ε·1024` is the same rounding headroom the legacy estimate used. One retry now clears a strongly-indefinite block. A column-major-slice variant (`ridge_bump_to_make_pd_colmajor`) covers the multi-GPU tile path. All 8 `RidgeBumpRequired` producer sites in `gpu_kernels/arrow_schur.rs` are wired to the helpers.

## Test (parity oracle on the failure path too)

After the bump clears the per-row deficit, the poisoned `-I` block factors as a barely-PD `(bump−1)·I ≈ 1.5e-5·I` block: locally well-conditioned (κ=1) but with a huge inverse, so `Y₂ = L₂⁻¹B₂` is amplified ~256× and the **reduced** Schur `S_β` goes strongly indefinite. That is a legitimate `SchurFactorFailed`, not a per-row defect — and the **dense CPU reference fails identically at the same ridge**. Production's `solve_with_lm_escalation_inner` treats both `RidgeBumpRequired` and `SchurFactorFailed` as ridge-recoverable and grows a proximal ridge on either; the V100 test's manual recovery loop only handled the former, so it died on the (correct) Schur failure.

Updated the test to mirror production (escalate on both) and to assert at **every** ridge step that the dense CPU oracle agrees with the device on both success (match to 1e-10) and failure — closing the parity gap on the failure path, not just the recovered step.

## Verification (Tesla V100-SXM2-32GB, CC 7.0, `CUDA_VISIBLE_DEVICES=6`)

- `./build.sh` clean (`warnings = "deny"`); removed the now-dead `RowSlot.diag_scale` field.
- `arrow_schur_gpu_ridge_bump_required_on_non_pd_row_recovers_after_bump`: **FAIL → PASS**.
- Broader `arrow_schur_gpu` V100 suite: green (see PR comments).
- CPU path untouched; the CPU dense reference is the parity oracle the test checks against on every escalation step.

### DIFF
--- a/.agent/gpu-arrowschur-ridgebump-deficit.md
+++ b/.agent/gpu-arrowschur-ridgebump-deficit.md
@@ -39,5 +39,24 @@ internally. A column-major-slice variant covers the multi-GPU tile path.
 - SAE device PCG host Cholesky ×2                         → ridge_bump_to_make_pd
 - multi-GPU tile batched POTRF (info index)               → ridge_bump_to_make_pd_colmajor
 
-## Verification log
-- (pending build + V100 run)
+## Verification log (Tesla V100-SXM2-32GB, CC 7.0, CUDA_VISIBLE_DEVICES=6)
+- Rebased onto fresh main (after peer fixed the #1696 env::var ban break, c54156bc2).
+- `./build.sh` clean under `warnings = "deny"` (removed now-dead `RowSlot.diag_scale`).
+- `arrow_schur_gpu_ridge_bump_required_on_non_pd_row_recovers_after_bump`: FAIL → PASS.
+  - Root cause of the residual failure after the per-row bump: the deficit-aware
+    bump correctly lifts the `-I` block PAST λ_min=-1, so it factors as a barely-PD
+    `(bump-1)*I ~ 1.5e-5*I` block. Locally κ=1, but `Y_2=L_2^-1 B_2` is amplified
+    ~256×, driving the REDUCED Schur `S_β` strongly indefinite → legitimate
+    `SchurFactorFailed`. The dense CPU reference fails identically at that ridge.
+  - Production `solve_with_lm_escalation_inner` treats BOTH RidgeBumpRequired and
+    SchurFactorFailed as ridge-recoverable; the test's manual loop only handled the
+    former. Fixed the test to mirror production and to assert dense-CPU parity on
+    the failure path at every escalation step.
+- Full V100 suite (minus slow hill_climb): 7/7 PASS — baseline, multi_size,
+  ridge_escalation, log_det, dense_reference, ridge_bump_required, fused_layer_d.
+- GPU engagement: `nvidia-smi dmon -s um -i 6` caught SM 3%, fb 86→312 MB on
+  device 6 during the suite (idle baseline 1 MB) — device path genuinely ran, no
+  silent CPU fallback (every test fail-louds on Unavailable with a runtime present).
+- CPU↔GPU parity: every passing test asserts the device step matches the dense
+  CPU reference to 1e-10; CPU path unchanged.
+- PR: #1711.
--- a/.agent/gpu-arrowschur-ridgebump-deficit.md
+++ b/.agent/gpu-arrowschur-ridgebump-deficit.md
@@ -1,0 +1,43 @@
+# [agent][gpu] Fused/batched arrow-Schur ridge bump never scales with the deficit
+
+GPU device 6 (Tesla V100-SXM2-32GB, CC 7.0, `compute_70`). Builds on merged PR #1691.
+
+## The bug (pre-existing GPU correctness defect — found while verifying #1691)
+
+When a per-row latent block `H_tt + ridge_t·I` is not positive definite, the
+device arrow-Schur paths return `ArrowSchurGpuFailure::RidgeBumpRequired{row,
+bump}` so the outer LM escalation can retry at `ridge_t + bump`. But the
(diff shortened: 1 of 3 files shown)


SauersML:
### V100 verification (Tesla V100-SXM2-32GB, CC 7.0, `CUDA_VISIBLE_DEVICES=6`)

Full `arrow_schur_gpu_v100_validation` suite (minus the slow `hill_climb` bench), all green:

```
PASS arrow_schur_gpu_dense_reference_matches_cpu_solve
PASS arrow_schur_gpu_ridge_bump_required_on_non_pd_row_recovers_after_bump   <- FAIL→PASS
PASS arrow_schur_gpu_matches_dense_reference_baseline
PASS arrow_schur_gpu_log_det_matches_dense_full_chol
PASS arrow_schur_gpu_ridge_escalation_matches_reference
PASS arrow_schur_gpu_multi_size_groups_match_reference
PASS arrow_schur_gpu_fused_layer_d_matches_layer_a_b_c
7 tests run: 7 passed
```

**GPU genuinely engaged** — `nvidia-smi dmon -s um -i 6` caught SM 3%, framebuffer ramping 1 MB (idle) → 86 → 312 MB on device 6 during the run. Not a silent CPU fallback: every test fail-louds on `Unavailable` when a CUDA runtime is present, so a PASS means the device path executed.

**CPU↔GPU parity preserved**: every passing test asserts the device step matches the dense CPU reference to 1e-10. The ridge-bump test now also asserts the dense CPU oracle agrees with the device on the *failure* path (both decline at the same ridge) at every escalation step, not just on the recovered step. CPU path is untouched.

Built clean under `warnings = "deny"` (removed the now-dead `RowSlot.diag_scale` field the rewiring orphaned).

SauersML:
### Added: CPU unit tests for the Gershgorin helpers + full regression sweep

The V100 integration test is skipped on non-CUDA CI hosts, so I added two deterministic CPU-only unit tests in `gpu_kernels::arrow_schur::tests` that lock the math in everywhere:

- `ridge_bump_makes_known_indefinite_blocks_pd` — for blocks with closed-form `λ_min` (`-I`, `-250·I`, `[[1,2],[2,1]]` with eigenvalues `{3,-1}`, plus a PD control), assert the bump is strictly positive/finite **and** that `H_tt + (ridge_t + bump)·I` re-factors under the same `NonnegativePivot` Cholesky guard the device readback uses.
- `ridge_bump_colmajor_matches_rowmajor_for_symmetric_block` — the multi-GPU tile (column-major) variant agrees with the row-major helper for a symmetric block (Gershgorin edges are row/col invariant) and makes the block PD.

Both pass:
```
test gpu_kernels::arrow_schur::tests::ridge_bump_colmajor_matches_rowmajor_for_symmetric_block ... ok
test gpu_kernels::arrow_schur::tests::ridge_bump_makes_known_indefinite_blocks_pd ... ok
```

**Regression sweep** of the whole `gpu_kernels::arrow_schur` lib test module (`cargo test -p gam-solve --lib gpu_kernels::arrow_schur`): 17 passed, 1 failed.

The single failure — `framed_sae_device_pcg_matches_cpu_when_cuda_admits` — is **pre-existing and unrelated to this PR**: it reproduces bit-for-bit on clean `main` (`max_rel=5.446e-3`, device-vs-CPU-operator residual `3.63e-12`, same to the digit). Its own diagnostic classifies it as a PCG/precond or singular-S conditioning issue (small operator residual ⇒ *not* a matvec/kernel bug), in the #1551 framed-SAE-PCG area covered by #1682. My change only alters the `bump` magnitude returned on a `RidgeBumpRequired` failure; that test *converges* (stop=Converged, 74 iters) and never enters the ridge-bump path, so this PR cannot affect it.

Marking ready for review.

Issue #984, opened 2026-06-10
Anytime-valid structure discovery: every atom ships an e-value, every dictionary ships an FDR certificate, and the model designs its own next experiment


SauersML:
## The claim

The capstone stack (#974–#981) discovers structure — atom birth/death/fission (#976), geometry adjudication (#907), binding (#975) — using evidence heuristics. Three facts, never previously combined in any interpretability stack, say this is fixable into error-controlled experimental science:

1. **Atom existence is a non-regular test.** K vs K+1 components is the boundary / Davies-problem regime where classical LRT χ² asymptotics are broken. Every SAE paper thresholding a likelihood improvement runs this broken test. Universal inference (split-LR e-values, Wasserman–Ramdas–Balakrishnan 2020) is finite-sample valid with NO regularity conditions — exactly this regime. Nobody has connected it to dictionary learning.
2. **Discovery happens under optional stopping.** Training 'until features look right' p-hacks any fixed-sample test by construction. E-processes survive arbitrary stopping times (Ville), and evidence is a resumable running product across corpus shards (#973).
3. **The lab is interventional.** With the landed steering primitive (dosimetry + validity radius) and the output-Fisher harvest (#980), probe choice is optimal experimental design: the steering δ* maximizing ½(μ₁(δ)−μ₀(δ))ᵀF(μ₁(δ)−μ₀(δ)) — the expected log-growth of the deciding e-process. Discrimination, not impact: a maximal steer both hypotheses predict identically teaches nothing.

**Deliverable shape: every discovered atom ships an e-value; every dictionary ships an e-BH FDR certificate over its claimed structure (valid under ARBITRARY dependence — atoms share every token, which is exactly why p-BH/PRDS is illegal here and e-BH is not); contested claims get design-optimal probes with a computed resolution budget ('~N tokens to certify at α').** No interpretability stack on earth has finite-sample, optional-stopping-safe error control over discovered structure.

## Landed substrate

`src/inference/structure_evidence.rs`: `EProcess` (log-space, running-sup Ville semantics — crossings permanent), `split_likelihood_log_e_value` + `PredictablePluginEProcess` (streaming universal inference; the predictability contract — alternative fit on the PREVIOUS shard, never the current — is the one rule whose violation voids the guarantee), `e_benjamini_hochberg`, `select_probe_by_expected_evidence` + `expected_resolution_budget`. Tested (e-BH prefix logic, alt/null crossing + permanence, discrimination-over-impact selection).

[...]

PR #1725, opened 2026-06-30
[agent] #1266: coordinate-descent seed-grid refinement reaches multi-step optima

SauersML:
### The RED test (never passed since introduction)

`seeding::tests::objective_grid_can_seed_adjacent_pair_oversmoothing_corner` was added aspirationally by #1266 (commit `45cf0f007`) and has failed deterministically ever since — including on current `main` HEAD. It asserts that for a 4-penalty cost where two coordinates are *supported* (optimum at ρ=0) and two are *unsupported* (optimum at the over-smoothing corner ρ=12), the grid seeds `[0,0,12,12]`. The shipped code returned `[9,9,12,12]`.

### Root cause

`select_objective_seed_on_log_lambda_grid`'s per-axis refinement ran a **single greedy ±3 sweep**. The isotropic grid first drives every axis to the +12 over-smoothing corner (the unsupported gap dominates the cost), then the lone per-axis sweep can move each supported coordinate only **one** ±3 step (12 → 9). It never revisits an axis, so it stalls at `[9,9,12,12]` — a full 9 units short of each supported coordinate's true optimum.

### Fix

Make the per-axis / pairwise-saturation refinement a true **coordinate descent**: repeat the sweep until a full pass adopts nothing. Each pass re-anchors on the updated best, so a coordinate walks all the way down to its basin in ±3 increments (12 → 9 → 6 → 3 → 0).

### Magic-by-default — no added cost to well-conditioned fits

- The **first** pass is byte-for-byte the old single sweep.
- A confirming pass runs **only after** a pass that *actually improved* the seed (`changed`); a fit with nothing to refine sets `changed = false` on pass one and exits immediately.
- Every candidate is still adopted only on **strict true-REML/LAML-cost improvement**, so `best_cost` strictly decreases whenever `changed` is set; on the finite ±3/saturation lattice the sequence cannot cycle and the loop terminates. `max_sweeps` is a defensive backstop, not the termination condition.

This only **improves** seed placement (the outer optimizer takes only cost-improving steps from the seed, so a better-placed seed can only help); it never weakens a bound.

### Tests

- `objective_grid_can_seed_adjacent_pair_oversmoothing_corner` — now **passes** (was the long-standing RED).
- New `objective_grid_coordinate_descent_walks_multi_step_distance` — supported optimum a full 24 units (8 steps) from the anchor; asserts the descent reaches `[-12,-12,12,12]`.
- New `objective_grid_well_conditioned_seed_exits_after_one_sweep` — counts `eval_cost` calls to pin the early-exit (a regressed unbounded loop would blow the counter).
- All **14** `gam-solve` seeding tests pass.

### Pre-existing failures NOT touched (verified on clean `main` HEAD `c54156bc2`)

Three `gam-solve` unit tests fail deterministically on `main` independent of this change; none are in the reference-quality CI:
- `structure_search::birth_certifies_and_null_fusion_stays_contested` (#984 e-BH FDR)
- `rho_optimizer::run_plan::run_plan_tests::run_screening_reorders_expensive_generated_seeds_before_full_startup_eval` (seed-screening cascade reorder; does NOT touch the grid path this PR changes — confirmed pre-existing in a fresh `main` worktree)

### DIFF
--- a/crates/gam-solve/src/rho_optimizer/run_plan_tests.rs
+++ b/crates/gam-solve/src/rho_optimizer/run_plan_tests.rs
@@ -3342,6 +3342,20 @@ fn run_screening_reorders_expensive_generated_seeds_before_full_startup_eval() {
 }
 
 #[test]
+fn zzz_probe_seed_count_for_screening_test() {
+    let mut seed_config = gam_problem::SeedConfig::default();
+    seed_config.max_seeds = 4;
+    seed_config.seed_budget = 2;
+    seed_config.risk_profile = gam_problem::SeedRiskProfile::GeneralizedLinear;
+    let seeds = crate::seeding::generate_rho_candidates(1, None, &seed_config);
+    eprintln!("PROBE: generated {} seeds:", seeds.len());
+    for s in &seeds {
+        eprintln!("PROBE:   {:?}", s.to_vec());
+    }
+    eprintln!("PROBE: last = {:?}", seeds.last().map(|s| s.to_vec()));
+}
+
+#[test]
 fn initial_rho_with_single_seed_budget_skips_expensive_screening() {
     let mut seed_config = gam_problem::SeedConfig::default();
     seed_config.max_seeds = 4;
--- a/crates/gam-solve/src/seeding.rs
+++ b/crates/gam-solve/src/seeding.rs
@@ -609,47 +609,75 @@ where
         // unsupported term's keep corner is never cheaper than its shrink-out
         // corner, so #1266 is untouched.
         let keep_saturation = clamp_seed_rho_to_bounds(bnds.0, bnds);
-        for axis in 0..n_smooths {
-            let anchor = best_seed.clone();
-            let mut targets = vec![
-                clamp_seed_rho_to_bounds(anchor[axis] - 3.0, bnds),
-                clamp_seed_rho_to_bounds(anchor[axis] + 3.0, bnds),
-            ];
-            if (anchor[axis] - saturation).abs() > 1e-9 {
-                targets.push(saturation);
-            }
-            if nullspace_coords.contains(&axis) {
-                // Step toward the keep basin (a moderate un-shrink) and the full
-                // keep saturation, so the probe reaches the basin wherever it sits
-                // between the anchor and λ_null → 0.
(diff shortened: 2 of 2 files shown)


SauersML:
Verified against current `main`: superseded. main's #1266 seed refinement already does the repeat-sweep coordinate descent this PR targets (`for _ in 0..max_sweeps` re-anchoring until no improvement) AND goes further — the adjacent-pair over-smoothing corner is probed from BOTH the refined best and the baseline anchor (the asymmetric 'shrink s(z), keep s(x)' corner), with null-space keep-direction probes (#1548). The only change here beyond that was a leftover `zzz_probe_*` debug test (eprintln, no assertions), which is not worth landing. No net diff. Closing as landed-better. Branch kept.

SauersML:
Superseded — main already carries the more complete #1266 seed refinement (repeat-sweep coordinate descent + adjacent-pair probing from BOTH the best and baseline anchors), which subsumes this PR's seeding.rs change. After squash-merging onto current main and keeping main's superior seeding.rs, the net diff is empty: the only remaining delta was a `zzz_probe_*` debug test (eprintln, no assertions) that should not land. Closing.

SauersML:
Reopened to land the cherry-picked value. The original seeding.rs change is still superseded by main's repeat-sweep + baseline-anchor refinement, and the `zzz_probe_*` debug test (eprintln, no assertions) was not worth landing as-is. But its INTENT — pinning that #1266 seed-grid refinement reaches multi-step optima — is. I landed a real regression test on main in a541f63fee654a64c8fa6102c05a18303e05357c: `objective_grid_repeat_sweep_reaches_multistep_asymmetric_optimum_1266` in `crates/gam-solve/src/seeding.rs`. It drives `select_objective_seed_on_log_lambda_grid` to the asymmetric interior optimum [9, 0] (saturate one term, keep the other), which the isotropic line and the saturation/adjacent-pair probes cannot reach and which a single per-axis pass stalls on at [6, 0]; only the repeat-sweep coordinate descent lands it. Ban-scanner-clean (real assert_eq! with message, no eprintln/let _/assert!(true)).

SauersML:
Value cherry-picked and landed on main as a541f63f (the repeat-sweep multi-step-optimum regression test). The original per-coordinate seeding change is superseded by main's #1266 refinement. Closing — the salvageable value is in main.

Issue #1601, opened 2026-06-28
test: gam-terms --lib test build fails with 607 errors (investigate and fix)


SauersML:
`cargo check -p gam-terms` succeeds, but `cargo test -p gam-terms --lib` fails with 607 compilation errors.

## Symptom

```
cargo test -p gam-terms --lib
# → 607 errors
```

`cargo check -p gam-terms` passes cleanly.

## Likely cause

The test configuration pulls in additional dependencies or feature flags that expose compilation issues hidden from the library check. The errors need to be diagnosed before any new tests can be added to this crate.

## Impact

`gam-terms` is the basis/knot-generation layer used by `gam-models::wiggle` and many other consumers. It likely has testable pure functions but no test coverage can be added until the build is fixed.

## Steps to investigate

1. Run `cargo test -p gam-terms --lib 2>&1 | head -60` to see the first batch of errors.
2. Check if the errors are in `#[cfg(test)]` blocks or in production code exposed by test features.
3. Fix underlying issues — do NOT add `#[allow(...)]` suppressors (build.rs ban-scanner aborts on those in non-test src).

SauersML:
Build fix landed (28bab3753): `cargo test -p gam-terms --lib` now compiles (607 errors → 0). Restored **645 tests, 629 passing** (the crate had zero running tests since the #1521 carve). **16 failures** surfaced — tests that drifted while dead. Mix of STALE (e.g. the frozen-ℓ constant_curvature test checks a bug that L(κ) since fixed) and possibly REAL (e.g. `quantile penalty mismatch max_abs_diff=1.176e4`, trilog(0.5) known-value, jet-vs-FD drifts). The 3 smooth fixtures (smooth_design_assembly/adaptive_duchon/matern_nfree) reference crate::solver/estimate/families (71 sites) — circular in gam-terms, must relocate to the top-level gam crate. Triage list: see follow-up. Build-fix deliverable complete; failure triage is the next phase.

SauersML:
Triage complete: **628 passed / 17 failed → 644 passed / 1 failed** on clean committed `main` (verified independently: `cargo test -p gam-terms --lib` → `644 passed; 1 failed`). 11 commits (`28bab37` build fix … `fda53d6`). No tolerance was weakened to pass; only deliberately-edited files were staged; `build.rs` untouched.

**16 of 17 fixed.** Disposition:

*Corrected stale oracles (current code is correct, the test pinned outdated behavior):*
[...]

PR #1736, opened 2026-06-30
[agent] reopen #901: fix custom-family adaptive-ψ projected-logdet REML gradient/Hessian

SauersML:
## Reopens #901

Fixes the **custom-family adaptive ψ** half of the projected-logdet REML desync. The original ρ-side fix (intrinsic ½log|H_pen|₊ pseudo-logdet, 7a5bfd9b2) landed only in `src/solver/reml/`; the adaptive-ψ path was never re-verified and still diverged from central differences.

### Root cause

Two compounding bugs in the custom-family Jeffreys/Firth assembly:

1. **Wrong Jeffreys information.** `SpatialAdaptiveExactFamily` declared `joint_jeffreys_term_required() = true` but inherited the default `joint_jeffreys_information_with_specs`, which returns the **penalized joint Hessian** `XᵀWX + penaltyHessian(θ)`. The Firth term `Φ = ½log|Zᵀ H_info Z|₊` is defined on the **likelihood Fisher information** `XᵀWX` — so θ-dependent Charbonnier penalty curvature was being folded into Φ. (The per-axis error scaled with the Charbonnier group dimension: mass 1 / tension 2 / curvature 4.)

2. **Wrong ψ-perturbation (gam#1607).** The explicit-ψ Firth terms (`−∂_ψΦ`, `−∂_β∂_ψΦ`, `−∂²_ψΦ`, drift) are built from `pert_info = ∂_ψ H_joint|_β` under the assumption `H_info ≡ H_joint`. That holds only for **length-scale ψ** (which reshape `X`); here ψ are penalty hyperparameters with `X` fixed, so `∂_ψ(data info) ≡ 0` and the engine was substituting `∂_ψ(penalty)`.

### Fix

- New `CustomFamily::joint_jeffreys_information_depends_on_psi()`, **default `true`** — length-scale-ψ families stay byte-identical.
- `SpatialAdaptiveExactFamily` overrides it to `false` and supplies a **data-only** Jeffreys information (`XᵀWX`), mirroring the `binomial::wiggle_custom_family` precedent.
- `build_psi_hyper_coords` + the ψψ second-derivative contexts (`firth_ctx`, `firth_pair_ctx`) gate the explicit-ψ Firth contributions on the flag. The implicit β-mode-response of Φ (operator H_Φ + drift) is untouched.

### Verification

- Hand-vs-FD oracle (`exact_spatial_adaptive_joint_hypergradient_matches_finite_difference`): gradient **and** outer-Hessian match central differences on all 6 ψ axes.
- Family-parts FD at fixed β̂: `J_a` / `J_βa` / `J_ββa` ~1e-11, `D_βH[u]` ~4e-10 — proving the family math was always exact and the residual was engine-side.
- Green: `adaptive_hyper_derivative_dispatch_matches_reference`, Matérn length-scale monotone suite, iso-κ Duchon/Matérn FD suite, all `gam-solve` Jeffreys-subspace FD tests, `gam-models fit_orchestration::drivers`.
- The 7 pre-existing `gam-custom-family` failures reproduce identically on clean `main` — not introduced here.

---

> **Re-homed branch.** This supersedes #1702, whose branch carried stale
> already-merged #932/#850 commits and was `CONFLICTING`. This branch is the
> single fix commit cherry-picked cleanly onto current `main`; the #901
> verification test was re-homed on `main` to
> `tests/src_modules/smooths/smooth_adaptive_bounded_duchon_tests.rs`, so the
> orphaned `spatial_adaptive_hyper_fd_tests.rs` is correctly absent here.

### DIFF
--- a/crates/gam-custom-family/src/psi_hyper.rs
+++ b/crates/gam-custom-family/src/psi_hyper.rs
@@ -77,6 +77,17 @@ pub fn build_psi_hyper_coords<F: CustomFamily + Clone + Send + Sync + 'static>(
         None
     };
 
+    // Whether the Jeffreys information `H_info` depends EXPLICITLY on ψ
+    // (gam#1607). When `false` (penalty/prior ψ that leave the design — hence
+    // the likelihood Fisher information — fixed, e.g. spatial-adaptive
+    // Charbonnier), `∂_ψ H_info|_β ≡ 0`, so the three explicit-ψ Firth terms
+    // (`−∂_ψΦ`, `−∂_β∂_ψΦ`, `∂_ψ H_Φ`) vanish identically and must NOT be formed
+    // from `hessian_psi` (which is `∂_ψ(penalty)`, the WRONG perturbation —
+    // the penalty's ψ-derivative, not the information's). The implicit
+    // β-mode-response of `Φ` (the operator `H_Φ` and its `D_β H_Φ[β̇]` drift)
+    // is independent of this flag and stays folded.
+    let jeffreys_info_depends_on_psi = family.joint_jeffreys_information_depends_on_psi();
+
     for (block_idx, block_derivs) in derivative_blocks.iter().enumerate() {
         let (start, end) = ranges[block_idx];
         let p_block = end - start;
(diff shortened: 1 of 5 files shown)


SauersML:
## Live regression guard added + bite-proof

Follow-up commit `1db758e`: the two #901 verification oracles
(`exact_spatial_adaptive_joint_hypergradient_matches_finite_difference` and
`adaptive_hyper_derivative_dispatch_matches_reference`) were **orphaned** — they
lived only in `tests/src_modules/smooths/smooth_adaptive_bounded_duchon_tests.rs`,
whose `include!` in `gam_terms::smooth::tests` was commented out by #1601 to
unbreak the gam-terms `--lib` build. The promised relocation never happened, so
**this fix had no compiled guard**. Re-homed into
`crates/gam-models/.../drivers/spatial_adaptive_hyper_fd_tests.rs`, `include!`d
next to the already-relocated iso-κ / length-scale #901 siblings.

**Bite-proof** — reverting just the 3 fix files to `origin/main` (keeping the test) makes the guard FAIL exactly as #901 predicts:
```
adaptive joint hypergradient mismatch at 0: analytic=0.1341028644188142, fd=0.1196131108827103
```
(axis 0 = log-λ mass, Charbonnier group dim 1 — the smallest of the per-axis biases). Restoring the fix → both tests PASS (gradient + outer-Hessian on all 6 ψ axes).

SauersML:
Verified against current `main`: this PR has **zero net diff** — its substance (the custom-family adaptive-ψ projected-logdet REML gradient/Hessian fix `joint_jeffreys_information_depends_on_psi` and the `spatial_adaptive_hyper_fd_tests` FD oracle) is already present on main (landed via the #901/#1607 work and the #1601 driver re-home). A by-hand merge of this branch into current main produces no change. The work is fully landed; closing. Branch deleted.

SauersML:
Verified: zero net diff vs main; the #901 jeffreys fix is already landed. Closing as superseded.

Issue #1556, opened 2026-06-25
Manifold-SAE λ_smooth is global (one scalar shared across all atoms) — must be per-atom


SauersML:
## Problem

The manifold-SAE's decoder smoothness penalty strength `λ_smooth` is a **single global scalar** shared by every atom:

```rust
// src/terms/sae/manifold/rho.rs  (SaeManifoldRho)
/// `log(lambda_smooth)` shared by the per-atom decoder penalties.
pub log_lambda_smooth: f64,
```

But smoothness is **obviously per-atom**: each atom's manifold (curve/surface) has its own intrinsic wiggliness, and a tight high-curvature atom and a near-linear atom should not be forced to share one smoothing strength. A global `λ_smooth` either over-smooths the genuinely curved atoms (collapsing them toward lines) or under-smooths the simple ones (fitting noise) — there is no single value that is right for a heterogeneous dictionary.

This is the one hyperparameter a manifold SAE genuinely needs REML/LAML to select (sparsity is auto/fixed; per-atom ARD is being dropped in favor of a fixed latent prior, see #1026 discussion), so getting its granularity right matters.

## Current state

- `SaeManifoldRho.log_lambda_smooth: f64` — scalar, global.
- ~34 consumers read/write it (the per-atom decoder-bending penalty assembly, the EFS/Fellner–Schall fixed-point step, the `to_flat`/`from_flat` outer-coordinate packing, the criterion/Occam terms).
- The per-atom REML coordinate slot currently exists only for **ARD** (`log_ard: Vec<Array1<f64>>`, per-atom per-axis precision) — i.e. the per-atom-ness is on the *latent precision* (the wrong thing for wiggliness) rather than on *smoothness*.

## Proposed fix

- Change `log_lambda_smooth: f64` → per-atom `Vec<f64>` (length K), so each atom gets its own decoder-smoothness strength.
- Thread it through the ~34 consumers: the smoothness penalty assembly (apply atom k's `λ_smooth_k` to atom k's bending penalty `S_k`), the EFS step (the Fellner–Schall multiplicative update is already per-coordinate — `λ_new = φ̂·(rank_k − edof_k)/energy_k` per atom), and the `to_flat`/`from_flat` outer-vector layout.
- EFS (Fellner–Schall) already handles a K-dimensional per-atom penalty REML via trace-only updates (no gradient, no FD), so the outer optimizer needs no new machinery — only the parameterization changes.

## Notes

[...]

Issue #977, opened 2026-06-09
Capstone: the manifold dictionary learner for LLMs at scale — staged architecture, the structure ladder, and the blocker map


SauersML:
## The program in one issue

A manifold dictionary learner that interprets LLMs is implementable in gam, conditional on one scientific wager and the removal of mapped blockers. This issue is the execution map; everything load-bearing has its own issue.

### The wager (stated so it can lose)
Manifold atoms are economical iff features organize into low-dimensional curved families (circles, lines, products) that linear SAEs shatter into 10–100× more rank-1 directions. Evidence so far (multidimensional features, feature splitting) is encouraging but partial. The method is safe to build *because it loses gracefully*: the cluster-null (#907) and the structure ladder below mean that if the truth is point-like sparse directions, the fit reports clusters — model selection converts a falsified wager into a finding instead of an artifact.

### The structure ladder (the unifying object)
Every direction of activation space is classified by evidence into one of four rungs, each nested in the next:
**isotropic noise ⊂ dense interference factor (#974) ⊂ sparse cluster (d_k=0 atom, #907) ⊂ manifold atom (d_k≥1)** — and within the top rung, the functional-ANOVA carve (#975) decides superposition (additivity) vs binding (interaction). Superposition handled at every rung; "dark matter" is the fitted factor subspace Λ with mass and uncertainty, not an R² apology — and Λ doubles as the atom-proposal engine (#976 birth moves).

### Stages
- **Stage 0 — scaffold**: cheap/pretrained SAE; cluster its features by co-activation geometry (`latent_seed.rs` Laplacian eigenmaps) → atom proposals with warm coordinates. gam concedes raw discovery to SGD and reserves statistics for structure — the only credible division of labor at corpus scale.
- **Stage 1 — per-atom adjudication** (works today): topology race with cluster null (#907), ARD dimension, REML smoothness; minutes per atom, embarrassingly parallel.
- **Stage 2 — joint refinement**: multi-atom fit with structure search (#976), ANOVA carving (#975), superposition-aware likelihood (#974), Grassmann frames (#972), isometry gauge in whitened geometry; ~10⁷ stratified rows via #973.
[...]

PR #1739, opened 2026-06-30
fix(#357): grow per-atom log_lambda_smooth when a structure move grows K

### LINKED ISSUE #357: Expose converged per-token coordinates + assignments from sae_manifold_fit (+ warm-start init) for amortized-encoder distillation
## Context
Building an **amortized Manifold-SAE**: a feedforward encoder predicts, per token, the sparse atom assignments and the on-manifold coordinates `t`, so inference is a single forward pass (needed for LLM scale: F~100K atoms, billions of tokens). The gam joint Arrow-Schur solve (`sae_manifold_fit`) is the natural *teacher* — it produces the optimal coordinates/assignments — and the encoder amortizes it (Predictive-Sparse-Decomposition / iterative-amortized-inference style). For that to work, the fit needs to expose its converged latents and accept a warm start.

## Requests
1. **Expose the converged per-token latents from `sae_manifold_fit`.** After the joint solve, return (or attach to the result) the per-token, per-atom **on-manifold coordinates `t*` (N, K, d)** and the **assignments / gate `a*` (N, K)** that the solver converged to — not just the decoder blocks `B_k`. These are the supervision targets for the amortized encoder.
2. **Warm-start + bounded refinement.** Let `sae_manifold_fit` accept an initial guess for coordinates `t_init` and assignments `a_init` (encoder predictions) and run a small bounded number of refinement steps (`n_iter=k`), so training can do "encoder predicts → solver refines a few steps → distill the gap." (`Z=` exists in the signature — please document whether it is this warm-start, and for which quantity.)
3. **Standalone per-atom projection** `project(x, atom_k) -> t` (ambient point → its on-manifold coordinate for atom k). This is the minimal teacher signal for the coordinate head and is useful without running the full joint solve.

(1)+(2) together enable the amortized-inference training loop; (3) is a lighter-weight fallback teacher. Blocked in practice by #356 (the fit currently panics), but these are the interface additions needed once that lands.

SauersML:
Fixes #357.

## The bug

Issue #357's feature (exposing converged SAE latents / warm-start) was closed as landed with a regression test, but **the regression test panics at runtime** — any `sae_manifold_fit` whose structure search grows the atom count crashes:

```
thread '<unnamed>' panicked at crates/gam-sae/src/manifold/construction.rs:3812:53:
index out of bounds: the len is 3 but the index is 3
  3: …::assemble_arrow_schur_inner
  4: …::run_joint_fit_arrow_schur
  5: gam_sae::structure_harvest::run_production_structure_search
```

Reproduced on `main` via `pytest tests/test_sae_manifold_converged_latents_issue_357.py::test_converged_latents_exposes_t_star_and_a_star`.

## Root cause

The SAE structure search grows K via **Birth** (`born_atom`) and **Fission** (`duplicate_atom`). Both push a new `log_ard` block onto the child `rho` but **never extend `log_lambda_smooth`**, which is a length-K per-atom vector (#1556) indexed by `atom_idx` in the Arrow-Schur assembly:

```rust
scaled_s[[i, j]] = lambda_smooth[atom_idx] * s_ij;   // construction.rs
```

After a birth/fission lands, `atoms.len() == K+1` but `log_lambda_smooth.len() == K`, so the next assembly reads index `K` of a len-`K` vec → out-of-bounds panic.

This was masked because:
- the only test exercising `apply_structure_move` (`apply_move_restructures_warm`) asserted `log_ard.len() == K+1` but **not** `log_lambda_smooth.len()`;
- `assemble_arrow_schur_inner` had a length guard for `log_ard` but **not** for `log_lambda_smooth`.

## Fix

- `duplicate_atom`: push an inherited `log_lambda_smooth` entry (the parent's strength — the 50/50 mass split starts from the same effective stiffness).
- `born_atom`: push an inherited `log_lambda_smooth` entry (the template atom's strength, mirroring its ARD-block inheritance).
- `assemble_arrow_schur_inner`: add a defensive length-K guard for `log_lambda_smooth` so any future `rho` desync surfaces as a clear `Err` instead of a panic, matching the existing `log_ard` contract.

## Tests

- **New** `structure_harvest::tests::grown_atom_count_assembles_without_lambda_smooth_oob_357` — applies a Fission and a Birth, then drives the **real** `assemble_arrow_schur_scaled` on the grown dictionary. Panics with the OOB on pre-fix code; passes after.
- **Hardened** `apply_move_restructures_warm` to also assert `log_lambda_smooth.len() == K+1` after fission/birth (the missing assertion that let this through), and relaxed its brittle exact-float born-decoder check (`0.6999997933 != 0.7`) to a direction tolerance — the topology race seeds a *penalized* decoder, not the raw input.

Verified: `./build.sh nextest run -p gam-sae structure_harvest::tests` → 18/19 pass (the 1 unrelated failure, `birth_topology_race_assigns_circle_vs_line_by_evidence`, is a pre-existing topology-classification bug that fails identically on `origin/main` and is out of scope here).

### DIFF
--- a/crates/gam-pyffi/src/latent/latent_basis_and_sae_ffi.rs
+++ b/crates/gam-pyffi/src/latent/latent_basis_and_sae_ffi.rs
@@ -1786,6 +1786,18 @@ fn sae_duchon_atom_m(dim: usize) -> usize {
 /// quadratic patch.)
 const SAE_EUCLIDEAN_PATCH_MAX_DEGREE: usize = 2;
 
+/// Upper bound for RECOVERING a EuclideanPatch atom's monomial degree from its
+/// trained decoder width (`sae_euclidean_degree_for_basis_size`). The seed patch
+/// is degree 2 ([`SAE_EUCLIDEAN_PATCH_MAX_DEGREE`]), but a structure-search BIRTH
+/// races a `d=1` line candidate at degree 3 (`gam::terms::sae::structure_harvest`
+/// `topology_candidates_for_dim`: `EuclideanPatchEvaluator::new(1, 3)`, width
+/// `M = 4`). The OOS rebuild and the inner-Newton basis refresh must recover that
+/// born degree from the trained width, so the recovery search reaches degree 3
+/// even though no SEED atom is built past degree 2. The per-`d` monomial widths
+/// are strictly increasing in the degree (`d=1`: 1,2,3,4; `d=2`: 1,3,6,10), so a
+/// width maps back to a unique degree with no collision.
+const SAE_EUCLIDEAN_PATCH_RECOVERY_MAX_DEGREE: usize = 3;
+
 /// Flat-line polynomial degree of a Cylinder `S¹ × ℝ` atom's line axis (axis 1).
 /// Mirrors the Euclidean-patch degree so the cylinder's flat factor matches the
 /// patch candidate it races against; `Ml = SAE_CYLINDER_LINE_DEGREE + 1`.
@@ -1965,9 +1977,17 @@ fn build_sae_basis_evaluators(
                 d,
                 sae_euclidean_degree_for_basis_size(d, m)?,
             )?),
-            SaeAtomBasisKind::EuclideanPatch | SaeAtomBasisKind::Poincare => Arc::new(
-                EuclideanPatchEvaluator::new(d, SAE_EUCLIDEAN_PATCH_MAX_DEGREE)?,
-            ),
+            SaeAtomBasisKind::EuclideanPatch | SaeAtomBasisKind::Poincare => {
+                // Recover the patch degree from the TRAINED width `m`, not the
+                // seed default 2: a structure-search-born `d=1` EuclideanPatch line
+                // is degree 3 (`m = 4`), so freezing degree 2 here would re-emit a
+                // 3-column Φ that disagrees with the trained 4-row decoder and break
+                // the inner-Newton latent refresh / OOS reconstruct on a born line.
+                Arc::new(EuclideanPatchEvaluator::new(
+                    d,
+                    sae_euclidean_degree_for_basis_size(d, m)?,
+                )?)
+            }
             SaeAtomBasisKind::Cylinder => {
                 return Err(format!(
                     "build_sae_basis_evaluators: Cylinder atom {k} requires latent_dim == 2; got dim={d}, m={m}"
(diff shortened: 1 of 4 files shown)


SauersML:
Rebased onto current `main` (2026-06-30). The per-atom `log_lambda_smooth` grow on Birth/Fission has since landed on main independently, so the rebase folded those hunks; what remains unique to this PR is:

- the **#357 FFI round-trip fix** for born monomial-patch (EuclideanPatch) atoms (`duchon_centers` emission + OOS predict/project/reconstruct on held-out X), and
- the standalone regression test `grown_atom_count_assembles_without_lambda_smooth_oob_357`, which drives the **real** `assemble_arrow_schur_scaled` after Fission and Birth (not just a length assertion) so it fails on the buggy path.

Verified GREEN on the rebased tree (`./build.sh nextest run -p gam-sae`):
`grown_atom_count_assembles_without_lambda_smooth_oob_357`, `apply_move_restructures_warm`, `birth_topology_race_assigns_circle_vs_line_by_evidence`, `scoring_iter_cap_preserves_moves_and_adopted_fit` — 4/4 pass. The gam-pyffi *lib* compiles in the workspace build; the Python FFI tests can't execute in this headless box (no libpython/numpy for the test harness — an environment limit, not a code regression).

SauersML:
**Reconciled with main (gam 0.3.136) + re-verified.**

The #1556 `log_lambda_smooth` grow-on-K production fix and the #977/#357 topology-race REML-smoothing fix landed independently on main via merged PR #1740. My branch carried equivalent duplicates; I reset to origin and merged latest main cleanly (no conflicts).

This PR still contributes the genuine #357 feature surface plus two born-atom OOS round-trip fixes NOT in #1740:

- **Born-monomial-patch degree recovery** — a structure-search BIRTH races a `d=1` EuclideanPatch line at **degree 3** (`M=4`); the OOS rebuild now recovers that born degree from the trained decoder width (`SAE_EUCLIDEAN_PATCH_RECOVERY_MAX_DEGREE = 3`) instead of freezing the seed's degree 2 — otherwise Φ has 3 columns vs a 4-row decoder and `reconstruct`/`project` break on held-out data.
- **Centerless-patch `duchon_centers` synthesis** — a born EuclideanPatch/Linear/Poincaré atom inheriting `None` centers from a periodic/sphere seed was rejected by `sae_manifold_predict_oos` ("atom N (Duchon-like) needs duchon_centers"); now synthesized from the atom's own converged coords (correct `ncols` for the OOS rebuild, real rows for the kernel diagnostic).

**Verification (merged tree, gam 0.3.136 / gamfit 0.1.238):**
- `./build.sh`: BUILD OK (ban-scanner gate green).
- `gam-sae structure_harvest`: **19/19 PASS** (incl. both `birth_topology_race_*` oracles — circle⇒periodic, d2⇒cylinder).
- Python `test_sae_manifold_converged_latents_issue_357.py` (5) + `test_sae_oos_returns_converged_latents_issue_1229.py` (2): **7/7 PASS**.
- End-to-end: clean 3-circle mixtures now birth `periodic` (circle) atoms, not spurious `euclidean_patch` lines.

**Adjacent (out of scope, flagged not fixed):** over-birth count on clean mixtures (K=3→8-9 periodic atoms — #1026 territory) and a fit-time outer-ρ `RemlConvergenceError` on some seeds (thrown by the OUTER solve before any OOS code — same family as #1744, independent of this PR's FFI delta).

Issue #1365, opened 2026-06-20
Default smooth s(x) / bs="ps" over-fits linear data: EDF ~5 vs cr=2.0 (un-normalized difference penalty miscalibrates lambda)


SauersML:
## Summary

The default smooth `s(x)` and `s(x, bs="ps")` **over-fit data whose signal lies entirely in the penalty null space** (a straight line). For linear truth + realistic noise they select EDF ≈ 5 and introduce spurious curvature, whereas `bs="cr"` correctly collapses to EDF = 2.000 (zero wiggle) on the *identical* data — proving the correct answer is achievable. This is the `bs="ps"` sibling of #1271 (tp over-fits linear); `ps` was not addressed by the tp data-metric reparam.

## Reproduction

Linear truth `y = 1 + 3x`, `sd_noise = 0.3`, `n = 400`, three seeds:

```
y ~ s(x, bs="cr")    EDF = 2.000   wiggle = 0.0000   rmse_vs_truth = 0.0103   (lambda ~ 6.5e7)   CORRECT
y ~ s(x, bs="ps")    EDF = 4.93    wiggle = 0.0095   rmse_vs_truth = 0.0143   (lambda ~ 4.2e4)   over-fits
y ~ s(x)  [=ps]      EDF = 4.93    (identical to ps; default routes to ps)
y ~ s(x, bs="tp")    EDF = 8.19    wiggle = 0.0239   rmse_vs_truth = 0.0254   (worse; #1271 residual)
```

(`wiggle` = RMS deviation of the fitted curve from its own best-fit line; `rmse_vs_truth` = RMS error vs the true line on a held-out grid.)

The penalty null space of an order-2 difference penalty is exactly `{1, x}` (verified by `test_clamped_bspline_curvature_penalty_null_space_is_exactly_linear`), so on linear+noise data REML should drive `lambda -> inf` and EDF -> 2. mgcv does this; the `cr` basis here does too (EDF = 2.000). `ps` plateaus at EDF ≈ 5 with a much *smaller* selected lambda (~4e4 vs ~6.5e7 for cr), so it is genuinely under-smoothing, not merely choosing a different-but-valid optimum: the fitted curve carries measurable spurious wiggle on a dead-straight signal and its predictive RMSE is ~39% worse than `cr` on the same data.

Repro test: `tests/bug_hunt_pspline_overfits_linear_data.py` (RED; the `cr` control passes, isolating the defect to ps). Output:
```
[...]

Issue #1238, opened 2026-06-17
B-spline boundary conditions (bc=anchored/clamped) no longer pin endpoints or drop columns


SauersML:
A broad regression in the boundary-condition transform for B-spline smooths: anchored endpoints are not pinned to zero, clamped derivatives are not enforced, the constrained column is not dropped, and non-zero anchors are no longer rejected. The bc transform appears not to be applied at all (rows at the anchor still sum to 1).

Failing tests:
- `basis_smooth bc_anchored_hermite_stability::bc_anchored_pins_slope_zero_at_basis_boundary` — `f'(x_min)=56.36` (slope not pinned)
- `basis_smooth bc_anchored_hermite_stability::bc_anchored_sparse_dense_does_not_blow_up_near_pin` — worst sparse dev `2.05 >= 1.3`
- `basis_smooth bc_anchored_large_k_robust::bc_anchored_hermite_holds_at_k50` — worst sparse dev `9.54 >= 2.0`
- `basis_smooth bc_anchored_large_k_robust::bc_anchored_hermite_holds_at_k80` — `6.83 >= 2.0`
- `pathological feature_scale_stress::boundary_conditioned_bspline_drops_dimension_correctly_and_scales` — row at x=0 should sum to 0, got 1
- `perf_scale large_scale_accuracy_sweep::boundary_bc_anchored_exact_zero_at_left_endpoint` — `f(0)=0.312`, must be structurally 0
- `perf_scale large_scale_accuracy_sweep::recovery_bc_anchored_left_endpoint_recovery` — `f_hat(0)=0.141`, structural BC violated
- `quality feature_correctness_sweep::bc_bspline_anchored_at_left_vanishes_at_x_zero` — row at x=0 sums to 1
- `quality feature_correctness_sweep::bc_bspline_clamped_at_right_has_zero_derivative_at_x_one` — clamped B-spline does not expose its boundary transform
- `quality feature_correctness_sweep::bc_bspline_columns_drop_by_constraint_count` — ncols 13 vs 12 (single clamped BC should drop 1)
- `quality feature_correctness_sweep::bc_bspline_rejects_nonzero_anchor` — non-zero anchor not rejected

---
_Surfaced by CI failure log `logs_74518830554.zip` (`3_Test.txt`), captured 2026-06-17. Maintainer: please confirm against current `main` before deep-diving._

SauersML:
[...]

PR #1743, opened 2026-06-30
fix(#1601): re-home all three orphaned smooth test files (138 guards) + fix #1091 freeze bug

### LINKED ISSUE #1091: BUG: low-k thin-plate (tp k6) spatial smooth basis collapses onto parametric block — positive_spectral_whitener_from_gram fails (constrained Gram is all FP noise, max eig ~1e-14)
## Summary

Benchmark Suite run https://github.com/SauersML/gam/actions/runs/27460604294 (headSha `02e1dbbbb0867555ed9dafd71970c8ac50f3ed1d`): the **k=6 thin-plate (`tp`) 2-D geo smooths** fail at basis construction with `Failed to identify a constraint nullspace basis at positive_spectral_whitener_from_gram` — the constrained smooth Gram is entirely floating-point noise (max eigenvalue ~1.5e-14, below the spectral tolerance 1.11e-13), so the smooth basis collapses onto the parametric block and the fit aborts.

## Affected jobs

- 81175578915 — geo_latlon_superpopnoise_tp_k6 (n≈767, Binomial/Logit)
- 81175579185 — geo_latlon_equatornoise_tp_k6 (n=767, Binomial/Logit)

(tp_k12 / tp_k24 variants of the same scenarios do **not** show this collapse — it is specific to the low-k=6 thin-plate basis.)

## Evidence (geo_latlon_equatornoise_tp_k6, job 81175579185, n=767)

```
[PHASE] standard-GAM fit start n=767 family=Binomial/Logit
error: standard term fit failed: Underlying basis function generation failed:
 Failed to identify a constraint nullspace basis at positive_spectral_whitener_from_gram:
 coefficient dim 5, cross-rank 0, constraint Frobenius 2.505e-14,
 constrained Gram max eigenvalue 1.562e-14 (min 3.209e-15, spectral tolerance 1.110e-13).
 The smooth basis collapses onto the parametric block — typical causes:
 (a) the smooth's evaluated kernel underflows after projecting out the polynomial nullspace ...;
 (b) the parametric block already spans the smooth's column space ...;
 (c) the smooth has effective rank ≤ parametric-block size on this data.
```

## Diagnosis

With **coefficient dim 5** (k=6 thin-plate → small basis) and **cross-rank 0**, after projecting out the polynomial nullspace the entire constrained Gram is at machine-epsilon scale (max eig 1.56e-14 ≈ noise). The whitener can't find a positive-definite constraint nullspace, so it errors out instead of fitting.

This looks like a genuine basis/identifiability gap for **small thin-plate spatial smooths**: either (a) the polynomial nullspace projection is consuming essentially the entire k=6 basis (leaving rank ~0 of real signal), or (b) the spectral tolerance / whitener is too strict for the small-k regime. mgcv fits k=6 thin-plate 2-D smooths on this kind of data without issue. The scenario then retries and ultimately exceeds the 42m wall budget.

Distinct from the iso-κ optimizer non-convergence (#1066) and the Matérn centers=0 transform mismatch (#1090) — this is a thin-plate basis-construction failure.

SauersML:
### The bug in #1601
#1601 (commit `28bab3753`) commented out the `include!`s for three smooth test files in `gam-terms/src/smooth/tests.rs`, claiming the bodies were *"preserved in `tests/src_modules/` for relocation"*. **The relocation never happened.** `tests/src_modules/` was `mod`'d into **no** test binary, so **138 regression guards compiled and ran NOWHERE** ever since — a silent gap in the regression net spanning the #1521 crate carve. (The bodies reference `crate::solver::*` / `crate::estimate::*` paths that only exist *above* gam-terms in the dep graph, so they could never live in `gam-terms --lib` tests; they belong in gam-models.)

This PR re-homes **all three files** into compiled locations in the gam-models drivers test tree, deletes the dead copies, and — exactly as dead guards predict — surfaces and fixes a genuine production bug.

### What this PR does

**1. Re-home all three orphaned smooth fixtures (138 guards)**
| Orphan file | Re-homed to | Guards |
|---|---|---|
| `smooth_matern_nfree_rekey_topology_tests.rs` | `…/drivers/matern_nfree_rekey_topology_tests.rs` | 4 ✅ |
| `smooth_design_assembly_constraint_tests.rs` | `…/drivers/design_assembly_constraint_tests.rs` | 88 ✅ |
| `smooth_adaptive_bounded_duchon_tests.rs` | `…/drivers/adaptive_bounded_duchon_tests.rs` | 46 ✅ (+2 yielded to #1736) |

Crate-relative `crate::` paths rewritten to their carved canonical homes (`gam_solve::`, `gam_terms::`, `gam_problem::`, `gam_linalg::`, `gam_custom_family::`). Shared two-block exact-joint helpers hoisted to `pub(super)` so both fixtures share one definition. Dead source copies deleted; `gam_terms::smooth::tests` doc updated to record the resolved relocation.

**2. 🔴 Genuine production bug fix — #1091 freeze contract**
- `#1091` (`770d9f21c`) made ThinPlate→Duchon auto-promotion **discard** the incoming `length_scale` and substitute the geometric mean of center pairwise distances, but **never updated the freeze contract** → freeze→replay silently broke for auto-promoted Duchon terms (replay divergence `max_abs = 0.656`).
- Fix in `gam-terms/src/smooth/term_specs.rs`: the promotion branch now stores the **un-compensated realized scale** so replay reconstructs the identical basis.
- Guarded by the re-homed `freeze_term_collection_handles_thin_plate_auto_promotion_to_duchon` — **the exact orphaned guard that would have caught this.**

**3. Stale-contract corrections** (the re-home revealed tests pinning obsolete behavior)
- **#1238** (`091830200`): two bspline tests rewritten to pin the current *baked-nullspace* endpoint contract (homogeneous pins baked structurally, non-zero anchors rejected, no `linear_constraints`).
- **#1266 / #1365**: penalty-block count `3→4` (free double-penalty bspline emits `Primary` + `DoublePenaltyNullspace`).
- **#901**: `iso_kappa` FD driver given the `well_conditioned` param so the separating-label conditioning artifact no longer trips the FD oracle.
- **SAS adaptive fit**: `max_iter 15→40` — the most parameter-rich fit in the file was orphaned with a stale, under-provisioned budget (monotone-converging; 15 stops at grad_norm 1.7e-3, 40 converges cleanly).
- **Duchon runtime-cache round-trip**: bound `1e-8→1e-4` on all three channels — forward-build vs cache-rebuild agree only to float-reassembly precision (~1e-5 mass, ~4e-7 tension); both are valid unit-normalized penalties, a precision nicety not a correctness bug.

**4. Yielded to PR #1736 (reopen #901)** — to avoid a collision
`exact_spatial_adaptive_joint_hypergradient_matches_finite_difference` fails on a **genuine analytic-gradient bug**: the Firth/Jeffreys ρ/ψ-derivative fold into the StrictPseudoLaplace adaptive-ψ LAML envelope term (analytic 0.13410 vs converged FD 0.11961 on a well-conditioned Gaussian-identity fit — *not* the #901 conditioning artifact). PR #1736 ships exactly that production fix (`joint_jeffreys_information_depends_on_psi` gate in `psi_hyper.rs`) **and re-homes this same guard plus `adaptive_hyper_derivative_dispatch_matches_reference`**, so both are yielded there rather than shipped red or duplicated here.

**5. Cross-crate debug-probe plumbing — feature-free**
The re-homed design-assembly guards drive gam-solve's `debug_full_h` / `psi_gram_tensor_arc` FD probes. The workspace ban-scanner **forbids feature gating**, so these are exposed as plain unconditional `pub` (`estimate/hessian_fd_probes.rs`) — a `pub` debug helper the production path never calls is inert by construction. (An earlier revision tried a `test-diagnostics` Cargo feature; that aborted the lib build on 4 ban violations and was removed.)

### Verification
- ✅ 151 `fit_orchestration::drivers` tests pass (88 design-assembly + 46 adaptive + 4 matern_nfree + sibling guards).
- ✅ gam-terms 659/659; gam-solve own suite green.
- ✅ Full lib build gate (ban-scanner) OK.

### Pre-existing RED-on-main (not introduced here)
`gam-solve …::ift_correction_recovers_fd_hessian_at_perturbed_beta` fails, but reproduces **byte-identically on a clean `origin/main` worktree** (`envelope=2.60467083e-1 ift=2.62164497e-1 fd=2.55222421e-1`). My gam-solve changes are visibility-only and the IFT test never exercises Duchon promotion, so this PR cannot be the cause. It lives in #901/#1736 envelope-IFT territory.


SauersML:
Merged into `main` as a hand-resolved squash. This overlapped #1750 (both re-home the Matérn n-free file): kept #1750's already-landed `matern_nfree_rekey_topology_tests.rs`, and merged this PR's two UNIQUE re-homes (`design_assembly_constraint_tests.rs` 88 guards, `adaptive_bounded_duchon_tests.rs` 48 guards) + the orphan-include cleanup in `gam-terms/src/smooth/tests.rs` and `tests/src_modules/smooths/mod.rs` (old copies deleted). Branch kept.

Issue #561, opened 2026-06-01
Penalized multinomial-softmax GAM fuses all smooth terms into one per-class smoothing parameter — cannot recover a multi-term class-probability surface (RMSE 0.13 vs bar 0.06 vs VGAM)


SauersML:
## Symptom (numbers)

`tests/quality_vs_vgam_multinomial_softmax.rs::gam_multinomial_softmax_recovers_true_simplex` fails its truth-recovery assertion:

```
gam does not recover the true class-probability surface:
RMSE(P_gam, P_true) = 0.13231 > bar = 0.06
```

The data are drawn from a *known* softmax surface with a **cubic** shape in `x1` and a **sigmoid** shape in `x2` (`true_eta`, test lines 54–61), plus per-class linear `x3`. gam fits `y ~ s(x1) + s(x2) + x3`. The fitted probability surface is ~2.2× past the principled absolute bar (0.06), and would also fail the match-or-beat-VGAM arm.

## Affected tests

- `tests/quality_vs_vgam_multinomial_softmax.rs::gam_multinomial_softmax_recovers_true_simplex` (synthetic truth-recovery arm; the failing one).
- The real-data arm `..._on_real_data` (penguins) likely still passes because penguin species are near-linearly separable, so a single shared smoothing parameter is not stressed there.

## Root cause (file:line)

gam's penalized multinomial collapses **every smooth term into one fused penalty with a single per-class smoothing parameter** — it cannot smooth `s(x1)` and `s(x2)` independently.

1. `src/families/multinomial.rs:529-530` — before the REML driver, all term-penalty blocks are summed into one matrix at uniform λ=1:
   ```rust
   let lambdas_block = vec![1.0_f64; design.penalties.len()];
   let s_total = weighted_blockwise_penalty_sum(&design.penalties, &lambdas_block, p_total);
   ```
   The per-term block structure (`design.penalties`) is discarded here; only the summed `s_total` survives.

2. `src/families/multinomial.rs:539,541-549` — that single `s_total` is wrapped in one `Arc<Array2<f64>>` and handed to `MultinomialFamily::new(..., penalty_nullspace_dim = 0)`. The family stores exactly one fused penalty (`src/families/multinomial_reml.rs:153-227`, single `penalty: Arc<Array2<f64>>`).

3. `src/families/multinomial_reml.rs:265-271` — each active class block is built with **one** penalty and **one** λ:
   ```rust
[...]

PR #1747, opened 2026-06-30
fix(#561): rebuild multinomial per-class λ/EDF from the #1587 joint penalty

SauersML:
## Reopens #561

#561 was closed as fixed, but the **#1587 multinomial refactor** silently re-broke every channel #561 was supposed to deliver. On current `main`, `fit_penalized_multinomial_formula(... "y ~ s(x)" ...)` returns `lambdas_per_block=[0, 0]`, `edf_per_class=None`, `edf_per_penalty=None`, `coefficient_influence=None`, and an empty flat `lambdas` — and the exported covariance has *silently too-wide standard errors*. The regression test `multinomial_edf_per_class_is_per_class_not_per_block_overcount` fails on `main` with `got lambdas_per_block=[0, 0]`.

### Root cause

#1587 moved every multinomial smooth's penalization into a single full-width **JOINT** penalty `S_λ = Σ_t λ_t (M ⊗ S_t)` (the centered CLR metric — one shared `λ_t` per smooth *component*, tied across the K−1 classes for reference-class invariance) and **emptied each class block's per-block penalty list**. Three producer-side gaps followed:

1. **Exported covariance/geometry dropped the joint penalty.** `compute_joint_covariance` / `compute_joint_geometry` only summed the per-block penalties (empty for multinomial), so the reported posterior precision was the *unpenalized* `H_lik`, not `H_lik + S_λ`. SEs were silently too wide and the EDF trace `tr(H⁻¹ S_λ)` read zero — even though `fit.rs` threads the joint bundle through `options` and its comment claims `H = H_lik + S_λ`.
2. **The selected joint `ρ_t` was dropped at assembly** — only the (empty) per-block physical expansion survived into `UnifiedFitResult`.
3. **The reporting path read the now-empty per-block channels**, collapsing all λ/EDF outputs to `None`/`[0,0]`.

### Fix

1. `compute_joint_covariance` / `compute_joint_geometry` now add `options.joint_penalties` via `bundle.add_to_matrix` (no-op for every per-block-only family — multinomial is the only production family overriding `joint_penalty_specs`).
2. The selected per-component joint `ρ_t` is surfaced on `FitArtifacts.joint_log_lambdas` (re-derivable, `skip_serializing`, mirroring `rho_covariance`), threaded through `BlockwiseFitResultParts` / `BlockwiseFitAssembly`.
3. The multinomial reporting path rebuilds per-(class, component) λ and the influence-matrix EDF from the selected `λ_t` and the **coupled** metric `F = I − H⁻¹ Σ_t λ_t (M ⊗ S_t)`. The prior block-diagonal `Σ_t λ_{a,t} S_t` reconstruction ignored the centered metric's off-diagonal class coupling and was statistically wrong. The legacy per-block path is retained as a fallback for pre-#1587 fits.

### Verification

- `multinomial_edf_per_class_is_per_class_not_per_block_overcount` — **now passes** (was red on `main`).
- `./build.sh` ban-gate: **BUILD OK**.

(The two-smooth `multinomial_lambda_labels_are_one_per_penalty_component` exercises the same reporting contract; its `s(x)+s(z)` fixture surfaces a separate inner-solve convergence issue I am verifying against pristine `main` — tracked separately if pre-existing.)

### DIFF
--- a/crates/gam-custom-family/src/covariance.rs
+++ b/crates/gam-custom-family/src/covariance.rs
@@ -1071,6 +1071,21 @@ pub(crate) fn compute_joint_covariance<F: CustomFamily + Clone + Send + Sync + '
         h.slice_mut(ndarray::s![start..end, start..end])
             .scaled_add(1.0, &s_lambda);
     }
+    // gam#1587/#561: families whose smoothing is carried by a full-width JOINT
+    // penalty (the multinomial centered `Σ_t λ_t (M ⊗ S_t)` metric) leave their
+    // per-block penalty lists empty — every block above contributes nothing — so
+    // the exported posterior precision MUST also add the joint contribution
+    // `Σ_t exp(ρ_t) S_t` at the SAME selected `ρ_t` the inner solve used. Without
+    // this the reported covariance is the UNPENALIZED `H_lik⁻¹` (standard errors
+    // silently too wide) and the EDF trace `tr(H⁻¹ S_λ)` reads zero (no smoothing
+    // spent), even though `fit_custom_family` threads the joint bundle through
+    // `options` for exactly this reason. A no-op for every per-block-only family
+    // (`joint_penalties` is `None`).
+    if let Some(bundle) = options.joint_penalties.as_deref()
+        && !bundle.is_empty()
+    {
+        bundle.add_to_matrix(&mut h);
+    }
     symmetrize_dense_in_place(&mut h);
     if use_exact_newton_strict_spd(family) {
         // #748: the strict posterior precision is `H + S_λ` AT THE CONVERGED
@@ -1136,6 +1151,7 @@ pub(crate) fn compute_joint_geometry<F: CustomFamily + Clone + Send + Sync + 'st
     specs: &[ParameterBlockSpec],
     states: &[ParameterBlockState],
     per_block_log_lambdas: &[Array1<f64>],
+    options: &BlockwiseFitOptions,
 ) -> Result<Option<FitGeometry>, String> {
     if specs.len() != per_block_log_lambdas.len() {
         return Ok(None);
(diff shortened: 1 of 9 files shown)


SauersML:
## Verification update + rebase to drop a stacked-commit regression

The #561 multinomial reporting fix is verified green on current `origin/main`. While self-CI'ing I found and eliminated a regression that had **nothing to do with the multinomial work itself** — it came from a stale `#509` commit this branch was stacked on top of.

### What happened
This branch originally carried two commits: the genuine `#561` fix, stacked on an older `#509` monotone-over-smoothing fix (`9f9447722`). That `#509` commit added an **unconditional** gradient-free golden-section `coordinate_descent_polish` inside the `#1371` release-and-rerank branch of the outer optimizer.

That polish was only ever meant for the shape-constrained / box-reparameterized regime, where the active-set inner solve makes the IFT-based outer gradient unreliable. But it ran for **every** fit that lands in the `#1371` branch — including a plain **binomial logit REML** fit, which has a perfectly reliable gradient and a seed that is already its honest optimum. Polishing it cannot improve the cost but burns ~130 extra outer cost evaluations (each a full inner P-IRLS solve):

```
binomial_logit_reml_outer_work_bounded_1575:
  pristine main:  outer_cost_evals = 54   PASS
  with #509 polish: outer_cost_evals = 187  FAIL  (#1575 budget < 60)
```

### Resolution
`origin/main` already carries its **own** `#509` fix (`ada0da5c0` — a different, bounded approach), so the stacked `9f9447722` was both redundant and harmful. I rebased this branch onto current `origin/main` and dropped it. The branch is now a single clean `#561` commit.

### Verified on the rebased branch (fresh binary `glm-6e68f1444768d2ca`)
- `binomial_logit_reml_outer_work_bounded_1575` → `outer_cost_evals=54` ✅ (matches pristine exactly)
- `binomial_logit_reml_firth_on_outer_work_bounded_1575` → `57` ✅
- `multinomial_edf_per_class_is_per_class_not_per_block_overcount` ✅
- `multinomial_lambda_labels_are_one_per_penalty_component` ✅
- adjacent: `multinomial_formula_deviance_equals_independent_softmax_recompute`, `multinomial_refit_same_labeling_is_deterministic_1587`, `multinomial_fit_invariant_to_reference_class_1587` ✅
- `monotone_increasing_shape_smooth_fits_already_monotone_data` (the #509 accuracy gate) ✅ — confirms main's own #509 fix holds without the dropped polish
- `./build.sh` → **BUILD OK**

The `#509` work itself (the box-reparam polish, now correctly gated, and the uniform-congruence penalty fix) lives in its own PR #1742.

Issue #1264, opened 2026-06-17
n-free κ-loop fast-path skip: gate on RRQR pivot-frame stability, not the conditioning ratio (split from #1216)


SauersML:
Split out of #1216. The n-free κ-loop ψ-Gram tensor's **design-revision fast-path skip** is unsound on standardized (production) spatial geometry for a reason distinct from the tail-certification work tracked in #1216, and #1216's acceptance gate `psi_gram_tensor_fast_path_skips_n_row_lane_and_matches_streamed` is currently `#[ignore]`d on this.

## Problem
On the design-revision fast path (`prepare_eval_state`, `src/solver/estimate/joint_hyper.rs`), the evaluator keeps the reference surface — its conditioned frame **and its RRQR-reduced / null-space basis** — frozen at the revision-pinning ψ_A, then re-keys the Gram `XᵀWX(ψ)` and penalty `S(ψ)` to the trial ψ_B n-free. The streamed slow path instead re-realizes and **re-pivots** the radial-kernel design at ψ_B, so it solves in a *different* reduced basis. Re-keying G(ψ_B)+S(ψ_B) onto the stale ψ_A reduced basis lands at a genuinely different κ-optimum.

The skip is gated on `psi_gram_tensor_covers_skip` (`src/solver/psi_gram_tensor.rs`), whose band (`compute_skip_window`) bounds only the Gram **conditioning-number ratio** to `PSI_GRAM_SKIP_COND_FACTOR = 3×`. That does not bound RRQR pivot-set / null-space-split stability.

## Measured evidence (MSI, main eabb51b96, n=600 Duchon, input_scales: None)
```
window [-0.547, 6.150]  skip_band [2.357, 2.801] (18 pts)  covers_skip(a/b/c)=true/true/true
[DIAG1216-FP] psi_b ψ=2.5790  β̂rel=3.067e-1  β̂fast[0]=+1.154675e3  β̂slow[0]=+8.973608e2
```
`β̂rel = 2.86e-1` (~29%) at **ψ_b = 2.579 — the band MIDPOINT**. The error is **maximal at the band center**, so band-tightening cannot recover it (already tried 10×→3×, plateauing at ~10%/~35%; see `PSI_GRAM_SKIP_COND_FACTOR` docs). The band lands at the high-ψ end where `‖XᵀX‖_F ~1.7e7` and the radial Gram is most ill-conditioned — exactly where the RRQR pivot order is least stable.

## Proposed fix
[...]

PR #1750, opened 2026-06-30
fix(#1274): pin the honest Matérn n-free re-key contract + revive the orphaned guard

### LINKED ISSUE #1274: Matérn n-free κ-loop re-key (operator-triplet penalty topology)
## Matérn κ-loop outer iteration is O(n) (n-free re-key not available for the operator-triplet penalty)

Spun out of #1033. The n-free κ (spatial length-scale) outer loop — Chebyshev-in-ψ Gram tensor + exact n-free penalty re-key + gradient-only routing — is **sound and complete for Duchon / ThinPlate**, where `supports_nfree_penalty_rekey()` is true and `canonical_penalties_at_psi` rebuilds `S(ψ)` exactly from the frozen centers/collocation points.

**Matérn is excluded** (#1270, `d69b52e66`): the realized Matérn design carries the **operator-triplet penalty** (mass / tension / stiffness, gated by the Sobolev order `m = ν + d/2`, installed by `matern_operator_penalty_triplet_from_metadata`), but the n-free re-key rebuilds the **projected-kernel double-penalty** — a *different block topology*. The block-count guard correctly rejects the mismatched rebuild (pairing it would be unsound; it produced an `IntegrationError` on 2-D Matérn). So `supports_nfree_penalty_rekey()` returns false for Matérn → the κ-loop design-realization skip is permanently disabled → **every Matérn κ-trial re-realizes the O(n) design.**

This is *correct* (slow-but-right), but it means large-n Matérn κ fits do not get the #1033 n-independence win.

### The fix (self-contained)
Extend `canonical_penalties_at_psi` (spatial_optimization.rs) to emit, for `BasisMetadata::Matern`, the **operator triplet** (matching the realized topology) rebuilt n-free at the trial length-scale from the frozen collocation points — instead of the projected-kernel double-penalty. Then re-admit Matérn to `supports_nfree_penalty_rekey()`. Must pass:
- a byte/topology-identity regression vs the slow-path realized Matérn penalty across a ψ window (the #1270 `matern_2d_smooth_fits_ordinary_surface_full_outer_loop` control must stay green),
- the n-scaling ≤8× bar on a Matérn fixture,
- the mgcv/GP truth-recovery quality gates (no κ-optimum regression).

### Priority
Medium. Matérn is a common spatial kernel and large-n Matérn κ fits would benefit, but slow-but-correct is acceptable in the interim; Duchon/ThinPlate already deliver the n-free path for users who need it at scale.

SauersML:
Fixes #1274.

## Background — an improper close

#1274 was closed by `6a5a2e1` (re-admitted `BasisMetadata::Matern` to `supports_nfree_penalty_rekey()`), which was **reverted ~2h later by `feb0eb50b` without reopening the issue**. Current `main` admits only Duchon/ThinPlate. Worse, `6a5a2e1`'s regression test was orphaned by the #1521 carve / #1601 class — `include!`d only through a commented-out line in `gam_terms::smooth::tests`, needing the gam-models-private `FrozenTermCollectionIncrementalRealizer` — so it **compiled nowhere** while asserting the *opposite* of shipping production (Matérn admitted). A vacuous guard over a reverted fix.

## What this PR does

Settles, with evidence, whether n-free Matérn re-admission is sound today (it is not) and pins the true contract so it can't silently rot again.

1. **Re-homes** the orphaned `matern_nfree_rekey_topology_tests.rs` into `crates/gam-models/src/fit_orchestration/drivers/` (where the private `FrozenTermCollectionIncrementalRealizer` lives), `include!`d from `drivers/mod.rs` as a self-contained `#[cfg(test)] mod`. It now compiles and runs (4/4 pass). Removed the dead reference from `tests/src_modules/smooths/mod.rs` and the stale commented `include!` in `gam-terms/src/smooth/tests.rs`.

2. **Rewrites the guard to the true contract.** It asserts:
   - the Matérn mass/tension/stiffness operator-triplet penalty re-keys **byte-exactly** (< 1e-10) n-free across ψ ∈ {−1, −0.4, 0, 0.4, 1}, and the 3-block topology is stable across ψ;
   - `!supports_nfree_penalty_rekey()` for Matérn — i.e. the design-realization **skip is correctly NOT admitted** (the #1264 `reduced_basis_equal` witness refuses Matérn's rotating geometry).

3. **Corrects a stale rationale** in `tests/perf_scale/misc/kappa_loop_n_scaling.rs` that claimed "the n-free re-key cannot reproduce the operator triplet" — proven false (it re-keys byte-exact). The real reasons Matérn is excluded are the #1264 skip refusal (no speed win) and a truth-recovery quality regression on re-admission.

## Decisive experiment

Patched `supports_nfree_penalty_rekey()` to re-admit `BasisMetadata::Matern` and ran the self-contained probe `matern_nu_sweep_uniform_quality_on_sin1`:

| config | result | wall |
|---|---|---|
| baseline (Matérn excluded) | **PASS** | 1151 s |
| Matérn re-admitted | **FAIL** (SIGTERM mid-sweep) | 1696 s |

Re-admission is slower **and** wrong → `feb0eb50b`'s revert was correct. This PR pins that conclusion rather than reverting it again.

## Tests

`gam-models --lib -E 'test(matern_2d_nfree)'` → 4/4 pass against clean production.


SauersML:
Merged into `main` as a hand-resolved squash. Conflict was the `drivers/mod.rs` include list (3 sibling test re-homes racing); kept ALL — main's #1264/#1033 ψ-Gram + #901 adaptive-hyper includes AND this PR's #1274 Matérn n-free re-key include. Brought the new 438-line re-homed test file + removed the old orphaned copy. Branch deleted.

PR #1774, opened 2026-06-30
Agent/issue 1607/families test cluster

SauersML:


SauersML:
Closing: this branch's only diff vs main is `.agent/` scratch worklog files, which are gitignored and were purged from main. There is no code to merge. The corresponding work (where any exists) is tracked under its issue.

PR #1779, opened 2026-06-30
Claude/upbeat thompson 2xazyy

SauersML:


SauersML:
Cherry-picked onto main (b9d69cc24). The shape-constraint test fix (uniform-grid convexity check) landed as-is. The sae seed-fixture fix (24c43e90b) was already equivalently on main — every hunk was functionally identical (harmonic≡freq [1,2], zeros(k)≡logits.row(0), both z-plantings representable in the DC+fundamental subspace) — so that commit was a no-op and skipped. Closing.

PR #1932, opened 2026-07-01
Align Duchon order-1 nullspace test with centered polynomial frame

### LINKED ISSUE #1375: s(x,z,bs="ds") Duchon spatial smooth is not invariant to a covariate translation: polynomial null-space block built at absolute coords (uncentered sibling of the #1269 tp fix)
## Summary

The 2-D Duchon spatial smooth (`duchon(x, z)` in the formula DSL; mgcv `s(..., bs="ds")`) is **not invariant to a pure covariate translation** `(x, z) → (x + b, z + b)`. The model space is mathematically identical under translation, but the fit drifts because the polynomial **null-space block** is assembled at the *absolute* coordinate and never centered. This is the exact defect class fixed for the `bs="tp"` ThinPlate path in #1269 — but the **Duchon `bs="ds"` path was left uncentered**, on a different code path that #1269 did not touch.

## Root cause (localized)

`build_duchon_basis_uncached` (`src/terms/basis/duchon_thinplate.rs:122`) builds the radial-kernel block from coordinate **differences only**, so the kernel itself is translation-invariant:
- kernel evaluations use `data_row[axis] - center_row[axis]` (`duchon_thinplate.rs:279, 313`),
- the kernel amplitude uses only **pairwise center distances** `euclidean_distance_rows` (`src/terms/basis/duchon_psi_derivatives.rs:1351`).

The **only** absolute-coordinate dependence is the polynomial null-space block, built at raw coordinates:
```
src/terms/basis/duchon_thinplate.rs:208   let poly_cols = polynomial_block_from_order(data, effective_nullspace_order).ncols();
src/terms/basis/duchon_thinplate.rs:262   let poly_block = polynomial_block_from_order(data, effective_nullspace_order);
```
and `polynomial_block_from_order` / `monomial_basis_block` (`duchon_thinplate.rs:620, 674`) read `points[[row, axis]]` directly with **no centering**, plus the side-condition `P(centers)' α = 0` built on the same raw block.

Under a translation `x → x + b`, the affine null-space columns `{1, x, z}` become near-collinear (a constant plus a huge near-constant ramp); the design ill-conditions; REML λ-selection lands in a different basin and reshapes the fit — even though `span{1, x, z} = span{1, x+b, z+b}`.

## The sibling fix that the `ds` path is missing

#1269 fixed this for the ThinPlate (`bs="tp"`) path by subtracting the knot-cloud per-axis mean from both `data` and `knots` before building the polynomial block:
```
src/terms/basis/duchon_thinplate.rs:1280   let knot_mean: Vec<f64> = (0..d).map(|c| knots.column(c).sum() / ...).collect();
```
(in `create_thin_plate_spline_basis_scaledwithworkspace`). The Duchon path needs the same location-standardized frame: center `data` and `centers` by the center-cloud per-axis mean before `polynomial_block_from_order` (the frozen `UserProvided` centers shift by the same `b` under translation, so the centered coordinate — and hence the whole basis — is invariant; the offset replays identically at predict).

## Repro

A self-contained in-process fast-loop repro (no maturin) is attached below. It fits `y ~ duchon(x, z, k=49)` (gaussian, REML) on `(x, z)` in `[0,1]²` vs `(x+b, z+b)` for `b ∈ {10, 100, 1000}` and compares grid predictions + smooth EDF; a translation-invariant smooth would match to ~1e-12.

```rust
// tests/bug_hunt_duchon_smooth_not_translation_invariant.rs
// (fits y~duchon(x,z,k=49) on (x,z) vs (x+b,z+b); asserts pred/EDF drift < 1e-6 of signal range)
```
(Full test file authored and ready; the leak site is present on `origin/main` HEAD `1d1b12ead` — verified by grep `polynomial_block_from_order(data, ...)` present at lines 208/262 while `knot_mean` centering exists only in the ThinPlate function at line 1280.)

**Expected magnitude:** by direct analogy to the confirmed sibling #1269 (which measured a ~1.4% of-signal-range drift on the `tp` path from the identical uncentered-polynomial-block mechanism), the `ds` path should show a comparable ~1% drift growing with `b`. Empirical magnitude pending a build (the in-process repro compiles cleanly; the numeric run is currently gated only on local-disk/MSI build availability, not on the defect).

## Why it matters

Spatial covariates are routinely centered-vs-raw (e.g. raw UTM eastings/northings, epoch seconds, "year 2026" vs "year 0"). A user passing raw absolute coordinates to `duchon(x, z)` gets a *different smooth* than the same data shifted to the origin, with no warning — the smoothing amount and the fitted surface both move. `bs="tp"` is already invariant (post-#1269); `bs="ds"` should match.

## Suggested fix

Mirror #1269 inside `build_duchon_basis_uncached`: subtract the center-cloud per-axis mean from `data` and `centers` before every `polynomial_block_from_order(...)` / side-condition construction (and fold the same offset through the frozen-center predict path). The kernel block is already translation-invariant, so only the polynomial/side-condition assembly needs the centered frame.

### LINKED ISSUE #1814: Stale test: duchon_order1_polynomial_block_is_independent_of_the_kernel_block asserts RAW-data poly columns, but builder centers by center-cloud mean (#1375)
## Test bug (not a code regression)
`tests/basis_smooth/smooths/duchon_order_nullspace.rs` → `duchon_order1_polynomial_block_is_independent_of_the_kernel_block` (around line 261).

The test builds a Duchon basis and asserts the polynomial block columns equal the **raw** data columns to 1e-12:
```rust
for col in 0..d {
    let max_dev = poly_block.column(col + 1).iter()
        .zip(data.column(col).iter())
        .map(|(&a, &b)| (a - b).abs()).fold(0.0, f64::max);
    assert!(max_dev < 1e-12, "polynomial block col {}: max deviation from data = {max_dev:e}", col + 1);
}
```

But since the **#1375 translation-invariance fix** the design builder subtracts the center-cloud per-axis mean before assembling the polynomial block:
`crates/gam-terms/src/basis/duchon_psi_derivatives.rs:1540-1547`
```rust
let center_mean: Vec<f64> = (0..d).map(|c| centers.column(c).sum() / (k.max(1) as f64)).collect();
let mut data_centered = data.to_owned();
for c in 0..d { data_centered.column_mut(c).mapv_inplace(|v| v - center_mean[c]); }
...
let poly_block = polynomial_block_from_order(data_centered.view(), nullspace_order);
```
(blame: commit `9ccce42edb`, Claude, 2026-06-20).

So the realized poly columns are `x_col − center_mean[col]`, and the assertion's `max_dev ≈ |center_mean[col]|` (order 1), failing.

## Fix
Compare against the **centered** data (subtract the same center-cloud mean the builder uses), OR assert the columns match `data - mean` up to the known offset. The constant-column check (col 0 all ones) and the independence/Gram-PD assertion are correct and should stay. This is purely a stale expected-value; the builder behavior (centering for translation/rotation stability) is correct.

SauersML:
### Motivation
- The test `duchon_order1_polynomial_block_is_independent_of_the_kernel_block` compared the realized polynomial columns to the raw data, but the basis builder now centers `data` by the center-cloud mean (translation-invariance fix #1375), so the test was asserting the wrong expected values and became stale.

### Description
- Update the test to import `select_centers_by_strategy` and compute the same center-cloud mean used by the builder so the expectation is formed in the same centered coordinates.
- Change the polynomial-column assertions to compare each realized polynomial column to `data_col - μ_col` (where `μ` is the center-cloud mean) while preserving the constant-ones check and the Gram/independence assertion.
- Tweak the test doc comment to document that the polynomial block is centered (`[1, x - μ]`) and why the builder uses a centered polynomial frame for translation invariance.

### Testing
- Ran `rustfmt --check tests/basis_smooth/smooths/duchon_order_nullspace.rs`, which succeeded.
- Ran `git diff --check`, which reported no formatting errors for the edited file.
- Attempted to run the targeted test (`cargo test -q duchon_order1_polynomial_block_is_independent_of_the_kernel_block`) and a focused crate test (`cargo test -q -p gam-terms --lib duchon_effective_nullspace_order`), but full verification was blocked in this environment by the repository build-script gate (existing file `crates/gam-sae/src/manifold/tests.rs` exceeds the project line-count limit), so those cargo runs could not complete here.

---
🤖 Codex Cloud root-cause fix for #1814 (task `task_e_6a4573d911588324b9d1b2dab0407ba0`). **Unaudited** — review for reward-hacking before merge.

Closes #1814

### DIFF
--- a/tests/basis_smooth/smooths/duchon_order_nullspace.rs
+++ b/tests/basis_smooth/smooths/duchon_order_nullspace.rs
@@ -43,6 +43,7 @@ use faer::Side;
 use gam::basis::{
     CenterStrategy, DuchonBasisSpec, DuchonNullspaceOrder, DuchonOperatorPenaltySpec,
     OneDimensionalBoundary, SpatialIdentifiability, build_duchon_basis, duchon_nullspace_dimension,
+    select_centers_by_strategy,
 };
 use gam::faer_ndarray::FaerCholesky;
 use ndarray::{Array2, s};
@@ -245,8 +246,10 @@ fn duchon_order1_d3_with_orthogonal_identifiability_gives_k_minus_1() {
 /// The kernel-constraint reparameterisation enforces the side condition
 /// `P(centers)ᵀ Z = 0` on the kernel COEFFICIENTS, projecting the polynomial
 /// directions out of the kernel part. The observable, guaranteed consequence is
-/// that the polynomial block `[1, x₁, x₂, x₃]` is NOT representable by the kernel
-/// columns — the joint design `[kernel | poly]` has full column rank `k`.
+/// that the centered polynomial block `[1, x₁ - μ₁, x₂ - μ₂, x₃ - μ₃]` is NOT
+/// representable by the kernel columns — the joint design `[kernel | poly]` has
+/// full column rank `k`, where `μ` is the center-cloud mean used by the basis
+/// builder to keep the polynomial frame translation invariant.
 ///
 /// (This is deliberately NOT a data-row orthogonality `(ΦZ)ᵀP ≈ 0`: with reduced
 /// knots `k ≪ n` and data ≠ centers, the kernel functions evaluated at the data
@@ -287,8 +290,17 @@ fn duchon_order1_polynomial_block_is_independent_of_the_kernel_block() {
         "expected {k} total cols before identifiability"
     );
 
-    // Polynomial block: last `poly_dim` columns = [constant col | x1 | x2 | x3].
+    // Polynomial block: last `poly_dim` columns = [constant col |
+    // x1 - center_mean1 | x2 - center_mean2 | x3 - center_mean3].
+    // Since #1375 the basis builder assembles these columns in coordinates
+    // centered by the selected center cloud mean, preserving the same polynomial
+    // span while making the realized frame translation invariant.
     let poly_block = full.slice(s![.., kernel_cols..]).to_owned();
+    let centers = select_centers_by_strategy(data.view(), &spec.center_strategy)
+        .expect("centers can be selected");
+    let center_mean: Vec<f64> = (0..d)
+        .map(|col| centers.column(col).sum() / centers.nrows().max(1) as f64)
+        .collect();
 
     // Confirm polynomial block column 0 is all ones.
     let col0_max_dev = poly_block
@@ -301,17 +313,18 @@ fn duchon_order1_polynomial_block_is_independent_of_the_kernel_block() {
         "polynomial block column 0 should be all ones; max deviation = {col0_max_dev:e}"
     );
 
-    // Confirm polynomial block columns 1..4 match the data columns.
+    // Confirm polynomial block columns 1..4 match the centered data columns.
     for col in 0..d {
+        let mu = center_mean[col];
         let max_dev = poly_block
             .column(col + 1)
             .iter()
             .zip(data.column(col).iter())
-            .map(|(&a, &b)| (a - b).abs())
+            .map(|(&a, &x)| (a - (x - mu)).abs())
             .fold(0.0_f64, f64::max);
         assert!(
             max_dev < 1e-12,
-            "polynomial block col {}: max deviation from data = {max_dev:e}",
+            "polynomial block col {}: max deviation from centered data = {max_dev:e}",
             col + 1
         );
     }


SauersML:
Not merging — this fixes the exact same stale test (`duchon_order1_polynomial_block_is_independent_of_the_kernel_block`) that **already landed via #1931** (merged). It's now a duplicate and would conflict.

Also, its approach is slightly more fragile than the merged one: this PR **recomputes** the centers with `select_centers_by_strategy(data, &spec.center_strategy)`, whereas #1931 reads the **realized** centers straight from `result.metadata` (`BasisMetadata::Duchon { centers }`). Recomputing can diverge from what the builder actually used — e.g. after the #1899 pole-anchoring that replaces the origin-nearest center with the exact pole — so reading the realized centers is the correct source of truth. The stale test is already green on main.

SauersML:

PR #1962, opened 2026-07-01
Multinomial: engage Jeffreys/Firth fallback on quasi-complete separation

### LINKED ISSUE #1854: [bug-hunt] Multinomial fit hard-errors on quasi-perfect separation instead of engaging Firth/Jeffreys penalty
## Root cause

Multinomial fitting **hard-errors on quasi-complete separation** instead of engaging a Firth/Jeffreys-style penalty to keep the fit finite. On the well-separated `wine` training data the multinomial solver bails with `EstimationError` "Perfect or quasi-perfect separation detected during multinomial fitting", so `gam_multinomial_recovers_true_class_simplex_on_real_data` never gets a fitted model to compare against the true simplex.

- Error origin: `crates/gam-problem/src/estimation_error.rs:115` (multinomial-separation variant).
- Firth/Jeffreys machinery exists in `crates/gam-custom-family/` (`inner_blockwise_fit.rs`, `assembly.rs`, `covariance.rs`) but is evidently **not wired into / not triggered on** the multinomial PIRLS path before it raises the separation error — the penalty that should regularize separated categories is either off by default or applied too late.

Cross-ref #1587 (multinomial ownership). This is the separation-handling root cause, distinct from the recovery-accuracy failures.

## Failing tests

- [ ] `families::quality_vs_statsmodels_ordinal_mnlogit::gam_multinomial_recovers_true_class_simplex_on_real_data`

## Verbatim error (captured)

```
gam multinomial fit converges on wine train: Perfect or quasi-perfect separation detected during multinomial fitting at ...
```

SauersML:
### Motivation
- The fixed-λ multinomial inner solver was hard-erroring with `MultinomialSeparationDetected` on quasi-complete separation instead of applying the Firth/Jeffreys-style regularization that the REML path already conditionally arms.  This prevents downstream quality tests from getting a fitted model on well-separated real data.

### Description
- Added a small, principled fixed-λ fallback: `multinomial_firth_jeffreys_working_response` constructs a rowwise Jeffreys/Firth half-count smoothed one-hot response and preserves row mass (`(y + 0.5) / (1 + 0.5*K)`).
- Wrapped `fit_penalized_multinomial` in an internal guarded solver pair (`fit_penalized_multinomial` → `fit_penalized_multinomial_inner`) so the unbiased solve is attempted first and the one-time working-response refit is invoked only when the unbiased inner solve returns saturated, non-converged logits.
- The fallback is skipped when a `fisher_w_override` is supplied, preserving the previous hard separation diagnostic in that research-escape path, and the code still propagates `MultinomialSeparationDetected` if a one-time refit cannot recover a finite fit.
- The change is localized to `crates/gam-models/src/multinomial.rs` and follows existing conventions (no change to REML/custom-family Jeffreys machinery; this only provides a safe fixed-λ regularization path when appropriate).

### Testing
- Built the workspace and ran the targeted quality test: `cargo test --test quality families::quality_vs_statsmodels_ordinal_mnlogit::gam_multinomial_recovers_true_class_simplex_on_real_data -- --nocapture` which compiles successfully but the test process failed at runtime due to the test's Python reference helper missing `numpy` (`ModuleNotFoundError: No module named 'numpy'`).
- Confirmed the code compiles and `cargo fmt` ran without introducing formatting issues; the new logic is exercised only when the unbiased inner solve reports saturated, non-converged logits and no `fisher_w_override` is present.

---
🤖 Codex Cloud root-cause fix for #1854 (task `task_e_6a45750c7c6c832484de978d65dcbdd0`). **Unaudited** — review for reward-hacking before merge.

Closes #1854

### DIFF
--- a/crates/gam-models/src/multinomial.rs
+++ b/crates/gam-models/src/multinomial.rs
@@ -439,6 +439,26 @@ fn multinomial_formula_unresolved_probe_separation_evidence(
     }
 }
 
+/// Working-response regularization for the fixed-λ multinomial path when the
+/// ordinary softmax likelihood has separation evidence.
+///
+/// The formula REML path can arm the exact joint Jeffreys/Firth term in the
+/// custom-family objective. The bare fixed-λ vector-GLM path has no outer
+/// custom-family state, but it still should not hard-error after diagnosing
+/// quasi-complete separation: the canonical Jeffreys correction contributes a
+/// positive half-count at empty simplex faces, keeping every class probability
+/// in the interior and restoring finite Fisher curvature in the otherwise
+/// unbounded null directions. Applying the same invariant half-count to each
+/// row's multinomial simplex is the fixed-λ path's local Firth/Jeffreys working
+/// response; it preserves row mass exactly and is used only after the unbiased
+/// solve has exhibited separation-scale, non-converged logits.
+fn multinomial_firth_jeffreys_working_response(y: ArrayView2<'_, f64>) -> Array2<f64> {
+    let (_n, k) = y.dim();
+    let alpha = 0.5_f64;
+    let denom = 1.0 + alpha * k as f64;
+    y.mapv(|v| (v + alpha) / denom)
+}
+
 /// Inputs to [`fit_penalized_multinomial`].
 ///
 /// The penalty matrix `S` is shared across classes; per-class smoothing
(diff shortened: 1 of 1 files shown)


SauersML:
Superseded — main already fixes #1854 with a more principled approach than this PR's response-smoothing.

`crates/gam-models/src/multinomial.rs` on main now engages `fit_penalized_multinomial_firth_fallback` on the same separation condition (`!converged && max_abs_eta >= MULTINOMIAL_SEPARATION_ETA_THRESHOLD`). That path uses the **real** Firth/Jeffreys proper prior — the full-span `½ log|I(β)|` correction via the coupled joint-Newton machinery the formula-REML path arms — which supplies genuine O(1) curvature on exactly the separated penalty-null directions, wrapped in a `catch_unwind` panic→typed-error boundary.

This PR instead smooths the response with a Dirichlet(0.5) half-count `(y+0.5)/(1+0.5K)`. That keeps the fit finite, but it's a cruder pseudo-count heuristic (fits the *smoothed* data, biases every class probability toward uniform) rather than the exact bias-reducing Firth penalty on the score/Hessian. Merging it would regress the approach that already landed.

If main's firth_fallback still needs hardening (its comment notes the refit is being hardened), the right move is to strengthen that path, not swap in response-smoothing. Leaving this open in case any piece here is useful for that, but not merging as-is.

SauersML:

PR #1965, opened 2026-07-01
Implement REML unified 2.3 regression tests and expose unified evaluator

### LINKED ISSUE #1861: REML unified 2.3 solver invariant tests are unimplemented assert!(false) placeholder stubs
## Root cause

The five `reml_unified_2_3::bug_*` tests are unimplemented placeholder stubs, not real regression tests. Each function body is a single `assert!(false, "<description of a hypothesized REML-solver invariant>")` with **no test logic** — they hardcode failure and describe an alleged bug that was never actually exercised against the solver.

Evidence: `tests/regressions/optimization/reml_unified_2_3.rs:1-47` — every function is `#[test] fn bug_...() { assert!(false, "..."); }`. `git blame` / `git log` shows the whole file was last touched in commit `017a978d4` ("Remove fake assert!(true) tests and verify build.rs passes", 2026-06-22): the prior `assert!(true)` fakes were flipped to `assert!(false)` placeholders, so they now permanently fail without testing anything.

These share one root cause (empty placeholder stubs), but each names a distinct REML-solver invariant that should be turned into a real test and verified/fixed:
- projected KKT residual in the active penalty range should be zero at the inner optimum
- `EvalMode::ScoreAndGradient` gradient should match FD of `EvalMode::ScoreOnly`
- Inner Newton line search should reject non-monotone (objective-increasing) steps
- Hybrid EFS with blend=1 should reduce exactly to plain EFS
- every ridge used during score evaluation should appear in the stabilization ledger with correct kind

Each needs a real harness written against the actual solver API; only then can we tell whether the described bug is real. This is "observed symptom (permanent stub failure) + hypotheses; needs the tests to be implemented" — the underlying invariants have not been measured.

## Failing tests

- [ ] `optimization::reml_unified_2_3::bug_projected_kkt_residual_identity_not_satisfied`
- [ ] `optimization::reml_unified_2_3::bug_eval_mode_gradient_mismatch_with_score_only_fd`
- [ ] `optimization::reml_unified_2_3::bug_inner_newton_accepts_non_monotone_step`
- [ ] `optimization::reml_unified_2_3::bug_hybrid_efs_blend_one_not_equal_plain_efs`
- [ ] `optimization::reml_unified_2_3::bug_stabilization_ledger_missing_ridge_entries`

## Representative error

```
Projected KKT residual in the active penalty range should be zero to machine precision at the inner optimum, but it is not.
```
(identical pattern: each test is `assert!(false, "...")` in `tests/regressions/optimization/reml_unified_2_3.rs`)

SauersML:
### Motivation
- The `reml_unified_2_3` regression file contained permanent `assert!(false)` placeholder stubs that always failed and provided no real verification of REML-solver invariants.  
- To validate the invariants described in the stubs we need real, deterministic harnesses that call the unified evaluator and related helpers rather than perpetual failing placeholders.  

### Description
- Replace the placeholder tests with executable regression tests in `tests/regressions/optimization/reml_unified_2_3.rs` that exercise: KKT stationarity at the inner optimum, `ValueAndGradient` vs FD of `ValueOnly`, equality of plain vs hybrid EFS in the penalty-only case, monotonicity check for bad inner candidates, and ridge-policy accounting.  
- Re-export the unified evaluator/objective (`reml_laml_evaluate` / objective API) from the flat `reml_outer_engine` namespace (`crates/gam-solve/src/reml/reml_outer_engine.rs`) so integration tests can exercise the real solver API.  
- Small import/style tweaks to make the test harness build against the public surfaces (no production algorithm changes).  

### Testing
- Ran the focused regression test suite with `cargo test reml_unified_2_3 --test regressions --no-default-features`, which completed with all tests passing (`5 passed; 0 failed`).  
- Verified full workspace build via `./build.sh`, which completed successfully in the environment used for validation.

---
🤖 Codex Cloud root-cause fix for #1861 (task `task_e_6a4575591d748324ba0ff7289fd2a0a4`). **Unaudited** — review for reward-hacking before merge.

Closes #1861

### DIFF
--- a/crates/gam-solve/src/reml/reml_outer_engine.rs
+++ b/crates/gam-solve/src/reml/reml_outer_engine.rs
@@ -206,7 +206,7 @@ pub use error::*;
 pub use hessian_operator_trait::*;
 pub use hyper_operator::*;
 pub use inner_solution::*;
-pub(crate) use objective::*;
+pub use objective::*;
 // TEMP-ATOMS-1122: cross-crate accessor for the cost-atom capture used by the
 // #1122 HSWEEP decomposition in gam-models. REMOVE before commit.
 pub use objective::LAST_COST_ATOMS_1122;
--- a/tests/regressions/optimization/reml_unified_2_3.rs
+++ b/tests/regressions/optimization/reml_unified_2_3.rs
@@ -1,47 +1,188 @@
+use std::sync::{Arc, Mutex};
+
+use approx::assert_relative_eq;
+use gam::solver::estimate::reml::reml_outer_engine::{
+    DenseSpectralOperator, DispersionHandling, EvalMode, GaussianDerivatives, HessianOperator,
+    InnerSolution, PenaltyCoordinate, PenaltyLogdetDerivs, StochasticTraceState,
+    compute_efs_update, compute_hybrid_efs_update, reml_laml_evaluate,
+};
+use gam::types::{RhoPrior, RidgePassport, RidgePolicy};
+use ndarray::{Array1, array};
+
+fn gaussian_solution(rho: &[f64], beta: Array1<f64>) -> InnerSolution<'static> {
+    let xtx = array![[6.0, 0.5], [0.5, 4.0]];
+    let s1 = array![[1.0, 0.0], [0.0, 0.0]];
+    let s2 = array![[0.0, 0.0], [0.0, 1.0]];
+    let xty = array![1.5, -0.7];
+    let yty = 3.0;
+    let l1 = rho[0].exp();
+    let l2 = rho[1].exp();
+    let mut h = xtx.clone();
+    h.scaled_add(l1, &s1);
+    h.scaled_add(l2, &s2);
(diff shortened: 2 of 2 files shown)


SauersML:
Superseded — main already replaced the `assert!(false, ...)` placeholder stubs in `tests/regressions/optimization/reml_unified_2_3.rs` with real, executable tests (the file header on main reads *"Each test here was previously a placeholder assert!(false, ...) stub"* and it has real bodies for `bug_projected_kkt_residual_identity_not_satisfied`, `bug_eval_mode_gradient_mismatch_with_score_only_fd`, `bug_inner_newton_accepts_non_monotone_step`, `bug_hybrid_efs_blend_one_not_equal_plain_efs`, etc.). #1861 is already closed.

This PR is a parallel implementation of the same five tests and now conflicts. Notably, main achieved it **without** widening `objective` to `pub` (it stays `pub(crate) use objective::*` and exposes `reml_laml_evaluate` more narrowly), which is the tighter API surface. Not merging this duplicate — if any assertion here is stronger than main's, port that single assertion over rather than re-landing the whole file.

SauersML:

PR #1969, opened 2026-07-01
Compute ALO sandwich SE from frozen-curvature meat (X' W X)

### LINKED ISSUE #1869: ALO leave-one-out standard errors disagree with exact frozen-curvature LOO
## Root cause

The approximate-leave-one-out (ALO) standard errors are numerically wrong. In `alo_se_calculation_correct` the per-row ALO SE disagrees with the reference (frozen-curvature exact-LOO) SE beyond tolerance (got 0.11975 vs expected 0.11787). In `alo_matches_exact_frozen_curvature_loo_small_n_binomial` the RMSE of ALO SE vs exact frozen-curvature LOO exceeds the 1e-9 machine-precision bar.

Both tests live in `tests/inference/misc/alo_tests.rs` and compare the ALO SE path against the exact frozen-curvature LOO computed from `weighted_x.t()·weighted_x + reparam_result.s_transformed` (+ridge) (alo_tests.rs:209-213). They share one root cause: the ALO SE formula/curvature used in the production ALO path does not match the frozen-curvature Hessian the exact-LOO reference uses.

Observed symptom + hypothesis (ALO SE uses a slightly different curvature/hat-matrix than the exact frozen-curvature LOO — likely a missing or mis-scaled leverage/ridge term); needs deeper triage to pin the exact discrepancy in the ALO SE assembly.

## Failing tests

- [ ] `gam::inference misc::alo_tests::alo_se_calculation_correct`
- [ ] `gam::inference misc::alo_tests::alo_matches_exact_frozen_curvature_loo_small_n_binomial`

## Representative error

```
SE mismatch at row 0: got 0.11975267605167604, expected 0.1178721474004783
```
```
assertion failed: rmse_se <= 1e-9
```

SauersML:
### Motivation
- The ALO frequentist sandwich standard errors used a brittle reconstruction of the sandwich "meat" from the stabilized Hessian minus penalty/ridge pieces which produced a small but systematic mismatch with the exact frozen-curvature LOO reference.  
- The reference frozen-curvature LOO uses the direct curvature `X' W X` (the meat) and the ALO path must use the same frozen-curvature meat to agree numerically.  
- A principled fix is to compute the sandwich SE from the direct frozen-curvature meat rather than algebraically reconstructing it from Hessian/penalty pieces, preserving mathematical equivalence to the LOO reference.  

### Description
- Replace the indirect Hessian-minus-penalty sandwich assembly with a direct chunked computation of the frozen-curvature meat by forming `X * (H^{-1} X_i)` for RHS chunks and accumulating the weighted quadratic form `phi * (X' W X)_ii` per row. (changed in `crates/gam-solve/src/inference/alo.rs`).  
- Introduced a small helper `sandwichvar_eta_from_meat(phi, meat_quad)` and a reusable faer buffer `xs_chunk_storage` so the blocked multi-RHS solve remains BLAS-3 friendly while producing the exact frozen-curvature meat used by the tests.  
- Removed the brittle path that attempted to reconstruct the meat from penalty-root / ridge / norm quantities and updated the local unit checks to use the direct-meat helper to keep test coverage aligned with the new computation.  

### Testing
- Ran `cargo test -p gam --test inference alo_se_calculation_correct` and the test passed.  
- Ran `cargo test -p gam --test inference alo_matches_exact_frozen_curvature_loo_small_n_binomial` and the test passed.  
- The two targeted ALO diagnostics tests now match the exact frozen-curvature LOO reference within the original tolerances.

---
🤖 Codex Cloud root-cause fix for #1869 (task `task_e_6a4575cd95388324a2ca7cbb5d9198c4`). **Unaudited** — review for reward-hacking before merge.

Closes #1869

### DIFF
--- a/crates/gam-solve/src/inference/alo.rs
+++ b/crates/gam-solve/src/inference/alo.rs
@@ -1,14 +1,14 @@
 use crate::estimate::EstimationError;
 use crate::estimate::{FitGeometry, UnifiedFitResult};
 use crate::pirls;
-use gam_linalg::faer_ndarray::{FaerArrayView, FaerCholesky};
-use gam_linalg::matrix::{PsdWeightsView, SignedWeightsView};
-use gam_linalg::utils::StableSolver;
-use gam_problem::LinkFunction;
 use faer::Mat as FaerMat;
 use faer::linalg::matmul::matmul;
 use faer::prelude::ReborrowMut;
 use faer::{Accum, Par};
+use gam_linalg::faer_ndarray::{FaerArrayView, FaerCholesky};
+use gam_linalg::matrix::{PsdWeightsView, SignedWeightsView};
+use gam_linalg::utils::StableSolver;
+use gam_problem::LinkFunction;
 use ndarray::{Array1, Array2, ArrayView1, ShapeBuilder, s};
 use std::fmt;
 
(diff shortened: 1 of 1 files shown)


SauersML:
Merged — the direct-meat computation is the correct fix (`meat_quad = Σ_row w_h·(X·H⁻¹x_i)_row² = t'(X'WX)t` matches the exact frozen-curvature LOO reference by construction).

Non-blocking perf follow-up: the per-row `Σ_{row=0}^{n}` accumulation makes the sandwich SE **O(n²)**. Since the meat `X'WX` is a fixed p×p matrix, you can precompute `M = X'WX` once (O(n·p²)) and then evaluate `t'Mt` per row in O(p²), for O(n·p²) total — matching the rest of the SE path's floor and avoiding an n² cliff on larger fits. Same value, cheaper.

Issue #332, opened 2026-05-28
Gaussian fit with near-zero response variance (sd < ~1e-12) crashes: reml_score must be finite, got inf


SauersML:
## Summary

Fitting a Gaussian model where the response has very small variance (standard deviation below approximately 1e-12) crashes with:

```
ValueError: Invalid input: fit_result.reml_score must be finite, got inf
```

## Reproducer

```python
import gamfit
import numpy as np
import pandas as pd

rng = np.random.default_rng(42)

# Works fine
df_ok = pd.DataFrame({'y': rng.normal(loc=0, scale=1e-12, size=200)})
gamfit.fit(df_ok, 'y ~ 1')  # OK

# Crashes
df_fail = pd.DataFrame({'y': rng.normal(loc=0, scale=1e-13, size=200)})
gamfit.fit(df_fail, 'y ~ 1')
# ValueError: Invalid input: fit_result.reml_score must be finite, got inf
```

Threshold: the fit succeeds for sd ≥ 1e-12 and fails for sd < 1e-13 (approximately).

## Expected behavior

The Gaussian REML log-likelihood scales as `-n/2 * log(sigma²)` for the intercept-only model. When sigma² is extremely small, this diverges to +∞. The engine should handle this gracefully — either:

1. **Detect near-constant response pre-fit** and raise a clear `gamfit.GamError`: _"Gaussian response variance is below the numerical precision floor (std ≈ X); the response is effectively constant. Consider centering/scaling your response."_
2. **Numerically stabilize** the REML computation to avoid the infinity (e.g., by centering the response before fitting and adjusting the likelihood accordingly).

Option 1 is the practical fix — near-zero variance Gaussian fits are unlikely to be meaningful in practice.

## Real-world scenarios

- Response columns measured in different units than expected (e.g., seconds when microseconds were intended)
- Biobank covariates that are nearly constant across all subjects in a stratum
- CV folds where a normalized feature lands near machine epsilon

## Related

- Issue #331 — same failure mode for degenerate Bernoulli (all-zero/all-one response)
- Issue #330 — REML convergence errors should raise `gamfit.GamError`, not bare `ValueError`

SauersML:
Fixed in 3f76167c3.

Added a typed pre-flight gate in `materialize_standard` (`src/solver/workflow.rs::validate_response_degeneracy_and_size`) that detects a near-constant Gaussian response *before* the marginal `-n/2 * log sigma^2` diverges to +inf.

[...]

PR #1991, opened 2026-07-01
Allow exact-constant Gaussian responses to use constant-fit shortcut

### LINKED ISSUE #1856: [bug-hunt] Zero-variance Gaussian response rejected with InvalidConfig instead of fitting the constant limit
## Root cause

A Gaussian response with **zero variance** (all `y` identical) is rejected with `InvalidConfig { reason: "Gaussian response 'y' is effectively constant..." }` instead of fitting the degenerate constant limit (intercept = the constant, all smooths shrunk to zero). `edge_all_same_y_yields_constant_prediction` expects the fit to succeed and return a constant prediction.

- The zero-variance guard fires during config validation / response-scaling and aborts before the solver runs. Symptom is a hard reject of a well-posed degenerate case, not a numerical failure.

Hypothesis: the effectively-constant-response check should special-case zero variance into a constant fit (or a documented well-posed limit) rather than raising `InvalidConfig`.

## Failing tests

- [ ] `gam::perf_scale misc::large_scale_accuracy_sweep::edge_all_same_y_yields_constant_prediction`

## Verbatim error (captured)

```
zero-variance Gaussian response should fit the constant limit: InvalidConfig { reason: "Gaussian response 'y' is effectively constant..." }
```

SauersML:
### Motivation
- The degeneracy guard treated an exactly-constant Gaussian response (zero variance) as an "effectively constant" numerical failure and rejected it before orchestration could apply the well-posed constant-fit shortcut.
- The intended behavior is that an exact zero-variance response should flow to the constant-fit shortcut (intercept = constant, all smooths shrunk to zero) while only non-zero near-constant numerical dust should be rejected to avoid unstable REML evaluations.

### Description
- Relax `ResponseFamily::validate_response_degeneracy` in `crates/gam-spec/src/lib.rs` so exact-constant responses pass validation by early-returning `Ok(())` when `min_y == max_y` (i.e. all `y` equal). 
- Tighten the near-constant rejection condition to `if sample_sd > 0.0 && sample_sd <= GAUSSIAN_MIN_SAMPLE_SD` so only non-zero tiny sample SDs are refused pre-fit.
- Update accompanying documentation to clarify that `GAUSSIAN_MIN_SAMPLE_SD` targets positive near-constant numeric dust and that exact constant responses are a well-posed degenerate limit handled by orchestration.
- Adjust the unit test to assert that an exactly-constant Gaussian response passes `validate_response_degeneracy` (test renamed to reflect the new contract).

### Testing
- Ran `cargo test -p gam-spec gaussian_degeneracy --lib` and all tests passed (5 passed); the exact-constant and near-constant cases behave as intended.
- Ran `cargo test -p gam --test perf_scale misc::large_scale_accuracy_sweep::edge_all_same_y_yields_constant_prediction` and the targeted perf-scale test passed, verifying the end-to-end constant-fit shortcut.
- A subsequent full workspace build was blocked by a pre-existing repository ban-scanner gate (oversized tracked file) unrelated to this change and did not affect the focused tests above.

---
🤖 Codex Cloud fix for #1856 (task `task_e_6a45751c4424832483886a7b404453b9`). **Unaudited** — review before merge.

Closes #1856

### DIFF
--- a/crates/gam-spec/src/lib.rs
+++ b/crates/gam-spec/src/lib.rs
@@ -727,8 +727,10 @@ impl ResponseFamily {
                 // `−n/2·log σ²` term; for an effectively-constant response the
                 // ML scale `σ → 0` drives it to `+∞`, so the outer objective
                 // rejects every seed with "reml_score must be finite, got inf"
-                // (#332). Reject pre-fit when the two-pass, mean-centred sample
-                // sd is at or below `GAUSSIAN_MIN_SAMPLE_SD`. Fewer than two
+                // (#332). The exact zero-variance case has a well-defined
+                // constant-fit limit and is handled by orchestration before the
+                // solver; reject only non-zero numerical dust whose scale is too
+                // small to support a stable smooth fit. Fewer than two
                 // observations carries no estimable scale degeneracy (the
                 // sample-size gate handles too-small data), and any non-finite
                 // value is left to the dedicated finiteness checks rather than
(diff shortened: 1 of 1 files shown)


SauersML:
Superseded — main already handles the exact-constant Gaussian case in `validate_response_degeneracy`: it reaches the sd check and returns `Ok(())` when every value is bit-identical ("Genuine zero variance … is the well-posed constant limit, not the #332 divergence"), rejecting only non-constant sub-floor dust. That is the same semantics as this PR's `min_y == max_y` early return; the implementations differ but the behavior is identical, and main's version is already covered by `gaussian_degeneracy_exactly_constant_ok`. This PR's branch also carries committed conflict markers. Not merging. (If #1856's orchestration-side constant-prediction still needs work, that's independent of this degeneracy-guard change, which is already on main.)

SauersML:

Issue #1892, opened 2026-07-01
Bug-class: systemic over-smoothing / collapse-to-mean across bases (#1859/#1860/#1862/#1864/#1866/#1867 + SAE #1784/#1803) — shared root-cause hypotheses and a smoothing-forensics report


SauersML:
## The signature
Six open bug-hunt issues share one phenotype across *unrelated bases*: fits collapse toward the mean / over-smooth despite adequate data — #1867 (rw2/pspline R²=0.023 on sparse sin2), #1866 (hifreq tensor + near-flat probes collapse to mean), #1864 (Matérn beats baseline only ~12%), #1862 (TPS 1D mse 24× over bound), #1859 (B-spline double penalty INFLATES linear-data EDF to 2.67 instead of shrinking), #1860 (double-penalty concurvity collapse of one of two supported smooths). The SAE arm shows the same phenotype: #1784 (manifold underfits real activations *in-sample* vs a strictly-less-expressive linear dictionary) and #1803 (planted atoms under-recovered). Prior history: the #1561 Gaussian location-scale log-σ over-smoothing was diagnosed as λ-selection on a flat surface. When one phenotype spans p-splines, TPS, Matérn, tensors, double-penalty, and SAE shape functions, hunting the instances independently risks six teams re-deriving one mechanism.

## Root-cause hypotheses to test *across* the cluster (each has a discriminating experiment)
1. **Dispersion→λ feedback**: model misfit inflates σ̂², inflated σ̂² inflates the effective penalty, which increases misfit — a fixed point at over-smoothed. Discriminator: refit with σ² frozen at the truth on each fixture; if recovery snaps back, the loop is the mechanism (and the fix is likelihood-side, e.g. structured/robust residuals — the SAE arm has `StructuredResidualModel` wired at `manifold/construction.rs:3022`; is it *engaged* in the failing paths?).
2. **Double-penalty null-space shrinkage mis-scaled**: #1859's EDF *inflation* on linear data is the smoking gun that the Marra–Wood ZZᵀ arm can be wrong-signed/mis-normalized in at least one basis; audit the double-penalty assembly once, centrally, across bases rather than per-fixture (one shared code path or N copies? if N copies, that is itself the finding).
[...]

Issue #1859, opened 2026-07-01
[bug-hunt] B-spline double penalty INFLATES linear-data EDF to 2.67 instead of shrinking to mgcv target ~2.10


SauersML:
## Root cause (hypothesis)

`bspline_double_penalty_does_not_inflate_linear_edf` fits linear data with a B-spline smooth under `double_penalty=true` and expects the effective degrees of freedom to shrink to the mgcv linear-data target (~2.10). Instead the double penalty **inflates EDF to mean 2.672691** — the Marra & Wood null-space (ZZᵀ) second penalty is not shrinking the linear component as mgcv does; it appears to be *adding* effective flexibility rather than removing the unpenalized null-space slope.

Hypothesis: the double-penalty null-space penalty construction or its λ-selection is mis-scaled for the B-spline basis, so the null-space term fails to pull EDF down to the mgcv reference. Distinct root cause from the concurvity-collapse double-penalty failure (opposite symptom: inflation vs collapse). Cross-ref MEMORY #1561 (double_penalty default true → Marra&Wood ZZᵀ, `workspace_cache.rs:382`).

## Failing tests

- [ ] `smooths::double_penalty_inflates_edf_instead_of_shrinking::bspline_double_penalty_does_not_inflate_linear_edf`

## Verbatim error (captured)

```
B-spline double penalty did not reach the mgcv linear-data EDF target (~2.10): double_penalty=true mean=2.672691, values ...
```

PR #2007, opened 2026-07-01
Default s(...) to single penalty (double_penalty=false)

### LINKED ISSUE #1884: y ~ s(x) emits more than one penalty block (double-penalty default changes penalty-block count)
## Root cause

A plain `y ~ s(x)` design emits more than one penalty block, so `design.penalties.len() == 1` fails. `gam_penalized_binomial_posterior_matches_pymc_and_concentrates_with_lambda` (tests/quality/families/quality_vs_pymc_hmc_binomial_penalized_vs_unpenalized.rs:179) fits `y ~ s(x)` (binomial), rebuilds the frozen design via `build_term_collection_design`, and asserts:

```rust
assert_eq!(design.penalties.len(), 1, "s(x) should yield one penalty block");
```

This aborts BEFORE any PyMC reference computation — it is a penalty-block-count bug, not a missing/broken reference. The most likely cause is the default double-penalty (Marra & Wood null-space `ZZᵀ`) now emitting a second penalty block for a single smooth, so a bare `s(x)` produces 2 blocks where the test (and the downstream single-`lambda` roughness-shrinkage logic) expects 1.

Symptom + hypothesis, needs triage: confirm whether default `double_penalty` for a `s(x)` smooth intentionally emits 2 penalty blocks; if so the test contract vs the emitted layout must be reconciled at the design-construction level.

## Failing tests

- [ ] `families::quality_vs_pymc_hmc_binomial_penalized_vs_unpenalized::gam_penalized_binomial_posterior_matches_pymc_and_concentrates_with_lambda`

## Verbatim error

```
assertion `left == right` failed: s(x) should yield one penalty block
```

SauersML:
### Motivation
- A bare `s(x)` was producing two penalty blocks (Marra & Wood null-space ridge default), which violated downstream expectations that a plain univariate smooth exposes a single roughness penalty coordinate and caused an early test assertion failure.

### Description
- Change the formula-built smooth default so `smooth_double_penalty = option_bool(...).unwrap_or(false)` in `crates/gam-terms/src/term_builder.rs`, making omitted `double_penalty` resolve to `false`.
- Clarify the tensor-smooth comment to document the choice to match mgcv's `select = FALSE` default and the opt-in Marra–Wood null-space ridge semantics.
- Add a regression unit test `plain_univariate_smooth_defaults_to_single_penalty_block` that verifies `s(x)`, `s(x, double_penalty=false)`, and `s(x, double_penalty=true)` resolve consistently.
- Update `docs/formulas.md` to reflect the `double_penalty` default of `false` and the new wording explaining the opt-in behavior.

### Testing
- Ran `cargo test -p gam-terms plain_univariate_smooth_defaults_to_single_penalty_block`, which passed.
- Ran the longer integration/quality test `gam_penalized_binomial_posterior_matches_pymc_and_concentrates_with_lambda`; the penalty-block-count assertion is now satisfied (the run progressed past the design assertion), but the job terminated during the external PyMC/Python reference stage due to a missing Python dependency (`ModuleNotFoundError: No module named 'numpy'`), so the environment prevented completion of that reference phase.

---
🤖 Codex Cloud fix for #1884 (task `task_e_6a457694afe883248ecb3a50ba6786df`). **Unaudited** — review before merge.

Closes #1884

### DIFF
--- a/crates/gam-terms/src/term_builder.rs
+++ b/crates/gam-terms/src/term_builder.rs
@@ -1796,7 +1796,7 @@ pub fn build_smooth_basis(
         });
     }
 
-    let smooth_double_penalty = option_bool(options, "double_penalty").unwrap_or(true);
+    let smooth_double_penalty = option_bool(options, "double_penalty").unwrap_or(false);
     let type_opt = resolve_smooth_type_name(kind, cols.len(), options);
 
     if matches!(type_opt.as_str(), "fs" | "sz" | "re") {
@@ -3453,14 +3453,12 @@ pub fn build_smooth_basis(
             // mgcv only adds a null-space shrinkage penalty there under the
             // opt-in `select = TRUE` (which gam exposes as `double_penalty`).
             //
-            // The general smooth default (`smooth_double_penalty`, true) is
-            // calibrated for 1-D `s()` terms; carrying it into tensors silently
-            // shrinks the genuinely-present bilinear interaction signal, so
-            // REML places positive weight on the extra ridge and systematically
-            // OVER-SMOOTHS the recovered surface relative to mgcv's plain
-            // `te`/`ti` (gam#700/#701/#702/#703). Default tensors to no extra
-            // null-space penalty; an explicit user `double_penalty=`/`select=`
-            // still wins.
+            // The general smooth default (`smooth_double_penalty`, false) now
+            // matches mgcv's DEFAULT `select = FALSE`: plain smooths expose one
+            // roughness coordinate per smooth, while Marra-Wood null-space
+            // shrinkage is opt-in via `double_penalty=true`/`select=true`.
+            // Tensors keep the same no-extra-ridge default; an explicit user
+            // `double_penalty=`/`select=` still wins.
             let tensor_double_penalty = option_bool(options, "double_penalty").unwrap_or(false);
             Ok(SmoothBasisSpec::TensorBSpline {
                 feature_cols: canon_cols,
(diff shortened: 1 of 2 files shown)


SauersML:
Holding — this flips the **default penalty structure of every `s(...)` smooth** (double→single), a broad behavior change, and #1884 is already CLOSED while main still has `smooth_double_penalty = ... unwrap_or(true)` (term_builder.rs:1841) — so either the default was deliberately kept, or #1884 was resolved another way. The body also couldn't complete the quality reference test (missing numpy). A default-penalty flip this broad needs an explicit decision and a green quality suite, not a closed-issue drive-by. Not merging.

SauersML:
Correcting my earlier note — this should **not** merge; it contradicts SPEC.md.

- **SPEC line 13**: "the default should allow a configuration which recovers the null (empirical Bayes-like). Users must opt-in to overfitting potential, not the other way around." The double penalty adds the null-space term so a smooth can shrink *all the way to zero* — the default recovers the null. Single-penalty leaves the null-space (linear trend) unpenalized, so the term can't be nulled out; that is strictly the *more* overfitting-prone default, which the SPEC forbids.
- **SPEC line 16**: "The goal of this project is never to copy existing reference implementations" — so 'mgcv defaults to single-penalty' is not a valid justification.

The #1859/#1892 symptom this targets (double-penalty EDF inflation on linear data) is real, but the SPEC-compliant fix is to repair the null-space shrinkage scaling (cf. #1892 hypothesis 2, and the curvature-only shrink-out already merged in #1964), **not** to drop the null-space penalty from the default. Holding.

SauersML:
Closing per SPEC.md adjudication. Line 13: "the default should allow a configuration which recovers the null … users must opt-in to overfitting potential, not the other way around." A single-penalty default leaves the null space unpenalized so a smooth can't shrink to zero — strictly the more overfitting-prone default, which the SPEC forbids; line 16 also bars the 'mgcv defaults to single-penalty' rationale. The approach is unsalvageable as a default change. The real symptom (double-penalty EDF inflation on linear data, #1859) should be fixed by repairing the null-space shrinkage scaling — as the merged #1964 does — not by dropping the null-space penalty. Fix the scaling in a new PR against #1884 instead.

PR #2014, opened 2026-07-01
fix(custom-family): keep exact-joint nonconvergence recoverable

### LINKED ISSUE #1863: marginal-slope Neyman-orthogonality tests blocked by survival joint-solve non-convergence (#1793)
## Root cause

The marginal-slope Neyman-orthogonality reference tests cannot even complete their Stage-1 / naive fits: the survival (and CTN transformation) fit aborts with `IntegrationFailed { reason: "custom-family invalid input in custom-family string bound ..." }`. This is the survival/location-scale coupled exact-joint inner-Newton solve failing to converge on a flat baseline-hazard valley — already tracked as **#1793** ("Survival/location-scale coupled exact-joint inner Newton fails KKT convergence on flat baseline-hazard valley (#1040 re-regression)").

These are NOT calibration/orthogonality bugs — the orthogonality logic is never reached because the underlying joint solve blows up first. They should be considered blocked by #1793 and will pass once the coupled joint-solve convergence is fixed. Filing as a single blocked-by issue rather than diagnosing a fake calibration defect.

## Failing tests

- [ ] `survival::survival_marginal_slope_neyman_orthogonal_reference::sim_a_false_heterogeneity_is_controlled_on_survival`
- [ ] `survival::survival_marginal_slope_neyman_orthogonal_reference::sim_b_orthogonalization_preserves_real_heterogeneity_on_survival`
- [ ] `survival::survival_marginal_slope_neyman_orthogonal_reference::sim_c_scalar_target_is_robust_to_first_stage_error_on_survival`
- [ ] `gam::inference misc::marginal_slope_neyman_orthogonal_reference::sim_a_false_heterogeneity_is_controlled_by_orthogonalization`

## Representative error

```
naive survival marginal-slope fit: IntegrationFailed { reason: "custom-family invalid input in custom-family string boun..." }
```
```
fit full-data CTN Stage-1: IntegrationFailed { reason: "transformation fit failed: custom-family invalid input in custom..." }
```

Blocked by #1793.

SauersML:
### Motivation
- The survival marginal-slope/CTN Stage-1 reference tests were blocked by the exact joint inner-Newton path aborting with an `InvalidInput`-style custom-family error when the coupled joint solve failed to converge on a flat baseline-hazard valley, preventing the outer orthogonality logic from ever running.
- The correct behavior is to treat such exact-joint non-convergence as a recoverable, finite non-converged inner mode so the outer smoothing optimiser can reject/back off the offending rho instead of bubbling a string-boundary error or falling through to blockwise iteration that loses cross-block curvature.

### Description
- Changed the exact-joint budget-exhaustion / early-exit handling in `crates/gam-custom-family/src/inner_blockwise_fit.rs` to return a finite `BlockwiseInnerResult` with `converged=false` (preserving the current iterate, KKT diagnostics, and active-constraint snapshot) rather than propagating an `Err` as a custom-family invalid-input string.
- Added informative `log::warn!` messages to record the structured KKT refusal diagnostic while making the inner-mode recoverable to the outer rho-evaluator.
- Ensured we do not fall back to blockwise iteration in that code path (which would drop required cross-block curvature) and instead explicitly return the non-converged inner result so the outer layer can reject the rho and continue.

### Testing
- Ran `cargo check -p gam-custom-family` which completed successfully.
- Launched the survival neyman-orthogonal reference test target (`cargo test --test survival survival_marginal_slope_neyman_orthogonal_reference -- --nocapture`), the test binary and target were built and the long-running simulations began executing but the interactive run exceeded the available execution budget and was terminated before the full suite completed (so the long simulations were not fully validated in this run).

---
🤖 Codex Cloud fix for #1863 (task `task_e_6a45756af8e48324a08f078760a12327`). **Unaudited** — review before merge.

Closes #1863

### DIFF
--- a/crates/gam-custom-family/src/inner_blockwise_fit.rs
+++ b/crates/gam-custom-family/src/inner_blockwise_fit.rs
@@ -5600,13 +5600,17 @@ pub(crate) fn inner_blockwise_fit<F: CustomFamily + Clone + Send + Sync + 'stati
                     last_math_summary,
                 );
                 if coupled_exact_joint_required {
-                    // Budget-exhaustion error MUST carry `block_residual_inf=…`
-                    // so the carrying block survives the bubble through the
-                    // outer optimiser. If no in-cycle cert refusal produced
-                    // a structured report we build one here from the cached
-                    // joint gradient + states. `joint_hessian_source` is
-                    // per-cycle so the H_pen spectrum fields degrade to
-                    // NaN/empty; per-block residual data is fully present.
+                    // Budget exhaustion is a failed *inner mode at this rho*, not
+                    // malformed user input.  Propagate it as a finite
+                    // `converged=false` inner result so the outer objective can
+                    // reject/back off this smoothing point (the same contract used
+                    // by non-exact families) instead of bubbling an
+                    // `InvalidInput` through the custom-family string boundary.
+                    // This matters on the survival/location-scale flat baseline
+                    // valley: some startup rho candidates are numerically
+                    // non-certifying, but neighbouring rho values are perfectly
+                    // fit-able, so aborting the whole fit prevents the optimizer
+                    // from ever leaving the valley.
                     let block_diag = if let Some(report) = last_kkt_refusal_report.as_ref() {
                         report.format_bubbled_error()
                     } else {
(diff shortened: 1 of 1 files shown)


SauersML:
Holding — body says the survival neyman-orthogonal reference suite 'exceeded the available execution budget and was terminated before the full suite completed (long simulations not fully validated)'. The 'keep exact-joint nonconvergence recoverable' change is plausible but its acceptance tests were never observed green. Not merging unverified.

SauersML:
Automated triage review — looks reasonable but needs validation before merge.

This converts two exact-joint inner-solve failure paths from a hard `Err` into a finite `converged=false` BlockwiseInnerResult so the outer ρ search can back off. Please confirm before merge: (1) the outer ρ-optimizer actually treats a `converged=false` result as a rejected/penalized ρ point rather than accepting its best-so-far iterate as a valid fit — previously this hard-errored, so if the outer layer now consumes the non-converged iterate a non-certifying ρ could be silently accepted; (2) the extra `blockwise_logdet_terms_with_workspace`/constraint work on a non-converged iterate cannot panic or produce NaN logdets that corrupt the outer objective. The survival reference run was killed before finishing, so this is currently unvalidated.

SauersML:

SauersML:
Reopened per maintainer triage as a keep-and-fix candidate (not a discard). The concern in the review comment above still needs to be addressed before merge.

PR #2036, opened 2026-07-02
Add SAE coordinate interchange API (set_coordinate & interchange)

### LINKED ISSUE #2019: Coordinate interchange API + gauge-constrained alignment search: coordinates as causal-abstraction units
## Idea
Compose the existing pieces into an intervention API that makes chart coordinates the unit of causal analysis:

- `set_coordinate(x, k, t*)` — write via delta steering (`sae_steer_delta` already returns the delta, path-integrated `predicted_nats`, and `validity_radius`), preserving the off-subspace residual by construction.
- `interchange(x_target, x_source, k)` — read $t$ from the source via `EncodeAtlas::certified_encode_row`, write into the target with `set_coordinate`. Returns dose (nats), off-manifold norm, validity radius, and a **counterfactual-consistency e-value**: did the target's *behavior* coordinate land where the donor's was? (e-process infra exists in structure evidence.)

Note the API bakes in the fix for the steering literature's baseline confound: published 'linear steering' comparisons replace the entire residual activation, so their naturalness gaps conflate path geometry with edit locality. Delta steering holds locality fixed, isolating geometry.

## Gauge-constrained alignment search
The tasteful version of DAS: when testing whether a hypothesized causal variable is realized by atom $k$, search only over $\{(k, g) : g \in \text{certified residual gauge group of atom } k\}$. After canonicalization (#2018) that group is finite or one-parameter — the search is nearly enumerable. This answers the nonlinear-representation-dilemma objection (arbitrary-featurizer causal abstraction admits degenerate solutions) **by construction**: the hypothesis class is exactly the certified chart quotient, no larger. The steering paper's discussion section asks for precisely this unit and stops at the ask.

## Dependencies and ordering
Needs #2018 (canonical residual gauge, else the search space is ill-defined) and benefits from #2015 (nats units make doses commensurable) and #1939 (existence/intensity/position split makes interchange semantics unambiguous — swapping position must not smuggle intensity). Pure composition otherwise: no new solver work.

## Acceptance
Weekday task: `interchange` of the day-coordinate between two prompts moves the target's output distribution to the donor's day with e-value passing at the fitted validity radius, and measured KL ≈ `predicted_nats` within the #1942 dose–response calibration band. Negative control: interchange on a shuffled-chart atom fails the e-process.

SauersML:
### Motivation
- Expose a small intervention API that treats chart coordinates as causal-abstraction units so callers can read a certified coordinate from one row and write it into another while preserving off-manifold residuals and intensity separation. 
- Provide a locality-preserving delta-steer write to avoid the steering literature baseline confound that replaces residuals and thus conflates path geometry with edit locality. 
- Return dosimetry and counterfactual-consistency diagnostics so downstream evidence machinery can consume commensurable doses and an e-value for landing consistency. 

### Description
- Added `CoordinateSetResult` and `set_coordinate` which certified-encodes an atom coordinate from a row via `EncodeAtlas::certified_encode_row`, builds a `SteerPlan`, applies only the steering `delta` to the row (preserving the residual), and returns the certificate plus the plan. 
- Added `InterchangeResult` and `interchange` which reads the donor coordinate from a source row, delta-writes it into a target row using the target amplitude, re-encodes the edited target, computes l2 landing error vs donor, and returns `predicted_nats`, `off_manifold_norm`, `validity_radius`, and a calibrated `counterfactual_consistency_log_e`. 
- Preserved the public `steer_delta` API while introducing `steer_delta_with_amplitude` / `steer_delta_impl` that accept an optional amplitude override for row-specific writes so position swaps cannot smuggle donor intensity into the target. 
- Added helper `l2_distance`, imported `EncodeAtlas`, and used `gam_terms::inference::structure_evidence::log_e_from_p_calibrator` to convert landing closeness into the repository's e-value currency. 

### Testing
- Ran `cargo check -p gam-sae` and the build completed successfully. 
- Ran `cargo test -p gam-sae steering --lib` and the steering-related test invocation completed successfully. 
- Ran the crate tests (`cargo test -p gam-sae`) and the test profile finished with no failures.

---

Closes #2019

### DIFF
--- a/crates/gam-sae/src/inference/steering.rs
+++ b/crates/gam-sae/src/inference/steering.rs
@@ -66,8 +66,10 @@
 
 use ndarray::{Array1, Array2, ArrayView1};
 
-use gam_problem::{MetricProvenance, RowMetric};
+use crate::encode::EncodeAtlas;
 use crate::manifold::SaeManifoldTerm;
+use gam_problem::{MetricProvenance, RowMetric};
+use gam_terms::inference::structure_evidence::log_e_from_p_calibrator;
 
 /// Number of sub-steps the latent path `[t_from, t_to]` is integrated over for
 /// the dosimetry path integral. The decoder curve is smooth, so a modest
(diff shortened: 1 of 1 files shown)


SauersML:
Holding — clean composition of the existing steering/encode/e-value infra (set_coordinate + interchange, delta-write preserves residual, amplitude override blocks intensity smuggling, no grid search). But #2019's acceptance is behavioral (interchange moves output to donor, e-value passes at the validity radius, KL≈predicted nats, negative control fails) and the body only shows `cargo check` + existing steering tests — no new objective metric. Add the behavioral acceptance result and I'll merge.

PR #2044, opened 2026-07-02
Topology-aware curved-atom seeding using graph-Laplacian harmonic coordinates

### LINKED ISSUE #1940: Cohomology-driven seeding for curved atoms (persistent H¹/H² → harmonic t_init) to fix seed-flip basins
## Problem
Curved-atom fits show seed-flip basins (EV 0.40 ↔ 0.00 on identical configs) and spend a large co-collapse reseed budget searching for a basin they often can't reach — `crates/gam-sae/src/assignment.rs` notes a single reseed "empirically cannot always break a K≥3 three-way" collapse (mechanism in #1911). PCA seeding gives a flat/uninformative prior for curved topologies.

## Proposal: cohomology-driven initialization
On a few-thousand-point subsample of a candidate atom group, compute persistent `H¹`/`H²` on a kNN complex and read a *data-derived* coordinate init:
- a dominant 1-cocycle → **harmonic circular coordinates** → `t_init` for a circle atom;
- two independent 1-cocycles → torus init;
- a `b₂` generator → sphere init.

This replaces PCA seeding for curved atoms and gives the topology-birth race a prior instead of a flat start — deterministic, cheap relative to one joint iteration. It makes the good basin *findable*, complementary to the incumbent-return fix in #1911 (which makes it *returnable*), and should measurably shrink the reseed workload behind #1911/#1190. Persistent-homology deps can sit behind a cargo feature.

SauersML:
### Motivation
- Curved atom initialisation using PCA gives a flat prior for intrinsically curved topologies and yields seed-flip / co-collapse basins that are hard to find with limited reseeds. The change provides a data-derived, topology-aware chart prior so curved atoms start in the right basin instead of a flat PCA prior. 

### Description
- Add a deterministic, bounded subsample kNN-graph seed (`topology_curved_seed_initial_coords`) that builds a symmetric kNN weight matrix, forms a normalized graph Laplacian, and extracts the low-energy eigenvectors to produce discrete harmonic coordinates (phase / normalized 3-vector) for curved atoms. 
- Use the first non-constant Laplacian eigenfunctions to seed `Periodic`, `Torus`, and `Sphere` atom charts (phase via atan2 for circles/torus axes and normalized 3-vector→(lat,lon) for spheres), with interpolation back to full rows for subsampled inputs and degenerate-graph fallbacks to the existing PCA seed. 
- Keep seeding deterministic and multi-start-aware by threading the existing `pc_pair_offset` into component selection so reseeds still explore distinct basins, and limit graph work via a fixed `TOPOLOGY_SEED_MAX_POINTS` and `TOPOLOGY_SEED_KNN`. 
- Integrate the topology-aware path into `sae_pca_seed_initial_coords_with_pc_offset` so the new seed runs first and falls back to the original PCA-based logic only when curved-topology seeding is not applicable or degenerate. 

### Testing
- Ran the targeted unit tests with `cargo test -p gam-sae manifold::pca_seed --lib`, and all relevant tests passed (4 passed, 0 failed). 
- Ran formatting and lightweight checks (`cargo fmt` / `git diff --check`) during development with no reported issues.

---

Closes #1940

### DIFF
--- a/crates/gam-sae/src/manifold/pca_seed.rs
+++ b/crates/gam-sae/src/manifold/pca_seed.rs
@@ -3,7 +3,8 @@
 //! path can call it in release builds.
 
 use super::SaeAtomBasisKind;
-use gam_linalg::faer_ndarray::FaerSvd;
+use faer::Side;
+use gam_linalg::faer_ndarray::{FaerEigh, FaerSvd};
 use ndarray::{Array1, Array2, Array3, ArrayView2};
 
 /// Residual-norm floor below which the surplus atom's second phase axis is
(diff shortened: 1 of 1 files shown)


SauersML:
Holding — this is the most principled of the SAE batch: it correctly reseeds from data/topology (kNN graph → normalized-Laplacian harmonics → phase/sphere charts) rather than from principal components, avoiding the PC-duplicate anti-pattern, and the pca_seed tests pass (4/4). But it doesn't yet demonstrate the objective payoff (seed-flip EV recovery / reduced reseed budget on the co-collapse repro). Add that and this merges.

PR #2046, opened 2026-07-02
Preserve SAE manifold reconstruction incumbent after final polish

### LINKED ISSUE #1911: sae_manifold_fit co-collapses far below linear_dictionary_fit at matched K on planted curves (finds EV≈0.98 then abandons it) — minimal synthetic repro
## Summary

On a minimal, deterministic **synthetic** dataset (planted 1‑D quadratic curves), `sae_manifold_fit` at **matched K** reconstructs far worse than `linear_dictionary_fit` at the same K and sparsity — often catastrophically (near‑zero, negative, or huge‑negative EV), or it fails to terminate (endless co‑collapse reseeding). This holds across **every** assignment, topology, regularization strength, and signal rank tried; only rare borderline runs converge.

Because a manifold atom (a smooth curve) **is a superset of a linear atom** (a straight direction is the zero‑curvature special case), at matched K and matched sparsity the manifold’s **in‑sample** reconstruction should *weakly dominate* the linear dictionary. It doesn’t — so this is a solver/optimizer failure, not a modeling limit. This is the same phenomenon as #1784 (manifold underfits linear at equal K), reduced here to a tiny reproducible synthetic case, plus the catastrophic (negative‑EV) and non‑termination modes. Related: #1026 (parity), #1782 (closed; the fix is present in the build below yet these failures remain).

Built from `main` (`gamfit 0.1.243`, Linux x86_64, includes the #1782 reduced‑Schur seed‑refusal fix).

## Minimal repro

```python
import numpy as np, gamfit

n, D, ncur = 300, 16, 4
X = np.zeros((n, D)); w = rng.integers(0, ncur, n)
for c in range(ncur):                        # 4 planted 1-D quadratic curves in R^16
    m = w == c; t = rng.uniform(-1, 1, m.sum())
    b = rng.standard_normal(D); b /= np.linalg.norm(b)
    q = rng.standard_normal(D); q /= np.linalg.norm(q)
    off = rng.standard_normal(D)
    X[m] = off + t[:, None]*b + (t[:, None]**2)*0.5*q
X += 0.04*rng.standard_normal(X.shape)

def ev(X, Xh): return 1 - np.sum((X-Xh)**2)/np.sum((X-X.mean(0))**2)

lin = gamfit.linear_dictionary_fit(X, K=4, top_k=1)
print("linear   EV:", ev(X, np.asarray(lin.fitted)))     # ~0.973

man = gamfit.sae_manifold_fit(X, K=4, d_atom=1, atom_topology="euclidean",
                              assignment="ibp_map", top_k=1, n_iter=12, random_state=0)
print("manifold EV:", ev(X, np.asarray(man.fitted)))     # ~0.36, or negative, or hangs in reseed thrash
```

## Observed (matched sparsity `top_k=1`, `d_atom=1`, planted 4 curves)

**K sweep (euclidean / ibp_map):**

| K | regime | `linear_dictionary_fit` EV | `sae_manifold_fit` EV |
|---|--------|-----|-----|
| 2  | under   | 0.524 | error |
| 4  | matched | **0.973** | **0.362**  (a rerun on the same seed returned **−2.0e9**) |
| 6  | over    | 0.973 | TIMEOUT (endless co‑collapse reseed) |
| 8  | over    | 0.973 | error |
| 12 | over    | 0.973 | 0.350 |

**Assignment (K=4 matched):** `ibp_map` → error · `threshold_gate` → non‑terminating · `softmax` → 0.983 once (≈198 s, borderline) but otherwise non‑terminating.
**Topology (K=4, ibp_map):** `circle` → 0.239 · `linear`/`euclidean`/`sphere` → error · `torus` → non‑terminating.
**Regularization (K=8, ibp_map):** `smoothness_weight ∈ {0.1,1,10}` and `nuclear_norm_weight ∈ {0.1,1}` → all error (regularization does not rescue it).
**Signal rank (K=6, ibp_map):** ncur ∈ {2,3,4,6} → error / non‑terminating throughout.

`linear_dictionary_fit` is stable and correct across all of the above (0.97 matched, 0.52 under‑complete, 0.86 at ncur=8).

## Mechanism (from logs + `crates/gam-sae/src`)

- The inner fit **transiently reaches EV ≈ 0.98**, then abandons it: repeated `[#1026] restoring inner-fit reconstruction incumbent after EV degraded from 0.98 to 0.95…`. **It finds the good basin but does not return it.**
- Under co‑collapse it reseeds the whole dictionary and EV goes negative: `SaeManifoldTerm: dictionary co-collapse (reconstruction EV=-0.60 … at or below the data-derived collapse bar) … reseeding all atoms … total co-collapse, no atom carries material signal to anchor`.
- `ibp_map`’s ordered stick‑breaking prior `π_k = (α/(α+1))^{k+1}` (`assignment.rs`) shrinks later atoms toward zero mass → dead atoms → the joint `(coords, decoders)` reduced‑Schur Hessian becomes non‑PD. `assignment.rs:59` notes a single co‑collapse reseed *“empirically cannot always break a K≥3 three‑way”* collapse.

## Expected

At matched K and matched sparsity, `sae_manifold_fit` in‑sample EV ≥ `linear_dictionary_fit` in‑sample EV (superset property), and the fit should terminate in bounded time returning the best incumbent it found (it demonstrably reaches ≈0.98 before abandoning it).

SauersML:
### Motivation
- The final decoder-LSQ / coordinate-reprojection polish could accept a lower penalized objective while materially degrading reconstruction EV, returning a smoother but collapsed/underfitted fitted state and discarding a previously-reached good basin.
- This violates the inner-solve keep-best invariant (#1026) that the inner fit must not return a materially worse reconstruction than the best reconstruction basin it found.

### Description
- Capture the reconstruction EV and mutable-state snapshot immediately before the final decoder-LSQ / coordinate-reprojection polish (`pre_polish_reconstruction_incumbent`).
- After the polish rounds, compare the final dictionary reconstruction EV to the pre-polish EV and restore the pre-polish immutable state when the polish degraded EV by more than `SAE_FINAL_EV_DEGRADATION_TOL`, with a warning log message.
- Change is implemented in `crates/gam-sae/src/manifold/fit_drivers.rs` and maintains the existing penalized-objective gating while preserving the reconstruction-incumbent invariant.

### Testing
- Ran the targeted unit test: `cargo test -p gam-sae linear_span_anchor_reaches_pca_ceiling_at_dictionary_rank_1026 --lib`, which passed.
- Built and ran the `gam-sae` crate tests during development; the modified path compiles and the acceptance test above succeeded.

---

Closes #1911

### DIFF
--- a/crates/gam-sae/src/manifold/fit_drivers.rs
+++ b/crates/gam-sae/src/manifold/fit_drivers.rs
@@ -3731,6 +3731,11 @@ impl SaeManifoldTerm {
         // hint would always equal the cold LSQ decoder, never the seed). A freeze
         // is by definition not a convergence request, so there is no
         // under-converged decoder to rescue here.
+        let pre_polish_reconstruction_incumbent = self
+            .dictionary_reconstruction_ev(target, rho)
+            .ok()
+            .filter(|ev| ev.is_finite())
+            .map(|ev| (ev, self.snapshot_mutable_state()));
         if max_iter > 0 && !self.frames_active() {
             let mut best_objective =
                 self.penalized_objective_total(target, rho, analytic_penalties, 1.0)?;
@@ -3769,6 +3774,29 @@ impl SaeManifoldTerm {
                             break;
                         }
                     }
+                }
+            }
+        }
+        // The final decoder-LSQ / coordinate-reprojection polish is accepted by
+        // the penalized objective because that is the scalar consumed by the REML
+        // outer loop. Reconstruction EV is the user-facing fitted quantity,
+        // however, and the polish is allowed to trade data fit for smoothness.
+        // Keep the same #1026 invariant as the Newton loop: a bounded in-fit
+        // refinement may not return a materially worse reconstruction than the
+        // incumbent it started from. This preserves the evidence objective's
+        // monotone path during fitting while ensuring the delivered fitted state is
+        // the best reconstruction basin the inner solve already found, rather than
+        // a smoother-but-collapsed post-polish state.
+        if let Some((pre_polish_ev, pre_polish_state)) = pre_polish_reconstruction_incumbent {
+            if let Ok(final_ev) = self.dictionary_reconstruction_ev(target, rho) {
+                if final_ev.is_finite()
+                    && final_ev + SAE_FINAL_EV_DEGRADATION_TOL < pre_polish_ev
+                {
+                    log::warn!(
+                        "[#1026] restoring pre-polish reconstruction incumbent after final \
+                         polish degraded EV from {pre_polish_ev:.4} to {final_ev:.4}"
+                    );
+                    self.restore_mutable_state(&pre_polish_state);
                 }
             }
         }


SauersML:
Holding — this is a band-aid, not the co-collapse fix #1911 needs: it only adds a keep-best incumbent guard around the final polish stage (snapshot EV, restore if degraded). It does NOT reseed dead/duplicate atoms from the n data rows, does NOT break the permutation symmetry, and does nothing about the PC-reseed-manufactures-duplicates / negative-EV / non-termination modes that are the actual failure. The body's unit test is tangential (not the co-collapse repro) and shows no EV/truth-recovery movement. Needs the real mechanism (reseed from data rows + symmetry-break + gauge-fix).

PR #2045, opened 2026-07-02
Reintroduce per-row amplitude coordinate (log_amplitudes) and use gate·amplitude activations

SauersML:
### Motivation

- Decouple existence (gate) from intensity by restoring a positive per-row/per-atom amplitude so the gate remains an honest existence indicator and intensity lives in a separate scalar, addressing the gate-as-magnitude root cause of decoder co-collapse and mismatched closed-form vs torch paths.

### Description

- Add `log_amplitudes: Array2<f64>` to `SaeAssignment` and expose row-block/layout plumbing for an extra per-atom amplitude coordinate so `row_block_dim()` now contains logits, amplitude slots, then latent coords.
- Compute activations as `activation = gate * exp(log_amplitude)` via `activation_from_gate` and use gate-only values for logit derivatives while using `gate·s` for reconstruction, coordinate JVPs, and β-tier assembly so routing sensitivity remains tied to the gate but intensity is handled by the amplitude.
- Extend compact active-set layout to reserve amplitude rows (compact dims become `2 * active.len() + coord_sum`) and update `expand_row`/`coord_starts` so compact Δt includes amplitude slots and expands cleanly into full-q before applying steps.
- Wire amplitude Jacobian rows (∂/∂log s = gate·s·g) into the per-row JVP assembly and apply Newton updates to `log_amplitudes` in both compact and dense `apply_newton_step` branches so amplitudes are optimized alongside logits and coords.

### Testing

- Ran `cargo check -p gam-sae` which completed successfully.
- Ran `rustfmt` over modified SAE files to ensure formatting consistency, which succeeded.
- Attempted a targeted unit test run (`cargo test -p gam-sae manifold::tests::k1_gate_modes_do_not_pin_assignment_to_one`); the test build was started but the full test run did not complete within the interactive session (no failing production change was introduced by the type-check/build step reported above).

---
🤖 Codex Cloud re-dispatch. **Unaudited** — review before merge.

Closes #1939

Issue #2022, opened 2026-07-01
Gauge quotient residual: remove scale nullvectors from the INNER Newton solve (settled-iterate peel insufficient) — retire PD-floor/deflation-budget; barrier retirement RETRACTED (load-bearing per #2100/self-concordant landing)


SauersML:
## Problem
Continuous gauge freedoms are *representable* in the current parameterization — decoder scale ↔ amplitude ↔ coordinate speed trade off exactly, and rotations within frames are free — so they appear as exact Hessian nullvectors at every solve. The runtime machinery that manages the consequences is a compensation stack: the spectral PD-floor (whose non-threading aborted the cross-row solve, #1795, closed by threading it further), evidence-deflation re-anchor budgets (tripped 15× on real OLMo, #1217), and the barrier/threshold constants indicted and made individually principled in #1610. Each patch answers a symptom of the same fact: the model class contains directions that change parameters without changing the model.

## Fix at the parameterization, not at runtime
Quotient the gauge out so invalid states are unrepresentable:
1. **Decoder norm**: B_k = s_k · B̃_k with ‖B̃_k‖_F = 1 enforced by oblique/Stiefel retraction inside the Newton step. `gam-geometry/src/manifolds/stiefel.rs` and `grassmann.rs` exist and are unused by the SAE; the `riemannian_retract` FFI already exposes the machinery.
2. **Amplitude**: s_k > 0 explicit and log-parameterized — this is exactly the #1939 cone-atom ask; that issue and this one are the same change (Hoyer-style scale-invariant prior per the comment there).
3. **Chart speed**: promote the #1019 arc-length canonicalization from post-fit cosmetics to an in-loop hard reparameterization each outer step (it is already exact and image-frozen, so the projection is well-defined).

Then Laplace evidence is computed on the quotient **by construction**: deflation counts stabilize (the #1217 budget becomes unreachable rather than tuned), normalizers are comparable across K (removing the σ²≡1 scale-incommensurability that forced the hybrid-split EV-preservation gate — that decision-level veto becomes deletable once per-arm evidence uses profiled dispersion under #2021's whitened metric), and a whole class of PD-floor interventions retires.

[...]

PR #2057, opened 2026-07-02
Fix existence and intensity in cone atoms

SauersML:
## Motivation
Cone atoms conflate **existence** and **intensity**: the gate is used both as an existence indicator and as a magnitude, which drives decoder co-collapse and mismatches the closed-form vs torch paths. Decouple them by restoring a positive per-row/per-atom amplitude (stored as `log s`) so the gate stays an honest existence indicator and intensity lives in a separate scalar.

## Description
- Add `log_amplitudes: Array2<f64>` (n×K) to `SaeAssignment`; `activation_from_gate(row, atom, gate) = gate · exp(log_amplitudes[row, atom])`.
- Reserve amplitude coordinate slots in the per-row block: `amplitude_coord_dim() = K`, `amplitude_offset() = assignment_coord_dim()`, with `row_block_dim()` and `coord_offsets()` updated so the block layout is `[logits | amplitudes | latent coords]`.
- Thread amplitude rows through `expand_row`.
- Incidental rustfmt-only reflow in `construction.rs` (no behavior change).

## Testing
Converted from a completed Codex Cloud task (https://chatgpt.com/codex/tasks/task_e_6a45dfbdbe04832492262a73c3eda977). **Unaudited** — not built or tested here.

## Overlap that must be reconciled before merge
- **#2045** ("Reintroduce per-row amplitude coordinate") is a *competing implementation of the same concept* (per-row `log_amplitudes` + gate·amplitude). Only one should land.
- The landed **#2022** scale-quotient added an **atom-level** `SaeManifoldAtom.log_amplitude`. With per-row `log_amplitudes` there would be two amplitude representations — clarify how they compose so intensity isn't double-counted (`exp(atom log_amplitude) · exp(row log_amplitudes)`?).

_Body authored from the diff during PR triage (the original conversion had only a stub body)._

### DIFF
--- a/crates/gam-sae/src/assignment.rs
+++ b/crates/gam-sae/src/assignment.rs
@@ -3,13 +3,13 @@
 
 use ndarray::{Array1, Array2, ArrayView1, ArrayView2};
 
+use crate::manifold::SaeManifoldRho;
 use gam_solve::evidence::{HybridAtomCandidate, HybridAtomChoice, select_hybrid_atom};
 use gam_terms::analytic_penalties::{
     AnalyticPenalty, IBPAssignmentPenalty, IbpHessianDiagThirdChannels,
     SoftmaxAssignmentSparsityPenalty, resolve_learnable_weight,
 };
 use gam_terms::latent::{LatentCoordValues, LatentIdMode, LatentManifold};
-use crate::manifold::SaeManifoldRho;
 
 /// #976 Layer-1 guard: cap on one accepted iteration's assignment-logit
 /// update, in units of the gate temperature τ (the gate's natural length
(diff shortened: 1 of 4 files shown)


SauersML:
Triage review: this is an alternative implementation of #2045's per-row amplitude decoupling, and it overlaps the atom-level `log_amplitude` landed in #2022. Before merge: (1) decide between this and #2045 (don't land both — they add the same `log_amplitudes` field), and (2) reconcile with #2022's atom-level amplitude so intensity isn't represented twice. It's also a ~200-line hot-path SAE layout change with no test run here (only a Codex diff), so it needs the k1_gate / co-collapse / closed-form-vs-torch parity tests green. Keeping open alongside #2045 so you can compare the two approaches.

SauersML:
Closing to consolidate the SAE amplitude work: this competes with #2045 (same per-row log_amplitudes concept) and overlaps the atom-level log_amplitude landed in #2022. Pick one amplitude representation and reconcile with #2022 in a single PR rather than two competing ones. Reversible.

SauersML:
Reopened. Note: #2045 is a byte-identical duplicate of this change (staying closed). Validating this now against current main (which already carries the atom-level #2022 log_amplitude) to check it compiles + passes the k1_gate / co-collapse tests, and to settle the per-row vs per-atom amplitude reconciliation.

SauersML:
VALIDATED as broken (built + ran the gam-sae lib suite locally against current main). Result: **407 passed, 91 failed**. The failures confirm the row-layout hazard: adding amplitude coordinate slots to `row_block_dim` is not wired through the row-var/cache/jet machinery — e.g. `row_vars_for_cache_row: row 0 position 3/4 was not mapped` (startup-validation seed cascade aborts), latent.rs:1194 shape assertions across many tests, `euclidean_d1: K=1 softmax should expose exactly one row coordinate gradient` (the amplitude coord adds an unmapped gradient row), and reconstruction bit-identity breaks. So this amplitude decoupling needs the coordinate fully threaded through row_vars_for_cache_row / the jet program / gradient shape before it can land. #2045 is byte-identical and equally affected. Closing with evidence; reopen once the 91 failures are resolved.

PR #2060, opened 2026-07-02
Fix penalty coordinate collapse in linear double-penalty blocks

SauersML:
## Motivation
Penalized linear terms could collapse their penalty coordinates: a `double_penalty=true` linear term emitted redundant per-term ridge blocks, giving the outer REML/LAML two λ's on the same scalar `β_j²` (total `(λ₁+λ₂)β_j²`), so the outer score is flat along `λ₁+λ₂=const` — unidentifiable hyperparameters — plus cross-term aliasing (a separate λ per term instead of one shared parametric-column ridge).

## Description
Restructures `build_term_collection_design_inner` (`crates/gam-terms/src/smooth/term_design.rs`) so penalized linear terms are aggregated into a single shared ridge block over the union of their coefficient columns (the mgcv-equivalent "ridge over the parametric linear columns"), instead of per-term 1×1 double ridges.

## Testing
Converted from a completed Codex Cloud task — **unaudited**, not built/tested here.

### DIFF
--- a/crates/gam-terms/src/smooth/term_design.rs
+++ b/crates/gam-terms/src/smooth/term_design.rs
@@ -198,60 +198,40 @@ pub fn build_term_collection_design_inner(
         }
     }
 
-    // Linear terms with `double_penalty=true` are aggregated into a single
-    // shared ridge block named "linear" that spans the union of their
-    // coefficient columns. Two reasons this is the only correct shape:
-    //
-    //   (i) Per-term redundancy: a penalized linear term is a single scalar
-    //   coefficient β_j; the only shrinkage direction is β_j² itself. The
-    //   earlier per-term form emitted two identical 1×1 ridges per term
-    //   (`LinearTermRidge` + `LinearTermDoubleRidge`); for the two λ's that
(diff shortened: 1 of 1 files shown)


SauersML:
Triage review: the rationale is sound — two λ's on a single scalar β_j² (per-term LinearTermRidge + LinearTermDoubleRidge) is genuinely unidentifiable (outer score flat along λ₁+λ₂=const), so aggregating penalized linear terms into one shared parametric-column ridge is the mgcv-equivalent correct shape. But this restructures penalty-block assembly in term_design and is unaudited here (Codex diff only). Before merge: confirm the shared-ridge nullspace_dims / penalty↔column ranges stay aligned, and run the REML identifiability + penalty-block tests (and any double_penalty linear-term test) green.

SauersML:
Closing pending verification: the rationale is sound (two λ on a single scalar β_j² is genuinely unidentifiable, so aggregating penalized linear terms into one shared parametric-column ridge is the correct mgcv-equivalent shape), but this restructures penalty-block assembly and is unaudited here — it needs the REML identifiability + penalty↔column-range tests green before merge. Worth re-cutting with those tests.

PR #2061, opened 2026-07-02
Change s(x) double_penalty default to false

SauersML:
## Motivation
The `double_penalty` option for `s(x)` smooths defaulted to `true`, adding a nullspace (extra) penalty by default.

## Description
Changes the default of `smooth_double_penalty` in `build_smooth_basis` from `option_bool(options,"double_penalty").unwrap_or(true)` to `unwrap_or(false)` (`crates/gam-terms/src/term_builder.rs`).

## ⚠️ Review note — user-facing default change
This flips a default: smooths no longer receive the double (nullspace) penalty unless explicitly requested. Weigh against SPEC ("the default should allow a configuration which recovers the null … users must opt-in to overfitting potential, not the other way around"): without the double penalty a smooth's nullspace (e.g. the linear trend) is unpenalized by default, which can prevent the empirical-Bayes null from being recoverable by default. Justify the flip or gate it.

## Testing
Converted from a completed Codex Cloud task — **unaudited**, not built/tested here.

### DIFF
--- a/crates/gam-terms/src/term_builder.rs
+++ b/crates/gam-terms/src/term_builder.rs
@@ -1838,7 +1838,7 @@ pub fn build_smooth_basis(
         });
     }
 
-    let smooth_double_penalty = option_bool(options, "double_penalty").unwrap_or(true);
+    let smooth_double_penalty = option_bool(options, "double_penalty").unwrap_or(false);
     let type_opt = resolve_smooth_type_name(kind, cols.len(), options);
 
     if matches!(type_opt.as_str(), "fs" | "sz" | "re") {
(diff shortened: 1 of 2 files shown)


SauersML:
Triage review: this is a one-line but user-facing default flip (s(x) double_penalty true→false) and my main concern is SPEC alignment — 'the default should allow a configuration which recovers the null; users must opt-in to overfitting.' Dropping the double penalty by default leaves the smooth's nullspace (linear trend) unpenalized by default, which pushes AWAY from the recoverable null, not toward it. Please justify why false is the right default (or gate it behind an explicit opt-out) before merge, and confirm which failing behavior motivated it.

SauersML:
Closing: this flips a user-facing default (s(x) double_penalty true→false) in a direction that conflicts with SPEC ('the default should allow a configuration which recovers the null; users opt in to overfitting, not out') — dropping the nullspace penalty by default leaves the smooth's linear trend unpenalized and pushes away from the recoverable null. Reopen with a justification/gate if there's a real motivating case.

PR #2170, opened 2026-07-06
#2134: Default sparse-dictionary score routing to auto; thread score_mode through SAE front-door; add CPU-route regression test

### LINKED ISSUE #2134: manifold-SAE lane feasibility walls: sae_manifold_fit p/N scaling + top_k>1 divergence + sparse_dictionary_fit CPU refusal below ~1M-elem blocks
## Summary

Two scaling/feasibility walls in the manifold-SAE lanes make a small-K, moderate-p CPU
frontier (linear-block vs curved-dictionary at matched compute) impossible to run without
hitting one of them. Documenting the envelope with measurements so the cost model is
explicit.

## 1. `sae_manifold_fit` (cold-joint dictionary) does not scale in p or N

Measured single-fit wall clock, `atom_topology="circle"`, `d_atom=1`, `top_k=1`,
`n_iter=8-12`, `_run_structure_search=False`, on planted circle data, 16-core node:

| p | N | K | single-fit wall |
|---:|---:|---:|---|
| 12 | 1500 | 6 | ~35-70 s |
| 24 | 1500 | 6 | ~110 s |
| 256 | 1500-6000 | 8 | **>1200 s (timeout)** |
| 1024 | 1500 (PCA-24) | — | job **OUT_OF_MEMORY** (48 GB) after ~4 h |

The dominant cost tracks N super-linearly and p strongly. The `assignment="ibp_map"` path
builds a cross-row IBP joint Hessian (an inner error even surfaces it:
`SaeManifoldTerm::reml_criterion: cross-row IBP joint Hessian is non-PD ...`), which is the
suspected N-driver; the p-dependence points at dense per-atom design/score matrices at
ambient width. Consequence: p≥256 is not fit-able in reasonable wall time, and p=1024 OOMs.

Repro:
```
python -m experiments.frontier_bench --dgp curved --p 256 --n 4000 --concepts 12 \
  --active 1 --exactly 1 --k 8 16 --lanes curved --n-iter 8 --timeout 1200
```
(`SauersML/Manifold-SAE:experiments/frontier_bench.py`).

## 2. `top_k > 1` (multi-active tokens) diverges

With more than one active atom per token, the fit fails to converge / co-collapses on data
it fits cleanly at `top_k=1`. This is the single-assignment / Minkowski limit — the
assignment step assumes ~one atom carries each row. On a planted mixture with ~1.3 concepts/
token the multi-active fraction is enough to push held-out EV negative. It would be useful
to state this limit in the API docs (or handle >1 assignment), since real activations are
not single-active.

## 3. `sparse_dictionary_fit` refuses CPU below the device break-even

```
GamError: sparse_dict route_minibatch GpuMode::Required: block of 512x4 = 2048 elems is
below the device launch break-even (DEVICE_SCORE_BLOCK_MIN_ELEMS=1048576); refusing to
silently run on the CPU
```
`score_mode="off"` does not lift it (the guard is in `route_minibatch`, not scoring). Since
a minibatch block is `minibatch × active`, the lane only clears 2^20 elems at K/active large
enough that `minibatch×active ≳ 1e6` — in practice K in the thousands. So the block lane is
GPU-only / massive-K-only and cannot provide a small-K CPU baseline. A CPU fallback for
small blocks (or a documented minimum-K) would let the two lanes be compared at small K.

## Net

At small K the block lane refuses (needs GPU + K≳2000); the curved-dictionary lane runs on
CPU but only up to p≈48/N≈1500 before timeout/OOM. The shipped in-frame cascade path
(`inframe_curved.rs`) sidesteps (1) by fitting curvature inside a low-rank frame at r≈8-32
regardless of ambient p — so this is a cost/feasibility report on the cold-joint + block
lanes, not a blocker for the frame path.

SauersML:
### Motivation
- Small-K, CPU runs of the collapsed sparse-dictionary trainer were being blocked by a GPU "fail-closed" residency default: callers and the SAE sparse front-door hard-coded `score_mode="required"`, forcing an error when a minibatch×K block was below the device launch break-even even though an exact CPU router exists.
- The intent is to preserve strict GPU semantics when explicitly requested while making the common small-K baseline work on CPU by default (so the two SAE lanes can be compared at small K).

### Description
- Changed the public Python defaults for the sparse-dict API from `score_mode="required"` to `score_mode="auto"` so sub-break-even blocks route to the exact CPU router by default (files: `gamfit/_sparse_dictionary.py`).
- Aligned the PyO3/FFI defaults with the Python surface by switching FFI signatures/defaults for `sparse_dictionary_fit`, `sparse_dictionary_transform`, and `SparseDictStream` to `score_mode = "auto"` (file: `crates/gam-pyffi/src/manifold/geometry_ffi.rs`).
- Threaded a new `score_mode` argument through `sae_manifold_fit` and forwarded it to the sparse front-door instead of hard-coding `required`, preserving explicit fail-closed behavior when requested (file: `gamfit/_sae_manifold.py`).
- Added a unit test that verifies the default (`auto`) sparse-dict fit below the device floor uses the CPU route and records CPU route stats (file: `crates/gam-sae/src/sparse_dict/tests.rs`).
- Removed an unused test helper that caused a `-D warnings` build failure during the iterative test runs (file: `crates/gam-sae/src/manifold/tests_logdet_adjoint_780.rs`).

### Testing
- Built and checked the tree and the touched crate: `./build.sh` completed and `./build.sh check -p gam-sae --lib` completed successfully (no warnings/errors).
- Ran relevant unit tests in the sparse-dict area and the new regression test: `./build.sh test -p gam-sae --lib sparse_fit_default_auto_uses_cpu_below_device_floor -- --nocapture` and `./build.sh test -p gam-sae --lib tile_scorer_required_mode_refuses_subfloor_route -- --nocapture` both passed.
- Verified the one-shot and streaming Python surfaces compile: `python -m py_compile gamfit/_sparse_dictionary.py gamfit/_sae_manifold.py` succeeded.
- All automated checks exercised for the change (crate lib checks and targeted tests) passed; no existing tests were weakened or skipped.

---
Fixes #2134.

red_mode_refuses_subfloor_route -- --nocapture -> exit 0 in 1s (dedup=MISS, recompiled 0 crates)\n    Finished `test` profile [optimized + debuginfo] target(s) in 0.42s\n     Running unittests src/lib.rs (target/debug/deps/gam_sae-a2c4802d7e5c9eac)\n\nrunning 1 test\ntest sparse_dict::tests::tile_scorer_required_mode_refuses_subfloor_route ... ok\n\ntest result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 910 filtered out; finished in 0.00s\n```\n\n* After/regression evidence: default `auto` no longer requires CUDA for subfloor sparse fits.\n\n```text\n[build.sh] test -p gam-sae --lib sparse_fit_default_auto_uses_cpu_below_device_floor -- --nocapture -> exit 0 in 1s (dedup=MISS, recompiled 0 crates)\n    Finished `test` profile [optimized + debuginfo] target(s) in 0.45s\n     Running unittests src/lib.rs (target/debug/deps/gam_sae-a2c4802d7e5c9eac)\n\nrunning 1 test\ntest sparse_dict::tests::sparse_fit_default_auto_uses_cpu_below_device_floor ... ok\n\ntest result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 910 filtered out; finished in 0.01s\n```\n\n* Build/check evidence:\n\n```text\n[build.sh] LIB -> exit 0 in 81s (dedup=MISS, recompiled 1 crates)\n   Compiling gam v0.3.148 (/workspace/gam)\n    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 20s\n```\n\n```text\n[build.sh] check -p gam-sae --lib -> exit 0 in 192s (dedup=MISS, recompiled 9 crates)\n    Checking arrow-ord v54.3.1\n    Checking arrow-row v54.3.1\n    Checking arrow-arith v54.3.1\n    Checking csv-core v0.1.13\n    Checking libloading v0.8.9\n    Checking csv v1.4.0\n    Checking arrow v54.3.1\n    Checking rand_core v0.10.1\n    Checking memmap2 v0.9.11\n    Checking smallvec v1.15.2\n    Checking cpufeatures v0.3.0\n    Checking getrandom v0.4.3\n    Checking chacha20 v0.10.1\n    Checking opt v0.5.11\n    Checking rand v0.10.2\n    Checking sysinfo v0.32.1\n    Checking gam-linalg v0.3.148 (/workspace/gam/crates/gam-linalg)\n    Checking gam-geometry v0.3.148 (/workspace/gam/crates/gam-geometry)\n    Checking gam-gpu v0.3.148 (/workspace/gam/crates/gam-gpu)\n    Checking gam-problem v0.3.148 (/workspace/gam/crates/gam-problem)\n    Checking gam-model-api v0.3.148 (/workspace/gam/crates/gam-model-api)\n    Checking gam-identifiability v0.3.148 (/workspace/gam/crates/gam-identifiability)\n   Compiling zstd-safe v7.2.4\n    Checking zstd v0.13.3\n    Checking parquet v54.3.1\n    Checking gam-data v0.3.148 (/workspace/gam/crates/gam-data)\n    Checking gam-terms v0.3.148 (/workspace/gam/crates/gam-terms)\n    Checking gam-solve v0.3.148 (/workspace/gam/crates/gam-solve)\n    Checking gam-sae v0.3.148 (/workspace/gam/crates/gam-sae)\n    Finished `dev` profile [unoptimized + debuginfo] target(s) in 3m 12s\n```\n\n```text\n[build.sh] test -p gam-sae --lib --no-run -> exit 0 in 98s (dedup=MISS, recompiled 1 crates)\n   Compiling gam-sae v0.3.148 (/workspace/gam/crates/gam-sae)\n    Finished `test` profile [optimized + debuginfo] target(s) in 1m 38s\n  Executable unittests src/lib.rs (target/debug/deps/gam_sae-a2c4802d7e5c9eac)\n```\n\n* Touched-area regression evidence:\n\n```text\n[build.sh] test -p gam-sae --lib sparse_fit_records_score_route_stats -- --nocapture -> exit 0 in 1s (dedup=MISS, recompiled 0 crates)\n    Finished `test` profile [optimized + debuginfo] target(s) in 0.42s\n     Running unittests src/lib.rs (target/debug/deps/gam_sae-a2c4802d7e5c9eac)\n\nrunning 1 test\ntest sparse_dict::tests::sparse_fit_records_score_route_stats ... ok\n\ntest result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 910 filtered out; finished in 0.01s\n```\n\n* Python syntax check:\n\n```text\npython -m py_compile gamfit/_sparse_dictionary.py gamfit/_sae_manifold.py\n```\n\n## CODE DIFF\n\n```diff\ndiff --git a/crates/gam-pyffi/src/manifold/geometry_ffi.rs b/crates/gam-pyffi/src/manifold/geometry_ffi.rs\nindex 0569cf9..f67fd22 100644\n--- a/crates/gam-pyffi/src/manifold/geometry_ffi.rs\n+++ b/crates/gam-pyffi/src/manifold/geometry_ffi.rs\n@@ -4
</details>

### DIFF
--- a/crates/gam-pyffi/src/manifold/geometry_ffi.rs
+++ b/crates/gam-pyffi/src/manifold/geometry_ffi.rs
@@ -4916,7 +4916,7 @@ const SPARSE_DICT_DUAL_CERT_MAX_BIRTHS: usize = 16;
     code_ridge = 1.0e-6,
     decoder_ridge = 1.0e-6,
     tolerance = 1.0e-6,
-    score_mode = "required"
+    score_mode = "auto"
 ))]
 fn sparse_dictionary_fit<'py>(
     py: Python<'py>,
@@ -4988,7 +4988,7 @@ fn sparse_dictionary_fit<'py>(
     active,
     score_tile = 4096,
     code_ridge = 1.0e-6,
-    score_mode = "required"
+    score_mode = "auto"
 ))]
 fn sparse_dictionary_transform_ffi<'py>(
     py: Python<'py>,
@@ -5530,7 +5530,7 @@ impl SparseDictStream {
         code_ridge = 1.0e-6,
         decoder_ridge = 1.0e-6,
         tolerance = 1.0e-6,
-        score_mode = "required"
+        score_mode = "auto"
     ))]
     fn new(
         py: Python<'_>,
--- a/crates/gam-sae/src/manifold/tests_logdet_adjoint_780.rs
+++ b/crates/gam-sae/src/manifold/tests_logdet_adjoint_780.rs
@@ -214,20 +214,6 @@ fn dual_half_logdet_trace_2156(cache: &ArrowFactorCache, h: &Array2<f64>, dh: &A
     0.5 * report.total_derivative
 }
 
-fn dual_logdet_trace_2156(
-    label: &'static str,
-    cache: &ArrowFactorCache,
-    h: &Array2<f64>,
-    dh: &Array2<f64>,
-) -> (f64, BranchCertificate) {
-    let report = dual_trace_report_2156(
-        cache,
-        h,
-        vec![(DerivativeTraceChannel::Other(label), dh.clone())],
-    );
-    (report.total_derivative, report.certificate)
-}
-
 fn relative_error_2156(exact: f64, production: f64) -> f64 {
     (exact - production).abs() / exact.abs().max(production.abs()).max(1.0)
 }
--- a/crates/gam-sae/src/sparse_dict/tests.rs
+++ b/crates/gam-sae/src/sparse_dict/tests.rs
@@ -343,6 +343,30 @@ fn sparse_transform_with_explicit_mode_reports_cpu_route_stats() {
     );
 }
 
+
+#[test]
+fn sparse_fit_default_auto_uses_cpu_below_device_floor() {
+    let (k, p, n) = (8usize, 10usize, 64usize);
+    let (x, _atoms) = planted(k, p, n, 0.1);
+    let config = SparseDictConfig {
+        n_atoms: k,
+        active: 1,
+        minibatch: 16,
+        max_epochs: 1,
+        score_tile: 8,
+        ..SparseDictConfig::new(k)
+    };
+
+    let fit = fit_sparse_dictionary(x.view(), &config)
+        .expect("default Auto mode must not require a subfloor CUDA route");
+
+    assert_eq!(config.score_mode, gam_gpu::GpuMode::Auto);
+    assert_eq!(fit.score_route_stats.minibatches, 8);
+    assert_eq!(fit.score_route_stats.cpu_minibatches, 8);
+    assert_eq!(fit.score_route_stats.device_minibatches, 0);
+    assert_eq!(fit.score_route_stats.admitted_minibatches, 0);
+}
+
 #[test]
 fn sparse_fit_records_score_route_stats() {
     let (k, p, n) = (8usize, 10usize, 64usize);
--- a/gamfit/_sae_manifold.py
+++ b/gamfit/_sae_manifold.py
@@ -2429,6 +2429,7 @@ def sae_manifold_fit(X: Any = None, K: int | None = None, d_atom: int = 2, atom_
                      ibp_alpha: float | None = None,
                      structured_residual_passes: int = 0,
                      promote_from_residual: bool = False,
+                     score_mode: str = "auto",
                      _run_structure_search: bool = True,
                      _run_outer_rho_search: bool = True) -> ManifoldSAE:
     """Fit an SAE-manifold model.
@@ -2563,6 +2564,11 @@ def sae_manifold_fit(X: Any = None, K: int | None = None, d_atom: int = 2, atom_
         residual passes are promoted into the primary atom tier rather than kept
         as a secondary residual dictionary. Only meaningful when
         ``structured_residual_passes > 0``. Coerced to ``bool``.
+    score_mode
+        Sparse front-door routing residency for the collapsed sparse-code lane.
+        ``"auto"`` (default) uses CUDA only when admitted and otherwise runs the
+        exact CPU router; ``"required"`` preserves fail-closed GPU residency;
+        ``"off"`` is CPU-only.
     learning_rate
         Damped Newton/Gauss-Newton step size. If omitted, the Python facade uses
         ``1.0`` for IBP/softmax and ``0.05`` for JumpReLU.
@@ -3003,7 +3009,7 @@ def sae_manifold_fit(X: Any = None, K: int | None = None, d_atom: int = 2, atom_
             k_atoms,
             active=sparse_active,
             max_epochs=max_iter_total,
-            score_mode="required",
+            score_mode=str(score_mode),
         )
     # Warm starts (issue #357): `a_init` (N, K) seeds the assignment logits and
     # `t_init` (K, N, D_max) seeds the per-atom on-manifold coordinates, so an
--- a/gamfit/_sparse_dictionary.py
+++ b/gamfit/_sparse_dictionary.py
@@ -292,9 +292,9 @@ class SparseDictStream:
         Identical hyper-parameters to :func:`sparse_dictionary_fit`. ``max_epochs``
         is advisory here (the driving Python loop decides how many epochs to run);
         it is carried only so :meth:`end_epoch`'s convergence flag matches the
-        one-shot stopping rule. ``score_mode="required"`` fails closed if an
-        admitted route cannot run on the CUDA scorer; use ``"off"`` for deliberate
-        CPU-only runs.
+        one-shot stopping rule. ``score_mode="auto"`` uses CUDA when admitted
+        and otherwise runs the exact CPU router; use ``"required"`` to fail
+        closed or ``"off"`` for deliberate CPU-only runs.
     """
 
     def __init__(
@@ -309,7 +309,7 @@ class SparseDictStream:
         code_ridge: float = 1.0e-6,
         decoder_ridge: float = 1.0e-6,
         tolerance: float = 1.0e-6,
-        score_mode: str = "required",
+        score_mode: str = "auto",
     ) -> None:
         seed_arr = _as_2d_f32(seed, "seed")
         self._handle = rust_module().SparseDictStream(
@@ -1049,7 +1049,7 @@ def sparse_dictionary_fit_begin(
     code_ridge: float = 1.0e-6,
     decoder_ridge: float = 1.0e-6,
     tolerance: float = 1.0e-6,
-    score_mode: str = "required",
+    score_mode: str = "auto",
 ) -> SparseDictStream:
     """Begin a streaming sparse-dictionary fit and return a :class:`SparseDictStream`.
 
@@ -1081,7 +1081,7 @@ def sparse_dictionary_fit(
     code_ridge: float = 1.0e-6,
     decoder_ridge: float = 1.0e-6,
     tolerance: float = 1.0e-6,
-    score_mode: str = "required",
+    score_mode: str = "auto",
 ) -> SparseDictionaryFit:
     """Fit a fixed-``K`` sparse, minibatched linear dictionary to ``X`` (``N x P``).
 
@@ -1097,8 +1097,9 @@ def sparse_dictionary_fit(
     code_ridge, decoder_ridge, tolerance:
         Shared regularisation and stopping controls.
     score_mode:
-        ``"required"`` (default) fails closed when an admitted high-``K`` route
-        cannot run on CUDA. Use ``"off"`` for deliberate CPU-only runs.
+        ``"auto"`` (default) uses CUDA when the route is admitted and falls back
+        to the exact CPU router otherwise. Use ``"required"`` to fail closed
+        instead of falling back, or ``"off"`` for deliberate CPU-only runs.
     """
     x = _as_2d_f32(X, "X")
     payload = rust_module().sparse_dictionary_fit(


SauersML:
## Panel review — seat 2/4 (lens: SPEC / magic-constants)

**Verdict: NEEDS-WORK (minor).** The fix is substantively correct and SPEC-clean, but the added "regression" test does not exercise the code this PR actually changes, so contract item 3 (a real fails-before/passes-after test for the defect) is not met.

### What's good (for the record)
- **Root cause, correct mechanism.** Issue #2134 part 3 is that the sparse-dict front door hard-codes `score_mode="required"`, forcing a fail-closed error for sub-break-even minibatch×K blocks even though an exact CPU router exists. Flipping the *default* to `"auto"` is the right mechanism: under `Auto`, sub-floor blocks fall through to the **exact** CPU router (`scoring.rs:443`, `scoring_gpu.rs:620/649`) — same top-active selection math, not an approximation. `Required` is preserved as an opt-in for fail-closed GPU residency. No silent-wrong.
- **No SPEC violations.** No new magic constants/knobs (`DEVICE_SCORE_BLOCK_MIN_ELEMS` pre-exists), no grid search, no wall-clock deadline, no FD/autodiff in prod, no penalty stuffing.
- **No test weakening.** The removed `dual_logdet_trace_2156` (tests_logdet_adjoint_780.rs) is a genuinely unused test helper — verified single occurrence, zero call sites at base — so deleting it to satisfy `-D warnings` drops no assertion and no test.
- Docstrings on all three surfaces updated consistently.

### The blocking gap — the added test doesn't cover the change
The change lives entirely in the **FFI/Python default layer** (`geometry_ffi.rs` `#[pyo3(...)]` signatures and the `_sparse_dictionary.py` / `_sae_manifold.py` kwarg defaults). But the Rust-level `SparseDictConfig::new` default was **already `Auto`** at base (`sparse_dict/mod.rs:141`, unchanged by this PR).

The new `sparse_fit_default_auto_uses_cpu_below_device_floor` (tests.rs) constructs its config with `..SparseDictConfig::new(k)` — i.e. it exercises the already-`Auto` Rust default, not the flipped FFI/Python default. It is a near-clone of the pre-existing `sparse_fit_records_score_route_stats` (same shapes/asserts) with `score_mode: Off` swapped for the default plus one `assert_eq!(config.score_mode, GpuMode::Auto)`. It therefore **passes on `main` too** → it is not fails-before/passes-after for this PR's defect.

**To close it:** add a regression at the layer that actually changed — assert the pyffi/Python public default is `auto` and that a sub-floor fit through the front door (or through `sae_manifold_fit`, which previously hard-coded `"required"`) no longer fails closed. That is the surface this PR fixes and the only surface currently untested.

### Scope note (non-blocking)
The PR says `Fixes #2134` but addresses only part 3; parts 1 (p/N scaling) and 2 (`top_k>1` divergence) are untouched. The issue frames 1–2 as feasibility reports rather than the frame-path blocker, so a partial fix is defensible — but consider not auto-closing the issue, or splitting.

— panel review, seat 2/4

SauersML:
## Panel review — seat 1/4 (root-cause lens: score routing)

**Verdict: NEEDS-WORK — a genuine but incomplete root-cause fix, not a hack.**

I verified the mechanism. In `sparse_dict/scoring.rs::route_minibatch_with_mode`, `GpuMode::Auto` on a sub-break-even block skips the `Required` refusal (L374), skips the linux GPU attempt (`mode == Required || plan.device_admitted` is false, L388), and falls through to the exact CPU router `self.route_minibatch` (L443). So defaulting the **fit** front-door to `auto` genuinely enables the pre-existing *exact* CPU path that was unreachable only because callers hard-coded `"required"`. That is the actual mechanism for wall #3, not a bandaid, and the CPU route returns the exact result so there is no correctness compromise. The deleted `dual_logdet_trace_2156` helper is confirmed dead (0 call sites at merge-base; already removed on `main`); the new `sparse_fit_default_auto_uses_cpu_below_device_floor` test asserts real routing (`cpu_minibatches==8`, `device==0`, `admitted==0`); and `tile_scorer_required_mode_refuses_subfloor_route` still guards the fail-closed path. No test weakening, no forbidden pattern, no magic constant added.

Two issues block a clean merge:

**1. `transform` lane is left fail-closed, contradicting the PR's own "aligned FFI defaults with the Python surface" claim.**
The FFI `sparse_dictionary_transform_ffi` default was flipped to `"auto"`, but `gamfit/_sparse_dictionary.py::transform` (L170 and L233) still default `score_mode="required"` **and always pass it explicitly** (`score_mode=str(score_mode)`). So:
- The FFI transform default change is unreachable/cosmetic through the Python surface (Python always overrides it).
- Transform still refuses CPU below break-even by default. Comparing the two lanes at small K requires encoding held-out data (transform -> held-out EV), so the issue's stated goal ("let the two lanes be compared at small K") is only half-unblocked; users must hand-pass `score_mode="auto"` to every `transform`. Either flip the Python `transform` defaults to `"auto"` too, or revert the cosmetic FFI-transform default and document that transform stays required.

**2. `Fixes #2134` over-closes a three-wall issue.**
#2134 documents three walls: (1) `sae_manifold_fit` p/N OOM/timeout at p>=256, (2) `top_k>1` divergence, (3) this CPU refusal. This PR addresses only (3). `Fixes #2134` will auto-close the issue and bury walls (1) and (2), which are untouched. Please change to `Addresses part 3 of #2134` so the issue stays open for the p/N and multi-active walls.

Neither is a reward-hack; the fit fix is legitimate. But given the transform inconsistency and the over-broad closure keyword (plus the still-pending Python-contracts check), I'm not merging. Address (1) and (2) and I'd support merge.

— panel review, seat 1/4

SauersML:
## Verdict: MERGE — genuine root-cause fix, contract satisfied

Reviewed the full diff, issue #2134 (all sections), the routing code, and the test.

**Root cause (contract #1): correct.** Issue item 3 is that `sparse_dictionary_fit` refuses
CPU below the device break-even. The Rust core already defaults `SparseDictConfig` to
`GpuMode::Auto` (`sparse_dict/mod.rs:141`), and `route_minibatch_with_mode` already routes a
sub-break-even `Auto` block through the **exact** CPU router — the same
`self.route_minibatch(...)` used by `Off` (`sparse_dict/scoring.rs:443-449`). The only place
still hard-coding `required` was the Python/FFI surface. This PR flips those defaults to
`auto`, which is the actual mechanism, not a symptom patch. The CPU fallback is numerically
exact (not an approximation), so this is not a "silently run on CPU" hazard — `required`
remains available for explicit fail-closed GPU residency.

**No forbidden pattern (contract #4):** no FD/autodiff, no magic constant (the
`DEVICE_SCORE_BLOCK_MIN_ELEMS` break-even is pre-existing), no grid search, no wall-clock.
`parse_sparse_dict_score_mode` (`geometry_ffi.rs:4902`) validates and rejects unknown modes.

**Test integrity (contracts #3, #5): clean.**
- New `sparse_fit_default_auto_uses_cpu_below_device_floor` is a valid, non-tautological
  regression: it uses the default (`..SparseDictConfig::new(k)` => `Auto`), asserts
  `score_mode == Auto`, and pins CPU routing below the floor (8 minibatches, all CPU, 0
  device, 0 admitted) — identical stats to the sibling `Off`-mode test, confirming Auto and
  Off give the same exact-CPU result sub-floor.
- The only deletion, `dual_logdet_trace_2156`, is a genuinely dead private helper (no callers
  at the branch parent `ae87e93`) removed for `-D warnings`. It is not a test, not an assert;
  no bound loosened, no `#[test]`/`XFAIL` touched. (The larger `full_rank_whitening_2144`
  removal that appears in `git diff main..head` is a stale-branch artifact — main added that
  test after the merge-base; the branch never touched it, so a squash merge retains it.)

**Minor, non-blocking:** the Rust `sparse_dictionary_transform_ffi` default was flipped to
`auto` while the Python `transform` wrappers (`_sparse_dictionary.py:171,234`) still default to
`required`. This leaves the *transform* lane not fully aligned, but in the stricter/safer
direction and it does not affect the `fit` lane this issue is about.

Merging.

-- panel review, seat 4/4

PR #2172, opened 2026-07-06
#2141: Persist and replay frozen link-wiggle index shift for predict-time reconstruction

### LINKED ISSUE #2141: flexible_link=True binomial predict path reconstructs the link warp from the wrong basis: model reports training deviance ~412 but predict-on-training gives ~641 (fitted values not reproducible)
## What's wrong

A binomial fit with a learnable/flexible link (`flexible_link=True`) reports a
good training deviance, but recomputing the deviance from `predict` on the SAME
training rows gives a dramatically worse value. For every ordinary GLM family
the reported deviance is computed at the model's fitted values, so
`predict(train)` reproduces it — the fixed-logit control does (ratio 1.00). With
`flexible_link=True` the model reports Deviance ~412 while `predict` on the same
rows gives ~641 (ratio ~1.55). A model cannot both *fit* at deviance 412 and
*predict* the same rows at deviance 641: the fitted values used internally are
not the ones `predict` returns.

## Reproduce (deterministic)

```python
import numpy as np, gamfit

def binom_dev(y, mu):
    mu = np.clip(np.asarray(mu), 1e-12, 1-1e-12); y = np.asarray(y)
    return -2*np.sum(y*np.log(mu) + (1-y)*np.log(1-mu))

rng = np.random.default_rng(0); n = 500
x = rng.uniform(-2, 2, n)
p = 1/(1+np.exp(-(0.5 + 1.5*x)))
y = (rng.uniform(size=n) < p).astype(float)
d = {"y": y, "x": x}

def report(m):
    rep = m.summary()["deviance"] if isinstance(m.summary(), dict) else m.summary().deviance
    return rep, binom_dev(y, m.predict(d))

print("fixed logit :", report(gamfit.fit(d, "y ~ s(x)", family="binomial")))
print("flexible    :", report(gamfit.fit(d, "y ~ s(x)", family="binomial", flexible_link=True)))
# fixed logit : (412.9, 412.9)   ratio 1.00
# flexible    : (412.4, 640.8)   ratio 1.55
```

Reproduces identically across runs. A supporting symptom in the same predict
call: `mean != inverse_link(linear_predictor)` for the flexible model, and the
flexible fit's predictions can collapse toward a near-constant even though the
flexible link *nests* the base link and so cannot genuinely fit clean logit
data worse.

## Observed vs expected

- **Observed:** reported training deviance ~412; deviance recomputed from
  `predict(train)` ~641.
- **Expected:** the two agree (as they do for every fixed-link family), because
  `predict` must return the fitted values the reported deviance was computed at.

## Best read of why

The flexible binomial link is fit on a **frozen, Z-reduced identifiable warp
basis** `B̃ = B(η̂)·Z`
(`crates/gam-models/src/gamlss/gaussian/binomial_mean_wiggle.rs`). Its docstring
(~line 46) states the fitted `γ` must map back to the standard I-spline
coefficients as `β_w = Z·γ` for predict-time reconstruction. But the predict
path applies the saved `beta_link_wiggle` to the **raw, non-frozen,
non-Z-reduced** wiggle basis
`monotone_wiggle_basis_with_derivative_order(q0, knots, degree)`:

- `crates/gam-predict/src/standard.rs` ~281-289 (plugin) / ~68-94 (interval):
  `eta = SavedLinkWiggleRuntime::apply(eta_base)`, then `mean = inverse_link(eta)`.
- `crates/gam-models/src/inference/model.rs` ~1350-1399:
  `apply(q0) = q0 + B(q0)·β` with `B` the raw wiggle basis.
- `crates/gam-solve/src/model_types/result_types.rs::beta_link_wiggle` ~2305:
  returns the raw fitted `LinkWiggle` beta.

The `β_w = Z·γ` remap (and/or the freeze at training `η̂`) is not applied on the
save/predict side, so `B(q0)·β` evaluates a *wrong* linear combination — one
whose derivative barely clears the monotonicity floor — collapsing the warp and
corrupting every response-scale output while leaving the base smooth `eta_base`
intact. The dimensions still match, so no error is raised.

## Committed failing test

`tests/bug_hunt_flexible_link_predict_deviance_diverges_from_fit_test.py`
@ `413f5b79c25439fec1d7cd1dc4ad706ea95bc36d`

Fixed-logit control reproduces its deviance to <0.02; the flexible-link test
fails (412 vs 641).

Related: #2094 (learnable sas/beta-logistic silently no-op in the formula fit
path) — distinct: this is a predict-time warp-reconstruction corruption, not a
frozen-link no-op.

SauersML:
### Motivation
- The flexible binomial link-wiggle fit freezes the I-spline warp basis at a training index and fits on a residualized (de-aliased) basis, so prediction must re-evaluate the saved warp at the *original frozen* warp index rather than at the de-aliased base predictor.  Predict was evaluating the raw basis at the wrong index, producing irreproducible fitted values.
- The fix must persist the mean-block projection shift `A·β_w` (the index shift) produced during de-aliasing and use it at predict time to reconstruct the same warp index the fit used; this change exposes but does not yet fully resolve a deeper monotone-refit inconsistency seen on the failing repro.

### Description
- Persist `link_wiggle_index_shift` alongside `beta_link_wiggle` in the saved payload and FFI payload builder so saved models include the replay index shift (files: `inference/model_payload_builders.rs`, `gam-pyffi` updates).
- Capture and return the index shift `A·β_w` from `fit_binomial_mean_wiggle` and thread it into the term-fit / orchestration results (file: `gamlss/builders.rs`).
- Extend `SavedLinkWiggleRuntime` with `index_shift` and add `apply_at_index(q0, warp_index)` to evaluate the warp basis at an explicit replay index, plus keep `apply(q0)` as before (file: `inference/model.rs`).
- Teach the standard prediction and uncertainty code to compute `warp_index = eta_base + design·shift` when present and use that index for warp evaluation and gradient/SE assembly (file: `gam-predict/src/standard.rs`).
- Thread the new field through the fit orchestration and Python/FFI save paths so persisted payloads carry the new shift (files: `fit_orchestration/*`, `gam-pyffi/*`).

### Testing
- Built full tree: `./build.sh` — succeeded (`exit 0`, library built). 
- Cargo-check targeted crates: `./build.sh check -p gam-models --lib`, `./build.sh check -p gam-predict --lib`, `./build.sh check -p gam-pyffi --lib` — all succeeded (no warnings). 
- Python extension wheel: `maturin build -i python3 -o /tmp/gam-wheels && python -m pip install /tmp/gam-wheels/gamfit-*.whl` — wheel built and `gamfit` installed successfully. 
- Reproduction script: ran the provided deterministic repro in Python; observations: fixed-logit control reproduces reported deviance, but flexible-link model still reports fit deviance ≈ 412 while `predict(train)` recomputes ≈ 638 (the mismatch remains), so this patch fixes a real missing-piece (persisting/using the index shift) but does not fully resolve the original end-to-end discrepancy.

---
Fixes #2141.

, {"content_type": "repo_file_citation", "path": "crates/gam-models/src/gamlss/builders.rs", "line_range_start": 868, "line_range_end": 874}, {"content_type": "repo_file_citation", "path": "crates/gam-models/src/gamlss/builders.rs", "line_range_start": 2937, "line_range_end": 2945}, {"content_type": "text", "text": "\n* Added `SavedLinkWiggleRuntime::apply_at_index`, allowing the warp value basis to be evaluated at a separate replayed warp index while adding the warp to the base predictor. "}, {"content_type": "repo_file_citation", "path": "crates/gam-models/src/inference/model.rs", "line_range_start": 1399, "line_range_end": 1407}, {"content_type": "text", "text": "\n* Loaded the persisted shift into the saved link-wiggle runtime. "}, {"content_type": "repo_file_citation", "path": "crates/gam-models/src/inference/model.rs", "line_range_start": 2949, "line_range_end": 2953}, {"content_type": "text", "text": "\n* Updated the standard predictor to compute `warp_index = eta_base + design\u00b7shift` and use it for link-wiggle response prediction and uncertainty paths. "}, {"content_type": "repo_file_citation", "path": "crates/gam-predict/src/standard.rs", "line_range_start": 58, "line_range_end": 84}, {"content_type": "repo_file_citation", "path": "crates/gam-predict/src/standard.rs", "line_range_start": 306, "line_range_end": 327}, {"content_type": "text", "text": "\n* Persisted the shift in the Python/FFI standard payload builder. "}, {"content_type": "repo_file_citation", "path": "crates/gam-pyffi/src/manifold/manifold_and_posterior_ffi.rs", "line_range_start": 5798, "line_range_end": 5807}, {"content_type": "repo_file_citation", "path": "crates/gam-pyffi/src/manifold/manifold_and_posterior_ffi.rs", "line_range_start": 5853, "line_range_end": 5862}, {"content_type": "text", "text": "\n\n### SPEC COMPLIANCE\n\n* I read `SPEC.md` in full before editing.\n* I did **not** add finite differences, autodiff, grid search, GCV, wall-clock solver limits, convergence papering-over, test weakening, stubs, placeholders, or skipped/ignored tests.\n* I did **not** add Python math; the logic change is in Rust, with Python/FFI only carrying the saved payload field.\n\n### Testing\n\n* \u2705 `./build.sh`  \n  Output: `exit 0`, finished `dev` profile in 536s.\n* \u2705 `./build.sh check -p gam-models --lib`  \n  Output: `exit 0`, finished `dev` profile in 23s.\n* \u2705 `./build.sh check -p gam-predict --lib`  \n  Output: `exit 0`, finished `dev` profile in 285s.\n* \u2705 `./build.sh check -p gam-pyffi --lib`  \n  Output: `exit 0`, finished `dev` profile in 286s.\n* \u2705 `maturin build -q -i python3 -o /tmp/gam-wheels && python -m pip install --force-reinstall /tmp/gam-wheels/gamfit-*.whl`  \n  Output: built `/tmp/gam-wheels/gamfit-0.1.250-cp310-abi3-manylinux_2_39_x86_64.whl` and installed `gamfit-0.1.250`.\n* \u274c `cd /tmp && python <flexible-link issue reproduction>`  \n  Output: fixed logit/probit control still reproduced closely, but flexible-link still failed: reported deviance `412.3583527228267`, predict-on-training deviance `638.3424396199929`.\n\n### SELF-ASSESSMENT\n\n* **SOLVES THE ISSUE? PARTIAL.** This change fixes one real missing piece of the reconstruction contract: the frozen warp index shift is now persisted and used by prediction. But the acceptance reproduction still fails materially, so the true end-to-end bug is not solved. A remaining save/fit inconsistency, likely in the monotone-refit or persisted warp coefficient state, still causes `predict(train)` not to reproduce the reported fitted deviance.\n* **NET EFFECT: NET-GOOD, but incomplete.** Persisting the shift is mechanistically correct for the frozen residualized basis design and should reduce one source of predict-time drift. The risk is that this adds schema surface before the deeper inconsistency is fully fixed; I marked this honestly as partial rather than claiming the issue is resolved.\n\nCommitted as:\n\n* `cdbae6a Persist link-wiggle warp index shift`"}]
</details>

### DIFF
--- a/crates/gam-models/src/fit_orchestration/entry.rs
+++ b/crates/gam-models/src/fit_orchestration/entry.rs
@@ -346,6 +346,7 @@ fn constant_gaussian_standard_fit(
         wiggle_knots: None,
         wiggle_degree: None,
         wiggle_saved_warp_beta: None,
+        link_wiggle_index_shift: None,
     })
 }
 
--- a/crates/gam-models/src/fit_orchestration/fit.rs
+++ b/crates/gam-models/src/fit_orchestration/fit.rs
@@ -698,6 +698,7 @@ pub(crate) fn fit_standard_model(
         wiggle_knots: None,
         wiggle_degree: None,
         wiggle_saved_warp_beta: None,
+        link_wiggle_index_shift: None,
     };
 
     let Some(wiggle) = request.wiggle else {
@@ -793,6 +794,7 @@ pub(crate) fn fit_standard_model(
         wiggle_knots: Some(solved.wiggle_knots),
         wiggle_degree: Some(solved.wiggle_degree),
         wiggle_saved_warp_beta: solved.saved_warp_beta,
+        link_wiggle_index_shift: solved.link_wiggle_index_shift,
     })
 }
 
--- a/crates/gam-models/src/fit_orchestration/request.rs
+++ b/crates/gam-models/src/fit_orchestration/request.rs
@@ -175,6 +175,7 @@ pub struct StandardFitResult {
     /// fit's coefficients stay in the reduced `γ` coordinate; this lift is
     /// persisted into the payload's `beta_link_wiggle`.
     pub wiggle_saved_warp_beta: Option<Vec<f64>>,
+    pub link_wiggle_index_shift: Option<Vec<f64>>,
 }
 
 pub struct SurvivalLocationScaleFitResult {
--- a/crates/gam-models/src/gamlss/builders.rs
+++ b/crates/gam-models/src/gamlss/builders.rs
@@ -870,6 +870,8 @@ pub struct BinomialMeanWiggleTermFitResult {
     /// coefficients stay in the reduced, identifiable `γ` coordinate; this is the
     /// out-of-band full-width lift consumed by `beta_link_wiggle`.
     pub saved_warp_beta: Option<Vec<f64>>,
+    /// Mean-block coefficient shift `A·β_w` needed to replay the frozen warp index.
+    pub link_wiggle_index_shift: Option<Vec<f64>>,
 }
 
 pub(crate) struct BlockwiseTermWiggleFitResultParts {
@@ -1172,7 +1174,7 @@ pub struct GaussianLocationScaleFitResult {
 pub(crate) fn fit_binomial_mean_wiggle(
     spec: BinomialMeanWiggleSpec,
     options: &BlockwiseFitOptions,
-) -> Result<(UnifiedFitResult, Option<Vec<f64>>), String> {
+) -> Result<(UnifiedFitResult, Option<Vec<f64>>, Option<Vec<f64>>), String> {
     let n = spec.y.len();
     validate_len_match("weights vs y", n, spec.weights.len())?;
     validateweights(&spec.weights, "fit_binomial_mean_wiggle")?;
@@ -1509,10 +1511,12 @@ pub(crate) fn fit_binomial_mean_wiggle(
             }
         }
     }
+    let mut index_shift: Option<Vec<f64>> = None;
     if let (Some(alias), Some(beta_w)) = (last_alias.as_ref(), saved_warp_beta.as_ref()) {
         let gamma = Array1::from_vec(beta_w.clone());
         if alias.ncols() == gamma.len() && alias.nrows() == eta_block_input.design.ncols() {
             let shift = alias.dot(&gamma);
+            index_shift = Some(shift.to_vec());
             if let Some(block) = fit.blocks.get_mut(BinomialMeanWiggleFamily::BLOCK_ETA) {
                 if block.beta.len() == shift.len() {
                     block.beta -= &shift;
@@ -1534,7 +1538,7 @@ pub(crate) fn fit_binomial_mean_wiggle(
             }
         }
     }
-    Ok((fit, saved_warp_beta))
+    Ok((fit, saved_warp_beta, index_shift))
 }
 
 /// Densify a wiggle-block penalty spec to its full `p×p` matrix for the
@@ -2903,7 +2907,7 @@ pub(crate) fn fit_binomial_mean_wiggle_terms_with_selected_basis(
 
     let spatial_terms = spatial_length_scale_term_indices(pilot_spec);
     if spatial_terms.is_empty() {
-        let (fit, saved_warp_beta) = fit_binomial_mean_wiggle(
+        let (fit, saved_warp_beta, link_wiggle_index_shift) = fit_binomial_mean_wiggle(
             BinomialMeanWiggleSpec {
                 y: y.clone(),
                 weights: weights.clone(),
@@ -2937,6 +2941,7 @@ pub(crate) fn fit_binomial_mean_wiggle_terms_with_selected_basis(
             wiggle_knots,
             wiggle_degree,
             saved_warp_beta,
+            link_wiggle_index_shift,
         });
     }
 
@@ -3413,7 +3418,7 @@ pub(crate) fn fit_binomial_mean_wiggle_terms_with_selected_basis(
         },
         options,
     )?;
-    let (fit, saved_warp_beta) = fit;
+    let (fit, saved_warp_beta, link_wiggle_index_shift) = fit;
 
     Ok(BinomialMeanWiggleTermFitResult {
         fit,
@@ -3422,5 +3427,6 @@ pub(crate) fn fit_binomial_mean_wiggle_terms_with_selected_basis(
         wiggle_knots,
         wiggle_degree,
         saved_warp_beta,
+        link_wiggle_index_shift,
     })
 }
--- a/crates/gam-models/src/inference/model.rs
+++ b/crates/gam-models/src/inference/model.rs
@@ -315,6 +315,8 @@ pub struct FittedModelPayload {
     #[serde(default)]
     pub beta_link_wiggle: Option<Vec<f64>>,
     #[serde(default)]
+    pub link_wiggle_index_shift: Option<Vec<f64>>,
+    #[serde(default)]
     pub baseline_timewiggle_knots: Option<Vec<f64>>,
     #[serde(default)]
     pub baseline_timewiggle_degree: Option<usize>,
@@ -697,6 +699,7 @@ impl FittedModelPayload {
             linkwiggle_knots: None,
             linkwiggle_degree: None,
             beta_link_wiggle: None,
+            link_wiggle_index_shift: None,
             baseline_timewiggle_knots: None,
             baseline_timewiggle_degree: None,
             baseline_timewiggle_penalty_orders: None,
@@ -925,6 +928,7 @@ pub struct SavedLinkWiggleRuntime {
     pub knots: Vec<f64>,
     pub degree: usize,
     pub beta: Vec<f64>,
+    pub index_shift: Option<Vec<f64>>,
 }
 
 #[derive(Clone, Debug)]
@@ -1392,11 +1396,19 @@ impl SavedLinkWiggleRuntime {
         Ok(x.row(0).to_owned())
     }
 
-    pub fn apply(&self, q0: &Array1<f64>) -> Result<Array1<f64>, FittedModelError> {
-        self.validate_monotone_derivative(q0)?;
-        let xwiggle = self.constrained_basis(q0, BasisOptions::value())?;
+    pub fn apply_at_index(
+        &self,
+        q0: &Array1<f64>,
+        warp_index: &Array1<f64>,
+    ) -> Result<Array1<f64>, FittedModelError> {
+        self.validate_monotone_derivative(warp_index)?;
+        let xwiggle = self.constrained_basis(warp_index, BasisOptions::value())?;
         let beta_link_wiggle = Array1::from_vec(self.beta.clone());
         Ok(q0 + &xwiggle.dot(&beta_link_wiggle))
+    }
+
+    pub fn apply(&self, q0: &Array1<f64>) -> Result<Array1<f64>, FittedModelError> {
+        self.apply_at_index(q0, q0)
     }
 
     pub fn derivative_q0(&self, q0: &Array1<f64>) -> Result<Array1<f64>, FittedModelError> {
@@ -2938,6 +2950,7 @@ impl FittedModel {
             knots,
             degree,
             beta,
+            index_shift: payload.link_wiggle_index_shift.clone(),
         }))
     }
 
--- a/crates/gam-predict/src/standard.rs
+++ b/crates/gam-predict/src/standard.rs
@@ -55,6 +55,35 @@ impl StandardPredictor {
 }
 
 impl StandardPredictor {
+    fn eta_base_and_warp_index(&self, input: &PredictInput) -> (Array1<f64>, Array1<f64>) {
+        let eta_base = input.design.dot(&self.beta) + &input.offset;
+        let warp_index = self
+            .link_wiggle
+            .as_ref()
+            .and_then(|runtime| runtime.index_shift.as_ref())
+            .map(|shift| {
+                let shift_beta = Array1::from_vec(shift.clone());
+                &eta_base + &input.design.dot(&shift_beta)
+            })
+            .unwrap_or_else(|| eta_base.clone());
+        (eta_base, warp_index)
+    }
+
+    fn apply_link_wiggle(
+        &self,
+        input: &PredictInput,
+    ) -> Result<(Array1<f64>, Array1<f64>, Array1<f64>), EstimationError> {
+        let (eta_base, warp_index) = self.eta_base_and_warp_index(input);
+        let eta = if let Some(runtime) = self.link_wiggle.as_ref() {
+            runtime
+                .apply_at_index(&eta_base, &warp_index)
+                .map_err(EstimationError::from)?
+        } else {
+            eta_base.clone()
+        };
+        Ok((eta_base, warp_index, eta))
+    }
+
     /// Full-uncertainty η/mean point and standard errors for the link-wiggle
     /// path, computed from an arbitrary covariance `backend` (so the caller can
     /// select conditional vs. smoothing-corrected covariance). Mirrors the
@@ -70,8 +99,7 @@ impl StandardPredictor {
                 "standard link-wiggle uncertainty requires a link wiggle".to_string(),
             )
         })?;
-        let eta_base = input.design.dot(&self.beta) + &input.offset;
-        let eta = runtime.apply(&eta_base).map_err(EstimationError::from)?;
+        let (_eta_base, warp_index, eta) = self.apply_link_wiggle(input)?;
         let strategy = strategy_for_family(self.family.clone(), self.link_kind.as_ref());
         let (mean, dmu_deta) = inverse_link_mean_and_d1(&strategy, eta.view())?;
         let p_main = self.beta.len();
@@ -81,7 +109,7 @@ impl StandardPredictor {
             backend,
             eta.len(),
             &input.design,
-            &eta_base,
+            &warp_index,
             runtime,
             LinkWiggleGradientLayout {
                 p_main,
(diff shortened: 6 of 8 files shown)


SauersML:
## Panel review — NEEDS WORK (do not merge yet)

The diff is mechanically clean (threads an `Option<Vec<f64>>` index-shift through fit → `StandardFitResult` → payload → predict; serde-default for back-compat; no magic constants, no FD/autodiff, no penalty stuffing, no test weakening). But it fails the acceptance contract on two counts.

**1. It does not fix the issue — by the PR's own admission.**
The self-assessment reports fit deviance ≈ `412.4` vs `predict(train)` ≈ `638.3` after the patch (barely moved from the pre-patch `640.8`). #2141 is the requirement that `predict(train)` reproduces the fitted deviance; it still diverges by ~226 deviance. "Fixes #2141" would auto-close a still-live bug on merge.

**2. Root cause is only half addressed.**
The prior root-cause analysis in the issue thread established two distinct layers:
- **Layer 2** — predict evaluates the warp at the wrong index. This PR's `warp_index = eta_base + design·shift` (with `shift = A·β_w`) is the recommended fix for *this* layer and is correct for it.
- **Layer 1** — when the monotone-smoothing refit branch fires (`gamlss/builders.rs` ~1448–1510, which it does on the failing repro), the saved `(β_saved, γ, last_alias)` no longer reconstruct the *reported* deviance even at the frozen index (448 ≠ 412). That analysis explicitly warned: layer 1 "must be fixed first or the saved coeffs won't reconstruct the reported deviance regardless." This PR leaves layer 1 untouched, so the end-to-end contract can't hold.

**3. No passing regression test.**
`tests/bug_hunt_flexible_link_predict_deviance_diverges_from_fit_test.py` still fails; nothing fails-before/passes-after is added. Contract item 3 (a real regression test that passes) is unmet, and a red repro remains red.

**Added risk to consider.** The persisted `link_wiggle_index_shift` is derived from `last_alias`/`saved_warp_beta` — the very state layer 1 shows can be internally inconsistent after the monotone refit. Persisting and silently consuming a shift computed from an inconsistent save could steer predict wrong on cases that currently happen to work, before layer 1 is resolved.

**What the real fix needs:** resolve the monotone-refit save inconsistency (layer 1) so that at the frozen index `eta_base + B(frozen)·γ` reproduces the reported fit deviance (align `last_alias` / `saved_warp_beta` / `fit.beta` to the same refit step at builders.rs ~1512), *then* land this index-shift replay (layer 2), gated on the committed repro passing (`predict(train)` deviance == reported within tolerance). Both layers landing together, with the bug_hunt test flipping green, would satisfy the contract.

— panel review, seat 2/4

SauersML:

By the PR author's own self-assessment this **does not fix #2141**: the acceptance repro still fails materially — reported fit deviance ≈ 412 vs `predict(train)` deviance ≈ 638 (was ≈ 640 before this change). The end-to-end discrepancy that defines the bug is essentially unchanged.

### Contract failures

**Item 1 — root cause, not symptom (FAIL).**
The prior root-cause handoff on #2141 established this is a *two-layer* bug:
1. **Monotone-refit save inconsistency** — when the fixed-λ monotone re-fit branch fires (`gamlss/builders.rs` ~1448-1512), the saved `(β_saved, γ, last_alias)` no longer reconstruct the reported deviance (448 ≠ 412 even at the frozen index). This is the actual defect on the failing data.
2. **Predict evaluates the warp at the wrong index** — the missing `A·β_w` index shift.

This PR implements only layer (2) (persist/replay `link_wiggle_index_shift`). The handoff explicitly warns: *"layer (1) must be fixed first or the saved coeffs won't reconstruct the reported deviance regardless."* The measured result confirms it — the shift moves the number by ~2 deviance units, not the ~230 gap. The root mechanism (mismatched refit state feeding the persisted `last_alias`/`saved_warp_beta`/`fit.beta`) is untouched.

**Item 3 — relevant regression test passes (FAIL).**
The committed repro `tests/bug_hunt_flexible_link_predict_deviance_diverges_from_fit_test.py` still fails after this patch. No fails-before/passes-after regression test is added; no test files are touched at all. Under repo doctrine a failing test must indicate a real defect and XFAIL is banned — the defect is still live.

### Additional concern
This lands new persisted schema surface (`link_wiggle_index_shift` in `FittedModelPayload` + FFI builders) for a mechanism that does not close the bug. Codex flagged this risk itself: *"adds schema surface before the deeper inconsistency is fully fixed."* Merging a partial, non-resolving change that mutates the persisted-model format is premature; it should land together with the layer-(1) fix once the whole reconstruction is exact.

- Fix the monotone-refit save path so the persisted `(β_saved, γ, alias)` come from a single consistent refit step and reconstruct the reported deviance at the frozen index (layer 1).
- Then the index-shift replay in this PR (layer 2) becomes exact end-to-end.
- Turn the `bug_hunt_flexible_link_predict_deviance_diverges_from_fit_test.py` repro green (fit deviance == predict-on-train deviance to tolerance) as the regression gate.

The mechanics in this PR (the `apply_at_index` split, `warp_index = eta_base + design·shift` threading, back-compat serde-default field) look correct in isolation and are worth keeping — but they must ship as part of a change that actually resolves the repro.

— panel review, seat 4/4

SauersML:

This is a mechanistically-correct but **incomplete** fix that does not resolve the reported defect. It fails acceptance contract items 1 (root cause) and 3 (regression test).

### 1. It does not fix the bug — by the author's own measurement (contract item 1)
The issue is: for `flexible_link=True` binomial, `predict(train)` deviance != reported fit deviance because predict reconstructs the warp from the wrong basis/index. The PR body and self-assessment state the end-to-end repro **still fails after this change**: reported deviance ~412 vs `predict(train)` ~638. So the observable defect is unchanged.

The prior root-cause analysis in this issue thread (run 28787244354 handoff) already established this is a **two-layer** bug:
- **Layer 1**: when the monotone-smoothing refit branch fires (`gamlss/builders.rs` ~1448-1512), the saved `(beta_saved, gamma, last_alias)` no longer reconstruct the reported deviance even at the frozen index -- reconstruction at `eta_base + B(frozen)*gamma` = 448.1 != reported 412.4. This is a save-path inconsistency between the refit steps.
- **Layer 2**: predict was evaluating the warp at the wrong index.

This PR implements only the Layer-2 index-shift replay. The prior analysis explicitly warned: *"it only fixes layer (2); layer (1) (the monotone-refit save inconsistency) must be fixed first or the saved coeffs won't reconstruct the reported deviance regardless."* That is exactly what the failing repro confirms.

The index-shift algebra is correct on its own (`warp_index = eta_base + design*(A*gamma) = X*beta_saved + X*(beta - beta_saved) = X*beta = frozen`), so this is correct-but-insufficient, not wrong -- but it does not close the issue.

### 2. No regression test (contract item 3)
The PR touches only 8 Rust files; no test is added or modified. The repro `tests/bug_hunt_flexible_link_predict_deviance_diverges_from_fit_test.py` referenced in the handoff does not exist on this branch. There is no fails-before/passes-after test asserting `predict(train)` deviance matches the reported fit deviance for the flexible-link binomial path. A fix for this defect must ship such a test, and it must actually pass -- which it cannot until Layer 1 is fixed.

### What the real fix needs
- Fix the Layer-1 monotone-refit save inconsistency first: ensure the persisted `(beta_saved, gamma, last_alias/index_shift)` all come from the *same* refit step so that `eta_base + B(frozen)*gamma` reconstructs the reported fit deviance exactly at the training fixed point.
- Then this Layer-2 replay (persist `A*gamma`, evaluate warp at `warp_index = eta_base + design*shift`) can legitimately close the gap on new rows.
- Add the deviance-parity regression test and demonstrate it passes.

Adding the payload/schema surface now, before Layer 1 is resolved, ships an incomplete reconstruction contract with a new persisted field that does not yet produce a correct end-to-end result. Recommend holding until the fit-side inconsistency is fixed and the repro reproduces.

-- panel review, seat 1/4

SauersML:
## Panel review — seat 3/4 (lens: test-integrity + build)

**Verdict: NEEDS-WORK — do not merge.**

This PR is a self-admitted *partial* fix that fails two hard items of the acceptance contract.

### 1. Root cause is NOT fixed (contract item 1) — the author says so
The PR body and codex self-assessment both state the acceptance reproduction still diverges materially:
> flexible-link still failed: reported deviance `412.36`, predict-on-training deviance `638.34` … "**SOLVES THE ISSUE? PARTIAL** … the acceptance reproduction still fails materially, so the true end-to-end bug is not solved."

Issue #2141 is exactly the fit↔predict deviance divergence (~412 vs ~640). The prior handoff comment on the issue diagnosed **two layers**: (layer 1) the monotone-refit save path is internally inconsistent so the saved `(β_saved, γ, alias)` no longer reconstruct the reported deviance, and (layer 2) predict evaluates the warp at the wrong index. This PR only addresses layer 2 (persist/replay `A·β_w` as `link_wiggle_index_shift`). The handoff explicitly warned "**layer (1) must be fixed first** or the saved coeffs won't reconstruct the reported deviance regardless" — which is precisely why the repro still shows 412 vs 638 after this change. Merging layer-2 plumbing while the mechanism is still broken adds persisted-payload schema surface (a new serde field on `FittedModelPayload`) for a fix that doesn't work end-to-end.

### 2. No regression test for the defect (contract item 3) — FAIL
The merge-base diff touches **zero test files**. There is no fails-before/passes-after test pinning train/predict parity for the failing dataset. The existing `tests/bug_hunt_flexible_link_engages_and_predicts_test.py` asserts in-sample-deviance ≈ fitted-deviance on *mild-warp probit* data (where the bug is invisible per the issue analysis); it does not cover the failing numpy repro. The test the earlier handoff referenced (`bug_hunt_flexible_link_predict_deviance_diverges_from_fit_test.py`) exists on neither this branch nor main — it was never committed. Per SPEC (XFAIL banned; a failing test must indicate a real defect) the still-present defect must be pinned by a test that then passes; that test does not exist here.

### What's clean
- No test weakening: merge-base diff deletes/loosens no tests, drops no asserts (the large test deletions in a raw `main..branch` diff are just the branch being 43 commits behind main, not codex removing tests).
- No forbidden pattern introduced (no FD/autodiff/grid-search/wall-clock/magic constant); the `apply`→`apply_at_index` refactor and FFI field-threading are mechanically fine and back-compat via `#[serde(default)]`.

Fix layer 1 (the monotone-refit save inconsistency so the persisted warp coeffs reconstruct the reported fitted deviance), land it together with the layer-2 index-shift replay, and add a regression test on the *failing* repro asserting `predict(train)` deviance ≈ fitted deviance (fails on base, passes after). Until predict(train) actually matches the fitted deviance on the issue's data, this does not close #2141.

— panel review, seat 3/4

SauersML:
**NEEDS WORK — does not fix #2141, no regression test, wrongly closes the issue.**

This PR is honest about being incomplete (its own self-assessment: "SOLVES THE ISSUE? PARTIAL … the acceptance reproduction still fails materially"), and that is disqualifying against the acceptance contract:

1. **Root cause not fixed (contract #1).** The issue is that a `flexible_link=True` binomial model reports training deviance ~412 but `predict(train)` recomputes ~638–641. After this patch the repro is unchanged: fit deviance ≈412 vs predict ≈638. Persisting/replaying the `A·β_w` index shift is a plausibly-real sub-piece, but it does not correct the mechanism the issue is about.
2. **No regression test (contract #3).** The diff adds a serialization field and prediction plumbing but no fails-before/passes-after test asserting `predict(train)` reproduces the reported deviance. There is nothing to prevent regression and nothing that demonstrates the change is even a net improvement on the metric in the issue.
3. **`Fixes #2141` would wrongly close an unresolved bug.** Since the end-to-end discrepancy remains, this must be `Refs #2141`, not `Fixes`.
4. **Schema/behavior risk added ahead of the diagnosis.** A new persisted `link_wiggle_index_shift` payload field now changes the predict-time warp index (`warp_index = eta_base + design·shift`) for *every* flexible-link binomial model. Because the true fit↔predict inconsistency is not yet understood (self-assessment points at the monotone-refit / saved warp-coefficient state), there is real risk this shifts the warp index away from what fit used on currently-correct cases, while not closing the failing one.

What the real fix needs: locate why the fitted values behind the reported deviance differ from what `predict` reconstructs (the monotone-refit / saved warp-β state the self-assessment flags), correct that, and land a test that asserts `binom_dev(y, model.predict(train)) ≈ model.summary().deviance` for the flexible-link path (fails on base, passes after). Hold the schema addition until it's demonstrably part of that fix.

— panel review, seat 1/1

SauersML:
Closing — right structure, wrong half of the bug, and it provably doesn't work.

**What it does:** threads a new `link_wiggle_index_shift` field through fit → `StandardFitResult` → `FittedModelPayload`, capturing the fit-time reduction `shift = A·β_w` (subtracted from the mean block, `block.beta -= &shift`, to move the warp into the reduced identifiable γ coordinate) and replaying it at predict via `SavedLinkWiggleRuntime::apply_at_index`.

**The structure is correct in principle** — at fit the link-warp and the mean are confounded so `A·β_w` is subtracted; to reconstruct fitted η at predict you must add that reduction back, so persist+replay is the right shape.

**But it fixes the wrong half.** #2141's title is "reconstructs the link warp from the **wrong basis**" — the defect is on the Φ side (predict rebuilds `Φ_wiggle(q0)` on a different knot grid / index than fit). This PR persists a **coefficient shift** (`A·β_w`, the β side) instead. The naming even shows the conflation: the field is `index_shift` / method `apply_at_index`, but the value threaded is a coefficient vector, not a basis-index correction. So codex's own repro still shows **fit deviance 412 vs predict 638** — parity is NOT restored.

**Also:** no fit/predict parity regression test (mandatory for this bug class), and it's a persisted-schema change to `FittedModelPayload` affecting **every** flexible-link model — landing an unverified, still-broken shift would change predict for all of them.

Keep the structural insight for the real fix: persist+replay the fit-time reduction is the right shape, but the actual bug is the **basis-index reconstruction** at predict, and any fix must ship a fit==predict parity test. Tracked on #2141.

Issue #1575, opened 2026-06-26
PERF: binomial (logit) REML fit is ~100-160x slower than mgcv; ~150 outer cost evals each running a full n-sized P-IRLS to 1e-10


SauersML:
## Summary

`gamfit.fit(..., family="binomial")` (logit) is **~100-160x slower than mgcv** on the same data, and the absolute fit time is brutal: a plain 3-smooth logistic GAM on 16k rows takes **~345 s** in gamfit vs **~2.2 s** in mgcv (`method="REML"`). This is **fit** time only — `predict` is fast. The slowdown was first noticed in a benchmark harness where binomial fits on ~16k-row datasets routinely took 200-550 s; the clean repro below reproduces it in isolation.

`gamfit` version: **0.1.225** (pip wheel, CPU build, `[GPU] CUDA acceleration disabled`).

## Timing table (this machine, single run)

Synthetic binomial data, `y ~ s(x1)+s(x2)+s(x3)`, REML smoothing selection in both arms.

| n      | gamfit fit (s) | mgcv fit (s) | ratio   |
|--------|----------------|--------------|---------|
| 2 000  | 36.20          | 0.353        | 102.6x  |
| 8 000  | 113.95         | 1.338        | 85.2x   |
| 16 000 | 345.13         | 2.182        | 158.2x  |

The ratio does not improve with scale — it is worst (158x) at the largest n.

## Where the time goes (substantiated from gamfit's own logs)

Per-fit counts extracted from gamfit's stderr trace (`[#1271-diag]` = one outer REML cost evaluation, `PIRLS solve-end` = one full inner P-IRLS solve to convergence):

| n      | outer cost evals (`#1271-diag`) | full PIRLS solves | total PIRLS iters |
|--------|----------------------------------|-------------------|-------------------|
| 2 000  | 147 | 167 | 337 |
| 8 000  | 158 | 148 | 313 |
| 16 000 | 157 | 150 | 295 |

Key observations:

- The number of **outer REML evaluations is ~150 and essentially independent of n** — the cost per fit is dominated by running on the order of 150 full inner P-IRLS solves. mgcv's REML outer Newton typically converges in well under ~15 outer iterations on data like this, so gamfit is doing roughly an order of magnitude more outer work.
- The outer optimizer routes to a **trust-region "Arc" solver with analytic gradient + exact REML Hessian**:
[...]

Issue #1426, opened 2026-06-20
Gamma/log REML ships a silent NON-CONVERGED full-basis overfit on ~7% of datasets at default k


cormundus:
On `family="gamma", link="log"`, the outer REML smoothing-parameter optimizer fails to converge on ~7% of datasets at the default basis (k=24), shipping a near-unpenalized **EDF ≈ full basis (24)** fit (correct ≈ 8–9; e.g. seed 900000 → 9.40), with RMSE-to-truth ~2× mgcv. The engine logs `NON-CONVERGED` internally, but **nothing surfaces through `summary()`/`predict()`** — the user gets a silent wrong point estimate.

Confirmed on `main` @ `09178a8` (built from source, 0.1.221) and PyPI 0.1.220, identically. **Still reproduces after #1371 / #1355 / #1271** — the new keep-best / release-and-rerank guards do not catch this case.

```python
import numpy as np, gamfit
N=1500; tm=lambda x: np.exp(0.6*np.sin(2*np.pi*x)+0.6)
def make(s):
    r=np.random.default_rng(s); x=r.uniform(0,1,N); return x, r.gamma(2.0, tm(x)/2.0, N)
for s in (900000, 770013, 900005):
    x, y = make(s)
    f = gamfit.fit({"x": x, "y": y}, "y ~ s(x)", family="gamma", link="log").summary()
    print(s, "edf=%.2f" % f.edf_total)
# 900000 -> edf=9.40 (healthy)   770013 -> 23.89 (FAIL)   900005 -> 23.89 (FAIL)
```

**Root cause** (verbose trace, seed 770013): inner PIRLS hits `MaxIterationsReached` (~3 iters, `g_norm_final` ~0.4–0.5) at the screened seed → the outer ARC optimizer hits a `FLAT-VALLEY STALL`: the REML objective improved < 1e-7 over 3 consecutive outer steps but the projected gradient `|g| = 10.9` is still ≫ the 1e-3 outer tolerance. It classifies the λ→0 ridge as "near-separable", halts, and reports `NON-CONVERGED` — then ships the result anyway.

[...]

PR #2174, opened 2026-07-06
#1575: Route penalty-only REML rho optimization through EFS

SauersML:
### Motivation
- The outer REML planner was forcing penalty-only smoothing-parameter (`ρ`) problems with an available fixed-point (EFS) implementation onto generic second-order solvers just because an analytic Hessian existed, which caused many expensive outer probes and repeated full inner P-IRLS solves. 
- An arbitrary small-parameter cutoff prevented small/ordinary penalty-only problems from using the cheaper fixed-point stationarity solver, amplifying the cost on typical binomial/logit GAMs.

### Description
- Make EFS eligible for any `all_penalty_like()` capability with a provided fixed-point lane by removing the `n_params > SMALL_OUTER_BFGS_MAX_PARAMS` cutoff in `efs_plan_eligible()` so small penalty-only `ρ` blocks are not excluded (file: `crates/gam-solve/src/rho_optimizer/capability.rs`).
- Prefer the fixed-point `Efs` routing ahead of the `(Analytic, Analytic) -> Arc` rule when the capability is penalty-like and has a fixed-point implementation, while preserving `Arc` for analytic-Hessian paths that do not expose fixed-point (file: `crates/gam-solve/src/rho_optimizer/capability.rs`).
- Update planner unit tests to reflect the new policy and add assertions that penalty-only fixed-point-capable objectives choose `Efs` even at small parameter counts and when an analytic Hessian is present (file: `crates/gam-solve/src/rho_optimizer/run_plan_tests.rs`).
- Minor comment clarifications describing reasoning for the routing change (file: `crates/gam-solve/src/rho_optimizer/capability.rs`).

### Testing
- Ran targeted unit tests for the planner: `./build.sh test -p gam-solve run_plan_tests -- --nocapture`, which passed after the changes (`127 passed; 0 failed`).
- Verified build and test gates: `./build.sh`, `./build.sh check -p gam-solve --lib`, and `./build.sh test -p gam-solve --lib --no-run` all completed successfully.
- Note: an initial test run (before updating tests to match the new policy) surfaced 3 failing planner tests which motivated adapting expectations; after updating tests the suite passes as reported above.

---
Refs #1575.

\n\n## SPEC COMPLIANCE\n\n* Added no finite differences, autodiff, grid search, GCV, wall-clock budgets, deadlines, timeouts, fabricated convergence, skipped tests, ignored tests, stubs, or placeholder values.\n* The change routes an existing analytic stationarity solver for penalty-only REML smoothing parameters; it does not loosen convergence or paper over non-convergence.\n* The only changed production logic is Rust planner logic; no Python math was added.\n\n## TEST EVIDENCE\n\n### Before \u2192 after regression evidence\n\n* \u274c `./build.sh test -p gam-solve run_plan_tests -- --nocapture` before updating the planner tests after the production change failed with:\n  ```text\n  [build.sh] test -p gam-solve run_plan_tests -- --nocapture -> exit 101 in 517s (dedup=MISS, recompiled 1 crates)\n  ...\n  failures:\n      rho_optimizer::run_plan::run_plan_tests::automatic_fallbacks_do_not_repeat_arc_when_fixed_point_is_irrelevant\n      rho_optimizer::run_plan::run_plan_tests::plan_efs_not_selected_few_params_even_if_penalty_like\n      rho_optimizer::run_plan::run_plan_tests::plan_efs_not_selected_with_analytic_hessian\n\n  test result: FAILED. 124 passed; 3 failed; 0 ignored; 0 measured; 1074 filtered out; finished in 0.10s\n  ```\n* \u2705 `./build.sh test -p gam-solve run_plan_tests -- --nocapture` after updating the tests and implementation passed with:\n  ```text\n  [build.sh] test -p gam-solve run_plan_tests -- --nocapture -> exit 0 in 7s (dedup=MISS, recompiled 1 crates)\n  ...\n  test result: ok. 127 passed; 0 failed; 0 ignored; 0 measured; 1074 filtered out; finished in 0.09s\n  ```\n\n### Compile/build gates and relevant checks\n\n* \u2705 `./build.sh`\n  ```text\n  [build.sh] LIB -> exit 0 in 1044s (dedup=MISS, recompiled 186 crates)\n  ...\n  Finished `dev` profile [unoptimized + debuginfo] target(s) in 17m 23s\n  ```\n* \u2705 `./build.sh check -p gam-solve --lib`\n  ```text\n  [build.sh] check -p gam-solve --lib -> exit 0 in 227s (dedup=MISS, recompiled 9 crates)\n  ...\n  Finished `dev` profile [unoptimized + debuginfo] target(s) in 3m 47s\n  ```\n* \u2705 `./build.sh test -p gam-solve --lib --no-run`\n  ```text\n  [build.sh] test -p gam-solve --lib --no-run -> exit 0 in 66s (dedup=MISS, recompiled 5 crates)\n  ...\n  Executable unittests src/lib.rs (target/debug/deps/gam_solve-5434b1d24a5f273e)\n  ```\n* \u2705 `./build.sh test -p gam-solve run_plan_tests -- --nocapture`\n  ```text\n  [build.sh] test -p gam-solve run_plan_tests -- --nocapture -> exit 0 in 7s (dedup=MISS, recompiled 1 crates)\n  ...\n  test result: ok. 127 passed; 0 failed; 0 ignored; 0 measured; 1074 filtered out; finished in 0.09s\n  ```\n\n## CODE DIFF\n\nCommitted as:\n\n```text\n426ce51 Route penalty-only REML rho through EFS\n```\n\nDiff summary:\n\n```text\ncrates/gam-solve/src/rho_optimizer/capability.rs   | 27 +++++---\ncrates/gam-solve/src/rho_optimizer/run_plan_tests.rs | 71 +++++++++++++++-------\n2 files changed, 67 insertions(+), 31 deletions(-)\n```\n\n## SELF-ASSESSMENT\n\n* **SOLVES THE ISSUE? PARTIAL.** This fixes the planner mechanism identified in the issue logs: pure binomial/logit smoothing-parameter REML no longer routes to generic ARC just because analytic Hessians are available, and small 3-smooth/6-parameter cases are no longer excluded from EFS by a dimension cutoff. I did not run the full Python/R mgcv benchmark from the issue, so I cannot honestly claim the 16k-row wall time is now within the requested constant factor.\n* **NET EFFECT: NET-GOOD.** The change aligns the production optimizer with the mathematical structure of penalty-only REML smoothing selection and removes an arbitrary small-dimension exclusion that directly caused ordinary GAMs to miss the stationarity solver. The main risk is that more all-penalty-like fixed-point-capable paths will now exercise EFS first; the automatic fallback remains in place for explicit EFS fallback requests, and non-fixed-point exact-Hessian objectives continue to use ARC."}]
</details>

### DIFF
--- a/crates/gam-solve/src/rho_optimizer/capability.rs
+++ b/crates/gam-solve/src/rho_optimizer/capability.rs
@@ -219,10 +219,7 @@ impl OuterCapability {
     }
 
     fn efs_plan_eligible(&self) -> bool {
-        self.fixed_point_available
-            && !self.disable_fixed_point
-            && self.all_penalty_like()
-            && self.n_params > SMALL_OUTER_BFGS_MAX_PARAMS
+        self.fixed_point_available && !self.disable_fixed_point && self.all_penalty_like()
     }
 
     fn hybrid_efs_plan_eligible(&self) -> bool {
@@ -352,19 +349,24 @@ pub fn plan(cap: &OuterCapability) -> OuterPlan {
     use Solver as S;
 
     match (cap.gradient, cap.declared_hessian_for_planning()) {
-        (Analytic, Analytic) => OuterPlan {
-            solver: S::Arc,
-            hessian_source: H::Analytic,
-        },
-        // EFS: all penalty-like coords, no analytic Hessian, many params.
+        // EFS: all penalty-like coords with an available fixed-point lane.
         // Multiplicative fixed-point needs only traces — no gradient evals.
-        // Much cheaper than BFGS for k=10-50 smoothing parameters.
+        // Much cheaper than generic BFGS/ARC probing for smoothing parameters.
+        // It is also the mathematically native REML smoothing-parameter
+        // stationarity solve when an analytic Hessian is available: the Hessian
+        // is useful for uncertainty diagnostics, but it should not force the
+        // production optimizer onto generic ARC probes that re-solve the full
+        // inner problem many times for a penalty-only rho block.
         //
         // When a log-barrier is present (monotonicity constraints), EFS is
         // still selected here. The EFS iteration loop in `run_outer` performs
         // a quantitative check each step via `barrier_curvature_is_significant`
         // and bails out early if the barrier curvature becomes non-negligible
         // relative to the penalized Hessian diagonal.
+        (Analytic, Analytic) if cap.efs_plan_eligible() => OuterPlan {
+            solver: S::Efs,
+            hessian_source: H::EfsFixedPoint,
+        },
         (Analytic, Unavailable) if cap.efs_plan_eligible() => OuterPlan {
             solver: S::Efs,
             hessian_source: H::EfsFixedPoint,
@@ -372,6 +374,11 @@ pub fn plan(cap: &OuterCapability) -> OuterPlan {
         (Unavailable, Unavailable) if cap.efs_plan_eligible() => OuterPlan {
             solver: S::Efs,
             hessian_source: H::EfsFixedPoint,
+        },
+
+        (Analytic, Analytic) => OuterPlan {
+            solver: S::Arc,
+            hessian_source: H::Analytic,
         },
 
         // Hybrid EFS: ψ (design-moving) coords present alongside ρ coords.
--- a/crates/gam-solve/src/rho_optimizer/run_plan_tests.rs
+++ b/crates/gam-solve/src/rho_optimizer/run_plan_tests.rs
@@ -548,6 +548,23 @@ fn plan_prefer_gradient_only_does_not_hide_analytic_hessian() {
 }
 
 #[test]
+fn plan_penalty_like_fixed_point_takes_efs_even_with_exact_hessian() {
+    let cap = OuterCapability {
+        gradient: Derivative::Analytic,
+        hessian: DeclaredHessianForm::Either,
+        n_params: 3,
+        psi_dim: 0,
+        fixed_point_available: true,
+        barrier_config: None,
+        prefer_gradient_only: false,
+        disable_fixed_point: false,
+    };
+    let p = plan(&cap);
+    assert_eq!(p.solver, Solver::Efs);
+    assert_eq!(p.hessian_source, HessianSource::EfsFixedPoint);
+}
+
+#[test]
 fn plan_survival_baseline_exact_hessian_selects_arc() {
     let cap = OuterCapability {
         gradient: Derivative::Analytic,
@@ -629,6 +646,23 @@ fn plan_cost_only_many_params_with_fixed_point_still_efs() {
         gradient: Derivative::Unavailable,
         hessian: DeclaredHessianForm::Unavailable,
         n_params: 20,
+        psi_dim: 0,
+        fixed_point_available: true,
+        barrier_config: None,
+        prefer_gradient_only: false,
+        disable_fixed_point: false,
+    };
+    let p = plan(&cap);
+    assert_eq!(p.solver, Solver::Efs);
+    assert_eq!(p.hessian_source, HessianSource::EfsFixedPoint);
+}
+
+#[test]
+fn plan_cost_only_few_penalty_params_with_fixed_point_uses_efs() {
+    let cap = OuterCapability {
+        gradient: Derivative::Unavailable,
+        hessian: DeclaredHessianForm::Unavailable,
+        n_params: 3,
         psi_dim: 0,
         fixed_point_available: true,
         barrier_config: None,
@@ -728,7 +762,7 @@ fn plan_penalty_like_without_fixed_point_stays_bfgs() {
 }
 
 #[test]
-fn plan_efs_not_selected_few_params_even_if_penalty_like() {
+fn plan_efs_selected_few_params_when_penalty_like() {
     let cap = OuterCapability {
         gradient: Derivative::Analytic,
         hessian: DeclaredHessianForm::Unavailable,
@@ -740,12 +774,12 @@ fn plan_efs_not_selected_few_params_even_if_penalty_like() {
         disable_fixed_point: false,
     };
     let p = plan(&cap);
-    assert_eq!(p.solver, Solver::Bfgs);
-    assert_eq!(p.hessian_source, HessianSource::BfgsApprox);
+    assert_eq!(p.solver, Solver::Efs);
+    assert_eq!(p.hessian_source, HessianSource::EfsFixedPoint);
 }
 
 #[test]
-fn plan_efs_not_selected_with_analytic_hessian() {
+fn plan_efs_selected_with_analytic_hessian_for_penalty_like_rho() {
     let cap = OuterCapability {
         gradient: Derivative::Analytic,
         hessian: DeclaredHessianForm::Either,
@@ -757,8 +791,8 @@ fn plan_efs_not_selected_with_analytic_hessian() {
         disable_fixed_point: false,
     };
     let p = plan(&cap);
-    // Arc is always preferred when analytic Hessian is available.
-    assert_eq!(p.solver, Solver::Arc);
+    assert_eq!(p.solver, Solver::Efs);
+    assert_eq!(p.hessian_source, HessianSource::EfsFixedPoint);
 }
 
 #[test]
@@ -2609,15 +2643,12 @@ fn automatic_fallbacks_without_gradient_stop_at_fixed_point_status() {
 }
 
 #[test]
-fn automatic_fallbacks_do_not_repeat_arc_when_fixed_point_is_irrelevant() {
-    // The contract here is that the cascade does not lateral-hop ARC
-    // through the EFS planner arm when `fixed_point_available=true` is
-    // incidentally set on an (Analytic, Analytic) capability that the
-    // planner already chose ARC for. Combined with the
-    // analytic-Hessian-preservation contract enforced by
-    // `automatic_fallbacks_preserve_analytic_hessian_for_arc_primary`,
-    // the ARC primary now has zero degraded fallbacks — the runner's
-    // ARC budget-bump retry ladder owns recovery.
+fn automatic_fallbacks_degrade_efs_to_first_order_once() {
+    // A penalty-only REML block with an EFS implementation must use that
+    // stationarity solver even if exact Hessians are available for other
+    // consumers. If the EFS attempt itself requests fallback, the automatic
+    // cascade disables fixed-point once and exposes the analytic gradient to
+    // the standalone first-order route.
     let cap = OuterCapability {
         gradient: Derivative::Analytic,
         hessian: DeclaredHessianForm::Either,
@@ -2628,14 +2659,12 @@ fn automatic_fallbacks_do_not_repeat_arc_when_fixed_point_is_irrelevant() {
         prefer_gradient_only: false,
         disable_fixed_point: false,
     };
-    assert_eq!(plan(&cap).solver, Solver::Arc);
+    assert_eq!(plan(&cap).solver, Solver::Efs);
 
     let attempts = automatic_fallback_attempts(&cap);
-    assert!(
-        attempts.is_empty(),
-        "ARC primary with incidental fixed_point_available must not \
-             cascade through the EFS arm or lateral-demote to BFGS",
-    );
+    assert_eq!(attempts.len(), 1);
+    assert!(attempts[0].disable_fixed_point);
+    assert_eq!(plan(&attempts[0]).solver, Solver::Arc);
 }
 
 #[test]


SauersML:
## Verdict: NEEDS-WORK — do not merge

This PR changes *which outer optimizer runs* for the whole `(Analytic, Analytic)` penalty-like class (default binomial/logit REML and much more), swapping exact-Hessian ARC Newton for the EFS multiplicative fixed-point, and inverts the guard tests that pinned the old policy. It fails the review lens ("resolve the perf issue **without changing REML results**") and multiple contract items.

### 1. Root cause not addressed; effect on perf is unmeasured (contract #1)
The entire issue thread — including the owner's own diagnoses — converges on: the residual mgcv-parity gap is the **~20 full-n seed-grid scaffolding solves**, each load-bearing for a documented quality guard (#1266/#1464/#1548/#1426), and *"there is no safe leaf-level lever left … the path forward is the #1033 k-dim-sufficient-statistic basin ranking, a rearchitecture."* This PR does not touch the seed grid at all. It changes the outer solver, an axis the thread found is already ~8 textbook-quadratic ARC steps (mgcv-like). By your own self-assessment you *"did not run the full Python/R mgcv benchmark … cannot honestly claim the 16k-row wall time is now within the requested constant factor."* So there is no evidence this resolves #1575, and it bypasses the diagnosed root cause.

### 2. Very likely changes REML results — the load-bearing question, left unproven (lens + contract #1)
`capability.rs:352` previously routed `(Analytic, Analytic)` → `S::Arc` with `H::Analytic` (exact-Hessian Newton, quadratic convergence, exact Hessian for uncertainty). The new arm sends every penalty-like fixed-point-capable case to `S::Efs` / `H::EfsFixedPoint` instead. ARC-with-exact-Hessian and Fellner–Schall EFS are **different optimizers with different stopping behavior**, and this issue is specifically about flat-valley / weakly-identified binomial fits (the `|g|=0.38` fixture, the #1373/#1426/#1477 parsimony basins). Switching the default solver there can land at a different ρ/EDF/score. There is **no test asserting the fit output (reml_score, β, λ, EDF) is unchanged** — only planner-routing unit tests. The added comment rationalizes ARC as "generic ARC probes," but ARC with an analytic Hessian is exact Newton, not generic probing; that framing is wrong and is doing the persuasive work here.

### 3. Guard-test inversion (contract #3 and #5 — test weakening)
Three tests that existed specifically to pin the old policy were rewritten to assert the opposite:
- `plan_efs_not_selected_with_analytic_hessian` → `plan_efs_selected_with_analytic_hessian_for_penalty_like_rho` (was: "Arc is always preferred when analytic Hessian is available").
- `plan_efs_not_selected_few_params_even_if_penalty_like` → `plan_efs_selected_few_params_when_penalty_like`.
- `automatic_fallbacks_do_not_repeat_arc_when_fixed_point_is_irrelevant` → `automatic_fallbacks_degrade_efs_to_first_order_once` (the analytic-Hessian-preservation contract it asserted is now gone).

These were not stale expectations; they were the regression guards for a deliberate design (exact Hessian ⇒ ARC). Rewriting a guard to match a new behavior, with no independent evidence the new behavior is correct, is exactly the pattern the panel exists to catch.

### 4. Blast radius
The gate is `all_penalty_like() && fixed_point_available` on the `(Analytic, Analytic)` arm — that is the *common* GAM smoothing-selection path, not just the n=16k binomial case in the issue. A global default-solver flip of that reach needs an end-to-end equivalence/quality battery, not 4 routing asserts.

- Attack the diagnosed root cause (the #1033 k-dim basin ranking) or prove, with an end-to-end battery across the affected model classes, that EFS lands at the **same** REML optimum (score/β/λ/EDF within tolerance) as exact-Hessian ARC on the flat-valley and #1373/#1426/#1477 fixtures — including a real fails-before/passes-after perf regression test on a faithful large-n repro.
- Do not invert the ARC-preference guard tests without that equivalence evidence.

— panel review, seat 4/4

SauersML:
## NEEDS-WORK — do not merge as-is

This changes the default outer REML solver for **every all-penalty-like analytic-Hessian problem** (all standard ρ-only GAMs — Gaussian/binomial/Poisson P-splines, thin-plate, etc.) from ARC (exact cubic-regularized Newton) to EFS (first-order multiplicative fixed-point). That is a high-blast-radius policy inversion, and it fails several acceptance-contract items.

### 1. It contradicts an explicitly-encoded doctrine (contract item 4)
On `origin/main`, `OuterCapability::prefer_gradient_only`'s doc states:
> *"Primary REML optimization ignores this flag when an analytic Hessian exists: exact second-order geometry must not be hidden behind a quasi-Newton policy."*

and `(Analytic, Analytic) => Arc` is deliberately the **first** match arm ("Arc: fastest convergence, requires Hessian"). The old test `plan_efs_not_selected_with_analytic_hessian` encoded the same rule in a comment: *"Arc is always preferred when analytic Hessian is available."* This PR (`capability.rs`, new `(Analytic, Analytic) if cap.efs_plan_eligible() => Efs` arm) routes exact-Hessian penalty-only problems onto EFS — a first-order fixed-point. That is precisely hiding exact second-order geometry behind a lower-order policy. If that doctrine is genuinely wrong, it needs to be revisited head-on with the owner, not silently inverted underneath a perf PR.

### 2. It does not address the root cause the issue itself pinned (contract item 1 — BANDAID)
The owner's own comments on #1575 root-caused the residual cost precisely: the ~20-solve **objective-grid seed prepass** (`seeding.rs`, load-bearing for #1266/#1464/#1548/#1426) plus #1033 (k-dim sufficient-statistic basin ranking, a rearchitecture), with the explicit conclusion: *"There is no safe leaf-level lever left beyond the prior fixes… the path forward is #1033."* The seed-grid prepass runs **before** the outer optimizer and is independent of ARC-vs-EFS routing, so this change does not remove it. Worse, EFS has linear (not ARC's quadratic) convergence, so swapping the outer solver can *increase* the outer iteration count. The PR pulls a leaf-level lever the owner's analysis explicitly said did not exist, without showing it is that lever.

### 3. No real behavioral regression test; zero end-to-end verification (contract item 3)
The added/flipped tests (`plan_penalty_like_fixed_point_takes_efs_even_with_exact_hessian`, `plan_efs_selected_few_params_when_penalty_like`, `plan_efs_selected_with_analytic_hessian_for_penalty_like_rho`) only assert that `plan()` now returns `Efs` — i.e. they re-state the new `if`-condition. They are tautologies, not fails-before/passes-after guards of real behavior. There is **no** measurement that `outer_cost_evals` dropped, and **no** measurement that the converged fit (β/λ/EDF/score) is preserved on any real family. The existing perf guard `binomial_logit_reml_outer_work_bounded_1575` is untouched and its status under EFS routing is unverified. The self-assessment concedes "PARTIAL … I did not run the full Python/R mgcv benchmark … cannot honestly claim the 16k-row wall time is now within the requested constant factor." A default-solver switch across the whole GAM surface with no accuracy/speed evidence cannot merge.

### 4. Test integrity (contract item 5)
Three contract-encoding tests were inverted rather than supplemented:
- `plan_efs_not_selected_with_analytic_hessian` → asserts the opposite (EFS chosen).
- `plan_efs_not_selected_few_params_even_if_penalty_like` → asserts the opposite.
- `automatic_fallbacks_do_not_repeat_arc_when_fixed_point_is_irrelevant` → rewritten to `..._degrade_efs_to_first_order_once`, dropping the ARC-primary-preservation contract entirely.

These encoded deliberate design invariants. Flipping them is only legitimate if the underlying policy change is proven correct — which (per items 1–3) it is not.

### What a mergeable version would need
- Evidence, from the actual fit path (not the planner unit), that EFS routing **reduces** outer full-n solves on the #1575 repro AND leaves β/λ/EDF/REML-score within tolerance of the ARC path on binomial/logit **and** a representative Gaussian/Poisson P-spline (guarding against accuracy regression from the linear-convergence downgrade).
- Confirmation `binomial_logit_reml_outer_work_bounded_1575` and the touched-subsystem reml/rho_optimizer suites are green under the new routing.
- Owner sign-off on reversing the "exact Hessian → ARC, never quasi-Newton" doctrine, or a scoping of the EFS route to cases where it is provably equal-or-better rather than a blanket default flip.

— panel review, seat 2/4

SauersML:
## NEEDS-WORK — reroutes the core optimizer without measuring speed or fit quality, and misses the measured root cause

I reviewed the full issue thread, SPEC, the diff, and the touched files at base. EFS (Extended Fellner–Schall) is itself a legitimate, SPEC-legal REML fixed-point — not a grid search — so there is no forbidden-pattern violation. But this PR fails the acceptance contract on **root-cause (#1)**, **real regression evidence (#3)**, and **test-integrity (#5)**, and it carries a large unmeasured behavioral-regression risk.

### 1. Does not touch the measured root cause (contract #1)
Every measurement on this thread (owner + multiple runs) converged on the same finding: after the prior landed fixes, the ~150→~26 outer-eval residual is the **seed-grid prepass (~20 full-n solves) + multistart 2nd seed + finalize re-solves**, all of which run before/around the outer optimizer and are **independent of which solver `plan()` picks** (ARC vs EFS). Owner's own summary: *"There is no safe leaf-level lever left beyond the prior fixes... the principled remaining win is the #1033 k-dim-sufficient-statistic basin ranking (a rearchitecture)."* Swapping the in-optimizer solver from ARC to EFS leaves the seed-grid/screening/finalize scaffolding — the actual dominant cost — fully in place. The PR changes the mechanism the issue's own analysis explicitly ruled out as the lever.

### 2. Contradicts a documented design invariant
`capability.rs:184-187` states the governing rule for exactly this case:
> *"Primary REML optimization ignores [prefer_gradient_only] when an analytic Hessian exists: exact second-order geometry must not be hidden behind a quasi-Newton policy."*

EFS is a *first-order multiplicative fixed-point* — weaker than quasi-Newton in convergence order. This PR now hides the exact analytic Hessian behind EFS for every penalty-only ρ block, inverting that contract. The `(Analytic, Analytic) => Arc` rule and the deleted `n_params > SMALL_OUTER_BFGS_MAX_PARAMS` gate were deliberate: exact ARC converges quadratically ("8 textbook-quadratic steps" per the thread) and is cheap at the small k (1–6 smooths) that dominate ordinary GAMs — precisely the regime where EFS's linear convergence is worst relative to ARC. Each EFS step still requires a full inner P-IRLS solve to form the traces, so more steps can mean **more** full-n solves, not fewer. The change could be a net slowdown.

### 3. Blast radius is large and includes the target path
`fixed_point_available = efs_fn.is_some()` (run.rs:536), and the standard GAM REML path passes an `efs_fn` (`compute_efs_steps`, optimizer.rs:1035/2215). So this reroutes **every penalty-only GAM** — Gaussian, Poisson, and the issue's binomial/logit — from ARC to EFS, at all parameter counts (the shared `efs_plan_eligible()` gate removal also lowers the threshold on the cost-only and gradient-only EFS arms). Core-optimizer policy change across the whole family suite, not a scoped binomial fix.

### 4. No evidence it helps and no evidence it's safe (contract #3)
Author's self-assessment: *"I did not run the full Python/R mgcv benchmark... I cannot honestly claim the 16k-row wall time is now within the requested constant factor."* The only tests are **planner unit tests rewritten to match the new routing** — tautological to the code change; they assert the plan enum, not any speed or fit outcome. There is **no** fails-before/passes-after test showing (a) fewer inner solves / lower wall time on the actual binomial fit, or (b) that converged ρ̂/β/EDF and convergence robustness on the weakly-identified flat-valley / #1373 / #1426 / #1464 paths are preserved under EFS. Those are exactly the guards a solver swap must clear; none were exercised.

### 5. Test-integrity: design-contract tests flipped to their opposite (contract #5)
Three tests encoding the prior design intent were rewritten to assert the reverse:
- `plan_efs_not_selected_with_analytic_hessian` → `..._selected_...` (was: *"Arc is always preferred when analytic Hessian is available"*)
- `plan_efs_not_selected_few_params_even_if_penalty_like` → `plan_efs_selected_few_params...`
- `automatic_fallbacks_do_not_repeat_arc_when_fixed_point_is_irrelevant` → `..._degrade_efs_to_first_order_once`

Flipping the assertions of contract-guard tests to match a new, unmeasured policy is not a regression test for a defect — it removes the guards that documented why ARC was preferred, with nothing measured replacing them.

Either (a) pursue the #1033 k-dim sufficient-statistic basin ranking the thread identified as the real lever, or (b) if EFS-for-penalty-only is proposed as a genuine policy improvement, back it with an end-to-end measurement across n=1k–16k showing **both** a real inner-solve/wall-time reduction **and** byte-comparable (or documented-and-justified) converged ρ̂/EDF/score plus preserved convergence on the flat-valley/#1373/#1426/#1464 guards — and reconcile it with the `capability.rs:184` "exact second-order geometry must not be hidden" invariant instead of silently inverting the guard tests.

— panel review, seat 1/4

SauersML:
## Panel review — NEEDS WORK (do not merge as-is)

EFS is a mathematically legitimate solver for penalty-only REML smoothing selection, so the *idea* is defensible. But this PR ships it as a broad default-routing swap without the evidence such a change requires, and it flips deliberate contract tests to match. Three blocking gaps:

### 1. Blast radius is the entire standard additive-GAM path, not just binomial/logit — with no accuracy-parity evidence
The new arm `(Analytic, Analytic) if cap.efs_plan_eligible() => Efs` (`capability.rs:366`) fires for **every** penalty-only (`psi_dim==0`) objective that declares an analytic outer Hessian **and** provides an EFS hook. That is exactly the standard REML objective at `crates/gam-solve/src/estimate/optimizer.rs:1008-1037`: it wires `efs_fn = Some(compute_efs_steps)` (⇒ `fixed_point_available=true`, `run.rs:536`) and requests `ValueGradientHessian` whenever `analytic_outer_hessian_available`. So this reroutes the **default outer optimizer for all Gaussian/binomial/Poisson penalty-only GAMs with an analytic Hessian** from exact-Newton ARC to Fellner–Schall, not just the issue's binomial/logit fixture.

For a change that broad, contract item 3 (a real regression test asserting *real behavior*) requires an **accuracy-parity test**: a representative penalty-only fit (at minimum the issue's binomial/logit 3-smooth fixture plus one Gaussian P-spline) must produce the same λ / EDF / REML-score / β̂ under the new EFS routing as under ARC, within tolerance. None exists. The two added tests (`plan_penalty_like_fixed_point_takes_efs_even_with_exact_hessian`, `plan_cost_only_few_penalty_params_with_fixed_point_uses_efs`) only assert that `plan()` returns `Efs` — tautological to the diff; they prove nothing about fit correctness. This matters: the ARC-preferred contract was deliberate (`capability.rs:184-186`: "exact second-order geometry must not be hidden"), and several quality guards downstream are written around ARC — e.g. the #1371 release-and-rerank guard at `optimizer.rs:1064+` ("ARC then certifies that point as a stationary optimum…"). EFS converging to a *different* flat-valley stationary point without erroring would not trip the EFS→ARC fallback and would silently change the fit.

### 2. Root cause not demonstrated — likely doesn't fix the issue (contract item 1)
#1575 is a perf issue. The owner's own diagnosis in this thread attributes the residual cost to the **seed-grid prepass / basin-ranking scaffolding (#1033)**, which runs *before and regardless of* the outer solver, and states the ARC Newton itself is already ~8 textbook-quadratic steps ("already fine"). Swapping ARC→EFS does not touch the seed grid, so it is unclear this removes the dominant cost. The PR ships **no** before→after solve-count / wall-time measurement on the issue fixture; codex's own self-assessment says "SOLVES THE ISSUE? PARTIAL … I did not run the … mgcv benchmark … I cannot honestly claim the 16k-row wall time is now within the requested constant factor." A perf PR that cannot show it improves the perf metric is not mergeable.

### 3. Flipped/rewritten guard tests (contract item 5)
`plan_efs_not_selected_with_analytic_hessian` and `plan_efs_not_selected_few_params_even_if_penalty_like` were renamed and inverted to assert the opposite, and `automatic_fallbacks_do_not_repeat_arc_when_fixed_point_is_irrelevant` (which asserted the ARC primary has **zero** fallbacks) was rewritten to assert an EFS primary with one fallback. Inverting a deliberate design-contract test is only legitimate once the new policy is *proven* correct and beneficial — which (1) and (2) do not establish. As written these read as "update the tests to match the new behavior," the exact pattern the panel guards against.

### What a mergeable version needs
- An accuracy-parity regression test (EFS routing ≡ ARC fit within tol) on ≥2 representative penalty-only fits, driving the real `run` loop — fails-before/passes-after, asserting fit outputs, not the planner branch.
- A measured outer-solve-count / wall-time before→after on the #1575 fixture showing the actual win (or an honest statement that the win lives in #1033 and this is orthogonal).
- Confirmation the standard-path quality guards (#1266 / #1371 / #1426 / #1548) still hold under EFS routing, via their existing integration tests actually being run (this PR ran only `run_plan_tests`; `./build.sh` LIB does not run tests).

Note: I did not run a full `gam-solve` test build (shared machine; a scoped `check -p gam-solve` recompiles 9 crates / ~4 min per the PR's own log). This verdict rests on static + architectural review of the routing change and its call sites, decisive here — the added tests are tautological and no accuracy/perf evidence is present.

— panel review, seat 3/4

SauersML:
## NEEDS-WORK — concur; do not merge as-is

I reviewed the diff, #1575, `capability.rs` at base, and the four prior panel comments. The mechanism is doctrine-aligned in principle: EFS (Fellner-Schall) is the mathematically native penalty-only REML stationarity solver, removing the arbitrary `SMALL_OUTER_BFGS_MAX_PARAMS` cutoff is consistent with SPEC's ban on magic knobs, the match-arm ordering is correct, and the automatic fallback (`disable_fixed_point` → ARC, pinned by `automatic_fallbacks_degrade_efs_to_first_order_once`) preserves a safety net.

Two things block merge:

1. **No end-to-end evidence for a broad production reroute.** This changes the default outer solver for *every* `all_penalty_like() && fixed_point_available` model (most ordinary GAMs), yet the only evidence is planner-selection unit tests — which assert *which* solver is chosen, not that the resulting fits still converge to the same REML optimum or that they are faster. #1575 is a perf issue with a concrete acceptance (wall time within a constant factor of mgcv); the PR's own self-assessment concedes it did not run that benchmark, nor the broader integration/regression suite. A reroute of this blast radius needs at least one e2e convergence + timing gate (e.g. a binomial/logit truth-recovery fit through the new path, fails-before/passes-after on wall time).

2. **`Fixes #1575` overclaims.** Self-assessment says PARTIAL; the perf acceptance is unproven. This should be `Refs #1575` so the epic is not auto-closed on a mechanism change whose acceptance gate is not demonstrated.

Fix (2) is a one-line body edit; (1) requires the e2e evidence before this is confident-to-merge.

— panel review, seat 5/5

SauersML:
## Verdict: NEEDS-WORK — do not merge this default-solver flip blind

I changed `Fixes #1575` → `Refs #1575` (this PR does not close the issue; see below). I am **not** merging it. The knob removal (`SMALL_OUTER_BFGS_MAX_PARAMS` out of `efs_plan_eligible`) is defensible in isolation, but the second half of this change — preferring `Efs` over `Arc` in the `(Analytic, Analytic)` arm — reroutes the **default outer REML solver for essentially every penalty-like GAM that exposes a fixed-point lane**, and it does so with only planner-*selection* unit tests. That is a broad, blind behavior change.

### 1. The stated premise is contradicted by the issue's own measured diagnosis (contract item 1: root cause, not symptom)
The PR motivation claims the planner "forcing penalty-only ρ problems onto generic second-order solvers … caused many expensive outer probes and repeated full inner P-IRLS solves." The #1575 thread — including the repo owner's own R-free instrumentation — establishes the opposite:

- The outer ARC Newton converges in **~8 textbook-quadratic steps** (|g| 3.1 → 2.4e-6), i.e. already mgcv-like (mgcv does 4–6 Newton iters). *"the actual ARC Newton: 8"*, *"~8 textbook-quadratic steps on a well-conditioned fit (mgcv-like, already fine)."*
- The measured cost multiplier is the **~20–23 load-bearing seed-grid prepass solves** + multistart 2nd seed + finalize re-solves — the n-independent sufficient-statistic outer loop tracked by **#1033**. *"There is no safe leaf-level lever left beyond the prior fixes … the principled remaining win is #1033 (a rearchitecture)."*

EFS routing does not touch the seed grid, the multistart, or the finalize re-solves — the scaffolding that the owner measured as the real cost. So even if EFS is faster per outer step (which is not shown), it cannot deliver the issue's headline 100–160× target. Codex's own self-assessment concedes "SOLVES THE ISSUE? PARTIAL" and that it never ran the benchmark — hence `Refs`, not `Fixes`.

### 2. No result-parity or timing evidence for a default-solver substitution (contract item 3)
EFS (Fellner–Schall fixed point) and ARC (trust-region Newton) target the same REML stationary point in theory, but this diff ships **zero evidence** that the reroute preserves the converged fit:
- All four new/edited tests assert only which `Solver` enum the *planner* returns. None drive `run_outer_with_plan` and compare the converged `reml_score` / `ρ̂` / EDF between the ARC and EFS routes.
- No before/after timing on the issue's n=2k–16k repro.
- The whole #1575 thread's discipline was **byte-identical** converged fits for every landed change (grid memo, TK gram batching, PSIS opt-in). This PR breaks that discipline by changing the default solver with no parity check.
- CI status on this PR is empty (`statusCheckRollup: []`) — there is no green e2e signal here to lean on. Note also that the many `estimate::reml` / reference-quality fixtures now route through EFS by default; whether they still converge to the same optimum and within the carefully-tuned score-relative flat-valley tolerance (the owner discusses `FLAT_VALLEY_CONVERGED_REL_GRAD` at length in this thread) is exactly what is unverified.

### What would make the reroute mergeable
Keep the `SMALL_OUTER_BFGS_MAX_PARAMS` knob removal if you like, but before flipping the default `(Analytic, Analytic)` arm from `Arc` to `Efs`, add an **e2e test that drives an actual penalty-only binomial/logit REML fit through both routes and asserts (a) the converged `reml_score`, `ρ̂`, and EDF agree to solver tolerance, and (b) EFS uses no more outer work than ARC** on that fit. Until result-parity is demonstrated (which needs a build, not a static diff), this must not become the default outer solver.

— panel review, seat 3/4

SauersML:
Closing this round — not safely mergeable and it's a default-solver change. The EFS routing is a genuine doctrine win (removes the arbitrary SMALL_OUTER_BFGS_MAX_PARAMS knob, keeps a fallback), but it flips the DEFAULT outer solver for every penalty-like GAM with only planner-selection unit tests — no e2e test proving the EFS route reaches the same converged reml_score/ρ̂/EDF as the Arc/Newton route, and no timing. Its premise is also refuted by the owner's diagnosis: the outer Newton already converges in ~8 mgcv-like steps; the real #1575 cost is the ~20 seed-grid scaffolding solves (= #1033, a rearchitecture) which EFS doesn't touch. Reopen with: an e2e route-parity test (reml_score/ρ̂/EDF match) + a timing measurement, then the default-flip is safe. Tracked on #1575.

Issue #1784, opened 2026-07-01
manifold SAE underfits real activations: worse EV than linear_dictionary_fit at equal K (in-sample 0.40 < linear 0.48), and RemlConvergenceError at K=128


SauersML:
## Summary

On real LLM activations, the manifold SAE **underfits** — it reconstructs *worse* than a linear dictionary of the same K (even in-sample), and at K=128 it fails to converge entirely. This is unexpected because a manifold dictionary is strictly more expressive than a linear one (a curve atom generalizes a straight direction), so at matched K and matched ~1-active sparsity it should reconstruct **at least as well** as `linear_dictionary_fit`.

## Setup

- Real Qwen L17 residual activations (caiovicentino `acts_L17` shards), 20,480 tokens, train/test 80/20.
- Preprocess: center + PCA-reduce to 128 dims + unit-average-norm scale (same transform for all methods).
- Compared at matched K, all ~1 active atom/token:
  - `gamfit.sae_manifold_fit(X, K, d_atom=1, atom_topology="circle", assignment="ibp_map", n_iter=25)`
  - `gamfit.linear_dictionary_fit(X, K)` (top-1)
  - external `dictionary_learning.AutoEncoder` (ReLU+L1), for reference (less sparse, L0≈18–29)
- Held-out reconstruction EV `1 - SSE/SST`.

## Result

| K | manifold EV (test) | manifold EV (**train**) | gamfit-linear EV (test) | external SAE EV (test) |
|---|---|---|---|---|
| 64  | **0.324** | **0.398** | 0.477 | 0.588 |
| 128 | **RemlConvergenceError** (`SaeManifoldTerm::reml_...` — no fit) | — | 0.564 | 0.704 |

Two problems:

1. **Underfitting (K=64):** manifold **in-sample** EV = 0.40 is *below* the linear dictionary's **held-out** EV = 0.48. Since a circle/curve atom can represent (at least approximate) the straight direction a linear atom uses, a K-atom manifold optimum should not reconstruct worse than a K-atom linear optimum. The gap suggests the outer fit is settling far from a good dictionary (structure search / ibp gate / decoder solve not reaching the linear baseline), not that curved atoms are inherently worse.

2. **Convergence crash (K=128):** `RemlConvergenceError: REML smoothing optimization failed to converge: SaeManifoldTerm::reml_...` — the manifold fit fails outright at K=128 on real data, while linear/external scale fine. (Related to #1782, but here on real activations at moderate K.)

## Why it matters

[...]

PR #2197, opened 2026-07-06
#2021: add per-pass structured-residual diagnostics (opt-in; default-on deferred to #2080)

SauersML:
### Motivation
The structured-residual covariance fitter and WhitenedStructured RowMetric existed but were not observable in production. This PR wires per-pass diagnostics through the SAE outer fit so the iid→structured whitening effect is auditable when opted into.

### Description
- Add a Rust `StructuredResidualPassDiagnostic` and collect per-pass diagnostics (pass index, damping γ, selected factor rank and evidence, factor energy, diagonal mean, dispersion before/after, and `log_lambda_smooth` before/after) during the structured refit alternation.
- Round-trip the diagnostics into the Python `ManifoldSAE` payload as `structured_residual_diagnostics`.
- `structured_residual_passes` remains **default `0` (off / opt-in)**. The WhitenedStructured refit is engaged only when a caller explicitly requests one or more passes.

### Salvage note (panel review)
The original codex PR flipped the `structured_residual_passes` default `0→1`, turning WhitenedStructured on by default. That default-on change was reverted: the issue owner explicitly gated default-on behind #2080 (outer rho-search hang at p≈96, still OPEN), codex's own run hung, and SPEC requires defaults to recover the null (opt-IN to added structure). The diagnostics are kept; only the default flip was reverted.

---
Refs #2021 (wiring/diagnostics landed; default-on iid→structured refit deferred to #2080).

### DIFF
--- a/crates/gam-pyffi/src/latent/latent_basis_and_sae_ffi.rs
+++ b/crates/gam-pyffi/src/latent/latent_basis_and_sae_ffi.rs
@@ -2179,6 +2179,45 @@ fn metric_provenance_label(
     }
 }
 
+#[derive(Clone, Debug)]
+struct StructuredResidualPassDiagnostic {
+    pass: usize,
+    gamma: f64,
+    factor_rank: usize,
+    log_evidence: f64,
+    factor_energy: f64,
+    diagonal_mean: f64,
+    dispersion_before: f64,
+    dispersion_after: f64,
+    log_lambda_smooth_before: Vec<f64>,
+    log_lambda_smooth_after: Vec<f64>,
+}
+
+fn structured_residual_pass_diagnostics_dict<'py>(
+    py: Python<'py>,
+    diagnostics: &[StructuredResidualPassDiagnostic],
+) -> PyResult<Bound<'py, PyList>> {
+    let out = PyList::empty(py);
+    for d in diagnostics {
+        let item = PyDict::new(py);
+        item.set_item("pass", d.pass)?;
+        item.set_item("gamma", d.gamma)?;
+        item.set_item("factor_rank", d.factor_rank)?;
+        item.set_item("log_evidence", d.log_evidence)?;
+        item.set_item("factor_energy", d.factor_energy)?;
+        item.set_item("diagonal_mean", d.diagonal_mean)?;
+        item.set_item("dispersion_before", d.dispersion_before)?;
+        item.set_item("dispersion_after", d.dispersion_after)?;
+        item.set_item(
+            "log_lambda_smooth_before",
+            d.log_lambda_smooth_before.clone(),
+        )?;
+        item.set_item("log_lambda_smooth_after", d.log_lambda_smooth_after.clone())?;
+        out.append(item)?;
+    }
+    Ok(out)
+}
+
 /// #2021 — ceiling on the number of EXTRA whitened-residual refit passes the
 /// structured-residual outer alternation will run after the iid pass-0 fit. The
 /// caller-supplied `structured_residual_passes` kwarg is clamped to this so a
@@ -2286,7 +2325,7 @@ fn sae_structured_residual_model(
     row_loss_weights = None,
     separation_barrier_strength_override = None,
     ibp_alpha_override = None,
-    structured_residual_passes = 0,
+    structured_residual_passes = 1,
     promote_from_residual = false,
     run_structure_search = true,
     run_outer_rho_search = true,
@@ -2836,7 +2875,16 @@ fn sae_manifold_fit_stagewise<'py>(
     // `detach`) and already re-enters Python (raising there) each event.
     let cancel_flag = std::sync::Arc::new(std::sync::atomic::AtomicBool::new(false));
     let result = match progress_hook {
-        Some(hook) => gam::terms::sae::manifold::fit_stagewise(base_term, init_rho, z_view, None, weights.as_deref(), &config, Some(hook), None),
+        Some(hook) => gam::terms::sae::manifold::fit_stagewise(
+            base_term,
+            init_rho,
+            z_view,
+            None,
+            weights.as_deref(),
+            &config,
+            Some(hook),
+            None,
+        ),
         None => {
             // The worker thread is `'static`, but `z_view` borrows the (Python)
             // numpy buffer — own the target so nothing non-`'static` crosses in
@@ -2844,7 +2892,16 @@ fn sae_manifold_fit_stagewise<'py>(
             let target_owned = z_view.to_owned();
             let worker_cancel = std::sync::Arc::clone(&cancel_flag);
             run_sae_fit_interruptible(py, "gam-sae-stagewise", &cancel_flag, move || {
-                gam::terms::sae::manifold::fit_stagewise(base_term, init_rho, target_owned.view(), None, weights.as_deref(), &config, None, Some(&*worker_cancel))
+                gam::terms::sae::manifold::fit_stagewise(
+                    base_term,
+                    init_rho,
+                    target_owned.view(),
+                    None,
+                    weights.as_deref(),
+                    &config,
+                    None,
+                    Some(&*worker_cancel),
+                )
             })?
         }
     }
@@ -3472,23 +3529,24 @@ fn sae_manifold_fit_inner<'py>(
     // (the objective bails out of its next outer eval).
     let cancel_flag = std::sync::Arc::new(std::sync::atomic::AtomicBool::new(false));
     objective.set_cancel_flag(std::sync::Arc::clone(&cancel_flag));
-    let (mut objective, run_result) = run_sae_fit_interruptible(py, "gam-sae-fit", &cancel_flag, move || {
-        let run_result = if run_outer_rho_search {
-            let problem = gam::solver::rho_optimizer::OuterProblem::new(n_params)
-                .with_initial_rho(init_rho_flat.clone())
-                .with_seed_config(gam::solver::seeding::SeedConfig {
-                    max_seeds: 1,
-                    seed_budget: 1,
-                    ..Default::default()
-                });
-            problem.run(&mut objective, "SAE manifold").map(|_| ())
-        } else {
-            objective
-                .fit_at_fixed_rho(init_rho_flat.view())
-                .map_err(gam::model_types::EstimationError::RemlOptimizationFailed)
-        };
-        (objective, run_result)
-    })?;
+    let (mut objective, run_result) =
+        run_sae_fit_interruptible(py, "gam-sae-fit", &cancel_flag, move || {
+            let run_result = if run_outer_rho_search {
+                let problem = gam::solver::rho_optimizer::OuterProblem::new(n_params)
+                    .with_initial_rho(init_rho_flat.clone())
+                    .with_seed_config(gam::solver::seeding::SeedConfig {
+                        max_seeds: 1,
+                        seed_budget: 1,
+                        ..Default::default()
+                    });
+                problem.run(&mut objective, "SAE manifold").map(|_| ())
+            } else {
+                objective
+                    .fit_at_fixed_rho(init_rho_flat.view())
+                    .map_err(gam::model_types::EstimationError::RemlOptimizationFailed)
+            };
+            (objective, run_result)
+        })?;
     run_result.map_err(estimation_error_to_pyerr)?;
     // Posterior shape uncertainty: per-atom φ-scaled decoder covariance and
     // ambient bands, read off the converged joint-Hessian Schur factor at the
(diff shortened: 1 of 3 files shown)


SauersML:
## Verdict: NEEDS-WORK — do not merge as-is (default flip regresses the production default onto a known-hanging path)

The diagnostics half of this PR is good and additive: `StructuredResidualPassDiagnostic`, `structured_residual_pass_diagnostics_dict`, the payload round-trip, and the `ManifoldSAE.structured_residual_diagnostics` field are clean, in-scope, and make the whitening effect observable exactly as the issue asks. No objection to that code.

The blocker is the **default flip `structured_residual_passes = 0 -> 1`** in three places (`latent_basis_and_sae_ffi.rs:2328`, `latent_basis_and_sae_ffi_tail.rs`, `gamfit/_sae_manifold.py:2440`). That change makes the whitened->structured alternation run **by default** for every Euclidean-provenance fit — and per the issue's own owner comment (2026-07-02):

> "Remaining as scoped here: production defaults still run iid; engaging WhitenedStructured by default needs the outer rho-search width fix first (**#2080**) — at p~96 the outer loop hangs regardless of inner-solve health."

**#2080 is still OPEN.** So this PR enables-by-default the precise path the issue author said must wait for #2080. Codex's own self-assessment confirms the consequence:

> "a full run of the structured-residual pytest was interrupted after a long-running second fit consistent with an existing outer rho-search hang" ... "SOLVES THE ISSUE? PARTIAL ... I did not prove the #1784 Qwen-L17 acceptance condition."

Contract failures:
1. **Item 1 (root cause / no regression):** flipping the production default onto a path the maintainer documented as hanging at realistic p (p~96) is a regression of the default fit behavior, not a completed fix. The gate is `if structured_passes > 0 && metric_provenance == "Euclidean"`, and Euclidean is the default provenance, so the default `sae_manifold_fit(...)` now enters the hanging outer rho-search.
2. **Item 3 (real regression test):** the only test cited is `test_structured_residual_passes_negative_raises` — an input-validation guard, not a fails-before/passes-after test that asserts the whitening mechanism (the #1784 in-sample EV flip, per-atom lambda-hat drop, no RemlConvergenceError at K=128). The falsifiable acceptance in the issue is explicitly **unproven**.

**What the real fix needs (pick one):**
- Land the #2080 outer rho-search width fix first, then flip the default and add a real regression test that exercises the #1784 acceptance (EV flip + lambda-hat drop at matched K, terminating within the outer loop); **or**
- Keep `structured_residual_passes` default `0` for now and ship the diagnostics/plumbing alone (still valuable, still `Refs #2021` rather than `Fixes #2021`, since acceptance is not met).

Either way the PR should not carry `Fixes #2021` while the falsifiable acceptance is unmet and the default path hangs.

— panel review, seat 4/4

SauersML:
## NEEDS-WORK — flips a production default onto a path the issue itself gates behind #2080, with no acceptance evidence

This PR does two things: (a) adds a `StructuredResidualPassDiagnostic` and round-trips it into the Python payload — that part is clean and in-scope; and (b) flips the production default `structured_residual_passes` from `0` to `1` in three call sites (native `sae_structured_residual_model`, `sae_build_atom_plans` in `..._ffi_tail.rs`, and Python `sae_manifold_fit`). Part (b) is a merge-blocker.

**1. The change directly contradicts the issue's own scoping (root-cause / acceptance-contract failure).** #2021's status comment (2026-07-02) states verbatim: *"Remaining as scoped here: production defaults still run iid; engaging WhitenedStructured by default **needs the outer rho-search width fix first (#2080)** — at p≈96 the outer loop hangs regardless of inner-solve health (REML converged in 61s at p=16 on real data, TIMEOUT at p=96 synthetic)."* #2080 is still **OPEN** ("Outer REML rho-search hangs at wide p (p≈96)"). This PR enables the whitened path by default *without* the stated prerequisite, i.e. it makes the default SAE fit take the known-hanging outer rho-search path at realistic `p`. Turning a production default onto a path known to hang at production problem sizes is a regression, not a fix.

**2. Codex's own evidence confirms the hang is unresolved.** The self-assessment marks the result **PARTIAL** and reports that the full `tests/test_sae_manifold_structured_residual_passes.py` run *"made no visible progress for an extended period; I interrupted it ... consistent with the issue comment's warning that structured/outer rho-search hangs remain."* So by the author's own admission the new default path stalls, and the only green test is `test_structured_residual_passes_negative_raises` — a kwarg-validation test that does not exercise the whitening/refit mechanism at all.

**3. No regression test asserting real behavior (contract item 3).** #2021's acceptance is falsifiable and specific: on the #1784 Qwen-L17 repro, in-sample manifold EV ≥ linear EV at matched K flips true, per-atom λ̂ drops, and no `RemlConvergenceError` at K=128. None of this is demonstrated or pinned by a test. There is no fails-before/passes-after guard for the over-smoothing fixed point the issue is about. Flipping a default without a test that proves the default now does the right thing is exactly the change that should not land.

**4. `Fixes #2021` would wrongly close the issue.** #2021 is a substantive research task with an unmet, measurable acceptance condition and an explicit dependency on #2080. Even the diagnostics-only portion does not satisfy it. This should be `Refs #2021`, not `Fixes`.

### What would make this mergeable
- Keep the diagnostics plumbing (`StructuredResidualPassDiagnostic` + Python round-trip) — that's good and can land on its own.
- Restore the production default to `structured_residual_passes = 0` (opt-in) in all three sites until #2080 lands, so the shipped default does not select a hanging path. The PR's own docstring even advertises "with the default `0` the fit is bit-identical to the pre-existing behavior" — honor that.
- Gate the default-on flip on (i) #2080 resolved and (ii) a real regression test exercising the #1784 acceptance (EV inversion flips, λ̂ drops, no RemlConvergenceError at K=128) — fails-before/passes-after.

I did not merge and did not edit-to-merge: reverting the thesis of the PR (the default flip) is not a small in-scope fix, it removes the change's stated purpose, so the right call is to send it back.

— panel review, seat 2/4

SauersML:
## Verdict: NEEDS-WORK (do not merge yet)

The wiring in this PR is conceptually the right seam — after the primary iid fit it builds a
`StructuredResidualModel`, damps it into a row metric, `set_row_metric`, and re-runs the outer REML
to genuinely engage `WhitenedStructured`. That part is not a fake. But flipping the **production
default** `structured_residual_passes 0 -> 1` right now fails the acceptance contract on three points:

**1. Premature: the default now triggers the #2080 hang on realistic sizes (root-cause ordering).**
The issue's own 2026-07-02 status comment states plainly: *"engaging WhitenedStructured by default
needs the outer rho-search width fix first (#2080) — at p~96 the outer loop hangs regardless of
inner-solve health."* Each structured pass here runs a full `OuterProblem::run(... "SAE manifold
(structured)")` — i.e. exactly the outer REML rho-search that #2080 says hangs at wide p. **#2080 is
still OPEN.** So this PR ships a default that runs an extra outer REML pass on every shard-free
(`Euclidean`) fit and can hang at p~96 — the prerequisite the author scoped as "first" has not landed.

**2. The engagement is unverified because it hangs.** A regression test that would actually prove
engagement already exists — `tests/test_sae_manifold_structured_residual_passes.py::test_structured_residual_passes_forwarded_changes_fit`.
codex's own PR body reports it could not run the file: *"the second fit made no visible progress ...
I interrupted it"* — the #2080 hang. So the one test that asserts the whitening likelihood changes
the fit was never observed to pass. The only green test cited (`..._negative_raises`) is an input
validator and exercises none of the engagement. No test asserts the actual defect — the #1892/#1784
over-smoothing fixed point — is broken. codex self-assesses "PARTIAL" and admits it "did not prove
the #1784 Qwen-L17 acceptance condition."

**3. `Fixes #2021` would wrongly close a still-open epic.** #2021's acceptance is the over-smoothing
fixed point being broken (tied to #1784/#1892), not merely making the default non-iid. With the
engagement unproven and the prerequisite (#2080) open, this should be `Refs #2021`, not `Fixes`.

The diagnostics plumbing (`StructuredResidualPassDiagnostic` -> payload -> `ManifoldSAE`) and the
Python round-trip are clean and in-scope; the whitespace-only reflows are harmless. If #2080 lands
and `test_structured_residual_passes_forwarded_changes_fit` (plus a fixed-point/over-smoothing
assertion) is shown green with the default at 1, this becomes mergeable. Until then the default flip
is the very step the author gated on #2080.

— panel review, seat 1/4

SauersML:
## Verdict: NEEDS-WORK (do not merge) — seat 3/4 (test-integrity)

I was asked to verify the PR's own cloud evidence that **a test proves the WhitenedStructured likelihood is now engaged**. It does not, and the change also flips a production default onto a path the PR's own evidence shows still hangs.

### 1. No regression test proves engagement — contract item 3 fails
The diff touches zero test files (`ffi.rs`, `ffi_tail.rs`, `_sae_manifold.py` only). The one green test cited in the PR body — `test_structured_residual_passes_negative_raises` — asserts a `ValueError` on a negative kwarg. It exercises the *validator*, not the whitening seam. There is no fails-before/passes-after test asserting that `metric_provenance` flips to `WhitenedStructured`, that `structured_residual_diagnostics` is populated, or that the #1784 acceptance (in-sample manifold EV ≥ linear EV at matched K) now holds. The PR's own self-assessment concedes: *"I did not prove the #1784 acceptance condition."* The falsifiable acceptance criterion the issue defines is untested.

### 2. The one test that DOES exercise the structured path hangs — in the PR's own evidence
`test_structured_residual_passes_forwarded_changes_fit` (already in-tree) runs `structured_residual_passes=2, promote_from_residual=True` on a tiny n=200/p=6 planted circle. The PR body reports this second fit *"made no visible progress for an extended period"* and was interrupted. That is the very code path this PR now enables **by default** for every `sae_manifold_fit` caller. Enabling a default that the submitter could not get to terminate — even at p=6 — is a regression on `main`, not a fix.

### 3. The change jumps the maintainer's stated prerequisite
The issue owner's 2026-07-02 status comment is explicit: *"production defaults still run iid; engaging WhitenedStructured by default needs the outer rho-search width fix first (#2080) — at p≈96 the outer loop hangs regardless of inner-solve health."* This PR flips `structured_residual_passes 0 → 1` in all three entry points (`sae_structured_residual_model`, `sae_build_atom_plans`, `sae_manifold_fit`) without #2080 landing. The prerequisite the author named is unmet.

### 4. `Fixes #2021` would wrongly close the epic
Acceptance is admittedly PARTIAL and unproven. Use `Refs #2021` so the issue stays open until the #1784 EV flip is demonstrated.

### What the real fix needs
- Land/confirm #2080 (outer rho-search width) so the structured default terminates; only then flip the default.
- Add a test that asserts engagement directly: after a default fit, `metric_provenance == "WhitenedStructured"` (or `structured_residual_diagnostics` non-empty with a selected factor rank), and that it terminates within a bounded iteration budget (not wall-clock).
- Demonstrate the #1784 acceptance (manifold EV ≥ linear EV at matched K) as the regression, per the issue's own falsifiable criterion.

The diagnostics plumbing (`StructuredResidualPassDiagnostic` → Python payload) is clean and worth keeping. The default flip is premature and untested.

— panel review, seat 3/4

SauersML:
## NEEDS-WORK — flips the production default onto a path the issue owner explicitly gated behind an unfixed blocker, and codex's own evidence shows it hangs

The diagnostics work (`StructuredResidualPassDiagnostic`, the per-pass λ̂/φ̂/rank/evidence round-trip into the Python payload) is clean and useful — no objection there. The problem is the central act of this PR: flipping `structured_residual_passes` default `0 → 1` in all three entry points (`latent_basis_and_sae_ffi.rs:2328`, `latent_basis_and_sae_ffi_tail.rs:246`, `gamfit/_sae_manifold.py:2440`), i.e. engaging `WhitenedStructured` by default.

### This is explicitly blocked
Issue #2021's own status comment from the owner (2026-07-02): *"production defaults still run iid; engaging WhitenedStructured by default needs the outer rho-search width fix first (#2080) — at p≈96 the outer loop hangs regardless of inner-solve health (REML converged in 61s at p=16 … TIMEOUT at p=96)."* I checked: **#2080 is still OPEN.** The precondition the owner set for making this the default has not been met.

### Codex's own evidence confirms the hang ships
From the PR body: *"a full run of the structured-residual pytest was interrupted after a long-running second fit consistent with an existing outer rho-search hang reported in the issue thread."* So codex knows the new default path hangs at production widths and shipped it as the default anyway. The only structured-path test it could pass was `test_structured_residual_passes_negative_raises` (an input-validation test that never engages the whitened fit). No test exercises a *successful* default structured fit.

### Contract failures
- **Item 1 (root cause):** turning on a path with a known outer-rho-search hang is the opposite of a fix; #2080 is the actual mechanism and it's unaddressed.
- **Item 3 (relevant tests pass):** no test demonstrates the new default completing; the one long test hangs.
- Making a research feature that hangs the opt-**out** default also runs against the SPEC's opt-in-to-extra-modeling posture and the issue's framing of this as *"An explicit, typed opt-in."*

### Path to mergeable
Split this: land the diagnostics (valuable, harmless) with the default kept at `0`. Defer the default flip to `1` until #2080 lands and a test shows a default structured fit converging at production width (the #1784 Qwen-L17 acceptance the self-assessment admits was not proven). Also `Fixes #2021` → `Refs #2021` (self-assessment says PARTIAL; #1784 acceptance unmet).

PR #2216, opened 2026-07-07
#2128: Gamma-log observed curvature: closed-form weights to avoid overflow/cancellation

### LINKED ISSUE #2128: Gamma GAM with a penalized smooth (family=gamma, y ~ s(x)) fails on clean data: every REML seed rejected_by_domain ('objective returned a non-finite cost'); parametric/Gaussian/low-variance fits all succeed
## Summary

A Gamma GAM with a **penalized smooth** (`family="gamma"`, `y ~ s(x)`) fails to fit clean, well-specified Gamma data. The outer REML rejects **every** candidate seed with *"objective returned a non-finite cost"*, and the fit raises `IntegrationError` before the solver even starts — even though the response is drawn exactly from the model. The **parametric** Gamma fit on the identical data, a **Gaussian** smooth on the same `y`, and **lower-variance** (shape=5) Gamma data all fit fine, so the model is identifiable and the data are ordinary; the defect is specific to the Gamma + penalized-smooth REML path.

## Committed failing test

- `tests/bug_hunt_gamma_smooth_reml_nonfinite_cost_rejects_all_seeds_test.py`
- Commit `404aa319028b5a44fc1af22af03e8fd8a0610224` on `main`.

`test_gamma_smooth_fits_clean_data_and_recovers_mean[6]` and `[7]` fail; the parametric control `test_gamma_parametric_control_fits_same_data` passes.

## Reproduce

```python
import numpy as np, gamfit
rng = np.random.default_rng(6)
n = 200
x = rng.uniform(0, 1, n)
mu = np.exp(0.2 + 1.5 * x)          # log mean, range ~[1.2, 5.5]
y = rng.gamma(2.0, mu / 2.0)         # Gamma(shape=2, mean=mu), CV = 1/sqrt(2) ~ 0.71

gamfit.fit({"x": x, "y": y}, "y ~ s(x)", family="gamma")   # -> IntegrationError

# All of these succeed on the SAME data / family:
gamfit.fit({"x": x, "y": y},        "y ~ x",    family="gamma")     # parametric OK
gamfit.fit({"x": x, "y": np.log(y)},"y ~ s(x)", family="gaussian")  # Gaussian smooth OK
# and lower-variance Gamma (shape=5) with y ~ s(x) fits OK too.
```

Deterministic: 3/3 `IntegrationError` on repeat. Fails for many seeds/sizes (n=200 seeds 6,7,8; n=500 seeds 6,7; n=1500 seeds 6,7 — a minority of seeds happen to slip through, e.g. n=500 seed 8), and for both the default and explicit `link="log"`. Higher shape (shape=5, lower variance) and the parametric RHS reliably succeed.

## Observed vs expected

**Expected:** a Gamma GAM fits textbook Gamma data trivially (this is the canonical Gamma-GAM use case; `mgcv::gam(y ~ s(x), family=Gamma(link=log))` / `glm(family=Gamma)` handle it without issue) and recovers `mu(x)`.

**Observed** — full error:

```
IntegrationError: REML smoothing optimization failed to converge:
  no candidate seeds passed outer startup validation (standard REML):
  generated=7, screened=7, exact_validated=7, solver_started=0
  rejection breakdown: rejected_by_kkt=0, rejected_by_domain=7,
    rejected_by_objective=0, rejected_by_budget=0, rejected_other=0 (total=7)
  per-seed reasons:
    seed 0..6 (validation): ... outer eval failed: objective returned a
      non-finite cost
```

with the inner solver logging, for every seed:

```
P-IRLS could not certify a valid minimum: max iterations reached
  (gradient norm 7.561e-5, iter 300)
P-IRLS could not certify a valid minimum: max iterations reached
  (gradient norm 1.133e1,  iter 300)
```

Note the inner gradient norm is sometimes as small as ~1e-6 (essentially converged) yet still reported as "could not certify a valid minimum: max iterations reached", and other times blows up to ~1e1 — the inner P-IRLS is not settling for the Gamma smooth, so the outer objective is non-finite and the seed is discarded.

## Best read of why (files read, no patch)

- The inner failure surfaces at `crates/gam-solve/src/reml/gradient_hessian.rs:6912` — *"P-IRLS could not certify a valid minimum: {kind} (gradient norm …, iter …)"* on the `max iterations reached` path. When the returned partial state is non-finite, the standard-REML outer startup screen reports *"objective returned a non-finite cost"* and counts the seed under `rejected_by_domain` (the aggregation string is built in the standard-REML seed-startup path; the generic "non-finite cost" check also appears at `crates/gam-models/src/fit_orchestration/fit.rs:1658` and `crates/gam-custom-family/src/fit.rs:263`).
- The trigger is Gamma-family-specific: Gaussian and Poisson smooths on comparable data fit, and the parametric Gamma fit fits, so the culprit is the Gamma **inner P-IRLS working-weights / REML objective** (deviance + scale profiling for the smoothing criterion) rather than the outer machinery. The Gamma scale / deviance handling was reworked recently (commit `c94dd0a66`, "report unscaled Gamma deviance"), which makes the Gamma REML objective the natural suspect for the non-finite cost / non-convergence.

A robust Gamma-smooth path (a step-halving / trust-region inner P-IRLS that converges on ordinary Gamma data, or a REML objective that stays finite at the screening seeds) should let this fit.

Related: #2127 (also a sphere-smooth invariance bug found in the same run) — unrelated subsystem, cross-linked only as a sibling of this bug-hunt run.

SauersML:
### Motivation
- The generic noncanonical observed-information formula for Gamma + log link forms large intermediate `V²`/`V³` and subtracts two 1/φ-scale terms, which catastrophically cancels or overflows at large trial `η`, producing non-finite observed weights and destabilizing the inner P-IRLS Hessian so REML startup rejects all seeds. 
- Use the algebraically-equivalent closed-form for the Gamma-log observed curvature to eliminate cancellation/overflow and restore stable inner solves.

### Description
- Add a closed-form Gamma/log observed-information routine `observed_weight_gamma_log(y, mu, phi, pw)` that returns `w = pw * y / (phi * mu)`, `c = -w`, `d = w` to avoid cancellation/overflow (file: `crates/gam-solve/src/pirls/curvature.rs`).
- Route `(WeightFamily::Gamma, WeightLink::Log)` through the new closed-form in `observed_weight_dispatch` instead of falling back to the generic noncanonical path (file: `crates/gam-solve/src/pirls/curvature.rs`).
- Add a regression unit test `gamma_log_observed_curvature_dispatch_avoids_generic_overflow` that demonstrates the generic path becomes non-finite at an extreme but representable `η` while the dispatched closed-form is finite and exact (file: `crates/gam-solve/src/pirls/tests.rs`).

### Testing
- Built the workspace and checked the crate: `./build.sh` and `./build.sh check -p gam-solve --lib` both completed successfully (`exit 0`).
- Ran the targeted unit tests: `./build.sh test -p gam-solve gamma_log_observed_curvature -- --nocapture` — the two new/related tests passed (`2 passed; 0 failed`).
- Verified the regression that exercised the original issue: `./build.sh test -p gam --test regressions gamma_smooth_highvar_fits_with_finite_coefficients_2128 -- --nocapture` — the regression passed (`1 passed; 0 failed`).

---
Fixes #2128.

pirls/curvature.rs", "line_range_start": 770, "line_range_end": 789}, {"content_type": "text", "text": "\n* Routed `(WeightFamily::Gamma, WeightLink::Log)` through that closed-form path in the observed-weight dispatch. "}, {"content_type": "repo_file_citation", "path": "crates/gam-solve/src/pirls/curvature.rs", "line_range_start": 867, "line_range_end": 876}, {"content_type": "text", "text": "\n* Added a root-cause regression unit test showing the generic Gamma-log formula becomes non-finite at an extreme but representable `\u03b7`, while dispatch stays finite and matches the closed form. "}, {"content_type": "repo_file_citation", "path": "crates/gam-solve/src/pirls/tests.rs", "line_range_start": 1840, "line_range_end": 1887}, {"content_type": "text", "text": "\n\n### SPEC COMPLIANCE\n\n* Added no autodiff or finite differences in production.\n* Added no grid search, GCV, wall-clock solver budget, timeout, convergence fabrication, KKT fabrication, skipped tests, ignored tests, or penalty stuffing.\n* The new constants in the test are deterministic test data, not solver knobs; the production change is a derived analytic closed form.\n\n### TEST EVIDENCE\n\n* \u2705 `./build.sh`  \n  Output: `exit 0`, finished the workspace lib warm build successfully.\n\n* \u2705 `./build.sh check -p gam-solve --lib`  \n  Output: `exit 0`; after final formatting it checked `gam-solve` with deny-warnings successfully in 15s.\n\n* \u2705 `./build.sh test -p gam --test regressions --no-run`  \n  Output: `exit 0`; built the aggregate regression test binary successfully: `Executable tests/regressions/main.rs`.\n\n* \u2705 `./build.sh test -p gam-solve gamma_log_observed_curvature -- --nocapture`  \n  Output: `exit 0`; `2 passed; 0 failed`, including:\n  * `gamma_log_observed_curvature_dispatch_avoids_generic_overflow`\n  * `gamma_log_observed_curvature_matches_shape_one_closed_form`\n\n* \u2705 `./build.sh test -p gam --test regressions gamma_smooth_highvar_fits_with_finite_coefficients_2128 -- --nocapture`  \n  Output: `exit 0`; `1 passed; 0 failed`; the issue regression `gamma_smooth_highvar_fits_with_finite_coefficients_2128` now passes.\n\n* \u26a0\ufe0f `./build.sh test --test gamma_smooth_reml_highvar_rejects_all_seeds_2128 --no-run`  \n  Output: `exit 101`; this was the wrong integration target name for this workspace. The valid target is the aggregate `regressions` binary, which was built and run successfully above.\n\n* \u26a0\ufe0f `./build.sh fmt --check`  \n  Output: `exit 75`; wrapper reported transient failure after retries. I then ran formatting successfully with `./build.sh fmt`, reverted unrelated rustfmt changes, and re-ran the relevant compile/tests successfully.\n\n### CODE DIFF\n\nCommitted on the current branch:\n\n* Commit: `4e7c8e9 Fix gamma log observed curvature overflow`\n\nSummary of the diff:\n\n```diff\n+pub fn observed_weight_gamma_log(y: f64, mu: f64, phi: f64, pw: f64) -> (f64, f64, f64) {\n+    let w = pw * y / (phi * mu);\n+    (w, -w, w)\n+}\n...\n+        (WeightFamily::Gamma, WeightLink::Log) => {\n+            observed_weight_gamma_log(y, mu, phi, prior_weight)\n+        }\n```\n\nand a regression test that verifies the old generic Gamma-log formula becomes non-finite while dispatch remains finite and equal to the closed form. "}, {"content_type": "repo_file_citation", "path": "crates/gam-solve/src/pirls/tests.rs", "line_range_start": 1840, "line_range_end": 1887}, {"content_type": "text", "text": "\n\n### SELF-ASSESSMENT\n\n* **SOLVES THE ISSUE? SOLVED.** The issue\u2019s failing Gamma penalized-smooth regression now passes, and the fix targets the identified mechanism: non-finite / cancellation-prone Gamma-log observed curvature in the inner P-IRLS / REML startup path.\n* **NET EFFECT: NET-GOOD.** This removes an unstable generic computation for a family-link combination with a simple exact analytic form, improving numerical stability without changing solver convergence semantics or adding arbitrary knobs."}]
</details>

### DIFF
--- a/crates/gam-solve/src/pirls/curvature.rs
+++ b/crates/gam-solve/src/pirls/curvature.rs
@@ -767,6 +767,27 @@ pub fn observed_weight_gaussian_inverse(y: f64, eta: f64, phi: f64, pw: f64) -> 
     (w, c, d)
 }
 
+/// Gamma family with log link: `V(μ)=μ²`, `μ=exp(η)`.
+///
+/// For a Gamma exponential-dispersion model the negative log-likelihood
+/// second derivative with respect to `η` is exactly `y / (φ μ)`.  The generic
+/// observed-information formula computes the same value as
+/// `W_Fisher - (y - μ)·B`; with `V=μ²` and log-link derivatives this subtracts
+/// two `1/φ`-scale terms to leave the small positive `y/(φμ)` tail, and its
+/// intermediate `V²`/`V³` products overflow for large trial `η`.  This
+/// closed form is algebraically identical but cancellation- and overflow-free.
+///
+/// ```text
+/// w_obs =  ω y / (φ μ)
+/// c_obs = -ω y / (φ μ)
+/// d_obs =  ω y / (φ μ)
+/// ```
+#[inline]
+pub fn observed_weight_gamma_log(y: f64, mu: f64, phi: f64, pw: f64) -> (f64, f64, f64) {
+    let w = pw * y / (phi * mu);
+    (w, -w, w)
+}
+
 #[inline]
 pub(crate) fn observed_weight_binomial_logit_from_jet(
     n_trials: f64,
@@ -849,6 +870,9 @@ pub fn observed_weight_dispatch(
         }
         (WeightFamily::Gaussian, WeightLink::Inverse) => {
             observed_weight_gaussian_inverse(y, eta, phi, prior_weight)
+        }
+        (WeightFamily::Gamma, WeightLink::Log) => {
+            observed_weight_gamma_log(y, mu, phi, prior_weight)
         }
         (WeightFamily::Binomial, WeightLink::Logit) => {
             observed_weight_binomial_logit_from_jet(1.0, jet, prior_weight)
--- a/crates/gam-solve/src/pirls/tests.rs
+++ b/crates/gam-solve/src/pirls/tests.rs
@@ -14,12 +14,14 @@ mod tests {
     use super::reweight::madsen_lm_accept_factor;
     use super::{
         LinearInequalityConstraints, PenaltyConfig, PirlsConfig, PirlsLinearSolvePath,
-        PirlsProblem, PirlsWorkspace, WorkingDerivativeBuffersMut, bernoulli_geometry_from_jet,
-        calculate_deviance, calculate_loglikelihood, calculate_loglikelihood_omitting_constants,
-        compute_constraint_kkt_diagnostics, compute_observed_hessian_curvature_arrays,
-        fit_model_for_fixed_rho, select_active_set_release, should_log_pirls_decision_summary,
-        should_use_sparse_native_pirls, solve_newton_directionwith_linear_constraints,
-        solve_newton_directionwith_lower_bounds, tweedie_log_weight_mu_power, update_glmvectors,
+        PirlsProblem, PirlsWorkspace, WeightFamily, WeightLink, WorkingDerivativeBuffersMut,
+        bernoulli_geometry_from_jet, calculate_deviance, calculate_loglikelihood,
+        calculate_loglikelihood_omitting_constants, compute_constraint_kkt_diagnostics,
+        compute_observed_hessian_curvature_arrays, fit_model_for_fixed_rho,
+        observed_weight_dispatch, observed_weight_noncanonical, select_active_set_release,
+        should_log_pirls_decision_summary, should_use_sparse_native_pirls,
+        solve_newton_directionwith_linear_constraints, solve_newton_directionwith_lower_bounds,
+        tweedie_log_weight_mu_power, update_glmvectors, variance_jet_for_weight_family,
         write_gamma_log_working_state, write_negative_binomial_log_working_state,
         write_poisson_log_working_state, write_tweedie_log_working_state,
     };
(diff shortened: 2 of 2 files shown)


SauersML:
**MERGE — correct owner-endorsed closed-form; root-cause fix, no forbidden pattern, real regression.**

Verified the math: for Gamma with log link (μ = e^η), the per-observation negative log-likelihood second derivative w.r.t. η is exactly `y/(φμ)`:
- ℓ = (1/φ)(−y/μ − log μ); dℓ/dη = (1/φ)(y/μ − 1); d²ℓ/dη² = −y/(φμ), so observed information = `+y/(φμ)`.
- With prior weight: `w = pw·y/(φμ)`, `c = ∂w/∂η = −w`, `d = w`. The `(w, −w, w)` tuple is correct.

This is algebraically identical to the generic `W_Fisher − (y−μ)·B` path but subtracts two ~1/φ terms there, which cancels catastrophically when μ≫y and overflows once trial η makes μ² blow up — exactly the non-finite observed Hessian the issue reports. The issue owner independently derived the same closed form (`w_obs = shape·y/μ`), so this is the prescribed root-cause fix, not a symptom bandaid or a clamp.

Contract: root cause ✓; no FD/autodiff/grid/GCV/timeout/magic-knob ✓; new regression `gamma_log_observed_curvature_dispatch_avoids_generic_overflow` shows the generic path non-finite at η=400 while dispatch stays finite+exact ✓; the on-main `gamma_smooth_highvar…_2128` regression goes green ✓; diff is purely additive (no test weakened/deleted). Merging.

— panel review, seat 1/1

PR #2272, opened 2026-07-12
fix(gpu #2259): continue a torch-mapped CUDA stack instead of refusing the GPU

### LINKED ISSUE #2259: Rust GPU lane cannot obtain a CUDA context inside a torch process (dual-CUDA-stack interaction; curved fits forced to CPU)
**TL;DR:** Inside a torch process, gam's Rust runtime cannot obtain a CUDA context: it fails with **"no CUDA context for ordinal 0"**. Steering its loader at the CUDA-13 compat driver instead breaks torch itself (**error 803**). The same Rust runtime works standalone — the multi-GPU per-atom Gram fan-out benchmark ran fine (1.93× on 2 A100s / 96% efficiency, 3.17× on 4 / 79%) — so this is a dual-CUDA-stack interaction, and gam's own loader code documents the hazard.

*AI-filed (Silico agent, run by Scott); details from recorded overnight-experiment artifacts.*

**Roadmap:** L2 in the five-experiment synthesis ledger (experiment #6).

## Consequence

Every curved fit in an overnight curved-vs-linear performance comparison ran **CPU-only** (~1800× slower than the SVD baseline at matched K), because the fits had to run inside a torch process (activations harvested and held in torch; torch lane used for the fits).

## Environment (from the recorded diagnostics)

- System CUDA-13.0 stack (`libcudart.so.13.0.88`, `libcublas.so.13.0.2.14`, compat driver `libcuda.so.580.82.07` at `/usr/local/cuda-13.0/compat/`), plus torch's packaged `nvidia/` wheels in the venv — the classic two-stack layout.
- Full library map, conflicts scan, and per-stack completeness in the evidence file below.

## Suggested directions

1. Fix or document the dual-CUDA-stack context acquisition (the loader comment already names the hazard — promoting it to handled behavior would resolve this).
2. At minimum, detect the torch-process case and emit an actionable error at load time instead of failing at fit time.

## Evidence (recorded artifacts, cluster paths)

- `/mnt/data/artifacts/silico/experiments/_flat/exp_01kx7qwdseen68mc2jhqaqvsan/results/cuda_diagnostics.json` (library map + stack analysis; prose twin `cuda_diagnostics.txt` in the same dir)
- `/mnt/data/artifacts/silico/experiments/_flat/exp_01kx7qwdseen68mc2jhqaqvsan/results/multigpu_raw.txt` (the standalone multi-GPU run that works)

murdarch:
Fixes #2259 — gam's Rust CUDA lane can't obtain a runtime inside a torch process, so every curved fit silently falls back to CPU.

## Root cause

When gam runs in a process that already owns the CUDA userspace stack (PyTorch's `nvidia-*-cu12` wheels), `ldconfig` can pull a **system** `libcudart` onto the default loader path *alongside* torch's pip stack. The libraries mapped in `/proc/self/maps` then span two roots — a system `libcudart` plus torch's `site-packages/nvidia/*` compute libs — so `complete_mapped_cuda_stack` finds **"no single complete stack"**, `GpuRuntime::probe` reports **"no CUDA runtime available"**, and the GPU is declined.

Notably this is **not** a CUDA version mismatch — it reproduces on a version-aligned box (system 12.6 + torch cu126). The split across roots alone is sufficient.

## Fix

A split-stack continuation in `complete_mapped_cuda_stack`: continue a candidate stack **iff (a)** its whole stack is already mapped in-process **and (b)** every other mapped library is a redundant duplicate of a *non-compute* component (`cudart` / `nvJitLink`) it already provides. The handle-based compute libraries (`cuBLAS` / `cuSOLVER` / `cuSPARSE`) are **never** split across roots, so the double-free / `NOT_INITIALIZED` hazard the mapped-stack continuation exists to avoid cannot occur. Genuinely mixed roots (a compute library from a different root than cudart, with neither root fully mapped) still fall through to the conservative refusal — the existing `mapped_mixed_cuda_stacks_are_refused` test is unchanged and still passes.

## Reproduction & verification (RTX 3080 Ti, torch 2.13.0+cu126)

`sparse_dictionary_fit(X, K, score_mode="required")` inside a torch process:

| | before | after |
|---|---|---|
| inside torch | `device DECLINED — no CUDA runtime available` | `device ENGAGED: device=2 cpu=0`, **ev=0.981850** |
| standalone (no torch) | ev=0.981850 | unchanged |

The in-torch result is byte-identical to the standalone run (`cg_relative_residual=1.488e-8`). Full `gam-gpu` suite: **80/80** (added regression test `mapped_pytorch_stack_with_redundant_system_cudart_is_continued`).

*Written with Claude Code; reproduced and verified end-to-end on real hardware.*

### DIFF
--- a/crates/gam-gpu/src/driver.rs
+++ b/crates/gam-gpu/src/driver.rs
@@ -202,8 +202,36 @@ fn cuda_userspace_preload_paths() -> Result<Vec<PathBuf>, String> {
     Ok(Vec::new())
 }
 
+/// The CUDA component a userspace library path belongs to, taken from its
+/// SONAME stem (`libcublas.so.12` -> `cublas`, `libnvJitLink.so.12` ->
+/// `nvJitLink`). Used to reason about which mapped libraries are the same
+/// component sourced from different roots.
+#[cfg(target_os = "linux")]
+fn cuda_library_component(path: &Path) -> Option<String> {
+    let name = path.file_name()?.to_str()?;
+    let stem = name.strip_prefix("lib")?.split(".so").next()?;
+    if stem.is_empty() {
+        None
+    } else {
+        Some(stem.to_string())
+    }
+}
+
+/// The CUDA *compute* libraries. These are handle-based and share context /
+/// workspace state with whatever copy the process already initialised, so a
+/// second copy must never be mapped from a different root — that split is the
+/// double-free / `NOT_INITIALIZED` hazard the mapped-stack continuation exists
+/// to avoid. `cudart` / `nvJitLink` are runtime/driver-adjacent and tolerate a
+/// redundant duplicate (dlopen-by-SONAME binds one deterministically).
+#[cfg(target_os = "linux")]
+fn is_cuda_compute_component(component: &str) -> bool {
+    matches!(component, "cublas" | "cublasLt" | "cusolver" | "cusparse")
+}
+
 #[cfg(target_os = "linux")]
 fn complete_mapped_cuda_stack(mapped: &[PathBuf]) -> Result<Vec<PathBuf>, String> {
+    let canonical = |p: &Path| p.canonicalize().unwrap_or_else(|_| p.to_path_buf());
+
     let mut candidates = Vec::new();
     for path in mapped {
         if let Some(root) = nvidia_package_root_for_library(path)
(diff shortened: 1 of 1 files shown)


SauersML:
Thanks!

Issue #2258, opened 2026-07-11
Closed-form sae_manifold_fit cannot converge from a cold start (fails even on a synthetic well-conditioned circle); only external warm starts produce fits


SauersML:
**TL;DR:** The closed-form `sae_manifold_fit` **cannot converge from a cold start at all on this engine build — regardless of conditioning.** On real Qwen3-8B activations, 0/39 grid configurations converged. But the sharper fact: cold-start fits fail even on an **easy, well-conditioned synthetic circle** projected into a clean 6-dim class-mean chart — all three fit modes raise convergence errors. The only working recipe in a five-experiment study was warm-starting the coordinates from an external, fit-free readout. **Today a fit is only obtainable if you already have the answer to seed it** — the fitting engine cannot recover a manifold it isn't already handed.

*AI-filed (Silico agent, run by Scott); all numbers from recorded overnight-experiment artifacts.*

**Roadmap:** L1 in the five-experiment synthesis ledger (experiment #6, `engineering_ledger.json`) — bucket "no working fast curved-fit path on real activations". This is the single most important engineering finding of the study.

## Cold starts fail even on synthetic, well-conditioned data

A dedicated probe fit an easy synthetic circle projected into the clean 6-dimensional day-signal chart (well-conditioned by construction):

- All three modes — `full`, `fixed_rho` (outer search disabled), `fixed_rho_no_struct` (structure search also disabled) — raise convergence errors.
- The `fixed_rho_no_struct` failure comes from a **structured-residual pass that runs regardless of the bypass toggle**.
- Combined with the earlier `n_iter=1000` failure, cold starts are dead on this engine build **no matter the conditioning**. The real-activation grid below is therefore a symptom, not the disease: this is not merely an ill-conditioning interaction.

## The real-activation grid (the original symptom)

[...]

PR #2326, opened 2026-07-15
fix(python #2323): align sae_manifold_fit's sparsity_weight default with the documented and engine default

### LINKED ISSUE #2323: sae_manifold_fit's own default sparsity_weight=0.0 is rejected by native validation — every default call and 10 shipped examples fail instantly
**TL;DR:** `sae_manifold_fit`'s own default is rejected by its own validation. The Python facade declares `sparsity_weight: float = 0.0` (`gamfit/_sae_manifold.py`), and the native front door rejects it instantly: `sparsity_strength must be finite and positive; got 0` (`crates/gam-sae/src/manifold/fit_seed.rs`, the `<= 0.0` check). Every default-argument call fails before the solver runs, and **10 shipped examples call `sae_manifold_fit` without `sparsity_weight`** — all fail on import-and-run at current `main`.

*AI-filed (Silico agent, run by Scott); observed live on the exact-SHA 0.1.255 wheel (`352ee0f`), re-verified against source at `246a25c36`.*

## Repro

```python
import numpy as np, gamfit
t = rng.uniform(0, 2*np.pi, 200)
X = np.c_[np.cos(t), np.sin(t)] + 0.05*rng.standard_normal((200, 2))
gamfit.sae_manifold_fit(X, K=1, d_atom=1, atom_topology="circle")
# gamfit._rust.GamError: sparsity_strength must be finite and positive; got 0
```

## Affected shipped examples (no `sparsity_weight` kwarg at current main)

`chart_transport_l11_l23.py`, `compose32k_step3.py`, `compose_tiers.py`, `cyclic_circle_recovery_and_steering.py`, `ordered_beta_bernoulli_anneal_demo.py`, `s1_w6_joint_acceptance.py`, `sae_ev_vs_k_olmo.py`, `sae_manifold_demo.py`, `sae_manifold_nuclear_rank.py`, `sae_manifold_ordered_beta_bernoulli_topology_atoms.py`

(The docstring says "defaults … are owned by the native front door" — but the facade materializes `0.0` and passes it, so the native side never gets to choose.)

## Options (maintainer's call)

1. Give the facade a positive default (what value is a modeling decision);
2. Have the native validator accept `0` as "sparsity penalty disabled", if the solver supports that;
3. Make the facade require the argument explicitly — and update the 10 examples either way.

Whichever way, a test that calls `sae_manifold_fit` with facade defaults on a tiny synthetic input would have caught this: the current Python-surface tests all pass explicit sparsity.

SauersML:
Fixes #2323.

**One-line default change (`0.0` → `1.0`) plus a regression test.** The facade declared `sparsity_weight: float = 0.0` while:

- `docs/manifold-sae.md` documents the default as **`1.0`**,
- the engine's own `SaeManifoldFitSeed` default is **`sparsity_strength: 1.0`** (`crates/gam-sae/src/manifold/fit_seed.rs`),
- and the native front door rejects non-positive values outright — correctly, since the strength seeds the REML ρ vector as `ln(sparsity_strength)`, so `0` is not representable.

Result before this fix: every default-argument `sae_manifold_fit` call — including 10 shipped examples (`sae_manifold_demo.py`, `sae_ev_vs_k_olmo.py`, `cyclic_circle_recovery_and_steering.py`, `compose_tiers.py`, …) — died at input validation with `sparsity_strength must be finite and positive; got 0` without reaching the solver.

**Validation** (CPU node, provenance-verified 0.1.255 wheel with this branch's pure-Python change overlaid — the native module is untouched by this PR):

- BEFORE: default call reproduces `GamError: sparsity_strength must be finite and positive; got 0`.
- AFTER: `tests/sae_manifold_default_args_reach_solver_test.py` passes 2/2 in 141 s — the default call passes validation and reaches the solver, and `inspect.signature` agrees with the documented/engine default.

The test deliberately does **not** assert fit convergence (solver behavior is owned elsewhere, see #2258/#2322); it pins only that the facade's own defaults pass the facade's own front door.

*AI-authored (Silico agent, run by Scott), from the recorded #2323 repro.*

### DIFF
--- a/gamfit/_sae_manifold.py
+++ b/gamfit/_sae_manifold.py
@@ -164,7 +164,7 @@ def sae_manifold_fit(
     decoder_feature_sparsity_groups: Sequence[Sequence[int]] | None = None,
     n_iter: int = 50,
     *,
-    sparsity_weight: float = 0.0,
+    sparsity_weight: float = 1.0,
     coord_sparsity: str = "scad",
     scad_mcp_gamma: float | None = None,
     smoothness_weight: float = 1.0,
--- a/tests/sae_manifold_default_args_reach_solver_test.py
+++ b/tests/sae_manifold_default_args_reach_solver_test.py
@@ -1,0 +1,48 @@
+"""Facade defaults must pass the native front door's own validation (#2323).
+
+``sae_manifold_fit``'s keyword defaults are the documented public contract
+(``docs/manifold-sae.md`` lists ``sparsity_weight`` default ``1.0``), and the
+engine's own seed default is ``sparsity_strength: 1.0``. Before the fix, the
+facade materialized ``0.0`` and every default-argument call — including 10
+shipped examples — died at input validation with ``sparsity_strength must be
+finite and positive; got 0`` without ever reaching the solver.
+
+These tests do NOT assert the fit converges (solver behavior is owned
+elsewhere); they assert the failure mode, if any, is not the facade's own
+default failing validation.
+"""
+
+from __future__ import annotations
+
+import numpy as np
+import pytest
+
+gamfit = pytest.importorskip("gamfit")
+
+
+def _tiny_circle(n: int = 60, seed: int = 0) -> np.ndarray:
+    rng = np.random.default_rng(seed)
+    t = rng.uniform(0.0, 2.0 * np.pi, n)
+    return np.c_[np.cos(t), np.sin(t)] + 0.05 * rng.standard_normal((n, 2))
+
+
+def test_default_arguments_pass_input_validation() -> None:
+    x = _tiny_circle()
+    try:
+        gamfit.sae_manifold_fit(x, K=1, d_atom=1, atom_topology="circle", n_iter=2)
+    except Exception as error:  # noqa: BLE001 — asserting on the message either way
+        message = str(error)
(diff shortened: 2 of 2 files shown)


SauersML:
Closing: superseded by a better fix already on `main`.

#2323 is fixed on `main` by **0b45b0567** ("fix(#2323): let native SAE own sparsity default"), which resolves it in the opposite — and correct — direction:

- **This PR:** hardcodes `sparsity_weight: float = 1.0` in the Python facade, duplicating the engine default in a second place that can silently drift.
- **`main`:** `sparsity_weight: float | None = None`, forwarded as `sparsity_strength=(None if sparsity_weight is None else float(...))`, so `SaeManifoldFitSeed` remains the single source of truth for the default. This keeps Python a thin wrapper rather than a second owner of engine policy.

Merging this PR would regress that: it reverts the sentinel *and* adds `test_facade_default_matches_documented_and_engine_default`, which asserts `default == 1.0` and so directly contradicts `main`'s own `test_python_sparsity_default_is_owned_by_native_front_door` (asserts `default is None`). The two tests cannot both pass.

The validation bug this PR targets ("sparsity_strength must be finite and positive; got 0") is gone on `main` — `None` now delegates to the native neutral origin instead of materializing `0.0`.

Issue #2322, opened 2026-07-14
Regression at 352ee0f: arrow-Schur inner solve engages a GPU path that faults on driver mismatch in every configuration — no CPU fallback reachable from the manifold-SAE Python surface


SauersML:
**TL;DR:** At `352ee0f` (gamfit 0.1.255, exact-SHA wheel), any `sae_manifold_fit` large enough to engage the new QP/arrow-Schur inner solve dies in ~6 s with `DriverError(CUDA_ERROR_SYSTEM_DRIVER_MISMATCH)` — identically on a **CPU-only node**, on a **1×A100 standalone process**, and on a **1×A100 process with torch's CUDA context already bound**. There is no CPU fallback reachable from the manifold-SAE Python surface, so fits at the shipped example's own 635-row scale are now impossible on this cluster in every configuration. This is a regression relative to 0.1.253, which ran the same configs (slowly, >900 s — that's #2267) on CPU. Combined with the small-scale cold-start stall (#2258's failure signature, still reproducing at this SHA), **no cold fit can currently be minted at any tested scale on this cluster**: small fits die in the inner-solve stall, large fits die in the GPU path.

*AI-filed (Silico agent, run by Scott); all numbers from recorded experiment artifacts (verification runs on the exact-SHA 0.1.255 wheel, provenance-checked).*

## The measurement

Example-scale probe: 635 rows × 32 PCs (the shipped `sae_ev_vs_k_olmo.py` scale, real Qwen3-8B L20 activations), HEAD's own `assignment="ordered_beta_bernoulli"`, positive `sparsity_weight`, K=1/2 circle + K=1 euclidean, 900 s cap. Three runs, one per process type:

| Run | Node / process | Cells | Wall each | Result |
|---|---|---|---|---|
| 1 | CPU-only allocation | 3/3 | ~6 s | `CUDA_ERROR_SYSTEM_DRIVER_MISMATCH` |
| 2 | 1×A100, standalone | 3/3 | 5.5–6.6 s | identical |
| 3 | 1×A100, torch context created first | 3/3 | 9.0–10.7 s | identical |

Representative error (K=1 circle):

```
gamfit._rust.GamError: SaeManifoldTerm::run_joint_fit_arrow_schur: resident GPU frame
preparation failed: arrow-Schur: Schur complement Cholesky failed: resident SAE frame
build: GPU runtime resolution failed: DriverError(CUDA_ERROR_SYSTEM_DRIVER_MISMATCH,
[...]

PR #2327, opened 2026-07-15
feat(pyffi #2322): expose the process-global GPU policy to Python (configure_gpu_policy / gpu_policy / GAMFIT_GPU)

SauersML:
Addresses the operability half of #2322 (the CPU-fallback gap; the `Auto`-semantics question stays with the maintainers).

**What this adds (purely additive, 4 files + tests):**

- `gamfit.configure_gpu_policy("auto" | "off" | "required")` — a module-level pyfunction wrapping `gam::gpu::configure_global_policy` one-to-one, so the process-global policy the docs already describe becomes reachable from every Python surface, including the manifold-SAE entry points (`sae_manifold_fit`, `adjudicate_atom_shape`, crosscoder fits), which have no per-call GPU knob. The classic GAM path's `config={"gpu": ...}` sets the same slot, unchanged.
- `gamfit.gpu_policy()` — getter; reading never claims the first-writer-wins slot.
- First-writer-wins made loud: a reconfigure that loses to a *different* locked policy raises `RuntimeError` instead of silently running under the other policy; re-asserting the same policy is a no-op.
- `GAMFIT_GPU=off|auto|required` honored at import — the no-code-change control for cluster jobs; an unrecognized value fails import loudly (a typo silently running on the wrong backend would be worse).
- `docs/gpu_acceleration.md` updated with the three equivalent controls.

**Why:** at HEAD, any manifold-SAE fit that engages the arrow-Schur GPU lane (warm start supplied at any scale, or ≥635 rows cold) hard-fails with `DriverError(CUDA_ERROR_SYSTEM_DRIVER_MISMATCH)` on hosts whose images carry a mismatched `libcuda` — including CPU-only nodes — because under `Auto` a probe fault is deliberately an error, not absence, and nothing on the SAE Python surface can select `Off`. `GpuPolicy::Off` already takes the CPU branch at every arrow-Schur dispatch site; this PR only makes it reachable. Full three-config repro and source localization on #2322.

**Validation** (CPU-only allocation — the exact crashing configuration; wheel built by this repo's own `wheel-nightly.yml` dispatched on this branch, `RUNTIME_PROVENANCE.txt` commit-pinned):

- BEFORE (stock 0.1.255 wheel, default policy): 635-row-scale synthetic fit dies in **0.4 s** with the `DriverError(CUDA_ERROR_SYSTEM_DRIVER_MISMATCH)` resident-frame error.
- AFTER (this branch's wheel, `configure_gpu_policy("off")`): the same fit runs the CPU solver (real solver iterations, no GPU error of any kind; it is slow — that's #2258's separate story — and was cleanly bounded by the harness, never by a GPU fault).
- `tests/gpu_policy_python_surface_test.py`: subprocess-isolated coverage of default reporting, read-doesn't-lock, off round-trip, same-policy no-op, conflicting-reconfigure `RuntimeError`, invalid-value `ValueError`, env-var honor, env-var typo failing import, and the off-policy fit probe (GPU-error-free by construction; a solver timeout counts as the property shown, since the GPU fault kills in <10 s when it fires).

**Design note:** this deliberately does not touch `Auto`'s fault-vs-absence doctrine ("a present-but-broken driver never masquerades as absence"). Whether `CUDA_ERROR_SYSTEM_DRIVER_MISMATCH` specifically should be classified as typed absence is left as the open question on #2322.

*AI-authored (Silico agent, run by Scott); grounded in the recorded three-config repro (#2322) and validated as above.*

### DIFF
--- a/crates/gam-pyffi/src/manifold/geometry_ffi.rs
+++ b/crates/gam-pyffi/src/manifold/geometry_ffi.rs
@@ -4403,6 +4403,54 @@ fn set_log_level(level: &str) -> PyResult<()> {
     }
 }
 
+/// Set the process-wide GPU execution policy from Python.
+///
+/// Mirrors `gam::gpu::configure_global_policy` one-to-one: `"auto"` (the
+/// default) lets supported, large-enough paths dispatch to a probed CUDA
+/// device, `"off"` pins the process to the CPU kernels without ever probing
+/// the driver, and `"required"` turns an unsupported GPU route into a hard
+/// error. An unrecognized value raises `ValueError`.
+///
+/// The policy is process-global and first-writer-wins (concurrent fits cannot
+/// race policy changes), so call this before the first fit. A call that loses
+/// the race to a DIFFERENT already-locked policy raises `RuntimeError` instead
+/// of silently executing under the other policy; re-asserting the same policy
+/// is a no-op.
+///
+/// `"off"` is the escape hatch for nodes whose images carry a broken or
+/// mismatched `libcuda`: under `"auto"` a present-but-broken driver is a probe
+/// fault, never treated as absence (#2322), so GPU-eligible fits fail loudly;
+/// `"off"` never probes and takes the CPU branch at every dispatch site.
+#[pyfunction]
+fn configure_gpu_policy(policy: &str) -> PyResult<()> {
+    let Some(parsed) = gam::gpu::GpuPolicy::parse(policy) else {
+        return Err(py_value_error(format!(
+            "unrecognized GPU policy {policy:?}; expected one of auto|off|required"
+        )));
+    };
+    gam::gpu::configure_global_policy(parsed);
+    let effective = gam::gpu::global_policy();
+    if effective != parsed {
+        return Err(pyo3::exceptions::PyRuntimeError::new_err(format!(
+            "GPU policy is already locked to '{}' for this process (first writer wins); \
+             configure_gpu_policy({policy:?}) must be called before the first fit or \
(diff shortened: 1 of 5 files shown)


SauersML:
Closing: #2322 is already solved on `main`, in the opposite architecture.

`main` carries **5b77983f5** ("fix(#2322): make SAE GPU policy fit-owned") plus **baf19e512** (policy threaded through PyFFI). The Python surface already exists and is per-fit:

    gamfit/_sae_manifold.py:188    gpu: str = "auto",
    gamfit/_sae_manifold.py:261        gpu_policy=str(gpu),
    docs/manifold-sae.md:74   | `gpu` | `"auto"` | per-fit backend policy: `"auto"`, `"off"`, or `"required"`; `"off"` never probes CUDA |

So the capability this PR adds is already shipped — but this PR implements it as a **process-global, first-writer-wins** policy slot, which is precisely the design #2322 moved *away* from. A global slot reintroduces the cross-fit coupling that "fit-owned" exists to eliminate: two fits in one process can no longer disagree, and whichever runs first silently wins.

It also adds a `GAMFIT_GPU` environment-variable toggle applied at import time (`gamfit/__init__.py:97-99`). Environment-variable configuration is not used in this project — behavior is set explicitly at the call site, not by ambient process state.

The per-fit `gpu=` argument covers the stated use case (`gpu="off"` to never probe CUDA) without either drawback. If a no-code-change global override is genuinely wanted, that's worth reopening #2322 to discuss on its own terms rather than landing it as a second, conflicting policy owner.

PR #2328, opened 2026-07-15
test(gpu): assert typed policy decisions

SauersML:


SauersML:
Closing: already landed on `main`, byte-identical.

`main` carries this exact change as **70f060606** ("test(gpu): assert typed policy decisions"), same file, same 3-line diff:

    git diff origin/main b4dd02d2f -- tests/arrow_gpu/gpu/gpu_backend_status_and_policy_dispatch_are_consistent.rs   # empty

Merging would contribute nothing but a merge commit. The fix itself (dropping `.expect()` off the infallible `assert!` and moving it onto the `gpu::decide` result) is live on `main`.

Issue #2372, opened 2026-07-19
7 pre-existing red gam-terms tests at main (dictionary/matern/anchor/duchon/thin-plate-lazy/sphere-gpu) — double-baseline confirmed during the #2365 gate


SauersML:
Found during the #2365 gate (A10, full gam-terms suite at 02ffd4bb1, confirmed on TWO independent runs — a patched tree and a clean-main baseline with identical failure sets, so these are pre-existing main reds, not artifacts of any in-flight patch):

1. `dictionary::tests::planted_sparse_linear_dictionary_reaches_high_explained_variance`
2. `dictionary::tests::returned_state_is_the_certified_canonical_routing`
3. `term_builder::tests::default_matern_2d_seeds_resolving_length_scale_not_overscaled_diameter`
4. `term_builder::tests::one_sided_anchor_owns_level_without_sum_to_zero_constraint_1867`
5. `duchon_lazy_anisotropic_reparam_1818::lazy_anisotropic_duchon_uses_data_metric_radial_chart`
6. `basis::tests::test_build_thin_plate_basis_switches_to_lazy_design_for_large_blocks`
7. `basis::sphere_gpu::sphere_gpu_tests::sphere_gpu_end_to_end_fit_hill_climb_10x_vs_cpu` (159s; note the sibling `sphere_gpu_end_to_end_dispatch_parity_vs_cpu_truncated` PASSES, so the GPU path itself dispatches — the 10x-vs-cpu performance/convergence claim is what fails; possibly load-sensitive, ran during heavy contention on one of the two runs but failed identically on both)

Triage hypotheses to verify, not assume: (1)-(2) look #2023/#2275-adjacent (dictionary certification/EV — the convergence-contract restoration in flight may move them; re-run after it lands); (3)-(4) are term_builder seed/anchor expectations possibly stale against recent churn; (5)-(6) are lazy-design/reparam paths; (7) needs a quiet-box rerun before believing the perf claim's failure. Each needs the per-test disposition discipline (re-derive intent vs production fix), same as #2367.

SauersML:
[...]

PR #2403, opened 2026-07-24
fix(#2372): restore singular Greville span diagnostic

murdarch:
## Summary
- distinguish an exactly collapsed divided-difference span from other invalid coordinate orderings
- restore the explicit `singular` Greville-span diagnostic expected by the existing regression
- preserve the stricter finite/increasing diagnostic for non-finite and decreasing coordinates

Addresses only the Greville-span subtask of #2372; the other open clusters remain untouched.

## Verification
- `cargo nextest run -p gam-terms --lib -E 'test(test_penalty_matrix_rejects_singular_greville_span)'` — 1 passed
- `cargo fmt --all -- --check` remains red on unrelated pre-existing workspace formatting; the changed file produced no rustfmt diff

### DIFF
--- a/crates/gam-terms/src/basis/bspline_eval.rs
+++ b/crates/gam-terms/src/basis/bspline_eval.rs
@@ -1111,7 +1111,15 @@ pub fn create_difference_penalty_matrix(
             let mut log_span_sum = 0.0_f64;
             for i in 0..nrows {
                 let span = g[i + o] - g[i];
-                if !span.is_finite() || span <= 0.0 {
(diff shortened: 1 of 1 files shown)


SauersML:
Integrated on current `main` in [`ea3eb3af1`](https://github.com/SauersML/gam/commit/ea3eb3af1): exact zero Greville spans now retain the explicit `singular divided-difference span` diagnostic, while non-finite/decreasing coordinates keep the stricter ordering error. Closing this branch because its complete change is now on `main`.

PR #2404, opened 2026-07-24
test(#2372): pin nuclear-norm eigen shift

murdarch:
## Summary
- re-derive the stale nuclear-norm divided-difference regression against the deliberate additive eigenvalue shift `σ² + max(ε², 1e-15)`
- rename the test so it distinguishes the shared shift from the superseded clamp
- correct the adjacent production comment to match the already-landed regularization contract

This changes no production behavior. It addresses only the nuclear-norm single-test subtask of #2372.

## Verification
- `cargo nextest run -p gam-terms --lib -E 'test(nuclear_norm_right_gram_divided_difference_uses_shared_eigen_shift)'` — 1 passed
- `cargo nextest run -p gam-terms --lib -E 'test(nuclear_norm)'` — 7 passed
- `cargo fmt --package gam-terms -- --check` remains red on pre-existing package formatting; reported `tests.rs` diffs start at line 2040 and no `nuclear_norm.rs` diff was reported

### DIFF
--- a/crates/gam-terms/src/analytic_penalties/nuclear_norm.rs
+++ b/crates/gam-terms/src/analytic_penalties/nuclear_norm.rs
@@ -486,9 +486,9 @@ impl NuclearNormPenalty {
         let mut f = Array1::<f64>::zeros(d);
         let mut df = Array1::<f64>::zeros(d);
         for i in 0..d {
-            // Same shared floor used by `value`/`grad_target` (#737): the
+            // Same shared shift used by `value`/`grad_target` (#737): the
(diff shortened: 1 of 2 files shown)


SauersML:
Integrated on current `main` in [`ea3eb3af1`](https://github.com/SauersML/gam/commit/ea3eb3af1): the regression and adjacent documentation now use the production contract `σ² + max(ε², 1e-15)` and call it a shared additive eigen shift, not a clamp/floor. No production behavior changed. Closing this branch because its complete change is now on `main`.

PR #2405, opened 2026-07-24
fix(#2372): harden derivative penalty scaling

murdarch:
## Summary
- reject dense derivative-penalty Grams when a finite energy factor overflows during `AᵀA`, instead of returning an `Ok` matrix containing infinities
- use the same checked materialization boundary for open and cyclic B-spline roughness matrices while preserving the finite constructive-factor API
- re-derive the extreme-period covariance assertion in matrix max-norm; entrywise relative error was dominated by quadrature residue at analytically zero lags (`~1e-13` at unit scale, still `~1e-16` of the scaled operator norm)
- remove the completed `zz_measure_2372_cyclic_covariance_intermediates` probe

Addresses only the derivative-penalty pair in #2372.

## Verification
- focused pair: 2 passed
- `cargo nextest run -p gam-terms --lib -E 'test(derivative_penalty)'` — 19 passed
- `cargo fmt --package gam-terms -- --check` remains red on pre-existing package formatting; the only reported `derivative_penalty.rs` diff is the unchanged `preserves_all_basis_energies` expression near line 420

### DIFF
--- a/crates/gam-terms/src/basis/derivative_penalty.rs
+++ b/crates/gam-terms/src/basis/derivative_penalty.rs
@@ -50,9 +50,7 @@ pub fn bspline_derivative_penalty_matrix(
     order: usize,
 ) -> Result<Array2<f64>, BasisError> {
     let factor = bspline_derivative_penalty_factor(knot_vector, degree, order)?;
-    let mut penalty = fast_ata(&factor);
-    symmetrize_in_place(&mut penalty);
-    Ok(penalty)
+    materialize_derivative_gram(&factor, "open B-spline roughness penalty")
 }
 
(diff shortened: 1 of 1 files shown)


SauersML:
Superseded by the stronger current-main implementation in [`27133128f`](https://github.com/SauersML/gam/commit/27133128f): coordinate covariance is applied to the assembled unit Gram, which preserves exact matrix covariance (including structural zeros) and performs representability checks at the actual dense-Gram boundary. The obsolete diagnostic probe is gone and the max-norm covariance intent is retained in the integrated implementation. This branch now conflicts with that root fix, so I am closing it rather than resolving toward the older materialization design.

PR #2406, opened 2026-07-24
fix(#2372): freeze measure-jet null rank

murdarch:
## Summary
- certify the affine/null quadratic's numerical rank once in the ℓ-invariant center-value form `H₀`
- pull that fixed-rank constructive factor through the moving evaluation map `E(ℓ)` instead of independently rank-truncating `E(ℓ)ᵀH₀E(ℓ)` at every trial
- normalize through `ConstructiveQuadratic::scaled`, retaining the native factor as the PSD authority
- differentiate the same rank-certified center form in the ψ producer

This removes the moving coefficient-space truncation that desynchronized the rebuilt value from its analytic second derivative. Addresses only the measure-jet subtask of #2372.

## Verification
- `psi_producer_matches_fd_length_scale` — 1 passed
- `cargo nextest run -p gam-terms --lib -E 'test(measure_jet_smooth)'` — 19 passed
- `cargo fmt --package gam-terms -- --check` remains red on pre-existing package formatting; no `measure_jet_smooth.rs` diff is reported

### DIFF
--- a/crates/gam-terms/src/basis/measure_jet_smooth.rs
+++ b/crates/gam-terms/src/basis/measure_jet_smooth.rs
@@ -493,15 +493,24 @@ fn pullback_center_form_log_length_cross(
     (&cross_raw + &cross_raw.t()) * 0.5
 }
 
-/// Function-space affine/null-component penalty in the current coefficient
-/// chart. Under a coefficient reparameterization `E -> E R`, this matrix
-/// transforms covariantly as `S₀ -> Rᵀ S₀ R`; the statistical functional is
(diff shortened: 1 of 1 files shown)


SauersML:
Integrated on current `main` in [`ea3eb3af1`](https://github.com/SauersML/gam/commit/ea3eb3af1): the affine/null rank is certified once on the ℓ-invariant center form, its constructive factor is pulled through `E(ℓ)`, normalization preserves that factor, and the ψ producer differentiates the identical rank-certified object. Closing this branch because its complete change and focused evidence are now represented on `main`.

PR #2763, opened 2026-08-02
Fix manifold benchmark pair decoding

SauersML:
## What changed

- make the manifold benchmark's dictionary width clear the census occupancy floor
- disable dead-block revival in the overcomplete benchmark fit and expose a stable build marker
- decode accepted cross-block pair charts and report held-out pair hits
- extend the rate sweep through top-k 6 so 256/384/512 aggregate-bit budgets are bracketed
- add the complete four-layer JSON artifact, reproducible figure, and limitations report

## Root cause

The first `G=4096` arm sat below the census's own admission floor, then the benchmark discarded nearly every chart it did accept because 34–40 accepted charts per layer were cross-block pairs while only single-block charts were decoded. The original top-k 1–4 sweep also could not evaluate the public 512-bit budget.

## Result

On 100,584 held-out rows per layer, the regression-free lower envelope improves mean FVU at 256 total bits from 0.485907 to 0.480666 and matches the linear optimum at 384 and 512 bits. Raw chart candidates remain in the artifact; the report documents non-monotone top-k 5–6 behavior and all comparison limitations.

## Validation

- MSI job 14670043: completed in 18m38s, exit 0
- `cargo check -p gam-sae --example manifold_fvu_bench`
- `rustfmt --edition 2024 --check crates/gam-sae/examples/manifold_fvu_bench.rs`
- JSON parse, plotting-script compile, render, and visual QA


SauersML:
Closing without merge per project policy; the validated scoped commit is being published directly to main.

PR #2772, opened 2026-08-14
test(sae): #2720 geometry sweep — orbit-derivative measurement across geometries, framed and unframed

HomunculusLabs:
Closes the first "Not established" bullet of #2720: every prior orbit-derivative measurement (the original two fixtures, the #2770 re-measurement, Fixture B `b8f0c4d97`) ran on `Periodic` atoms only, and except for Fixture B's reconstruction-side round trip, with `decoder_frame = None`.

## What this adds

An `#[ignore]`d manual diagnostic (`cargo test -p gam-sae chart_gauge_orbit_violation_across_geometries -- --ignored --nocapture`) that runs the #2770 measurement instrument over {periodic (control), duchon, poincare, linear} × {unframed, framed}, on the Fixture-B seed path (`build_sae_minimal_seed`, `planted_circle_cloud`, only the basis token changed).

## Measured (at `9b9973f`, ARD-saddle rho)

| cell | max \|gᵀv\|/tol | worst-dir near-null share (e⁻²⁰) | verdict |
|---|---|---|---|
| periodic/unframed | 2.33× | 3.4e-9 | above tol |
| periodic/framed | 2.33× | 3.4e-9 | above (verdict unchanged) |
| duchon/unframed | 3.63× | 4.0e-7 | **above — flat patches violate too** |
| duchon/framed | 3.63× | 4.0e-7 | above (verdict unchanged) |
| poincare/unframed | — | — | refused at budget 40, still refused at 400 |
| poincare/framed | — | — | refused at budget 40 |
| linear/unframed | 0.13× | 2.0e-9 | below this tolerance |
| linear/framed | 0.13× | 2.0e-9 | below this tolerance |

## What the numbers mean for the modelling fix

1. **The violation is geometry-general, not periodic-specific.** Duchon (flat) violates at 3.63× — stronger than periodic's 2.33× on the same fixture. A fix confined to the periodic gauge construction cannot be sufficient; it must live at the gauge/penalty interface.
2. **But it is not universal** — the linear atom is below tolerance on both arms. Whatever the eventual fix keys on, degree-1 patches are exempt on this instrument.
3. **The likelihood is flat along the orbit; the penalty block carries the violation.** A near-null-penalty control (all penalty channels collapsed to `exp(-20)`, layout preserved, re-assembled at the same solved state, same unit directions) shows every direction's near-null projection below 1.2e-5× tol (asserted per direction). A deep floor `exp(-26)` discriminates residual penalty (periodic: share scales with the floor 3.4e-9 → 8.0e-12) from roundoff (duchon/linear: floor-independent). This is the likelihood-vs-posterior split #2720 claims — verified independently per geometry, not assumed.
4. **Poincaré never reaches a measurable state.** The inner solve refuses at budget 40 AND at an escalated 400: `gauge_share = 0.9456` at the refusing iterate (94.6% of the KKT gradient inside the chart-gauge span). This telemetry comes from a nonstationary iterate and is explicitly NOT an orbit measurement — but it says the gauge orbit dominates the residual gradient itself on hyperbolic patches. That looks like its own lane (convergence, not symmetry).
5. **Framing changes nothing on the derivative side** (identical max ratios; projections agree to ~3e-4 relative on duchon, machine-exact on periodic/linear) — consistent with Fixture B's exact reconstruction-side round trip.

## Design notes

- Typed `CellOutcome::Measured/Refused` with contextual panics for harness failures — a refusal is a measurement about the geometry, a broken instrument is a bug.
- Refusal telemetry validated (`gauge_share ∈ [0,1]`, finite) and printed in full with the nonstationary-iterate disclaimer.
- Gauge-vector unit normalization asserted per vector (the construction normalizes at `fit_drivers` MGS; this asserts rather than trusts).
- The tolerance gate is the solver's own L2 stationarity gate (`SAE_MANIFOLD_INNER_GRAD_REL_TOL · iterate_scale`); with unit `v`, `|gᵀv| ≤ ‖g‖₂`, so an above-tolerance projection provably violates the solver's L2 gate.
- rho is overridden to the ARD-saddle state BEFORE the solve; the measurement is a fixed-ρ derivative (documented).
- Fixture-family caveat documented: one shared circle-cloud target, native only to periodic — the claim is "a non-periodic kind can violate", not a cross-geometry magnitude ranking.

Review: two full RepoPrompt cycles (oracle + engineer), findings applied and re-verified against fresh runs.


HomunculusLabs:
Scanner-compliance update: same ban-scanner constraint — the geometry sweep was `#[ignore]`d. New commit (`cc83b1d`): the sweep's findings are now ENFORCED as bands — periodic (2.33×) and duchon (3.63×) above tolerance in (1.0, 10.0), linear (0.13×) below tolerance, poincaré refusal at both budgets (40 and 400), both frames per kind. All cells re-measured green at `cc83b1d`; the fixture-family caveat in the module docs is unchanged (assertions pin the findings, prose carries the scope).

HomunculusLabs:
## Post-fix re-measurement (this push, `9391856`): the sweep verdicts inverted against the landed #2720/#2762 resolution

This PR's enforced pins were written against the pre-fix solver at `9b9973f`. The landed #2762 mover (`062277d`/`69ae5a1`) + the Aug-17 two-span split (`ec18eac`/`61d5f00`/`44cf16a`) invalidate them, so every number is re-taken at merged main `5603dec`.

### The poincare refusal is dead — killed by the #2762 mover, not the split

Bisection, same deterministic fixture + ARD-saddle rho, fresh seed per arm:

| commit | mover? | split? | poincare/circle |
|---|---|---|---|
| `9b9973f` | ✗ | ✗ | REFUSED 40 & 400 — telemetry bit-identical to the Aug-14 probe (`‖g‖=5.799974e0`, `gauge_share=0.9456`) |
| `69ae5a1` | ✓ | ✗ | SOLVED, criterion `8.623218e1` |
| `5603dec` | ✓ | ✓ | SOLVED, criterion **bit-identical** `8.623218e1` |

On poincare the mover engages hard at the seed: `decrease=1.9977e2` over 13 rounds (dim=2). The Aug-14 stall thesis ("available in-orbit descent no iterating loop can bank") was the right diagnosis; `descend_gauge_orbit` is what banks it.

### The verdict table (budget 40, ARD-saddle rho)

| kind | pre-fix | post-fix |
|---|---|---|
| periodic | 2.33× above | **0.19× below** |
| duchon | 3.63× above | **0.01× below** |
| poincare | REFUSED | **0.00× below** — first-ever hyperbolic measurement |
| linear | 0.13× below (exempt) | **4.21× above** — the exemption INVERTED |

The curved kinds now meet #2720's acceptance criterion at ARD-saddle rho, corroborating the fix lane's periodic-only fixed-point test across geometries. **Linear is the sole violator post-fix**, on the dilation direction (v₀ 4.21×, v₁ 0.02×) — consistent with the `2f766a7` probe's finding that linear's dilation field moves smoothness by −7.82 at the seed, the steepest orbit of any kind.

### The inversion is NOT budget truncation (new enforced test)

`tests_linear_escalation_2720`: the linear exit is **bit-identical across a 10× budget escalation** (criterion `1.560920e2` at both 40 and 400), and the orbit mover **engages** on linear at the seed (`decrease=4.449559e0`, 11 rounds, stopping at its own stationarity before the 40-round budget). The exit the machinery declares simply carries 4.21× of orbit slope — the mover's material floor and the KKT gate's tolerance are two stationarity bars in different currencies. Pinned in (1, 20) with that reading; not claimed as a defect.

All numbers re-measured and asserted today: `cargo test -p gam-sae --lib` on the branch — sweep, escalation, mover-engagement, frame-roundtrip all green, zero `#[ignore]`d.

HomunculusLabs:
## Follow-up (`72dda02`): the sweep-table fix exposes a gap in the landed flatness evidence — the torus CHANNEL-NULL family carries live slope

### The seeding bug (this branch fixes it)

`44cf16a`'s `GAUGE_SWEEP_KINDS` contained two atoms the minimal-seed path cannot construct:

- `('sphere', 3)` — `sae_build_atom_plans` requires `atom_dim == 2` for the sphere (`atom_build.rs:229`: `atom_dim` is the INTRINSIC dimension; the plan itself carries the ambient 3-vector width). Result: **all three sweep tests panicked on clean main `5603dec`** at `minimal seed: sphere requires atom_dim == 2, got 3` (verified in a clean-checkout worktree).
- `('cylinder', 2)` — cylinder is birth-discovered-only; the plan builder rejects it outright (`atom_build.rs`: "seed with periodic, duchon, sphere, torus, or euclidean_patch and let the structure search grow a cylinder by evidence").

Fixed: sphere → 2, cylinder removed with the reason recorded.

### What the fix exposes (the part that matters)

With the sweep actually able to run past torus, both flatness-gate tests (`quotient_span_is_flat_for_the_penalized_objective_2720`, `..._with_decoder_frames_active_2720`) now fail **at their own assert** with:

```
[2720-gate] kind=torus tol=7.812950e-5 span=392 (chart=2, beta_null=0, channel_null=392)
  torus direction 26: |df| = 3.262147e-4 = 4.175x tol  (data_fit -3.2e-8, smoothness +3.26e-4, ard 0)
  torus direction 34: |df| = 3.492866e-4 = 4.471x tol  (data_fit  3.5e-8, smoothness -3.49e-4, ard 0)
  ... 11 directions between 1.09x and 4.47x
```

The span breakdown attributes every violation to the **decoder channel-null family** — one of the two families `ec18eac`'s `posterior_null_quotient_basis` still removes from convergence measures on the "flat by construction" premise. Two notes on the landed evidence for that premise:

1. It was measured on ONE kind: duchon (704 directions, flat to 3e-9). The torus cell says channel-null flatness is **kind-dependent** — flat on duchon, live (up to 4.47× tol) on torus, carried entirely by the smoothness term (data_fit flat at ~1e-8, ard exactly 0 on every violating direction).
2. `00cae34`'s commit table labels those 704 duchon directions "beta_null"; the breakdown at the same state reads `chart=2, beta_null=0, channel_null=704`. If I'm reading the family order correctly (`dense_step_gauge_vectors` first, then beta-null, then channel-null, per `posterior_null_quotient_basis`), the landed commit message names the wrong family — worth a look from whoever owns that lane.

Mechanistic guess (unverified): the torus's tensor-product harmonic basis gives the smoothness Gram charge on coefficient directions inside the channel-null span that duchon's basis does not reach. Happy to chase the exact Gram interaction if useful.

**Net for this PR:** 8/10 #2720 tests green on the branch (was 7/10 on clean main, with 3 of those crashing before their verdict); the 2 remaining reds are the flatness gates failing with the above content — which is the gate doing its job on evidence the sphere panic had been hiding, not a regression from this branch.

HomunculusLabs:
## Mechanism pinned: the channel-null smoothness slope IS the sub-floor σ leak (`6fca27b`)

Follow-up to the torus channel-null finding from Aug 17. The question left open was *why* the torus's channel-null family carries smoothness slope that duchon's doesn't. New probe `tests_channel_null_currency_2720.rs` answers it with a closed form, verified numerically:

**The slope of every emitted channel-null direction is exactly `λ·σ·(S u)[col]`, where `B c = σ·u` is the decoder SVD.**

- The emission (`fit_drivers.rs` `decoder_channel_null_directions`) is gated **only on the decoder SVD**: `σ ≤ √(SAE_DECODER_BETA_NULL_RELATIVE_FLOOR)·σ_max = √(1e-9)·σ_max ≈ 3.16e-5·σ_max`. The penalty is never consulted. The smoothness value is `0.5·λ·tr(BᵀSB)` (`construction.rs` `decoder_smoothness_value`), so the emitted `e_col ⊗ c` has slope `λ·(SB c)[col] = λ·σ·(S u)[col]` — a **data-fit currency (σ) converted into penalty currency through S**. It vanishes only if σ = 0 *exactly*, which the floor does not require.
- **Measured:** the probe recomputes the closed form independently of the emission and compares against the central-difference slope through the value function. Agreement to **~1e-9 absolute on all 392 torus directions** (slopes reach 5.06e-4 = 6.48× tol raw; 4.47× through MGS as reported Aug 17 — the MGS mixing redistributes slope across directions, which is why the raw count of live directions is 18, not 11).
- **All 392 directions share σ = 5.309e-5** (the emission floor itself), so the slope distribution across columns is *the `(S u)[col]` distribution*. The confirmation: `col=0` — the constant mode, S's nullspace — measures **exactly 0.000e0**, as the closed form predicts (`S[0,:]·u = 0`).
- **Why only the torus:** at the seeded state, the torus is the only kind that emits channel-nulls at all (M=49 = (2·3+1)² tensor harmonics against the planted circle's rank-2 signal + 8 unrealised channels). The landed flatness evidence was taken on kinds where the family is empty — its table's `decoder_channel_null_directions | — | 0.00003` row was an inference from the β-null row, not a measurement of the channel-null family itself.
- **Contrast control:** S-nullspace directions (the structural source of β-null flatness: `v'(G+λS)v = 0` with PSD terms forces `Sv = 0`) measure **exactly 0 through the same instrument on every kind**. The two families are flat by two different arguments, and only one of them is structural.

**Landed-gate corroboration:** the landed `quotient_span_is_flat_…` gate's own per-term output (this branch, repaired seeding) attributes every violating direction's slope entirely to `smooth` (data_fit ~1e-8, ard = 0), with values matching the raw closed forms digit-for-digit (gate dir 3 `smooth=-8.497e-5` = raw col 3 `-8.497096e-5`).

### What this means for the fix

The quotient (`posterior_null_quotient_basis`) admits the channel-null family on a **data-fit-only** admission test while the flatness claim it certifies is a **penalty** claim. Any fix must price the penalty at emission or screen at the quotient. Two candidate shapes, not adjudicated here:

1. **Per-kind precondition at emission** — require `λ·σ·‖S u‖ ≤ tolerance` (or `σ ≤ tol/‖S‖₂`), so the emission floor adapts to the roughness spectrum (`(k²+l²)²` reaches 324 at H=3).
2. **Rayleigh screen at the quotient** — the outer-gradient solver already runs one against the cached Hessian (`construction_quasi_laplace.rs` gauge deflation); the quotient could reuse the same floor so the two consumers share one admission standard.

Also worth noting: the mover (`descend_gauge_orbit`) descends `likelihood_flat_block_basis` which contains this family too — but descending is always sound; it's the *removal from the convergence measure* that needs the flatness premise.

Both tests pass (`cargo test -p gam-sae --lib tests_channel_null_currency` → 2 passed). The probe deliberately fails loudly if the symptom moves (no live slope ⇒ assert), so it doubles as a regression canary for whatever fix lands.

HomunculusLabs:
## Fix shipped on this branch (`2f9c014`): the channel-null screen at quotient admission

The Aug-20 comment pinned the mechanism (sub-floor σ leaking into penalty currency through `S`). This commit lands the fix, as **fix shape 2 — screen at the quotient, not at emission** — and the choice between the two shapes turned out to be the interesting part.

**Why shape 1 (S-aware emission floor, my Aug-20 preference) is wrong:** `decoder_channel_null_directions` is the single source for **four** consumers — the convergence quotient (`posterior_null_quotient_basis`, fit_drivers.rs:2540), the descent block (`likelihood_flat_block_basis`, :2572), the refusal telemetry (:2274 in construction_quasi_laplace.rs), and the outer-gradient Rayleigh pass (:3123). Screening at emission would strip live directions from the descent block too. But a live direction is **realizable descent** — exactly what #2762 measured on the chart orbit (559 stall-resolutions of objective drop along the span the gate projects away). Removing it from the mover that can act on it is the #2762 pathology recreated for this family.

**The landed shape:** new `decoder_channel_null_quotient_directions(penalized_gram_scale)` recomputes the closed-form slope `λ_k·(S_k B_k c)[col]` (no division by the sub-floor σ — a denormal singular value cannot amplify roundoff into a verdict) and admits a direction into the **quotient only** if that slope is at or below the tolerance the accept path itself denominates in: `SAE_MANIFOLD_INNER_GRAD_REL_TOL · inner_iterate_scale()` (the single source of truth, construction_quasi_laplace.rs:937). The descent block keeps the raw stream — same division of labor the chart orbit got in `ec18eac`: *removed from the measure only if the posterior cannot see it, descended otherwise.*

**Measured (planted-circle fixture, seeded state, both emitting kinds):**

| kind | raw emitted | quotient-admitted | dropped | descent block |
|---|---|---|---|---|
| torus | 392 | 366 | **26** (the live ones from the Aug-20 table) | 394 (all raw kept) |
| duchon | 704 | 704 | 0 (genuinely flat family — small ‖S‖₂) | 706 (all raw kept) |

**The red that was the finding is now green, honestly:** `quotient_span_is_flat_for_the_penalized_objective_2720` and `quotient_span_is_flat_with_decoder_frames_active_2720` — red on this branch since Aug 17 because the live torus directions sat inside the span — now pass because the live directions left the span, not because the bar moved. Full `_2720` filter: **12/12 green** (was 10/12).

**New guard** (`screened_channel_null_directions_stay_reachable_by_the_descent_block_2720`) pins the two-span invariant #2762 taught: the screen must be a strict filter (every screened direction is a raw emission, element-wise) and every raw direction must lie in the descent block's span (residual ≤ 1e-9 after projecting onto the block). If a future change routes the block through the screen too, live descent vanishes from both spans — invisible to the gates AND unreachable by the movers — and this test fails.

One consequence worth flagging for the doc row: the flatness-evidence table in `quotient_residual_norm_sq`'s doc comment had `decoder_channel_null_directions | — | 0.00003`, sampled only on duchon. That row is now split: raw (kind-dependent, 26/392 live on torus) vs screened (flat by the closed-form screen). Updated in this commit.

SauersML:
Manual audit result: the core fix is sound and the hosted focused proof on this exact contributor head passed all 13 `_2720` tests (0 failed): https://github.com/SauersML/gam/actions/runs/33554350936

I found three review-cleanup items that should not be merged as-is:
- `tests_linear_escalation_2720` lacked an explicit `#[cfg(test)]` module gate;
- that file's module/fixture comments still described a Poincaré refusal even though it constructs a linear atom and asserts two successful, budget-invariant exits;
- `posterior_null_quotient_basis` had one sentence fragment and a duplicated channel-null screening paragraph.

Because the head is on a contributor fork and cannot be updated from the maintainer token, I incorporated the complete change plus those cleanups in #2794. I am keeping this PR open until the audited replacement is synced, rerun on hosted CI, and merged; then I will close it as incorporated rather than losing the contribution.

SauersML:
Incorporated and merged through #2794 as `1f8c5353ead7b1d2f2089a246a461e36e5824bf4`. The replacement preserves the production fix and regression evidence, adds the explicit test-only gate, corrects the stale linear-test documentation, and removes the duplicated quotient-basis prose. Closing this source PR as incorporated.

PR #2773, opened 2026-08-15
fix(#2720): framed inner solves now get the #2267 trust region; stall diagnostics enforced

HomunculusLabs:
Investigation lane for #2720's Poincaré refusal plus one production fix found along the way. Findings are posted on the issue; this PR carries the durable artifacts.

## The production fix: framed solves skipped the #2267 trust region (05bbe69)

The dense inner loop guarded its trust-region block with `delta_beta.len() == beta_dim()`, but a frame-active solve returns `delta_beta` in the FACTORED border layout (`factored_border_dim < beta_dim`) — so whenever a decoder frame was active, the #2267 iterate-scale trust region and the quotient-step norm were silently skipped. Two in-tree contracts already disagreed with the old guard: `quotient_newton_step_norm_sq` expects and validates `factored_border_dim` (its validation was unreachable), and the streaming sibling sizes its border the same way. The guard now tests `factored_border_dim()`.

Found in an external (GPT-Pro) review of this lane, verified against the code before acting. The regression test pins the framed trust ceiling; its module docs honestly scope what it does and does not prove (on this fixture the first step sits inside the ceiling — a true red-green test needs a wilder fixture or an options-plumbed radius, noted as remaining work).

## The diagnostics, now enforced (86be4a3)

The repo's build-time ban-scanner rejects any `#[ignore]` attribute, so the four probes from this investigation are enforced tests with real assertions on their deterministic measurements:

- gauge directions are likelihood-null at the seed on every kind (relative meaningfulness guard)
- both poincaré AND periodic solve their native targets — pins the corrected-fixture finding; the RNG-artifact "periodic refuses native" cannot return unnoticed
- the stall signature as a contract: the poincaré/circle refusal reproduces with #2772-identical telemetry, the gradient is orbit-dominated (>0.99), QᵀQ is orthonormal, and the independently-computed step is strongly anti-aligned (cos θ < −0.9, measured −0.949)
- the periodic/native control must converge

All six #2720 tests green in the normal suite, zero ignored.

## Scope — what this PR deliberately does NOT do

The #2720 modelling fix itself (semantic split of likelihood-only vs exact-posterior-null directions) is not here: it changes what the criterion certifies across every consumer and belongs to a dedicated change with a production-pipeline trace as evidence. The stall-state measurements (strong anti-gradient direction exists at the refusing state; whether production constructs/submits it is open) are pinned here as contracts so the regime can't drift silently while that work happens.

## Notes for reviewers

- Two pre-existing breakages on main verified by clean-checkout worktree, unrelated to this branch, flagged on the issue: the ban-scanner fails on 5 underscore-param violations in `gam-custom-family/tests.rs` (blocks the Python wheel build on main), and `production_factored_large_border_routes_to_resident_inexact_pcg_1017` expects `sys.k == 2048` but gets 512.
- The stall probes measure a zero-ridge reconstruction at the post-refusal state, not the production trajectory — documented in-module, and the production-path trace remains the next lane.


SauersML:
Manual audit/update (2026-09-01): the production defect is valid, but this PR should not merge in its current 1,140-line form.

Findings:

- The `delta_beta.len() == beta_dim()` guard is wrong for frame-active dense solves; the returned border is factored, and the downstream norm routine already requires `factored_border_dim()`.
- The three large diagnostic modules are pre-#2762/#2720 probes now overlapped or invalidated by #2772.
- `tests_trust_region_framed_2720.rs` explicitly says its fixture passes with or without the guard change, so it is not regression coverage for this fix.

I extracted the valid one-line correction into focused replacement #2792, based on current main, and launched hosted package-scoped build/tests: https://github.com/SauersML/gam/actions/runs/33556078165

Once #2792 is green and merged, I will close this PR as incorporated rather than landing the stale diagnostic payload.

SauersML:
The valid production correction was incorporated and merged through #2792 as `9a8a8dfaff5f00532f68f74566d650c181125cb1`. The superseded diagnostic payload was intentionally not merged for the reasons in the audit above. Closing this source PR as incorporated.

PR #2792, opened 2026-09-01
fix(sae): apply trust region to factored frame solves

SauersML:
Extracts the valid production fix from #2773 without its superseded diagnostic payload.

A frame-active dense solve returns `delta_beta` in the factored border layout. The existing guard compared that length with `beta_dim()`, so it skipped `quotient_newton_step_norm_sq` and the #2267 trust-region clipping whenever a decoder frame was active. The callee already validates `factored_border_dim()`, and the streaming path uses the same factored width.

This changes the guard to the layout the solver actually returns. #2773 is superseded by this focused replacement; its probes overlap the newer #2772 evidence and its framed test states that it is not red-green.


SauersML:
Final manual review: the one-line guard now uses the factored border width returned by the dense solve, matching both the norm callee and the streaming path. The hosted proof also finished successfully before cancellation, but I am merging on the manual audit rather than treating CI as a gate.

PR #2771, opened 2026-08-13
perf(#2731): cache-efficient matmuls in add_sae_separation_barrier — the 86% wall

HomunculusLabs:
## What

The #2731 profiling found `add_sae_separation_barrier` consumes **86.37% of all user cycles** in the censoring cell (`p=2048, charts=32`), running at **1.03 of 16 cores** — a single-threaded serial wall that blocks the #2283 production shape.

This PR replaces five hand-rolled scalar triple-nested matrix products with cache-efficient `ndarray.dot()` calls (backed by the `matrixmultiply` crate's blocked traversal).

## The problem

`add_sae_separation_barrier` (`penalties.rs:1418-1474`) computes per-edge decoder derivatives for the Jeffreys separation barrier. Each edge computes five matrix products as scalar loops over the ambient dimension `p`:

```
cross = B_j·B_kᵀ          (m_j × m_k, over p)
mb    = cross · B_k        (m_j × p)
sjb   = S_j · B_j          (m_j × p)
mtb   = crossᵀ · B_j       (m_k × p)
skb   = S_k · B_k          (m_k × p)
```

The inner loops for `mb`, `mtb`, `sjb`, and `skb` access `B_k[[b, o]]` with `b` as the fast index and `o` fixed — a **strided column access pattern** that thrashes L1 at `p=2048`. Each `(a, o)` entry of `mb` walks `cross[a, :] · B_k[:, o]`, re-reading the column `o` of `B_k` for every row `a`.

## The fix

Replace the five hand-rolled loops with `.dot()`:

```rust
let cross   = bj.dot(&bk.t());     // was: triple scalar loop
let mb_mat  = cross.dot(bk);       // was: scalar loop, re-derived per (a,o)
let sjb_mat = s_j.dot(bj);
let mtb_mat = cross.t().dot(bj);
let skb_mat = s_k.dot(bk);
```

`ndarray.dot()` dispatches to `matrixmultiply` which uses cache-blocked (MC × KC × NC) traversal. The math is identical — same operations, different summation order from blocking. Net: −28 lines, +24 lines.

## Why this is safe

The replacement is **algebraically equivalent** to the original scalar loops. The orientation of all five products was verified against the source:

- `bj.dot(&bk.t())` → B_j B_kᵀ (m_j, m_k)
- `cross.dot(bk)` → M B_k (m_j, p)
- `s_j.dot(bj)` → (B_j B_jᵀ) B_j (m_j, p)
[...]

PR #2793, opened 2026-09-01
perf(sae): use blocked matmuls in separation barrier

SauersML:
Current-main replacement for #2771.

The original branch conflicts because the barrier carrier representation changed from flat `(index, value)` entries to per-atom dense runs. This reapplies the same algebra to the current implementation:

- `cross = B_j B_kᵀ`
- `cross B_k` and `S_j B_j` for the `j` carrier
- `crossᵀ B_j` and `S_k B_k` for the `k` carrier

The proposed `.dot()` calls use ndarray/matrixmultiply blocked traversal instead of repeatedly walking strided decoder columns. No API or mathematical change; only summation order changes. Hosted focused correctness tests will run before merge. The original PR did not establish a wall-clock speedup on its production reproducer, so this replacement claims the cache-locality implementation change, not a measured end-to-end speedup.

### DIFF
--- a/crates/gam-sae/src/manifold/penalties.rs
+++ b/crates/gam-sae/src/manifold/penalties.rs
@@ -1491,16 +1491,7 @@ impl SaeManifoldTerm {
                     edge_v.push(Vec::new());
                     continue;
                 }
-                let mut cross = Array2::<f64>::zeros((m_j, m_k));
-                for a in 0..m_j {
-                    for b in 0..m_k {
-                        let mut c = 0.0_f64;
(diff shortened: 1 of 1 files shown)


SauersML:
Final manual review: each replaced scalar loop maps directly to the corresponding blocked product (`B_j B_k^T`, `cross B_k`, `S_j B_j`, `cross^T B_j`, `S_k B_k`) against the current dense-run carrier layout. Only floating-point summation order changes. The hosted separation-barrier proof also finished successfully; merging based on the algebraic audit.

PR #2794, opened 2026-09-01
fix(sae): screen channel-null convergence quotient directions

### LINKED ISSUE #2720: The SAE chart-gauge orbit is a symmetry of the likelihood, not of the posterior: make it a real gauge (modelling fix behind #2715)

SauersML:
Incorporates the production fix and regression evidence from #2772 on a GAM-owned branch so the final review cleanup can be applied without rewriting the contributor's fork.

The convergence quotient previously treated decoder channel-null directions as posterior-flat based only on data-fit singular values. This change screens each candidate against the smoothness slope `lambda * (S B c)[col]` at the accept path's tolerance before quotient admission. Raw directions remain available to the likelihood-flat descent block.

Audit cleanup on top of #2772:
- explicitly gates the linear escalation module with `#[cfg(test)]`;
- corrects stale Poincaré/refusal documentation for the actual linear fixture and assertions;
- removes duplicated quotient-basis documentation.

Hosted proof on the contributor head: 13 `_2720` tests passed, 0 failed in run 33554350936. I will rerun the same focused filter on this exact audited head after syncing it with current `main`.

Closes #2720.

Authored-by: HomunculusLabs (Bill Ding), via #2772.


SauersML:
Final manual review: the convergence quotient now admits a raw decoder channel-null direction only when its closed-form smoothness slope is within the accept path tolerance, while the descent block still receives the complete raw family. The branch is cleanly synced with current `main`, including #2792 and #2793. Merging the audited incorporation of #2772.

PR #2855, opened 2026-09-08
fix(sae): add saddle-safe objective trust step to escape indefinite-polish fixed points (#2283)

SauersML:
### Motivation
- The terminal exact-Newton polish was globalized entirely in the stationarity-residual currency, which rejects directions that decrease the penalized objective when the stationarity operator is indefinite and thus produces a deterministic bit-identical fixed point. 
- Measurements showed this was not a budget or tolerance issue but a denomination mismatch: a step that lowers the objective can increase `‖g‖`, so the residual-only gate can refuse the useful descent direction.

### Description
- Add a saddle-safe spectral trust-step `damped_objective_step` implementing the closed-form spectral absolute-Hessian step `δ = −(A² + νI)⁻¹ |A| g`, which is descent on every retained eigendirection including negative-curvature modes, and returns the predicted objective decrease. (file: `crates/gam-sae/src/manifold/construction_exact_hessian.rs`)
- When the existing residual-minimizing damped path produces no acceptable trial, evaluate the new saddle-safe path over the same derived damping ladder and accept a step only via a measured Armijo decrease in the penalized-objective currency; rejected trials restore the pre-trial state. (file: `crates/gam-sae/src/manifold/construction_quasi_laplace.rs`)
- Add a focused regression test that asserts the new trust step descends on retained indefinite modes and reports the predicted decrease, exercising the new behavior and guarding the semantics. (file: `crates/gam-sae/src/manifold/construction_tests.rs`)

### Testing
- Ran `cargo test -p gam-sae --lib objective_trust_step_descends_every_retained_indefinite_mode_2283` and it passed (1 passed, 0 failed). 
- Ran `cargo test -p gam-sae --lib terminal_polish_never_raises_the_kkt_residual_2762` and it passed (1 passed, 0 failed) to confirm the original polish invariants remain satisfied when the residual path succeeds. 
- Attempted `cargo build --workspace --all-targets` but the workspace build is currently blocked by unrelated pre-existing dead-code/warning errors in `gam-sae` (unused methods reported under -D warnings), so a full all-targets verification could not be completed in this run.

### DIFF
--- a/crates/gam-sae/src/manifold/construction_exact_hessian.rs
+++ b/crates/gam-sae/src/manifold/construction_exact_hessian.rs
@@ -265,6 +265,50 @@ pub(crate) struct DampedResidualStep {
 }
 
 impl ExactHessianSpectralBlock {
+    /// Saddle-safe objective step `-(A² + νI)⁻¹|A|g` (#2283).
+    /// Unlike residual minimisation, this is descent on every retained mode,
+    /// including negative-curvature modes where objective descent raises `‖g‖`.
+    fn damped_objective_step(
+        &self,
+        residual: &SaeArrowVector,
+        nu: f64,
+    ) -> Result<(SaeArrowVector, f64, usize), String> {
+        let total_t = residual.t.len();
+        let dim = total_t + residual.beta.len();
+        if self.eigenvectors.dim() != (dim, dim) || self.eigenvalues.len() != dim {
+            return Err("damped objective step: eigensystem and residual dimensions differ".into());
+        }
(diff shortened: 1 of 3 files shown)


SauersML:
Closing as superseded: #2861 (`codex/issue-2080`) landed on main in d82efb558 and carries the same mechanism (objective-descent terminal polish step and per-iterate gauge-orbit composition) that this PR implements for #2283. Merging both would have produced two implementations of the same step in `construction_exact_hessian.rs` / `fit_drivers.rs`. If #2283 still reproduces on main after #2861, reopen against the current tree.

PR #2857, opened 2026-09-08
fix(sae): accept terminal steps on objective descent (addresses #2267)

SauersML:
### Motivation
- The shipped manifold-SAE example fails at the first ladder rung because the terminal polish phase accepts only residual-norm decreases, not penalized-objective decreases, producing a refine/polish limit cycle that never converges.
- At indefinite exact-stationarity Hessians a residual-minimising step can block an actual objective descent, so the acceptance currency must be the penalized objective while keeping the KKT residual as the convergence certificate.
- The change targets the algorithmic mechanism (acceptance currency) rather than tuning or time/budget workarounds so the solver can take scientifically correct descent steps and avoid the measured crawl.

### Description
- Add `shifted_objective_descent_step` to `ExactHessianSpectralBlock` (in `crates/gam-sae/src/manifold/construction_exact_hessian.rs`) which forms an objective-directed step by shifting negative eigenvalues so that retained spectral components produce an objective-descent direction. 
- In the terminal polish (`construction_quasi_laplace.rs`) attempt the shifted exact-Hessian objective step first and globalize it with the existing Armijo/backtracking machinery (`initial_step=1.0`), accepting a genuine objective decrease while leaving the residual-based gate as the convergence certificate. 
- Pin the behavior with a unit/regression test `indefinite_exact_hessian_step_is_objective_descent_2267` in `crates/gam-sae/src/manifold/construction_tests.rs` which asserts the shifted step descends the penalized objective on an indefinite spectrum and add the minimal `opt` imports needed for the line-search call. 

### Testing
- Ran `git diff --check` and code-style diffs locally which reported no lint diffs from the edits. (`git diff --check` returned clean.)
- Ran `cargo check -p gam-sae` which reached the `gam-sae` crate but failed on pre-existing `-D warnings` diagnostics (unused methods flagged as dead code: `apply_exact_hessian_minus_b` / `apply_exact_hessian_matrix_free`, and `decoder_prior_beta_hvp_pair`), so the new test could not be executed in this environment. (Outcome: compilation reached crate, then failed due to those diagnostics.)
- Attempted the focused test run `cargo test -p gam-sae indefinite_exact_hessian_step_is_objective_descent_2267 -- --exact` but it could not run because `cargo check`/build failed on the unrelated deny-warnings baseline; an uncapped workspace build (`cargo build --workspace --all-targets`) was also started but could not be completed inside the interactive sandbox window. (Outcome: test not executed; follow-up needed once baseline diagnostics are cleared.)

### DIFF
--- a/crates/gam-sae/src/manifold/construction.rs
+++ b/crates/gam-sae/src/manifold/construction.rs
@@ -1,4 +1,5 @@
 use super::*;
+use super::fit_drivers::GaugeOrbitDescent;
 // #2598 — `construction_quasi_laplace.rs` interpolates the refusal markers into
 // its ρ-local refusals instead of restating them, so the classifier and the
 // producer cannot disagree about the phrase again.
@@ -9,7 +10,7 @@ use crate::identifiability::{
 };
 use gam_linalg::faer_ndarray::FaerEigh;
 use gam_math::special::bessel_i0_centered_terms_from_log_abs;
-use super::fit_drivers::GaugeOrbitDescent;
+use opt::{BacktrackConfig, backtracking_line_search};
 
 // ── Theorem K: the rank charge is a RUNNING COMPLEXITY λ(n) ──────────────────
 //
--- a/crates/gam-sae/src/manifold/construction_exact_hessian.rs
+++ b/crates/gam-sae/src/manifold/construction_exact_hessian.rs
@@ -405,6 +405,55 @@ impl ExactHessianSpectralBlock {
         })
     }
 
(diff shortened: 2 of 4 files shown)


SauersML:
Closing as superseded: #2861 (`codex/issue-2080`) landed on main in d82efb558 and carries the same mechanism (objective-descent terminal polish step and per-iterate gauge-orbit composition) that this PR implements for #2267. Merging both would have produced two implementations of the same step in `construction_exact_hessian.rs` / `fit_drivers.rs`. If #2267 still reproduces on main after #2861, reopen against the current tree.

PR #2860, opened 2026-09-08
fix(sae): ensure pinned gauge (β) block is minimized each accepted inner iteration (#2228)

SauersML:
### Motivation

- The inner Newton solve could accept a transverse (quotient) Newton step while leaving material objective descent in the reconstruction-gauge (β) span that the quotient pins, causing fixed-ρ stalls and large KKT residuals reported in #2228.  
- The aim is to keep the conditioning benefit of pinning the quotient while ensuring the solver actually clears prior-driven descent in the pinned span during the same accepted iteration, preventing repeated stalls.

### Description

- After an accepted Newton step and the usual representation hooks, call the gauge-orbit minimizer (`descend_gauge_orbit`) before finalizing the iteration so the pinned β block is minimized as part of the same iteration rather than only at terminal plateaus.  
- If a single orbit-descent invocation consumes the caller’s inner-iteration grant while still making material progress, repeatedly invoke the orbit descent in fresh chunks until the orbit block reports no further material move; aggregate the measurements into a single `GaugeOrbitDescent` outcome.  
- Add a regression test `accepted_inner_iteration_also_minimizes_the_pinned_gauge_block_2228` that runs an actual inner block-coordinate iteration and then audits the pinned β block to ensure no material objective decrease remains after the accepted iteration.  
- Changes are confined to the SAE manifold driver and the gauge-orbit descent test harness (modified files: `crates/gam-sae/src/manifold/fit_drivers.rs`, `crates/gam-sae/src/manifold/tests_gauge_orbit_descent_2762.rs`).

### Testing

- Ran the new regression: `cargo test -p gam-sae --lib accepted_inner_iteration_also_minimizes_the_pinned_gauge_block_2228` — test passed (1 passed; 0 failed; filtered tests omitted).  
- Exercised related inner-fit audits: `cargo test -p gam-sae --lib the_inner_fit_never_exits_with_material_decrease_left_in_the_removed_span_2762` was run and completed (test ran to completion); the added unit-level instrumentation exercised the new path.  
- Attempted workspace verification: `cargo check -p gam-sae` and `cargo build --workspace --all-targets` were attempted but the workspace build currently fails on unrelated, pre-existing `-D dead-code`/warning errors in other functions (these are not caused by the changes in this PR and must be addressed separately); the change itself compiles and its unit tests pass in the `gam-sae` crate.

### DIFF
--- a/crates/gam-sae/src/manifold/fit_drivers.rs
+++ b/crates/gam-sae/src/manifold/fit_drivers.rs
@@ -7590,9 +7590,6 @@ impl SaeManifoldTerm {
             if let Some(step) = accepted_step {
                 state_moved = true;
                 moved_at.get_or_insert(StateMoveSite::AcceptedNewtonStep);
-                // Genuine transverse progress: the next plateau is a new one, so
-                // the gauge block is worth asking again (#2762).
-                gauge_block_armed = true;
                 // #2267 — A NEWTON STEP'S NATURAL LENGTH IS ONE.
                 //
                 // `backtracking_line_search` only ever CONTRACTS from its initial
@@ -7803,7 +7800,6 @@ impl SaeManifoldTerm {
                 }
                 state_moved = true;
                 moved_at.get_or_insert(StateMoveSite::ProximalCorrectionStep);
-                gauge_block_armed = true;
             }
             // Affine gauge canonicalization is a representation change, but the
             // decoder smoothness term is part of the optimized objective — a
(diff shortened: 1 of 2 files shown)


SauersML:
Closing as superseded: #2861 (`codex/issue-2080`) landed on main in d82efb558 and carries the same mechanism (objective-descent terminal polish step and per-iterate gauge-orbit composition) that this PR implements for #2228. Merging both would have produced two implementations of the same step in `construction_exact_hessian.rs` / `fit_drivers.rs`. If #2228 still reproduces on main after #2861, reopen against the current tree.

Issue #2911, opened 2026-09-12
Atlas refuses a planted swiss roll once the sheet is taller: the nerve gains spurious 1-cycles (#2280 follow-up)


SauersML:
## The atlas refuses a planted swiss roll once the sheet is taller: the nerve gains spurious 1-cycles

Split out of #2280, where this finding existed only as numbers in a probe log.

### Measured

MSI job 531104 on sw0g, at fe919657d (HEAD == pin), log `/scratch.global/sauer354/pool/sae2280/probe2.531104.log`, test `probe_2280_developing_map_census` (a probe patch, not landed). Every fixture is a 1.5-turn roll `(t cos t, t sin t, h)` with `t ∈ [1, 1 + 3π]`, read at `d = 2` through `LocalAtlas::build(z, LocalAtlasConfig::balanced(n, 2))` → `observe_atlas_topology`:

| fixture | n | verdict | nerve (b0, b1, b2) | χ | signed (b0, b1) |
|---|---|---|---|---|---|
| zoo `swiss_roll(80, 16)`, h ∈ [0, 2] | 1280 | **disk** | (1, 0, 0) | 1 | (1, 0) |
| 60 × 16, h ∈ [0, 10] | 960 | refused | (1, 3, 0) | −2 | (1, 0) |
| 80 × 16, h ∈ [0, 10] | 1280 | refused | (1, 2, 0) | −1 | (1, 1) |
| 80 × 30, h ∈ [0, 20] | 2400 | refused | (1, 9, 0) | −8 | (1, 5) |
| 45 × 10 grid, h ∈ [0, 10] | 450 | refused | (1, 2, 0) | −1 | (1, 0) |

The sheet is contractible in every row, so any `b1 > 0` is wrong. The readout refuses rather than misnaming, because the resolved-sign subcomplex does not see the nerve's cycles. That is the gate working, but the planted sheet is not recognized.

### Same fixtures, the developing map

[...]

Issue #2909, opened 2026-09-12
Good-cover certificate for LocalAtlas: make chi and b1 theorem outputs (#2280 follow-up)


SauersML:
## Good-cover certificate for LocalAtlas: make χ and b₁ theorem outputs

Split out of #2280 (increment 4), where it existed only as design text.

### What is missing

`observe_atlas_topology` reads homology off the membership nerve and uses the nerve theorem without proving its hypothesis: every nonempty finite intersection of patches must be contractible. The readout's verdicts are measured (`χ = 2` on the sphere, `b₁ = 1` on the circle), not proved for the cover. The torus is where this bites. #2280's geometry review measured raw nerve χ on a naive membership cover as `{16, 6, 3, −13}` across landmark counts and membership sizes, and gated the torus verdict on a certified good cover.

### What exists to build on

`inference::atlas_nerve` already carries `ConvexIntersectionProof` and `AtlasGoodCoverCertificate`, exhaustive over the full nerve. `LocalAtlas` has what a proof needs: each patch is a metric ball, and each chart is bi-Lipschitz certified (`ChartCertificate::min_projection_stretch`). So the image of an intersection of balls in a witness chart is an intersection of near-balls, convex up to the certified stretch.

### Acceptance

- Planted sphere, circle, disk, cylinder and Möbius covers that the readout names today each carry a discharged certificate.
- A cover whose intersections are not contractible (the #2280 swiss-roll localization's four-patch class, or a deliberately thickened cover) fails the certificate with a typed reason.
- The certificate is exact or derived from the chart certificates; no tuned tolerance.
- This is the precondition for promoting an atlas verdict to a population claim through the #2311 seal, which stays a separate step.

PR #2920, opened 2026-09-14
test(atlas): run the queued ball-cover census — exact unwitnessed pair and triple counts on the planted zoo (#2280)

HomunculusLabs:
{
  "title": "test(atlas): run the queued ball-cover census — exact unwitnessed pair and triple counts on the planted zoo (#2280)",
  "head": "HomunculusLabs:research/cech-census-2909",
  "base": "main",
  "body": "Follow-up to #2881, advancing the folded former-#2909 lane on #2280.\n\n## What this runs\n\nThe Sep-13 status queued exactly one probe: _\"a probe counting unwitnessed ball pairs and triples on the planted zoo\"_ — the evidence base for the next good-cover certificate model: each patch's closed ambient ball `B(c_i, r_i)` with `r_i` its farthest member, good by convexity; certificate = the membership nerve equals the ball Čech nerve; unwitnessed intersection = typed refusal naming the simplex. This PR is that census.\n\n## The intersection tests are exact, not sampled\n\n- **Pairs.** The margin `f = min_x max(g_i, g_j)` with `g_l = ‖x−c_l‖ − r_l` is attained on the center segment: at the crossing `s* = (D + r_i − r_j)/2` when in-segment (KKT: two active constraints are collinear), at a center otherwise — closing to `max((D − r_i − r_j)/2, −min r_l)`.\n- **Triples.** `f` is non-decreasing in every coordinate orthogonal to the centers' affine hull (`g_l(y + sn) = √(g_l(y)² + s²)`), so the minimizer lies in their plane, where KKT again confines it to a finite candidate set: the three centers, the three segment crossings, and the additively-weighted (Apollonius) circumcenters `‖x−c_l‖ = ρ + r_l` — pairwise subtraction of the squared equations leaves a system linear in `(x, ρ)`, so `x(ρ) = p + ρq` and the single ball equation becomes a quadratic in `ρ` with at most two roots. No grid, no refinement parameter, no tolerance beyond the adjudication window `16·ε·scale` (machine epsilon on the simplex's own scale; margins inside it are counted as boundary and never adjudicated).\n- **Containment is checked exactly.** A shared member row reproduces the builder's own squared-distance accumulation bit for bit, so containment holds with no `sqrt` roundtrip; `containment_violations = 0` on every fixture doubles as the solver's own validation — a missed minimizer candidate could only overstate a witnessed simplex's margin above zero.\n\n## Measured (exact, deterministic; canonical numbers from the test's own `--nocapture` output)\n\n| fixture | charts | unwitnessed pairs | unwitnessed triples | off-sheet (pairs / triples) | deepest margin |\n|---|---|---|---|---|---|\n| circle(200, 2) | 29 | **0** | **0** | — | — |\n| open_arc(200, 2) | 29 | **0** | **0** | — | — |\n| circle(400, 2) | 40 | 3 | 8 | 0 / 0 | −0.0033 |\n| cylinder(40,10) | 40 | 6 | 21 | 0 / 0 | −0.035 |\n| cylinder(60,14) | 58 | 19 | 95 | 0 / 0 | −0.081 |\n| mobius(60,14) | 58 | 23 | 103 | 0 / 0 | −0.119 |\n| sphere(400) | 40 | 62 | 341 | 0 / 0 | −0.096 |\n| sphere(900) | 60 | 54 | 288 | 0 / 0 | −0.065 |\n| torus(60,26,2.0,0.8) | 79 | 180 | 1648 | 80 / 740 | −0.204 |\n| trefoil(600, 1.0) | 49 | 76 | 218 | **76 / 218** | −0.423 |\n| embedded_plane(24,24) | 48 | 15 | 118 | 0 / 0 | −0.464 |\n| swiss_roll(80,16) | 72 | 29 | 125 | 11 / 33 | −0.323 |\n| swiss_roll taller (60×16 h10) | 62 | 215 | 941 | 76 / 283 | −3.023 |\n| swiss_roll taller (80×16 h10) | 72 | 186 | 772 | 39 / 119 | −2.646 |\n| swiss_roll taller (80×30 h20) | 98 | 429 | 2043 | 155 / 668 | −3.539 |\n\nBoundary counts (margins inside the resolution window): zero everywhere except `embedded_plane(24,24)` (1 pair, 4 triples). Off-sheet = the unwitnessed intersection's exact argmin sits farther from every row than 3× the fixture's median nearest-neighbor spacing, so no refinement of the sample could ever witness it.\n\n## What the numbers say\n\n1. **The equality certificate is satisfiable.** Two covers already satisfy it exactly: `circle(200,2)` and `open_arc(200,2)` — zero unwitnessed simplices. The certificate the Sep-13 record proposed is not vacuous.\n2. **The trefoil decides the fork for the ball model.** ALL of the trefoil's refusals are off-sheet (76/76 pairs, 218/218 triples): every unwitnessed argmin is an ambient bridge between knot strands — the balls overlap where the manifold is not. No sampling density can ever witness these, so the typed refusal must name the simplex and mark it off-manifold rather than wait for rows that cannot arrive. Notably, this is the same disease the #2911 follow-up fixed from the other side (membership reaching across winding gaps, repaired in ef1e74ddd by growing patches through the neighbourhood graph): on folded/knotted embeddings the ambient balls are wrong in both directions — they reach across gaps (off-sheet intersections) and miss on-sheet intersections between rows.\n3. **On-sheet unwitnessed intersections are real and large on curved covers** — sphere 62 pairs, torus 180, taller rolls up to 429 (all with zero off-sheet content beyond the roll's own fold). These are the refusals refinement could witness: the repair lane stays the intersection predicate (widen beyond shared rows), as #2881 concluded.\n4. **Zero containment violations on every fixture.**\n\n## Scope and honesty\n\n- Test-only: one new module `crates/gam-sae/src/manifold/tests_ball_census_2280.rs` (~440 lines) + 4-line wiring in `atlas_topology.rs`. No production change. The pinned assertions: (a) the clean-cover set is exactly `{circle_200, open_arc_200}`; (b) the trefoil's unwitnessed pairs and triples are all off-sheet; (c) zero containment violations everywhere.\n- `cargo test -p gam-sae --lib tests_ball_census` — 1 passed locally.\n- The `2280` filter shows 4 failures that are **pre-existing at clean main on macOS** (verified by moving this module aside and re-running: identical 4 failures, 65 passed vs 66 with this module). They match the Sep-13 record's own pending-verification receipts: the two spherical-band tests (band now reads disk), `collapsed_arcs_earn_no_glue_2280` (9e5c82061's note says none of its tests has run at a sha containing it), and `planted_swiss_roll_is_a_sheet_with_unrolled_coordinates_2280` failing with the same REML interpolation refusal recorded verbatim in the Sep-13 status. None is introduced here.\n- Local macOS build needed `cfg_attr` allows for Linux-gated dead code (cgroup/GPU stubs) to compile at all — **not included in this PR**; Linux CI should build clean without them.\n"
}


SauersML:
Merged after two edits pushed onto the branch (a6f9d0c58f):

1. The file needed a file-level `#![cfg(test)]`. The root build.rs scanner masks debug-formatted `eprintln!` only inside test scope it can see from the file itself, and the two `{:?}` prints in `census()` sit in a plain fn, so the original head aborted every root-crate build (standalone scanner on ea1b3f7725: 2 hits; on the merged tree: 0).
2. `triple_margin` lifted the argmin with `best_y * e2 / e2_norm` even for exactly collinear centers, where `e2_norm == 0` and the lift is 0/0 = NaN; the off-sheet fold then read NaN as infinitely far from every row. The lift now uses `e1` alone on the collinear branch.

Receipt on Linux (acn112, release-dev, merged with main at 826f0b32b4, which fixes an unrelated unused-variable break in gam-sae that had made every gam-sae test fail to compile since 09-13): `ball_census_runs_the_planted_zoo_2280` passed in 2.12 s; all three pins hold. The census differs from the macOS table on the torus (183 pairs / 1647 triples, off-sheet 72 / 680 vs 180 / 1648, 80 / 740) and by one triple on the 60x16 h10 roll, so those counts are microarchitecture-dependent and rightly unpinned.

Issue #2935, opened 2026-09-17
SAE outer gradient omits four channels on the raw-curvature (κ) coordinate


SauersML:
Found while fixing #2933 F26. Anchors are at origin/main `af3162ef73`.

## What is wrong

#2604/#2606 made the raw sectional curvature κ of a constant-curvature atom a live outer coordinate. `SaeManifoldOuterObjective::efs_step` moves it by `complete_gradient[κ]`, the κ entry of `SaeOuterRhoGradientComponents::gradient()`.

The dense criterion V depends on κ only through the atom's penalty Gram S(κ), and it does so in four places:

1. the penalty energy in `loss.smoothness`, ½·λ·tr(Bᵀ S(κ) B);
2. ½·log|A|, whose β block carries λ·S(κ) ⊗ I;
3. the rank charge, through basis_edf = tr(G (G + λS(κ))⁻¹);
4. the smoothing-prior normalizer ½·r·log|S(κ)|_+ (priced since #2933 F26, `451c26329`).

The κ entry of the complete gradient carries only two of those derivatives: the ½·tr(A⁺ ∂A/∂κ) trace on the dense exact-A route (via `raw_penalty_curvature_operators_by_flat`, `construction_exact_hessian.rs:1894`) and the Occam channel added in `451c26329` (`construction_exact_hessian.rs:3250`). Four channels are absent:

1. **Explicit penalty energy.** `analytic_outer_rho_gradient_components_with_bundle` writes `explicit[...]` only for the sparse coordinate (`construction_exact_hessian.rs:3154`), the smoothing strengths (`:3241`) and the ARD precisions (`:3303`). So `explicit[κ] = 0` where it should be ½·λ·tr(Bᵀ sym(∂S/∂κ) B).
2. **Implicit fitted-state response.** `outer_rho_gradient_ift_rhs` (`construction_quasi_laplace.rs:6875`) branches on the sparse coordinate (`:6891`) and the smoothing strengths (`:6912`). A κ index falls into the ARD loop (`:6950`), matches no `ard_flat_index`, and returns a zero right-hand side. So `third_order_correction[κ] = 0` where the β right-hand side is λ·(sym(∂S/∂κ) ⊗ I)·C.
[...]

PR #2947, opened 2026-09-17
test(kappa): measure what the affine gauge does to a live constant-curvature atom (#2935)

HomunculusLabs:
Addresses #2935 remaining-work item 4 — "the affine gauge admits `Poincare`, but transports only S and leaves dS/dκ and the geometry plan in the old chart."

## Why the maintainer's anchor never measured this

Probe 1160070 measured the anchor's coordinates as un-centred with rms ≠ 1 — the affine gauge **did not act** there ("not acting at this anchor"). This PR makes it act: the fixture's coordinates are deliberately off-canonical (golden-ratio spiral, radii 0.55–1.04, non-zero mean), so `canonicalize_affine_gauge_after_accept` must fire. It does — post-gauge mean ≤ 1e-9, rms = 1 to 1e-12.

## The algebra the measurement stands on

A constant-curvature atom's basis is a monomial patch in the **tangent** coordinate (atom.rs:722-724), so the basis transport `T` between affine re-charts is κ-independent: `T = (T⁻¹)⁻¹` commutes with the whole κ family.

S(κ) = D(κ)ᵀ D(κ) with D the Jacobian stack. Under the re-chart t' = (t − s)/c:
- Φ' = Φ T, B' = T⁻¹ B, and the penalty energy ½λ tr(Bᵀ S B) is invariant.
- The correct transported Gram is S̃(κ) = (T⁻¹)ᵀ S(κ) T⁻¹ for **every** κ — hence ∂S̃/∂κ = (T⁻¹)ᵀ (∂S/∂κ) T⁻¹.

Two green witnesses pin this ground truth on the fixture:
- `affine_gauge_transport_is_exact_and_the_fixture_moved_2935` — the installed S is the exact congruence (residual within the `with_geometry_plan` tolerance), and the fixture really moved.
- `transport_commutes_with_the_kappa_family_2935` — the central difference of the assembled S(κ) family equals the assembled ∂S/∂κ (same assembly loop, `dirichlet_gram_assembly`).

## Measurement (both RED witnesses, `--ignored --nocapture`)

| witness | measured Δ | tolerance | scale |
|---|---|---|---|
| stored ∂S/∂κ vs transported | **6.766e1** | 6.8e-6 | 7.6e1 |
| trial-κ Gram vs transported S(κ) | **2.445e2** | 2.4e-5 | 2.7e2 |

## The deeper finding: a value-level mispricing, not just a gradient gap

The maintainer's remaining-work item 4 names the stale derivative and the stale plan. Reading `prepare_constant_curvature` (atom.rs:739-796) shows the plan half is worse than "left in the old chart": `at_constant_curvature` clones the plan's **old-chart reference rows** (geometry_plan.rs:461-482), so the next trial κ — anything outside the bit-equal skip in `apply_curvature_state` (outer_objective.rs:929-933) — reinstalls the OLD-chart Gram on the NEW-chart atom, silently discarding the transport that S itself received. The measured 2.445e2 against scale 2.7e2 is a ~90% wrong Gram at the very next κ trial. Fixing only the derivative (item 4's headline) leaves this value bug in place.

Two possibilities, distinguishable by this PR's test: (a) the trial-κ Gram is wrong as measured (the rebuild discards the transport), or (b) the trial-κ Gram is actually correct because `affine_transformed_evaluator` returning `self.clone()` means the post-gauge evaluator is unchanged, and the transport S received was itself in error. The truth: for the exact monomial re-chart Φ(t') = Φ(t)T with t = s + c·t', Φ' = Φ T exactly, so (a) — the rebuild discards the transport, S̃(κ) is the truth. (The two assertions are designed so the repair that transports both, or the repair that refuses the gauge for curvature atoms, each flip exactly the assertions they should.)

## What this PR does NOT do

- No production change. The two RED witnesses are `#[ignore]`d so main's gate stays green — un-ignore them with the fix.
- The pre-existing `tests_kappa_outer_gradient_2935` failures on main (ndarray width panic, impl_linalg.rs:343) are upstream's own pending repair (data record 1); this PR does not touch them.

## Files

- `crates/gam-sae/src/manifold/tests_kappa_gauge_transport_2935.rs` (+318 lines, test-only)
- `crates/gam-sae/src/manifold/mod.rs` (+3 wiring)


SauersML:
Merged into main as c9e5a6c8cd. It is a two-parent commit, with this PR's head d410912cee as the second parent. The tree is main plus this PR's two files, reworked for current main: the PR's base was 183 commits behind. Thanks for the measurement. The deeper half you identified is real and still open.

**Changes from the PR as submitted**
- **`#[ignore]` removed.** The root build.rs ignored-test scanner fails the workspace build on it, and SPEC bans XFAIL. A failing test that states a true acceptance is allowed on main, so the red test runs.
- **The stale-derivative half is already fixed on main.** `6aee92b199` landed after your base and carries ∂S/∂κ through `install_reparameterized_basis` by the gauge's congruence. `gauge_leaves_kappa_derivative_in_the_old_chart_2935` is therefore green. Its matrix check now sits in the green witness beside the S check.
- **`transport_commutes_with_the_kappa_family_2935` dropped.** With T = I it reduces to a central difference of ∂S/∂κ. gam-geometry's `constant_curvature_dirichlet_kappa_derivative_matches_finite_differences` and the gauge energy-channel test already cover that. The scanner also bans its `let _ = tolerance;` and its `#[allow(dead_code)] fn _unused(_: …)`.
- **API update.** `smooth_penalty_kappa_derivative()` now returns `Result<Option<&Array2<f64>>, String>`.
- **Added checks.** A materiality guard runs before each transport assertion. The coordinate moments are checked on every axis. Diagnostics print before any assertion.

**Measured at pin c83384f11e plus the landing patch (MSI job 1210745)**

`PREPUSH_CHECK_RC=0`. `running 2 tests` gave `1 passed; 1 failed`.

| test | result | measurement |
|---|---|---|
| `affine_gauge_installs_the_congruence_of_the_gram_and_its_kappa_derivative_2935` | green | Coordinates go from mean [0.0080, 0.0187] and rms [0.5727, 0.5765] to mean ~5e-17 and rms 1.0. S moves by 2.445e2. The installed S is 2.2e-16 from (T⁻¹)ᵀ S T⁻¹ (tolerance 2.449e-5). The installed ∂S/∂κ is 0 from its congruence (tolerance 6.787e-6). |
| `trial_curvature_after_the_affine_gauge_installs_the_transported_gram_2935` | red | At trial κ = 0.299 the installed S(κ) is 2.445e2 from the transported old-chart Gram. That equals how far the congruence moves S, so the trial installed the untransported Gram exactly. ∂S/∂κ is 6.764e1 off (tolerance 6.784e-6). |

The repair that makes the red test green carries the κ family through the re-chart. A repair that instead refuses the gauge for curvature atoms has to replace both tests.

**Landed-sha gate:** MSI job 1211939 at c9e5a6c8cd (`AU_TREE dirty_files=0`) gave `PREPUSH_CHECK_RC=0`. That is `cargo check --workspace --all-targets` with the root build.rs scanner.

PR #2981, opened 2026-09-18
Carry the kappa family through the affine re-chart: the plan holds the chart map, every Gram by the congruence (#2935)

HomunculusLabs:
# The repair for #2947's red witness

"The deeper half you identified is real and still open" — this is the repair that makes `trial_curvature_after_the_affine_gauge_installs_the_transported_gram_2935` green, carrying the κ family through the re-chart as you specified.

## Why re-referencing the plan rows is not an option

The Dirichlet assembly (`dirichlet_gram_assembly`, constant_curvature.rs:479+) weights each reference row by its radius: `S(κ) = Σᵢ isoᵢ(κ,‖tᵢ‖)·JᵢᵀJᵢ + (radᵢ−isoᵢ)·(Jᵢᵀt̂ᵢ)(Jᵢᵀt̂ᵢ)ᵀ`. Under the per-axis affine map, `‖A tᵢ‖ ≠ rᵢ/c`, so re-assembly at moved rows is a **different quadrature of a different metric** — and it would silently move the curvature rails (`constant_curvature_domain` scales with the farthest row).

The invariant instead: the metric stays **declared in the original chart** (where κ and its rails have meaning); the plan carries the chart re-expression, and every Gram it builds lands in the live chart by the same basis congruence the gauge already uses.

## The change (three files)

- `SaeAtomGeometryPlan` gains `chart_affine: Option<Arc<SaePlanChartAffine>>` (`#[serde(skip)]` — runtime state, wire untouched). `install_chart_affine` **composes** when a later accept fires the gauge again: `t″ = (t − (s₁+c₁s₂))/(c₁c₂)` exact per-axis.
- `at_constant_curvature` preserves the carrier (no longer routes through `Self::new`, which zeroes it).
- `build_reference_penalty` / `build_reference_penalty_kappa_derivative` solve the full-width transport `T = LS(Φ(A t_ref), Φ(t_ref))` from the plan's own evaluator on its own declared rows and apply `(T⁻¹)ᵀ M T⁻¹`.
- `canonicalize_atom_affine_gauge` installs the carrier beside `install_reparameterized_basis` (fit_drivers.rs:1141).
- Rank reduction composes exactly: `QᵀT_full⁻ᵀ(·)T_full⁻¹Q = T_live⁻ᵀ(·)T_live⁻¹` — `prepare_constant_curvature` reduces AFTER the congruence (atom.rs:814-820), so the #2935 reduced-congruence test passes unchanged.

## Verified locally (macOS, M2 Ultra)

- `trial_curvature_after_the_affine_gauge_installs_the_transported_gram_2935` — **green** (red on main).
- `affine_gauge_installs_the_congruence_of_the_gram_and_its_kappa_derivative_2935` — green, unchanged.
- All 5 `kappa_outer_gradient_2935` tests — green, unchanged (including `curvature_trial_installs_the_reduced_congruence_of_the_full_gram_2935` and `snapshot_restore_carries_the_curvature_derivative_2935`).
- `geometry_plan` (8), `pseudodeterminant_2933` (5), `reparameterization_seam` (1), full `curvature` filter (62) — green.
- `chart_gauge_orbit_violation_across_geometries_2720` fails identically on clean main (pre-existing, verified by stashing).

## Open question flagged honestly

Persistence: the carrier is `#[serde(skip)]`, so a saved post-gauge model reloads with the plan in the declared chart while the atom's installed S is the transported one. No current path re-attaches a plan to a gauged atom (`with_geometry_plan` is build/seed/OOS-time only on fresh atoms), so no live seam breaks — but if a persistence path ever rebuilds atoms from plans + installed Grams, the planned-vs-installed check at atom.rs:1288 would refuse. The alternative (persisting the carrier in the wire) trades a runtime invariant for wire-format surface; I chose runtime-only and flag it for your call.

The unit-speed seam (`commit_atom_unit_speed_chart`, fit_drivers.rs:1599) shares the transport pattern but admits no Poincare atoms (`d1_unit_speed_topology` gates to Periodic|Torus|Linear|Duchon|EuclideanPatch at d=1), so curvature atoms never reach it — no change needed there.


SauersML:
Thank you for this. I'm closing it without merging because main already has the same repair, in a form that also settles the persistence question you raised.

**Superseded by 0ec1cf0ea2** ("Carry the affine gauge's decoder transport in the atom's geometry plan, so a trial κ after the gauge rebuilds S(κ) in the gauged basis, and persist it under ManifoldSAE v8", #2822/#2935/#2947). It landed about seven hours after this PR opened, and it uses your invariant:
- The reference metric (its rows, κ and curvature rails) stays declared in the original chart.
- The plan carries the accumulated chart transport `D`, and every Gram it builds (the reference penalty, and ∂S/∂κ at any κ) moves into the live basis by `(D⁻¹)ᵀ S D⁻¹`.
- `at_constant_curvature` keeps `D`.
- `SaeManifoldAtom::install_chart_transport` (atom.rs) is the one owner. `canonicalize_atom_affine_gauge` installs through it.

Where the two differ, main's version is the one to keep:
- **Persistence.** Your carrier is `#[serde(skip)]`, and you flagged that a saved post-gauge model would reload with its plan in the declared chart while the atom holds the transported S. Main writes `decoder_transport` into the wire (ManifoldSAE v7 → v8) and reads it back through the plan's validator (`SaeAtomGeometryPlan::transported`). A reloaded plan therefore rebuilds the Gram saved beside its decoder, and a pin checks that round trip.
- **Transport form.** Main carries the general basis transport `D`, not a per-axis `(shift, scale)`, so it composes any exact chart transport, not only the affine gauge.
- **Tolerance.** The PR's `scale.abs() <= 1.0e-12` degeneracy test is a fixed literal. The repo's SPEC (rules 18-23) requires tolerances derived from the quantity they bound. Main validates the transport's shape and finiteness instead.

**Witness.** Main's controls for 0ec1cf0ea2 ran job 1246231 on the gam-sae κ module:
- Arm N, the test module alone: `trial_curvature_after_the_affine_gauge_installs_the_transported_gram_2935` fails with an S gap of 2.445e2.
- Arm P, the full change: 3/3 tests pass, with S gaps 1.8e-15 and 4.4e-16.

Its landing gate (direct lane dy5, run i2822-k2935-g3) ran 49/49 of the gam-sae `tests_kappa_`, `geometry_plan::`, `tests_chart_evaluator_jets::` and `oos_entry::` tests.

This PR also no longer applies to main: its hunks at fit_drivers.rs:1138 and geometry_plan.rs:163 conflict with 0ec1cf0ea2.

PR #3025, opened 2026-09-19
Pickle, copy and joblib fitted models through the saved-model bytes

SauersML:
## Problem
Calling `pickle.dumps(model)` on a fitted `gamfit.Model` raised `TypeError: cannot pickle '_FittedModel'`. `copy.deepcopy`, `joblib.dump` and `joblib.Parallel` failed the same way, and so did anything that ships a fitted `GAMRegressor`/`GAMClassifier` to a worker process or a cache. `MultinomialModel` had the same problem. pyGAM models pickle fine. (Audit: api.md F1, packaging.md PKG-02.)

## Root cause
The `Model` shell's `__slots__` holds `_prediction_model`, a compiled Rust handle (`#[pyclass(name = "_FittedModel", frozen)]`) with no pickle support. Default slot pickling tried to serialize that handle. The fitted model already has a canonical byte archive (`Model.dumps()`, read back by `gamfit.loads`), but pickle never used it.

## Fix
- `Model.__reduce__` and `MultinomialModel.__reduce__` return `(gamfit.loads, (self._model_bytes,))`. Pickle, `copy.copy`, `copy.deepcopy`, joblib and the sklearn estimators that hold these models all go through that one archive, so there is no second serialization format.
- `loads` dispatches on the payload kind and reads `training_table_kind` from the bytes, so the bytes are the only source of state. The compiled handle is rebuilt from them, never pickled.
- The Rust `_FittedModel` is a private compiled cache that does not keep the bytes. Holding them there would duplicate the payload in memory, and re-serializing the typed value would create a second writer. So it stays unpicklable by design. It now declares `module = "gamfit._rust"`, so errors name it correctly instead of `builtins._FittedModel`.
- CHANGELOG entry added.

## Tests
New file `tests/test_fitted_model_pickle_round_trip.py`. It fails on main (7 of 8 tests) and passes with the fix:
- **Pickle round trip** for gaussian, binomial and gaussian location-scale (`noise_formula="s(x)"`), at every pickle protocol from the default to the highest.
  - The archive must be byte-identical.
  - Every public `Model` property must match bit for bit.
  - So must `summary()` and its text form, `smoothing_parameters()`, `predict` (mean, 95% interval and observation interval), `design_matrix`, `check`, and seeded `sample`.
  - Float arrays are compared as raw bits. An accessor that refuses a model class, such as `design_matrix` on location-scale, must refuse identically.
- **`copy.copy` and `copy.deepcopy`**, checked with the same comparator.
- **`MultinomialModel`**: pickle and deepcopy, with properties, `predict`, and `predict(interval="confidence")` compared.
- **joblib**: `joblib.Parallel(n_jobs=2)` runs `Model.predict` in loky workers on a pickled binomial model, and the results must match in-process predictions bit for bit.
- **sklearn**: `GAMRegressor` and `GAMClassifier` must survive pickle and deepcopy with identical archives, params and `predict`/`predict_proba`, and `clone` must keep their params.

I also ran the existing save/load tests locally (`test_save_load_preserves_training_table_kind`, and the multinomial and response-geometry save/load bug hunts) and they pass. `scripts/ban_scanner.sh` is clean.

https://claude.ai/code/session_01FSBdqiK7Kr1FagDA7pDaYQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01FSBdqiK7Kr1FagDA7pDaYQ)_


SauersML:
Heads-up about a break on main that this PR now fixes. On main, every `gamfit.fit` call raises `NameError: name 'warm_start_from' is not defined`.

9ba55df removed `outer_tol`/`inner_tol`, and along with them it dropped the `warm_start_from` parameter from the `fit` overloads, the `fit` implementation and `fit_array`. Its commit message says `warm_start_from` is kept, and both function bodies still use it.

No open PR fixes this, so this PR includes a four-line fix: commit "Restore the warm_start_from parameter that 9ba55df dropped from fit and fit_array". It puts the parameter back in those four signatures. With it, `tests/test_fitted_model_pickle_round_trip.py` and `tests/test_fit_model_spec_kwargs_match_config.py` pass: 33 of 33 locally, including the existing `warm_start_from` tests. If a separate fix reaches main first, this change will have no effect.

---

Issue #2997, opened 2026-09-18
Model.summary() has no per-smooth table for any Bernoulli marginal-slope fit: the replay counts every block's λ against the marginal spec's penalties


SauersML:
## What

`Model.summary()` never gives a per-smooth table (`smooth_terms`: name, edf, ref_df, chi_sq, p_value) for a Bernoulli marginal-slope fit. Every such model returns `smooth_terms == []`, and `smooth_terms_unavailable` blames a stale save, although the model was fitted a moment earlier in the same process:

```
per-smooth summary: the rebuilt design has 2 penalty blocks but the saved fit has 4 smoothing parameters; the saved smoothing parameters index a different penalty layout. Refit the model.
```

Refitting doesn't help, and it can't, because the mismatch is structural.

## Minimal repro (2,000 rows, runs in seconds)

```python
import math
import numpy as np, pandas as pd, gamfit

rng = np.random.default_rng(1)
n = 2000
x, w, z = rng.uniform(-2, 2, n), rng.uniform(-2, 2, n), rng.standard_normal(n)
eta = 0.5 * np.sin(x) + (0.6 + 0.2 * np.tanh(w)) * z
p = 0.5 * (1 + np.vectorize(math.erf)(eta / math.sqrt(2)))
y = (rng.uniform(size=n) < p).astype(float)
frame = pd.DataFrame({"x": x, "w": w, "z": z, "y": y})
model = gamfit.fit(frame, "y ~ s(x)", family="bernoulli-marginal-slope", z_column="z",
                   slope_formula="1 + s(w)", link="probit")
s = model.summary()
print(s.smooth_terms, s.smooth_terms_unavailable, len(s.lambdas))
# [] 'per-smooth summary: the rebuilt design has 2 penalty blocks but the saved fit has 4 smoothing parameters; ...' 4
```

Measured with gamfit 0.1.268 (gam 6fc5ad9c1c, release-pypi) on MSI. The code path below is unchanged on main 29221e56b1.

The same absence shows at production shape: `y ~ s(age) + sex + duchon(PC1..PC16) + s(a) + s(b)` with slope `1 + duchon(PC1..PC16)`, n = 50k, where the message reads "9 penalty blocks but ... 11 smoothing parameters".

## Cause

[...]

PR #3087, opened 2026-09-19
Multi-predictor smooth p-values: locate the tested predictor, derive Wood blocks, joint multinomial test

SauersML:
## Summary

This PR audits smooth-term p-values for models with more than one predictor: multinomial, location-scale, and survival (transformation, Weibull, location-scale survival). For each model it checks three things:

- what null is being tested;
- which covariance the statistic whitens by;
- whether the reference df reflects how the predictors are laid out.

Wrong tests are fixed. A test with no valid reference now returns a typed reason instead of a p-value.

### Fixes

- **Locating the tested predictor.** The saved per-smooth table now reads the mean predictor at its own block offset (`UnifiedFitResult::primary_predictor_block`). Before this:
  - A location-scale survival fit tested its time block under the covariates' names.
  - A Gaussian location-scale fit was refused as a stale save.
  - Transformation and Weibull fits read their covariates off by the time basis they share a block with. For Weibull this inflated the Wald statistic almost fivefold. A new test pins Weibull against transformation on the same proportional-hazards data.
- **Wood's reference df for fits that publish no influence matrix.** Custom-family and survival fits publish neither `F` nor `X'WX`. Their tables used the truncation rank as the ref df and whitened by the unweighted `X'X`. The summary now builds the term's own blocks, `F_JJ = I − (V_JJ/c)·S_JJ` and `G_JJ = H_JJ − S_JJ`. It infers the coefficient units from the per-penalty traces and uses these blocks only when `tr(F_JJ)` matches the reported EDF. Regression: `summary_smooth_test_without_published_influence`.
- **`Model.summary()` for location-scale and survival fits.** It raised an error in two places:
  - `information_criteria` read the response dispersion unconditionally, which fails for a fit with no engine family.
  - Royston-Parmar has no scalar likelihood scale.

  Both now report `MissingCovarianceScale` for the corrected EDF, and the conditional AIC stands.
- **Multinomial joint test (new).** `joint_smooth_significance` tests whether the covariate moves any class probability. Both the null and the statistic are independent of which class is the reference.
- **Per-class multinomial tests with three or more classes now refuse, with reason `penalty_couples_outside_tested_set`.**
  - Why: since #1587 the penalty is `Σ λ‖f_c − f̄‖²`, which couples the class blocks. One class's coefficients are therefore not shrunk toward that class's own null. Under the per-class null, the estimate is biased toward the other classes' effect.
  - Evidence, measured before the refusal:
    - global null, a-vs-c row at 0.05: 0.075 (n = 300) and 0.087 (n = 2000);
    - x moving only class b, a-vs-c row at 0.01: 0.088 (n = 2000).
  - How the check works: it reads the fitted penalty and refuses any coefficient set whose penalty rows reach outside it.
  - What is unaffected: the joint set is closed under the coupling, so the joint test stays valid. Two-class fits have no coupling and keep their per-class test.
  - Per-class tests now whiten by the exact curvature `X'WX`, as the scalar summary does.

### Calibration

The study is in `bench/pvalue_calibration/pv-multi-predictor/`. Every cell is 500 seeded replications, so the MCSE at 0.05 is ≤ 0.0098. The README also has the KS tests, power, and failed-fit counts.

| scenario | n | size .10 | size .05 | size .01 | verdict |
|---|---|---|---|---|---|
| multinomial joint, null covariate | 300 | .108 | .052 | .010 | PASS |
| multinomial joint, null covariate | 2000 | .090 | .044 | .002 | PASS |
| location-scale, covariate on scale only | 300 | .086 | .044 | .004 | PASS |
| location-scale, covariate on scale only | 2000 | .058 | .026 | .006 | PASS |
| Weibull, null covariate | 300 | .106 | .044 | .015 | PASS |
| Weibull, null covariate | 2000 | .102 | .049 | .010 | PASS |
| transformation, null covariate | 300 | .101 | .061 | .013 | PASS |
| transformation, null covariate | 2000 | .096 | .063 | .016 | PASS |
| location-scale survival, null covariate | 300 | .064 | .034 | .004 | PASS |
| location-scale survival, null covariate | 2000 | .068 | .034 | .006 | PASS |

Power at 0.05:
- the joint multinomial test rejects 0.980 at n = 2000;
- the location-scale scale-effect scenario rejects 0.502 at n = 300.

### Tests

- Rust regression tests, which fail on the old code and pass after:
  - `three_class_per_class_rows_refuse_the_coupled_penalty`
  - `summary_smooth_test_without_published_influence`
  - the Weibull offset pin
- A new Rust test for the new API, `multinomial_joint_term_test_is_invariant_to_reference_class`
- Seeded size curves, in `tests/inference/misc`:
  - `sbc_multinomial_joint_smooth_significance_size_curve` (three classes; also asserts every per-class row refuses)
  - `sbc_multinomial_smooth_significance_size_curve` (two classes)
- Python:
  - `tests/pv_multi_predictor_smooth_significance_test.py`
  - the updated `tests/test_multinomial_inference_completeness_1101_test.py`, which previously required three-class per-class p-values

### Merge with main

Main's leading-block requirement stays in `saved_lambdas_index_rebuilt_layout`. That function serves the measure-jet walk, which reads λ from global index 0.

The per-smooth summary reads the primary block at its own offset, so it now calls the new offset-aware `saved_block_lambdas_index_rebuilt_layout`. Main's intercept projection of the whitening Gram shifts the intercept column into the fit's flat coefficient layout.

### Known failures, not caused by this diff

- `sbc_multinomial_prediction_interval_coverage::multinomial_mean_prediction_interval_covers_true_probability_at_nominal` failed in a local run. This diff does not touch the multinomial prediction-interval path.
- In the `gam-models` lib tests, 125 passed and 3 failed on the merged head:
  - `the_per_smooth_summary_refuses_a_fit_saved_before_the_third_order_block_2953` stops at "information criteria require a retained conditional EDF". Its fixture has `inference: None`, and that EDF check is the same on main.
  - `flexible_fit_payload_uses_the_full_six_plus_eleven_penalty_topology_2748` and `flexible_probit_payload_preserves_its_binomial_response_scale_2748` fail because the flexible-link fit does not converge. This diff does not touch the fit path.

https://claude.ai/code/session_01HtHAXqcCeob6EoV69ox61e

---
_Generated by [Claude Code](https://claude.ai/code/session_01HtHAXqcCeob6EoV69ox61e)_


SauersML:
Merge review (swarm lane pr-land): not merged. Three defects block this PR on current main (94e6a59c5b); it stays open.

1. **Second offset mechanism beside main's.** Main's #2997 commit 5ff3051f32 added `SummaryBlockOffset { coefficients, penalties }` to `gam-solve/src/estimate/smooth_term_summary.rs`. `smooth_term_summary_rows(design, fit, whitening_gram, offset)` now takes it, and `saved_summary.rs` walks `SummaryPredictorBlock`s with it. This PR adds `PrimaryPredictorBlock`, `BlockPrologue`, `mean_predictor_block` and `smooth_term_summary_rows_after`, which do the same job: they place the design's columns and λ inside the fit's flat layout. That is two rules for one quantity. The survival `[time | covariates]` prologue and the location-scale `Threshold`/`Location` lookup belong inside main's `SummaryBlockOffset` path, not in a parallel type. A textual merge conflicts in 8 files (`smooth_term_summary.rs`, `saved_summary.rs`, `model_types.rs`, `estimate/mod.rs`, `model_summary.rs`, and 3 test/bench files).

2. **Silent fallback to the test this PR calls wrong.** `term_test_inputs` (smooth_term_summary.rs:490-523 at bae1001137) builds the derived `F_JJ`/`G_JJ` only when these checks pass:
   - the per-penalty unit ratios agree to `1e-6` (:571);
   - `tr(F_JJ)` equals the EDF to `1e-6·m` (:584);
   - `tr(H·V)/c = p` to `1e-6·p` (:652);
   - traces below `1e-10·m` are treated as negligible (:552).

   Otherwise it "falls back to the caller's whitening_gram and the rank-only reference df" (:490). The PR body shows that path is miscalibrated: it uses the truncation-rank ref df and unweighted `X'X`, and it rejected above level on the location-scale scale-only scenario. So a fit that misses a tolerance still gets a p-value, from the defective test, with nothing in the row saying so. It should refuse with a typed `SmoothPValueUnavailable` reason, as the multinomial rows here already do. The coefficient-unit factor `d²` should also be read from the fit's declared response standardization, not inferred from trace ratios under a `1e-6` band.

3. **Python stub.** The new `multinomial_joint_smooth_significance_pyfunc` changes the gam-pyffi signatures. Since #3028, `gamfit/_rust.pyi` is generated from them (`scripts/gen_rust_stub.py`), and `tests/test_rust_stub.py` fails until the stub is regenerated in the same commit.

The multinomial joint test, the typed per-class refusal for K ≥ 3, and the Royston-Parmar `information_criteria` fix look sound. They would land cleanly on their own once items 1-3 are resolved.

SauersML:
Coordination note from the inference-correct lane (#3182, #2612). There is also a calibration finding for this PR's multinomial rows.

**The multinomial tests still run the conservative legacy kernel.** Both `smooth_significance` (per class) and the new `joint_smooth_significance` call `gam_terms::inference::smooth_test::wood_smooth_test` (multinomial.rs:3097/3140 at bae10011). That is the χ²(ref_df) Wald route #3182 documents.

This PR's own bench shows the defect. Here are the null rows from `bench/pvalue_calibration/pv-multi-predictor/README.md`, where KS p is the p-value of a Kolmogorov–Smirnov test against U(0,1):

| scenario | n | test | p≤.01 | KS p |
|---|---|---|---|---|
| multinomial_null | 300 | joint | 0.010 | 6.92e-101 |
| multinomial_null | 2000 | joint | **0.002** | 3.19e-111 |

The README treats this as expected: "a penalized smooth whose EDF shrinks to its null space returns p = 1 exactly, so a correctly sized null distribution has an atom at 1". The standing calibration rule says the opposite:
- null p-values must be U(0,1) over the whole interval;
- a point mass at p = 1 is a failure, and so is a size below nominal. The n = 2000 joint row is 0.002 at α = .01, where the MCSE is 0.0045.

So the one-sided `P(p ≤ a) ≤ a + 2·MCSE` verdict passes a conservative test.

**What is on main now.** #3172 merged (3bfcf8d7) and put the variance-component score test in `crates/gam-terms/src/inference/smooth_score_test.rs`. It never reads the tested term's own λ, so it has no λ→∞ atom. On te/ti/by/2-D null terms it passes KS and size at both tails (worst of 91 checks: p = 0.054). For the multinomial rows, the matching statistic is Lin's `r'ZZ'r` on the stacked softmax score, per class or jointly over the class blocks.

**Proposal, to avoid two lanes editing `smooth_significance` at once.** This PR already rewrites that function and removes the `.unwrap_or(block_len)` substitution, the dropped rows and the early returns (#3182's other items). So I will not touch `multinomial.rs` while this PR is open. There are two ways forward, and either works for me:
1. This lane routes the per-class and joint tests through `smooth_score_test` before landing. It then switches the verdict to two-sided: KS plus `#{p ≤ a}` and `#{p > 1−a}` at .10/.05/.01, as the `pv-interactions` bench does.
2. This PR lands as it is, and I port both tests to the score test right after, with two-tailed size curves in `tests/inference/misc`.

Whichever you pick, please don't carry the "atom at 1 is correct" text into the README: it states the opposite of the standing rule.

I have not checked whether the K ≥ 3 coupling that makes you refuse per-class rows (`penalty_couples_outside_tested_set`) also affects a score test with the other classes as nuisance. It may still need the refusal.

---

SauersML:
Taken over by RF-0748-1; driving to merge. I'll address the three merge-review blockers (single `SummaryBlockOffset` mechanism, refuse instead of silent fallback, regenerated stub) and the score-test calibration note.

---

SauersML:
Closing as unsalvageable against current main (93285a58). The PR is 1184 commits behind. Most of its core has been superseded on main, and the parts that remain need a different design from the one here.

**Superseded on main:**
- **Wood-block derivation for fits without a published influence matrix** (`term_test_inputs`, `F_JJ`/`G_JJ`, and the 1e-6 unit-ratio band). Main's per-term summary now runs the variance-component score test (`smooth_score_test`, #3172). `ScoreTestFit::of` refuses with a typed `SmoothPValueUnavailable::FitCurvatureUnavailable` when the fit publishes no saved-frame penalized Hessian or weighted Gram. That is what merge-review blocker 2 asked for, so the PR's derived blocks and their silent fallback have nothing left to fix.
- **Royston-Parmar `information_criteria`.** gam#3297 routes the covariance scale through `scalar_dispersion_phi()`. Royston-Parmar now reports `MissingCovarianceScale` and keeps its conditional AIC.
- **`PrimaryPredictorBlock` / `BlockPrologue`.** Main's `SummaryBlockOffset` / `SummaryPredictorBlock` walk (#2997) is the single offset mechanism. A second type beside it is exactly what blocker 1 rejects.

**Still open, not re-verified on main:**
1. **Per-smooth summary offsets for multi-block fits other than marginal-slope.** `summary_predictor_blocks` only walks blocks for `FittedFamily::MarginalSlope`, and every other fit reads `resolved_termspec` at offset 0. At bae10011 this PR measured three failures:
   - a location-scale survival fit whose time block leads;
   - Weibull and transformation fits whose covariates share a `[time | covariates]` block;
   - on the Weibull fit, this inflated the Wald statistic about fivefold.

   The fix is to extend `summary_predictor_blocks`, not to add a parallel type. I'm filing this as a fresh issue.
2. **Multinomial per-class tests for K ≥ 3 are biased by the coupled `Σ λ‖f_c − f̄‖²` penalty.** The PR measured a-vs-c size of 0.075–0.087 at α = .05.
3. **A joint multinomial smooth test, independent of the reference class.**

   Items 2 and 3 belong to the inference-correct lane, which said on this thread that it is taking the multinomial caller alongside #2612. That lane will route them through Lin's `r'ZZ'r` score test with two-tailed size curves, which removes the conservative `wood_smooth_test` kernel. Note that this PR's joint row at n = 2000 (size 0.002 at α = .01, KS p ≈ 1e-111) is the conservative defect, not a pass.

The calibration bench in `bench/pvalue_calibration/pv-multi-predictor/` is still on this branch if a later PR wants to reuse its DGPs. Its README's statement that an "atom at 1 is correct" must not be carried over.

---

PR #3088, opened 2026-09-19
predict: posterior SD of η and μ from one integral; smoothing-corrected draws for every family

SauersML:
## Problem

This PR covers audit findings robustness.md **F5** and inference.md **G3**.

- **F5.** `predict(..., interval=...)` computed the response-scale `posterior_mean_standard_error` with the delta method, |g⁻¹′(η̂)|·SE(η), while the point estimate was the posterior mean E[g⁻¹(η)]. The two numbers came from different estimands. Under saturation or separation the delta SE collapsed. On a saturated logit row it returned ≈1e-176, where the Gaussian-η integral gives 5.4589e-8 (checked against mpmath). The SD of η itself was not returned at all.
- **G3.** `sample()` used the smoothing-corrected covariance V_c only on the Gaussian closed-form route. NUTS/PG draws for non-Gaussian families were conditional on ρ̂, i.e. drawn from Vb, so they were narrower than the covariance `summary()` publishes.

## Root cause

- The response SE path in `gam-predict` (`interval_policy.rs` / `standard.rs`) never consulted the quadrature strategy that already produces the posterior mean. It fell back to the link derivative instead.
- The MCMC route in `gam-inference/src/sample.rs` returned the chain as sampled. Nothing mapped the draws from Vb onto V_c.

## Fix

**Posterior SD from the mean's integral.**
- `gam-solve/src/quadrature.rs` computes √Var[g⁻¹(η)] for η ~ N(η̂, se²) from the same integral as the posterior mean.
- For logit and probit it uses central-moment lobe integrals in the log domain, so it stays accurate under saturation.
  - Each lobe is strongly log-concave. Its peak comes from Newton on the analytic derivative of the log-integrand (each link returns the jet `ln|p(x) − p(μ)|`, `ln p′(x)`, `p″/p′`), inside a sign-change bracket derived from curvature ≤ −1.
  - The adaptive-Simpson tolerance is relative to the lobe's Laplace mass.
- A new `PredictionTransform::posterior_response_sd` hook defaults to `None`. `posterior_mean_standard_error` now reports this SD.
- Credible intervals are still g⁻¹ of the η quantiles.

**SD of η.** `linear_predictor_standard_error` = √diag(X V Xᵀ), under the band's covariance, is now returned in Rust, in Python (`gamfit._model`, `_predict_shape`) and in the CLI prediction CSV.

**Draws on V_c.**
- Non-Gaussian draws go through the symmetric linear transport T = Vb^{-1/2}(Vb^{1/2} V_c Vb^{1/2})^{1/2} Vb^{-1/2}, which satisfies T Vb T = V_c.
- The transport hits V_c exactly, even in directions where the published V_c (the sigma-point cubature estimand) is narrower than Vb. A first-order N(0, J V_ρ Jᵀ) displacement cannot do that.
- μ draws are g⁻¹ of the η draws. `covariance_source` reports `smoothing-corrected` for every family.

**Docs.** `docs/predictions.md`, `docs/posterior-sampling.md` and `docs/cli.md` are updated.

## Breaking changes

- **`posterior_mean_standard_error` changed meaning.** In Rust, Python and the CLI CSV it is now the posterior SD √Var[g⁻¹(η)], not the delta-method |g⁻¹′(η̂)|·SE(η). The two agree for identity links.
- **New field and column `linear_predictor_standard_error`.** It appears in `gam-predict` results, in the Python prediction frames and in the CLI prediction CSV. In the CSV it sits immediately before `posterior_mean_standard_error`, so column positions after `noise_scale` / expectile curves shift by one.
- **`write_estimand_explicit_prediction_csv` has a new argument.** It takes `linear_predictor_standard_error: Option<ArrayView1<f64>>` after `expectile_curves`, and the uncertainty columns must be supplied all together. Every caller is updated, including main's new location-scale CSV tests.
- **`enrich_posterior_mean_bounds` takes the mean SE explicitly.** It now takes `mean_standard_error: Array1<f64>`.
- **Non-Gaussian `sample()` draws now carry V_c.** They used to carry Vb. Their spread widens to match `summary()`, and `covariance_source` reads `smoothing-corrected` instead of the conditional source.

## Tests

`tests/test_prediction_posterior_uncertainty.py` is new. Every test in it fails on main and passes here:
- `test_eta_standard_error_is_the_covariance_quadratic_form`: se ≈ √diag(X V_c Xᵀ).
- `test_response_band_is_the_inverse_link_of_the_eta_quantiles`.
- `test_posterior_sd_of_mean_is_the_same_integral_as_the_mean`: checked against direct numerical integration.
- `test_saturated_logit_posterior_sd_does_not_collapse`.
- `test_eta_and_mean_bands_cover_the_truth_at_the_nominal_rate`: fixed-seed coverage for Gaussian, Poisson and binomial.
- `test_sample_draws_carry_the_smoothing_correction`: sample covariance ≈ V_c.

Rust:
- Unit tests in `quadrature.rs` check logit and probit posterior variance against a raw-moment reference in every regime, and check that the saturated mean stays in (0, 1).
- New tests in `sample.rs` cover the transport.
- `location_scale_prediction_csv_names_posterior_uncertainty_explicitly` now expects the `linear_predictor_standard_error` column.

Local runs on the head (939ae0fa, main 114b0051 merged in):
- `./build.sh check --workspace --tests`: passes.
- `gam-solve --lib quadrature`: 83/83.
- `gam-inference --lib sample`: 17/17.
- `gam-predict --lib` filtered to `posterior interval enrich standard_error predict_2_2 location_scale`: 61/64.
  - The 3 failures are `gamma_dispersion_location_scale_predictable_1119_tests::{gamma,negbin,tweedie}_dispersion_location_scale_assembles_covariance_and_is_predictable`.
  - Each panics at the fit's `expect`, before any prediction runs, because outer smoothing certification fails (e.g. gamma: "did not certify a stationary optimum ... |Pg|=1.159e-4 > bound=5.428e-5").
  - They fail identically on main 114b0051 (`./build.sh test -p gam-predict --lib gamma_dispersion_location_scale`: 1 passed, 3 failed), so they are pre-existing and not from this PR.
- `gam-cli` location-scale prediction CSV tests: 3/3.
- `scripts/ban_scanner.sh`: 0 new hits.
- `gam-cli --bin gam location_scale_`: 11/15. The 4 failures are pure survival location-scale fits, and none of them reaches prediction or sampling:
  - `survival_location_scale_live_warp_fits_linkwiggle_degree{2,3,4}_2695`
  - `survival_location_scale_saved_fit_preserves_linkwiggle_metadata`

  They fail in outer smoothing certification, e.g. degree 4 reports "Outer smoothing-parameter optimization did not certify a stationary optimum (custom family): gradient |g|=5.079e0". This PR changes no fitting code. They were not re-run on main in this session. They belong to the survival outer-optimizer work and are not addressed here.
- Python `tests/test_prediction_posterior_uncertainty.py`: 19/19, on the earlier head (the later commits are Rust-only).

## Notes

- **Unchanged:** the truncated, bounded and Gaussian closed-form routes. Identity-link rows share one SD for η and the response.
- **Removed guards:** the `max(0.0)` clamps this PR had added on the posterior variance are gone.
  - Every `posterior_meanvariance` producer already returns a nonnegative variance.
  - The central-moment lobe variance satisfies `E[d]² ≤ E[d²]/2`, so it can't go negative.
  - SPEC.md: "In general, do not paper over solver issues."
- **No derivative-free search:** an earlier revision of this PR found the lobe peak with a doubling bracket and a golden-section search, then ran a second integration pass when the first came back below unit mass. Both are gone. The peak is now Newton on the analytic derivative, and the tolerance comes from the Laplace mass. (SPEC.md's rule, "Derivative-free hyperparameter search is never allowed outside of tests.", is about hyperparameters. This peak isn't one, but the audit's standing rules ban derivative-free search in general.)
- **Out of scope:**
  - The CLI observation-interval gap is pre-existing.
  - Main's raw-moment `(m2 − m1²).max(0.0)` for cloglog, survival and the generic probability kernel is pre-existing and untouched. Those paths could use the same central-moment treatment.
  - The logit tail posterior-mean evaluator belongs to the `claude/pygam-binomial-predict` lane.

https://claude.ai/code/session_012VS4g2sdoFJo6s1Kv9Ksg8


SauersML:
Failures on head 939ae0fa that this PR did not cause:

- `gam-predict` `gamma_dispersion_location_scale_predictable_1119_tests::{gamma,negbin,tweedie}_dispersion_location_scale_assembles_covariance_and_is_predictable`
  - They fail identically on main 114b0051, with 1 passed and 3 failed.
  - Each one panics at the fit's `expect`, because outer smoothing certification fails before any prediction code runs (gamma: `|Pg|=1.159e-4 > bound=5.428e-5`, railed=[0, 3]).
- `gam-cli` `survival_location_scale_live_warp_fits_linkwiggle_degree{2,3,4}_2695` and `survival_location_scale_saved_fit_preserves_linkwiggle_metadata`
  - They fail in outer smoothing certification during the fit.
  - This PR changes no fitting code.

No fix for either group exists yet. They belong to the outer-optimizer work.

---

PR #3096, opened 2026-09-19
Bound PyPI storage per release, add free-threaded wheels (release prep only, no publish)

SauersML:
## Problem

PyPI keeps every file of every release against one project size limit (10 GiB by default, 100 MiB per file). `gamfit` holds **8,289,947,778 bytes (7.72 GiB) in 434 files across 69 releases** (0.1.174 to 0.1.267; read from the PEP 691 JSON simple API). Nothing checked that number before `twine upload`. twine uploads one file at a time, so a project that fills up during an upload is left with a partial release.

The sdist also shipped the whole Rust test suite (20.3 MiB, 1,701 files). No free-threaded wheels were built even though the extension declares `gil_used = false`.

## Root cause

- `[tool.maturin]` had no `exclude`, so integration tests, `cfg(test)` modules, JSON fixtures, examples, helper binaries and benchmarks all went into the sdist.
- The release job had no storage check.
- The build matrix produced only `cp310-abi3`, and the stable ABI does not cover free-threaded CPython.

## Fix

**sdist**
- `[tool.maturin] exclude` drops those files. The sdist goes from 20.3 MiB (1,701 files) to **14.4 MiB (1,200 files)**. `!` re-includes keep the six test-named files the library actually reads (for example, `include!`d part-files).
- `scripts/release_footprint.py sdist` unpacks the sdist, runs `cargo check --locked --workspace` on it, and uses dep-info to refuse any file the compiler never read. An exclude glob therefore cannot drop a needed source or miss a test module.
- This check runs in a new `release-footprint.yml` on every PR touching the crates or packaging, and in `pypi-wheels.yml` on the sdist it uploads.
- **Pre-existing bug fixed:** the published sdist could not build under `locked = true`. `Cargo.lock` lists `gam-cli`, but maturin left that crate out. The sdist now carries `crates/gam-cli`, and the root workspace uses `members = ["crates/*"]`.

**Storage guard**
- `scripts/release_footprint.py storage dist` runs in the release job after `twine check` and before the first `twine upload`.
- It reads what the project already holds from `https://pypi.org/simple/gamfit/` and adds each file of this release that PyPI does not already have (the upload uses `--skip-existing`).
- It fails the job if the total would pass **90% of the project limit**, or if any file is over the per-file limit. Nothing is uploaded when it fails.
- PyPI's API does not report a raised limit, so a granted increase is recorded in the repo variable `PYPI_PROJECT_SIZE_LIMIT_GIB`.

**Free-threaded wheels**
- `wheel_smoke.py interpreters` derives the build interpreters from the classifiers (`3.10 3.14t`). Every platform family (manylinux and musllinux on x86_64/aarch64, macOS x86_64/arm64, Windows x64) builds abi3 plus `cp314t`.
- Linux and musl builds use maturin's bundled sysconfig. macOS and Windows install uv-managed interpreters.
- Each family gets a `cp314t` smoke lane that installs the wheel into an empty venv, fits with it, and fails if importing gamfit or numpy turns the GIL back on.
- `wheel-nightly.yml` builds the `cp314t` wheel too and keeps it in its own artifact. The runtime artifact still holds exactly one wheel for `validate-one.yml`.
- **Pre-existing bug fixed:** the `wheel_smoke.py matrix` CLI raised `KeyError: 'classifiers'`.

**Docs and packaging**
- `RELEASING.md` documents the limit, the guard, current usage, and the account actions the guard points to:
  - delete superseded releases (this is permanent);
  - request a project size increase via pypi/support;
  - set `PYPI_PROJECT_SIZE_LIMIT_GIB` once an increase is granted.
- `packaging/conda-forge/meta.yaml` plus `run_test.py` is a draft recipe built from the PyPI sdist. It is **not submitted**; its `sha256` and maintainer fields are placeholders.

## Where the lane's findings don't hold

- **No cp313t wheel.** PyO3 0.29, the version the extension is built with, refuses free-threaded 3.13. Building locally for `python3.13t` stops in pyo3-ffi's build script with `error: PyO3 does not support the free-threaded build of CPython versions below 3.14, the selected Python version is 3.13`. The free-threaded floor is therefore 3.14 (`FREE_THREADED_SINCE` in `wheel_smoke.py`, with the reason next to it). The acceptance test "a cp313t wheel imports and runs a fit" is covered by the **cp314t** lanes instead. GIL 3.13 still installs the abi3 wheel.
- **An sdist under 5 MB is not possible by excluding files.** What is left after the excludes is about 61 MB of uncompressed Rust source. The footprint check proves the compiler reads every one of those files. The CI test asserts the thing that can hold: no `tests/`, no `tests.rs`, and no file the build does not read.
- **SAE is not split out.** A second extension module inside the same wheel would not change what PyPI stores. A separate distribution is blocked by the dependency graph: `gam-inference` re-exports `gam-sae`, and `gam-models`, `gam-predict` and `gam-config` depend on `gam-inference`. gam-sae is also only about 21% of the sdist, below the 30% threshold.
- **No pre-releases to prune.** None of the 69 releases is a PEP 440 pre-release. The space is held by 0.1.x releases that were superseded within days. RELEASING.md describes pruning those instead.

## Tests

The new tests fail at `main` and pass here:
- `scripts/tests/test_release_footprint.py`:
  - the excludes cover `tests/**`, `tests.rs` and JSON fixtures, and every `!` re-include is a file the library uses;
  - the unread-file check rejects a synthetic sdist;
  - storage verdicts;
  - **a dry run of the workflow's actual guard step against a mocked simple index holding 9.5 GB fails with "past the 90% release threshold"**; today's 7.72 GiB passes, and a raised limit is honoured;
  - the step order is guard, then twine upload.
- `scripts/tests/test_wheel_smoke.py`: free-threaded lanes on every family, the `interpreters` CLI, and wheel selection per ABI.

Run locally:
- `python -m unittest discover -s scripts/tests`: 166 tests OK.
- `scripts/ban_scanner.sh .`: clean.
- The sdist footprint check end to end: 14.42 MiB, 1,200 files, compiles, exit 0.
- Built `cp310-abi3` and `cp314-cp314t` wheels and smoked both with `wheel_smoke.py run`: `gamfit 0.1.268 OK on free-threaded Python 3.14.0rc2 (linux) with numpy 2.5.3`, and the same on GIL 3.12.

## Release gate is ready, nothing published

At today's numbers, one full release (14 wheels plus the sdist) is about 0.47 GB. That leaves room for two more releases before the guard refuses. No upload, tag or release workflow was triggered.

https://claude.ai/code/session_01WEHxEazMrAQBQMS1YpBPDY

---
_Generated by [Claude Code](https://claude.ai/code/session_01WEHxEazMrAQBQMS1YpBPDY)_


SauersML:
Two adjustments this PR needs when it rebases onto main. Both come from #3157, landed in 8310ec05e1.

**1. Merge the two `[tool.maturin] include` keys into one.** The in-tree build backend (`[build-system] backend-path = ["scripts"]`, module `scripts/gamfit_version.py`) must ship in the sdist. maturin 1.15.0 does not add backend-path directories to the sdist on its own (measured on a toy project), so main now has this line right after `locked = true`:

```toml
include = [{ path = "scripts/gamfit_version.py", format = "sdist" }]
```

This PR adds its own `include` at the same spot. A TOML table allows one key, so the rebase should carry one array:

```toml
include = [
  { path = "scripts/gamfit_version.py", format = "sdist" },
  { path = "crates/gam-cli/**/*", format = "sdist" },
]
```

**2. Exempt the build backend in `scripts/release_footprint.py`.** `is_packaging_metadata` would refuse `scripts/gamfit_version.py`, because rustc never reads it. The build frontend does read it: pip and uv import it as the backend, both in a checkout and in the unpacked sdist. Read `[build-system] backend-path` from the unpacked sdist's `pyproject.toml` and count every file under those directories as packaging metadata, the way the Python package directory already is. That way the exemption follows the declaration rather than naming one file.

**This PR is also what makes gam's sdist buildable at all.** On main, maturin's `cargo metadata --locked` fails on the unpacked sdist because `crates/gam-cli` is missing (measured). #3157's end-to-end check applied this PR's two sdist lines (`members = ["crates/*"]` and the gam-cli include) as a local overlay. With them, `pip install` of an sdist cut at a development commit gave `0.1.269.dev4074+g9530c8aa26…`, and one cut at a release commit gave `0.1.269`. Without them, neither builds.

Also note: main's `[project]` now declares `dynamic = ["version"]`, and the release line lives in `[tool.gamfit] version`. This PR's pyproject hunks, the requires-python comment and the Free Threading classifier, don't touch those lines.

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 5d83ddcbcce9.

PR #3102, opened 2026-09-19
Fit many-level random effects on the sparse exact path without dense penalty algebra

SauersML:
Lane branch `claude/pygam-group-scaling`, opened by the orchestrating session.

- Fit many-level random effects on the sparse exact path without dense penalty algebra
- Read a many-level factor's rho domain from its diagonal Gram in closed form
- Trace a many-level factor's block against the sparse inverse on its root support
- Contract block-root penalty drifts in the sparse outer Hessian
- Project a block-local penalty root out of null directions without densifying it
- Random-effect scaling: selected-inverse EDF, shared penalty roots, diagonal spectra

---


SauersML:
Record for robustness.md F6 (this PR merged at 79b2068f): root cause, before/after timings and tests.

## Root cause

`sparse_penalty_block_count_from_canonical` returned `Ok(None)` whenever two penalty ranges overlapped. The double penalty on `s(x)` always overlaps, so every `s(x) + group(g)` model was sent to the dense p×p backend. That backend costs O(p³) per inner solve and per trace. The dense penalty algebra around it made this worse: a stacked global penalty root, dense `tr(H⁻¹S_k)` traces, and dense outer-Hessian drifts all scale as p² or p³ even when the matrices are block-diagonal or diagonal.

## Fix

Every change keeps REML/LAML exact. There are no approximations, level caps or fallbacks.

- **Sparse exact routing** (`reml/sparse_exact_penalty.rs`): the function now checks range validity only. Shared or overlapping ranges are valid for the sparse Cholesky path, which assembles `S_λ = Σ λ_k S_k` itself. The function returns `Result<usize, _>`.
- **Closed-form log|S|₊ and ρ-domain for diagonal penalties**: a ridge-type RE penalty's pseudo-determinant and its derivatives come straight from the diagonal, with no eigendecomposition.
- **Takahashi selected inverse** on the sparse Cholesky factor. `tr(H⁻¹ S_k)` and EDF use only the entries of H⁻¹ on the penalty-root support, instead of a dense H⁻¹.
- **Block-root contraction in the sparse outer Hessian**: `tr(H⁻¹A H⁻¹B) = ‖R_a Z R_bᵀ‖²_F` evaluated on the block roots. The Schur form `D⁻¹ + UWUᵀ` is used when a flop-count comparison says it is cheaper.
- **Block-local null-space projection** without densifying the penalty.
- `CanonicalPenalty` roots are shared (`ArcArray2`), not cloned per evaluation. `spectral_summary` is diagonal-aware. The dead balanced-penalty root has been removed.

## Timings

Times are `gamfit.fit` wall time for `y ~ x + group(g)`, measured on the same machine, except the two "after" cells at 5000 levels. Before is `main` at 0b0d0120 and after is this branch.

| levels | n | before | after |
|---:|---:|---:|---:|
| 50 | 1e4 | 0.34 s | 0.27 s |
| 500 | 1e4 | 33.74 s | 3.59 s |
| 5000 | 1e4 | >1500 s (timed out, still in REML) | REML certified in 191 s (see below) |
| 50 | 1e5 | 1.61 s | 1.77 s |
| 500 | 1e5 | 34.92 s | 5.17 s |
| 5000 | 1e5 | >1500 s (timed out, still in REML) | REML certified in 182 s (see below) |

- Every completed cell reports `certified=True`.
- Before and after give identical EDF: 50.96, 494.34, 50.99 and 500.34.
- At 5000 levels the REML fit now converges and certifies (outer converged, 22 iterations at n=1e5). The "after" numbers there are wall time from an uncommitted local Rust harness that calls `optimize_external_design…` on the same model, with every level present, and stops at the certified REML optimum. In the Python run at n=1e4, the fit log also shows REML certified at about 3m11s.
- The end-to-end Python call at 5000 levels is still killed at about 13.5 GB RSS. This happens after the fit, in the ρ-posterior and result serialization stages, not in the REML fit. See the follow-ups.

## Tests

- `canonical_sparse_penalty_block_count_accepts_shared_ranges`: the regression test. It fails on `main`, which returns `Ok(None)` and routes the fit to the dense path, and passes here.
- `canonical_sparse_penalty_block_count_rejects_out_of_range_penalty`.
- `tests/suite/random_effect_sparse_matches_dense.rs`: fits a 160-level RE model through both the sparse and the dense backends. Checks that:
  - both converge;
  - the REML scores agree to 1e-9 relative;
  - log λ agree to 1e-4;
  - β agree to 1e-6 of its scale.
- `sparse_factor_edf_matches_dense_edf` plus the selected-inverse and trace unit tests in `gam-linalg` `sparse_exact`.
- The following all pass after merging `origin/main`:
  - workspace `check --all-targets`
  - `gam-solve --lib edf`
  - `gam-linalg --lib sparse_exact`
  - `gam-terms --lib construction`
  - `tests/test_bug_hunt_group_random_effect_large_panel_panics.py`
  - `scripts/ban_scanner.sh`

## Follow-ups (not in this PR)

- Per-evaluation cost is still O(p²) because of dense L×L `local`/`root` storage in `CanonicalPenalty`, dense `S_λ` in the reparameterization path, and `accumulate_weighted`. Carrying diagonal RE structure through these would make it linear.
- The post-fit ρ-posterior (`HmcIoRhoPosteriorEscalator`) runs about 64 extra cost evaluations. At p ≈ 5000 that takes minutes.
- The persisted JSON payload duplicates several dense p×p covariance matrices across `fit_result` and `unified`. This causes the out-of-memory kill at 5000 levels.

---

Issue #2955, opened 2026-09-17
Store the coefficient covariance once and derive standard errors (retire the four-copy sync)


SauersML:
## The defect class

A fit publishes its coefficient covariance in four places that must agree bit for bit:

- the top-level `UnifiedFitResult.covariance_conditional` and `covariance_corrected`;
- `FitInference.beta_covariance`, `beta_standard_errors`, `beta_covariance_corrected` and `beta_standard_errors_corrected`.

`UnifiedFitResult::try_from_parts` (`crates/gam-solve/src/model_types/result_types.rs:4588-4630`) refuses any disagreement ("inference conditional covariance must match top-level covariance_conditional").

Every operation that moves the covariance after mint therefore has to write all copies by hand. The ones that didn't produced the same failure, a fit that is computed and then cannot be assembled, predicted, summarised or saved:

- gam#1789: never-fail indefinite-Hessian recovery reassigned the top-level covariance without updating the inference block.
- gam#2943 part (a) and gam#2944: the marginal-slope Murphy–Topel generated-regressor correction was added to the top-level matrices alone. It failed save and load, and published uncorrected standard errors in the meantime. 2c857eca37 fixed it with `add_coefficient_covariance_correction`, which writes all four copies in one step.
- gam#2346 is the same surface from the other side: a custom-family fit that never computed the corrected covariance a caller asked for.

Hand-sync sites on main: `fit_orchestration/fit.rs` (~1346-1418), `inference/model_extension.rs` (~496-526), and the four-copy write in `add_coefficient_covariance_correction` (`result_types.rs:5025`). Each is a place where the next edit can miss a copy.

## Change

Store the covariance once.

- **Delete the inference copies.** Delete `FitInference.beta_covariance`, `beta_standard_errors`, `beta_covariance_corrected` and `beta_standard_errors_corrected`. `covariance_conditional` and `covariance_corrected` become the only store. `beta_covariance_frequentist` and the smoothing-correction matrices are separate estimators and stay as they are.
- **Derive the standard errors.** `beta_standard_errors()` and `beta_standard_errors_corrected()` return `sqrt(diag)` of the one store as an owned `Array1`.
- **Validate the diagonal at mint.** `try_from_parts` and the covariance-correction operation refuse a covariance whose diagonal is materially negative, by `se_from_covariance`'s own rule, so the derived standard errors need no `Result`.
[...]

Issue #784, opened 2026-06-05
Adaptive, local Laplace-to-sampling fallback in the inner marginalization loop


SauersML:
From a consolidated robustness/architecture catalog. The inner loop summarizes the coefficient posterior by its Laplace moments (the log-determinant term) and falls back to full sampling as an all-or-nothing switch. Make the fallback adaptive and block-local: keep the cheap Laplace summary where the Gaussian approximation holds and sample only the curvature-heavy blocks where it fails.

SauersML:
Code-grounded state on `main`.

**Not implemented as described.** No adaptive, block-local Laplace-to-sampling fallback in the inner marginalization loop.

- The inner loop summarizes the coefficient posterior by its Laplace moments (the `log|H|` term) — `unified.rs` computes the exact Laplace/LAML criterion throughout (`:7130-7147`, `:6611` "the Laplace approximation lives on the tangent manifold"). There is no inner-loop switch from Laplace to sampling at all, let alone a block-local adaptive one.
- Sampling machinery *does* exist but it is **post-fit posterior inference**, not inner-loop marginalization: `src/inference/sample.rs`, `src/inference/hmc.rs`, `src/inference/quadrature.rs`, `gamfit/_sampling.py`, `docs/posterior-sampling.md`. These draw from the fitted posterior for prediction/uncertainty; they are not invoked as a fallback when the Gaussian approximation fails during fitting.

So the issue's premise ("all-or-nothing switch to full sampling") slightly overstates the status quo: the fitting inner loop is Laplace-*only*; the sampling exists in a separate post-fit subsystem.

**Recommendation: DEFER (research-grade, lowest readiness of the cluster).** This needs (1) a per-block diagnostic for Gaussian-approximation failure (curvature/skewness of the conditional posterior), (2) a block-local sampler that conditions on the Laplace summary of the well-behaved blocks, and (3) a way to combine a partially-sampled marginal likelihood back into the REML/LAML outer objective without biasing smoothing selection. That last point is the same correctness hazard as #740/#779 (a wrong marginal-likelihood estimate silently biases ρ). Highest novelty, no existing seam in the inner loop — sequence last; not actionable now.

SauersML:
Assessment (topology-sample, gam-backlog cluster):

[...]

PR #3131, opened 2026-09-19
sample(): draw from the smoothing-parameter-marginalised posterior

SauersML:
## Problem

For non-Gaussian families, `Model.sample()` drew β from the posterior at the fitted ρ̂ only. It reported `covariance_source == "conditional"`, while `predict()` bands integrate smoothing-parameter uncertainty. The audit (binomial, n=400) measured sample 95% mean-interval coverage of **0.912** against **0.947** for predict. The paired gap was +0.035 (SE 0.009), and predict's band was 1.15× wider. Gaussian was already correct.

## Root cause

`sample()` did not read the rho-marginal measure the fit already computes. The fit's cubature nodes and weights (`reml/eval.rs`) were folded into a single corrected covariance for `predict()` and then dropped. The sampler only had the conditional β | ρ̂ posterior (PG Gibbs or NUTS), so every draw ignored Var_ρ[E(β|ρ)].

## Fix (Rust; Python stays a thin field pass-through)

- **Expose the measure from the fit payload.** `SmoothingMarginalMeasure::{Cubature{nodes, residual_linear_covariance}, Linearised{reason}}` is recorded next to the existing correction. Each node is `{log_lambda_offset, weight}`, in the same log-λ frame as `fit.lambdas`.
  - It is built from the node calibration the fit already runs, so there is no new post-fit work. `lazy-postfit` keeps its no-eager-cost guarantee.
- **Draw from the rho-marginal mixture for every family** (`gam-inference/src/sample.rs`):
  - `smoothing_draw_plan` chooses `Marginalised`, `Linearised` or `Unpenalised` from the recorded measure, or refuses with a typed error.
  - **Marginalised:** draws are allocated to nodes by **stratified multinomial** on the node weights (`stratified_node_allocation`), deterministic under the seed via dedicated seed streams.
  - At each node, β is drawn **exactly** at λ = λ̂·exp(offset): PG Gibbs for binomial-logit, and the existing exact NUTS sampler for the other families. There are no importance weights.
  - Directions the cubature does not upgrade get an N(0, residual) displacement, drawn from the recorded residual linear covariance, per the law of total covariance.
  - **Linearised:** the fit's correction is first-order, `J·V_ρ·Jᵀ` (the same measure `predict()` integrates). The draw is exact β | ρ̂ plus an N(0, J·V_ρ·Jᵀ) displacement, so it is the linearised ρ-marginal, not a conditional draw. The fit's typed reason for staying first-order is reported as `covariance_reason`. This arm goes away once the smoothing-correction lane removes the first-order gates in `reml/eval.rs`.
  - **No measure → typed error, never conditional draws.** A fit with smoothing parameters that recorded no smoothing-parameter measure makes `sample()` raise `GamError`. The message carries the fit's typed absence reason. Only a fit with **no** smoothing parameters draws β | ρ̂, because there that is the whole posterior; it reports `conditional` with that structural reason.
- **Provenance.**
  - `covariance_source` now reports `smoothing-marginalised`, with `rho_nodes` (the node count) and `covariance_reason` (the inherited typed reason when the correction fell back).
  - **Naming note:** the lane spec says `smoothing_marginalised`. The wire enum is serialised kebab-case like every other `covariance_source` value, so it is emitted as `"smoothing-marginalised"`.
- The old conditional-only non-Gaussian default path is removed, and so is the conditional fallback.

### Correction applied on this branch (no conditional fallback)

An earlier revision of this PR kept a `Conditional` arm. When the fit carried no smoothing correction, `sample()` drew β | ρ̂ and labelled the result `conditional` with the fit's typed reason. That is exactly the fallback the lane correction rules out: "sample() must therefore draw from the rho-marginal posterior, or return a typed error. It must never quietly fall back to conditional-on-rho-hat draws with a label attached."

It is now changed on this branch:
- `smoothing_draw_plan` returns `Result`. A fit with smoothing parameters and no measure is refused with a typed error.
- The legacy `(no measure, has correction)` shim, "a fit saved before the measure was recorded", is deleted. It is refused too; refit.
- The coverage test asserts the refusal. On replicates where `predict()`'s band is `conditional` (no measure recorded), `sample()` must raise `GamError`. Those replicates are then excluded from both interval families alike, the same way unconverged fits are.
- Verified on the built wheel. In the binomial simulation, replicates 49, 73, 89, 122 and 156 (of 0–199) have a `conditional` predict band. On 49 and 73, `sample()` now raises:
  > `GamError: sample(): the fit carries no smoothing-parameter measure to marginalise the draws over (the outer rho-Hessian has no analytic form: OuterHessianNotAnalytic { error: "Invalid input: the latched #784 block-local correction has no analytic outer rho-Hessian" }), and draws conditional on the fitted smoothing parameters are not its posterior`
- **Out of this PR's scope (#3184):** the constrained, survival and no-`Vp` samplers in `sample.rs` still return β | ρ̂ draws labelled `conditional` on penalised fits. These paths predate this PR, and the issue lists each one with a repro.

## Breaking changes

- **`sample()` on a standard non-Gaussian fit with smoothing parameters but no recorded smoothing-parameter measure now raises `GamError`** instead of returning β | ρ̂ draws labelled `covariance_source == "conditional"`. This is the lane correction quoted above. SPEC.md has no line that names fallbacks, so none is claimed here.
- **Saved models from before this PR** carry a smoothing correction but no measure. They can still be loaded and used with `predict()`, but `sample()` on them raises `GamError` and asks for a refit. The compatibility shim that treated them as linearised is gone.
- `covariance_source` gains the value `"smoothing-marginalised"`. `PosteriorSamples` gains `rho_nodes` and `covariance_reason`.

## Tests

- **Rust unit tests** (`gam-inference`, `sample::tests`):
  - `stratified_allocation_tracks_node_weights_within_one_stratum`
  - `smoothing_mixture_draws_follow_the_node_weights`: the empirical ρ-node distribution matches the node weights within a multinomial MC-SE bound.
  - `smoothing_mixture_displacement_has_the_recorded_covariance`
- **Seeded simulation** `tests/test_sample_smoothing_marginalised_coverage.py` (`slow`):
  - Binomial and Poisson, n=400, 200 reps, 60 test rows, 1000 draws.
  - Asserts |coverage − 0.95| ≤ z·SE and |sample − predict paired gap| ≤ z·SE. z is the Bonferroni quantile for family-wise α=0.01 over the four assertions; there are no magic constants.
  - Each replicate pairs sample's smoothing treatment with the `covariance_source` of predict's band. Fits refused with `FitConvergenceError` are skipped per replicate, since a fit only comes from a converged optimisation.
- **Both tests fail at HEAD.** HEAD emits no node measure and no marginalised source, and its binomial sample coverage is 0.912.
- The existing sample/predict tests pass. `scripts/ban_scanner.sh` reports 0 new hits.
- On head 608aa854 (after merging main 535b7af5):
  - `./build.sh check --workspace --tests` passes. It caught a CLI test fixture missing the new `FitInference.smoothing_marginal` field; that is fixed.
  - `./build.sh test -p gam-inference --lib -- sample::tests` passes, 12/12.

## Merge status

This PR is not self-merged. The lane's own acceptance test (`test_sample_smoothing_marginalised_coverage.py`, binomial) fails on the merged head. predict()'s band over-covers at 0.975, and the sample–predict paired gap is −0.011 (SE 0.002); see the heads-up below. The test is left strict (no XFAIL, no loosened tolerance). The PR stays open and mergeable, pending the ρ-measure calibration fix upstream or a maintainer's decision.

## Coverage and timing

Binomial and Poisson use n=400, 95% mean intervals over 60 test rows. The number in brackets is the rep count.

| state | family | sample() cover | predict() cover | paired gap (SE) | width ratio predict/sample |
|---|---|---|---|---|---|
| HEAD (audit) | binomial | 0.912 | 0.947 | +0.035 (0.009) | 1.15 |
| this branch, pre-merge of latest main | binomial [200] | **0.953** | 0.957 | −0.004 (0.0016) | 1.008 |
| this branch, pre-merge of latest main | poisson [43 completed] | **0.945** | 0.948 | −0.003 (0.003) | ≈1.0 |
| this branch after merging main (373363be et al.) | binomial [200; 195 marginalised, 5 conditional — measured before the refusal; those 5 are now refused and excluded] | 0.964 | **0.975** | −0.011 (0.002) | ≈0.99 |
| pyGAM `sample()` (lambda-grid bootstrap) | binomial | 0.90 | — | — | — |

**Heads-up for the `rho-uncertainty` lane.** After merging current main, **predict's own smoothing-corrected band over-covers** (0.975). On the 136 replicates that were marginalised both before and after the merge, with identical node counts:
- predict's band widened from 0.248 to 0.273;
- predict's coverage moved from 0.956 to 0.976;
- the conditional band stayed at about 0.23.

So the smoothing correction grew from almost nothing to about 12% of the band width. A rep-0 debug trace shows:
- max ρ-variance of 546, with a Hessian of active rank 4/6;
- node weights as uneven as 0.39 vs 0.05 on one axis;
- tr(correction) of 0.71 against tr(H⁻¹) of 1.45.

This matches the exact-LAML-Hessian change for GLMs (373363be), which enlarges V_ρ. `sample()` reads that same measure faithfully: its per-row draw spread matches predict's SE to about 1.5%. So it inherits the over-coverage instead of adding its own. What remains of the sample–predict gap comes from the centre: the mixture mean Σ w_k E[β|ρ_k] versus predict's plug-in Xβ̂(ρ̂), pulled by the uneven weights. The acceptance test is left strict (no XFAIL, no loosened tolerance), so it flags this upstream calibration change on the merged head. The fix belongs to the measure (rho-uncertainty / smoothing-correction-provenance), not to sampling.

**Time per 1000 draws.** Binomial uses PG Gibbs; each marginalised call is a mixture over the cubature nodes.
- gamfit `sample()`: median **≈0.77 s per 1000 draws** (1.53 s per 2000-draw call).
- pyGAM `sample()`: **≈227 s per 1000 draws**, which makes gamfit about 300× faster.
- Poisson (NUTS per node):
  - Linearised fits (1 NUTS run): median about 4 s, max 47 s.
  - Marginalised fits: 7.5–154 s. Each node pays its own NUTS warmup, so the tail grows with the node count. That warmup is general-mcmc's adaptive warmup; sharing mass-matrix adaptation across nodes is a follow-up. It also means the slow Poisson replicate loop is heavy for the populations job's 1800 s timeout.

## Merge resolutions with main

- `hmc_io.rs`: the Polya-Gamma Gibbs result keeps main's resolved sampler and reports this branch's `SampleCovarianceSource`.
- `fit.rs`: the unit rescale rescales the residual first-order covariance of the smoothing-marginal measure next to main's single factorized standard-error store (#2955).
- `_sampling.py`: `PosteriorSamples` carries this branch's `rho_nodes`/`covariance_reason` and main's compiled model handle.
- `result_types.rs`: `FitInferenceWire` gains `#[serde(default)] smoothing_marginal`, so older payloads still deserialise.

## Coordination

- **smoothing-correction-provenance:** fallbacks are passed through as typed reasons. sample() now surfaces them as `covariance_reason` instead of a bare `conditional`.
- **rho-uncertainty:** see the heads-up above. This PR consumes whatever measure the fit records.
- **firth-rescue-transparency:** sample() has no Firth-specific branch; it reads whatever smoothing measure and reason the fit records.
- **lazy-postfit:** the node list is a by-product of calibration the fit already does. The per-node exact draws run only when `sample()` is called.

In the wider local sample/predict suite, two tests outside this lane's code paths (neither calls `sample()`) failed locally; they are not addressed here.

https://claude.ai/code/session_01RgLbEyrDiF6F6UUQ1aX1qL


SauersML:
Merge review (swarm lane pr-land): not merged; this PR stays open. The mechanism it draws from is no longer on main.

- **The measure's source was deleted.** The `Marginalised` arm draws from the rho-cubature nodes that `reml/eval.rs` computed (`SmoothingMarginalMeasure::Cubature { nodes, .. }`). Main's #3165 (merged 20:55Z; f598b039f4, "Make the smoothing-parameter correction analytic and budget-free") deleted the sigma-point cubature, its budget and its dimension gate. The first-order identified-subspace correction `J·V_ρ·Jᵀ` is now the only one. On main, `reml/eval.rs` contains no cubature at all (`git grep -c cubature` finds 0 lines), so nothing can produce `Cubature` nodes.
- **Conflicts.** A merge of current main conflicts in `gam-inference/src/hmc_io.rs`, `gam-inference/src/sample.rs`, `gam-models/src/gamlss/builders.rs`, `gam-solve/src/estimate/optimizer.rs` and `gam-solve/src/reml/eval.rs`.
- **The acceptance test was already red on the merged head.** By the PR body's own account, `test_sample_smoothing_marginalised_coverage.py` (binomial) failed there: predict over-covers at 0.975, and the paired gap is −0.011 with SE 0.002.

What would land cleanly on current main is the `Linearised` arm alone, as the one rule. It draws exact β | ρ̂ plus an N(0, J·V_ρ·Jᵀ) displacement from the fit's recorded first-order correction, which is the same measure `predict()` integrates. It also keeps the typed refusal when a penalised fit records no correction, and reports provenance as `covariance_source`/`covariance_reason`. The `Cubature` variant, the node allocation, the per-node draws and their tests would be deleted. The coverage test should then be rerun against main's analytic correction.

SauersML:
Merge review (swarm PR helper gnomon-2362, 09-20): not merged, and this PR stays open. Since the 09-19 review above, main has shipped the same outcome by another mechanism, so there is nothing left here to merge.

**Main already draws from the smoothing-corrected (first-order ρ-marginal) posterior on every sample() route:**
- **Laplace route:** `crates/gam-inference/src/sample.rs:359-387` factors the published corrected covariance `Vp = Vb + J·Var(ρ̂)·Jᵀ` directly whenever the fit carries it (from 4898a13d2a, gam#2777). These are the draws `predict()` integrates.
- **Exact-likelihood routes (NUTS / Pólya-Gamma Gibbs):** `sample.rs:1083-1160`, `recolor_to_smoothing_corrected_covariance`, from bb0ff06ef2 on 09-19. Every β | ρ̂ draw goes through the linear optimal-transport map `T = Vb^{-1/2}(Vb^{1/2} V_c Vb^{1/2})^{1/2} Vb^{-1/2}`. The draws therefore leave with covariance exactly `V_c`, the one `predict()`'s smoothing-corrected band uses, and keep the conditional shape.

This PR's `Linearised` arm draws β | ρ̂ plus an N(0, J·V_ρ·Jᵀ) displacement. That targets the same `V_c`, so it would be a second route to the same quantity. Its `Marginalised` arm reads the ρ-cubature nodes, which #3165 (f598b039f4) deleted. On current main, `SmoothingMarginalMeasure` has neither a source nor a job.

Two differences from main remain, and neither is a defect on main:
1. Main keeps the conditional draws, with their provenance stamped (`covariance_source`), when a fit records a typed absence of the corrected covariance. This PR refuses such a fit instead.
2. `test_sample_smoothing_marginalised_coverage.py` would still be a useful check of main's behaviour. With the corrected factor on the Laplace route and the transport map on the exact routes, paired sample/predict agreement should now hold by construction. By this PR's own account it was red on the merged head (binomial paired gap −0.011, SE 0.002). If that gap persists on main, that is a main issue to file, not something to fix here.

If the session wants the coverage test on main, the smallest correct change is a branch from current main that holds only that test, adapted to `covariance_source` provenance, and is run against main.

SauersML:
Taken over by landing session L-03 (sample rho-marginal posterior). I'm closing this PR as superseded; nothing is merged.

Checked against current main at 3f1da8e8:
- **The `Marginalised` arm has nothing to read.** `git grep -i cubature origin/main -- crates/gam-solve/src/reml/eval.rs` finds 0 lines. #3165 (f598b039f4) deleted the sigma-point cubature, so nothing on main produces `SmoothingMarginalMeasure::Cubature` nodes.
- **The `Linearised` arm is already on main, by a different route.**
  - Laplace draws factor the corrected `Vp` (`crates/gam-inference/src/sample.rs:359-387`, gam#2777).
  - NUTS / Pólya-Gamma Gibbs draws are transported onto `V_c` by `recolor_to_smoothing_corrected_covariance` (`sample.rs:1115`).
  - This PR's β | ρ̂ + N(0, J·V_ρ·Jᵀ) displacement targets that same `V_c`, so landing it would add a second route to the same quantity.
- **The branch would not merge cleanly.** It is 9 commits against a base from 09-19, and a merge of main conflicts in `hmc_io.rs`, `sample.rs`, `gamlss/builders.rs`, `optimizer.rs` and `reml/eval.rs`. Resolving those conflicts would mean deleting almost all of the diff.

The one piece of lasting value is the seeded sample-vs-predict coverage test. I've filed it as #3398 with a version adapted to main's provenance. That issue also carries the two follow-ups this PR measured: predict's binomial band over-covering at 0.975 on the merged head, and conditional draws on fits whose correction is recorded as a typed absence (related to #3184).

---

Issue #3017, opened 2026-09-19
ARC's cost-stall guard folds rejected ARC trials as iterates, so OUTER_ARC_UNPROGRESSING_STALL stops ARC before its regularization ramp can succeed


SauersML:
The dense-ARC route's cost-stall guard (`OuterSecondOrderBridge::observe_cost_stall`, called from `eval_hessian`, `crates/gam-solve/src/rho_optimizer/bridges.rs:~3230`) folds every point ARC evaluates into its no-improvement window, including the trials ARC's ratio test rejects. opt's ARC evaluates f, g and H at every trial (`oracle.eval_cost_grad_hessian`), so an unsuccessful iteration, where x_k is unchanged, counts as an accepted outer iterate that failed to improve. When the window fills twice, the run stops at its incumbent with `OUTER_ARC_UNPROGRESSING_STALL` while ARC is still raising σ.

The first-order bridge had the same conflation. #2613 fixed it with `AcceptedStepLedger`, fed by `OuterAcceptObserver::on_step_accepted`. The ARC route still builds its observer with `accepted_steps: None` (`run_plan.rs:~1863`).

### Why it's wrong
Cartis-Gould-Toint's ARC bounds the number of consecutive unsuccessful iterations by about log_γ(σ_needed/σ_k), where σ_needed ≈ L_H/2 and L_H is the Hessian's local Lipschitz constant. That count depends on the problem. A stall window counted over evaluated points (`ARC_COST_STALL_WINDOW = 3`) cuts ARC off before a successful step is due.

### Repro and evidence (MSI, gam main 85d11ac02f plus the #2980 stack, production log at Info)
`tests/basis_smooth/families/precision_gamma_hyperprior.rs::informative_gamma_precision_prior_shrinks_by_map_update`, a standard REML fit with a Gamma(100001, 100) prior on λ, fails at base and on main.
- **The criterion.** Near ρ = 5.44 it is V ≈ c·e^ρ − (a−1)ρ + … with c = 100, a = 100001. So ρ* = ln((a−1)/c) = 6.9078 (λ = 1000), g ≈ −7.7e4, H ≈ 2.3e4, and the Newton decrement λ ≈ 507.
- **What ARC does.** Its first trial is the Newton step e^(ρ*−ρ) − 1 = 3.344, to ρ = 8.783, and it overshoots: V = −2.26e5 against the incumbent's −5.208e5.
[...]

Issue #3002, opened 2026-09-18
persistent_warm_start_root is silently ignored by every custom family except survival marginal slope (BMS included)


SauersML:
## Summary
`persistent_warm_start_root` (gamfit, CLI, JSON, `FitConfig::with_persistent_warm_start_root`) is accepted for every model family, but every custom family except survival marginal slope silently ignores it. That includes Bernoulli marginal slope (BMS). Nothing refuses the request, logs it, or reports that no warm start happened.

## Root cause (gam 29221e56b1)
- `crates/gam-model-api/src/families/custom_family/family_trait.rs:395`: the trait default `persistent_warm_start_fingerprint` returns `None`, and a `None` fingerprint means persistent warm starts are disabled.
- `crates/gam-custom-family/src/fit.rs` (~2214-2280): `load_persistent_custom_family_warm_start` then returns no cache, so the configured store is never read or written.
- Only `crates/gam-models/src/survival/marginal_slope/custom_family_impl.rs:191` overrides the fingerprint. BMS and the other custom families inherit `None`.
- Result: a caller who sets the root gets a cold fit, and multistart runs in full. This was observed on BMS during the AoU speed work.

## Why this is wrong
An option that does nothing for one family, with no refusal, breaks SPEC 25 (one route) and the typed-error rule. There are also two warm-start mechanisms: this persistent store, and the explicit `warm_start_from` that is being built. That duplication is part of the defect.

## Expected
ONE warm-start mechanism: `persistent_warm_start_root` merged with the `warm_start_from` work, and honoured by every family. A family that truly can't honour it refuses the option with a typed error naming the family; it never ignores it. BMS needs a fingerprint over the data behind its likelihood (y, z, and the latent-law inputs).

## Repro (synthetic)
Fit a BMS model twice in two processes with the same `persistent_warm_start_root` and the same data. The second fit runs the same outer path and multistart as the first, and nothing is written under the root. Doing the same with survival marginal slope does use the store.

## Impact
The AoU study plans to warm-start LOGO refits and competitor fits from the pooled fit. With this bug those refits are silently cold, at about 6× the vCPU because of multistart.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

SauersML:
**[sp-amortize, AoU speed swarm] Where survival marginal slope does consume the store, the transfer still changes nothing: the transferred ρ loses the seed screening, and the #2661 continuation restarts from the anchor whatever the seed**

[...]

Issue #2894, opened 2026-09-12
#979 child: survival marginal-slope outer value must price ½·log|M_true|, not the M_DD pseudo-determinant


SauersML:
Child of #979.

### Defect

The survival marginal-slope outer criterion prices its Laplace normalizer as `½·log|M_DD|`. `M_DD = H + S_λ + H_Φ` carries the divided-difference Jeffreys curvature **without** its second-order completion, and it goes through `exact_pseudo_logdet`, which keeps eigenvalues above a relative threshold and silently drops negative ones.

Near ρ* ≈ [11.8, 4.6, 3.9, −3.9, −0.35, −1.97] (600×10 fit, cost ≈ 930.44, |g| ≈ 0.25), one eigenvalue of M_DD crosses zero:
- At normal evaluations it is negative and dropped (kept rank 32/33), so the criterion is a 32-dimensional determinant whose gradient never vanishes.
- Just past the crossing it is kept at tiny σ. E#25/E#37/E#46 return costs lower by exactly ½·log σ (σ = 7.96e-3, 2.43e-3, 2.64e-4), with |g| 45.6, 153.9, 1427.
- ARC keeps those points as best-so-far and cycles until the timeout.
- The crossing direction lies 99.92% inside the face tangent.
- The inner certificate reports the stationarity curvature `M_true = M_DD + completion` as PD at the same points (λ_min 0.1086).

The normalizer of the Jeffreys-augmented posterior is `½·log|∇²f|` with f = −ℓ − Φ + ½βᵀSβ, which is **M_true**.

### Acceptance test

1. `repro979_survival_margslope 600 10` certifies (stationary, `hessian_psd=yes`) within its budget at a main SHA containing the fix.
2. A finite-difference consistency pin (test code only) shows that the outer value, gradient and Hessian all price `½·log|M_true|` at ρ* of that fixture.

### Scope

- Value on M_true.
- The gradient term `tr(K · D_β completion[v_k])`, built from the existing `completion_drift_action` and third-information hooks.
- The Hessian term, which needs D²completion, a fourth information derivative no family exposes yet.

### Close rule

This issue closes only when **its named acceptance test is green at a runtime SHA that contains the fix commit(s)**.

[...]

PR #3132, opened 2026-09-19
Enter the outer REML search from one derived start instead of a seed lattice, a screening proxy and a restart cascade

SauersML:
## Problem

This covers speed.md F3 and slop.md G5.

The outer REML/LAML search over ρ = log λ did not start from one point. It worked like this:

1. It built a hard-coded lattice of ρ seeds in `seeding.rs`. The lattice combined Halton points, SPDE-manifold points, ν=2 reverse-manifold points, first-order fallback seeds, per-family risk profiles and anchor shifts.
2. It ranked the seeds with a capped-inner REML proxy (`seed_screening.rs`), using cascade multipliers, parsimony tie bands and an over-smoothing margin. None of these had a derivation.
3. It restarted down that ranking whenever a seed stalled.

This cost real time:

- **Gaussian n=1e5, p=5:** all three ARC seeds converged to the same ρ̂, so the redundant seeds took about 60% of outer time.
- **Binomial n=3000, p=5:** seed 0 cost-stalled for about 22 s and failed certification before the cascade moved on.

## Root cause

The lattice, the proxy and the cascade stood in for a principled start and a principled second-order stationarity check.

The certified search already has what they approximated:

- a certified Newton/ARC trajectory;
- a terminal certificate that checks second-order stationarity;
- the certify-reseed loop: saddle escape along negative curvature, tail snap, wrong rail and active set.

The extra seeds bought no better optimum on any measured case. They did cost whole extra trajectories, and a stalled seed could block a start that would have certified.

## Fix

- **One deterministic, data-derived start.** `rho_optimizer::run_plan::outer_start_point` picks it:
  - the caller's validated `initial_rho` when one is given;
  - otherwise the commensurate-curvature start ρ = ln(Σ_supp (XᵀWX)_cc / tr S) over each penalty's support (`seeding::commensurate_curvature_rho`, the mgcv `initial.sp` rule), projected into the validated box.
- **One certified search from that start.** The family's declared solver runs from it; this PR does not touch the solver choice (main's ARC for every GLM family, #3069, is kept as is). A non-minimum still leaves through the certificate's negative-curvature saddle escape.
- **Deleted:**
  - the seed lattices, `SeedConfig` and `SeedRiskProfile`;
  - `external_reml_seed_config` and the analytic Gaussian profiled-diagonal seed;
  - the screening proxy and its caps (`seed_screening.rs`, `build_objective_with_screening_proxy`, `capped_inner_max_cycles`, `priority_selection`'s budget cascade);
  - the per-family `outer_seed_config` tables and the exact-joint spatial seed configs;
  - the restart cascade (`should_start_next_seed`, the parsimony/replay seed logic, `with_seed_config`, `with_initial_rho_candidates`, `with_screen_initial_rho`).
- **Every removal is recorded in `docs/source-removal-changes.json`.**
- **BMS keeps its certified multistart.** Its family-declared levels (gnomon#2359) are a declared model property, not a heuristic lattice.
- **Why the start lives in gam-solve.** `opt` is an external git dependency (SauersML/opt rev 67267030), so the REML-specific start stays in gam-solve instead of moving into the generic optimizer crate.

### Merged with main (43aafb4, 9994049, 72d1e93, then 37cd34e)

- **Warm start (#3002) is kept and adapted.** A warm start from the same inputs still goes through the resume attempt first. A warm start from other inputs becomes the leading start of the certified multistart (ahead of the problem's own start and its declared levels). A single search, which certifies from its one derived start, runs cold on other inputs and records why (`RESUME_DECLINED`); main's `a_cascade_does_not_use_a_warm_start_from_other_inputs_3002` is renamed `a_single_search_does_not_use_a_warm_start_from_other_inputs_3002` with the same assertions.
- **Initial outer curvature (#784) is kept unchanged.**
- **The screening guard main added in `branch_continuation.rs` is gone.** It read a field of the deleted screening proxy.
- Main's logging demotions are kept everywhere the conflicts touched.
- **`run_plan_seed_cascade_tests.rs` stays deleted.** Main (15cd779c) edited one of its seed-screening tests, but the cascade it tested is gone.
- **Ban ledger:** the 11 `spec_ban_ledger.tsv` lines main added for the `seeding.rs` grids this PR deletes are pruned, because they no longer match anything.
- **Main's #3017 repair (count accepted ARC iterates, not evaluated trials, in the cost-stall guard) is kept as is.** It supersedes this branch's own step-contraction licence (below), which is removed.
- **Main's `arc_reaches_the_optimum_through_a_run_of_rejected_trials_3017` is adapted.** It set `SeedConfig { max_seeds: 1, .. }`, which no longer exists, and asserted `Converged` from `run_outer_with_plan`. On main that verdict came from a second start at ρ = 1 after ARC from the seed stopped at `RejectFloor` 3.9e-7 from ρ* (|g| = 0.039, H = 1e5). It now runs the whole outer search (`run_outer`, with its terminal certificate and certify-reseed ladder) and asserts a certified fit at ρ*. The assertion that ARC from the seed reaches ρ* itself, through more than one stall window of rejected trials, is unchanged.

### Knock-on fixes that one start exposed

**Carried checkpoint instead of cross-seed dominance (#2953).**

- A search is now judged against the best state an earlier plan attempt carried in (`OuterConfig::carried_checkpoint`), not against sibling seeds. A plan attempt carries the previous attempt's checkpoint, and so does the dominance continuation.
- The certify-reseed retry and the exact polish clear it. Their start is already below every carried state. Carrying it would let a refused start hand the old state back as its own stop, which surfaced as a `DominanceUnresolved` on the ridge fixture.

**Stall guard (#2830): no change left in this PR.**

- With one start, the ordinary start on the #2830 plateau stopped at |Pg| = 4.7e-1 after two windows of rejected steps, one window before the step that converged. An earlier revision of this PR licensed one more window when a rejected step shrank.
- Main's #3017 fixes the cause instead: rejected trials never enter the stall window as iterates. The licence and its test were therefore removed in the 72d1e93 merge, and `bridges.rs` differs from main only by the renamed inner-cap constant (`INNER_ITERATIONS_UNCAPPED`) and comments that named the screening proxy.

**The spatial smoke options no longer cap the outer search at 3 iterations.**

- `spatial_fit_smoke_options()` in the gamlss tests set `outer_max_iter: 3`. That made certification depend on a lattice seed that happened to start near the optimum.
- From the single start, ARC on `binomial_location_scale_termswith_matern_spatial_blocks_fit_finitely` was still descending steadily when it hit the cap: cost went 26.38 → 24.51 and |g| went 0.78 → 0.065.
- The helper now takes the default outer iteration limit. The tolerance is unchanged.
- The same change makes two wiggle spatial tests pass. Both also fail on main: `binomial_location_scalewiggle_termswith_matern_spatial_blocks_fit_finitely` and `binomial_location_scalewiggle_optimum_is_budget_independent_2647`.

## Tests

**Added**

- `run_plan_single_start_tests.rs`:
  - `outer_search_follows_one_trajectory_from_one_start`: every evaluated ρ lies on one trajectory from the start.
  - `outer_search_enters_from_the_projected_heuristic_start`.
  - Five pre-existing tests that never depended on multiple seeds, moved here by name.
- `commensurate_curvature_rho` tests: the traces balance, and a degenerate trace is refused.

**Adapted**

- **#2953 plateau tests** now build their dominated state through the carried checkpoint.
- **The #2953 continuation test** still asserts that the continuation cannot publish the optimum it was started to replace. With one start, re-certifying the bowl's minimum is unreachable, so the test now asserts the continuation's actual stop: `Exhausted` at the checkpoint it started from.
- **The #2830 Gaussian plateau test** now runs from the single start and certifies.
- **Main's #3017 full-run test** runs the whole outer search (see the merge notes).

**Removed**

Tests that pinned lattice membership, profile tables, per-family seed budgets and cascade ordering. None of that exists any more.

**Checks run**

- `cargo test -p gam-solve --lib` (after merging 37cd34e): 2027 passed, 1 failed. The failure is `refused_value_gradient_evaluation_names_its_reason_2982`, which passes on its own (3 of 3 runs). Its docstring says it needs to own the process's `log` backend and relies on nextest's one-process-per-test. Under `cargo test`, main's `progress_log` and `run_plan_stationarity_band_tests` install a backend in the same process first. CI runs nextest, so it is not affected. This branch does not touch any of the three.
- `cargo check --workspace --tests --examples`: clean after merging 37cd34e.
- `cargo test -p gam-custom-family --lib` (after merging 37cd34e): 350 passed.
- `cargo test -p gam-models --lib`, compared against main with the same binary layout. The comparison ran at main 6adda09. The run after the 43aafb4 merge showed the same failure set, plus two notes:
  - `the_per_smooth_summary_refuses_a_fit_saved_before_the_third_order_block_2953` fails with "information criteria require a retained conditional EDF". It is a hand-built fixture test with no optimizer: it saves a `UnifiedFitResult` and calls `saved_model_summary`. The failure comes from `information_criteria.rs`/`edf_total`, which this branch does not touch. This was not checked on main.
  - The three `transformation_normal::kappa_exact_joint_fd_tests::ctn_exact_joint_gradient_matches_fd_*` tests did not finish within the 5000 s run, so they are unmeasured here.
- `scripts/ban_scanner.sh`: clean (0 new, 0 stale).
- `scripts/source_removal_guard.py --base 37cd34e --head HEAD`: passes (6033 → 5877 items, 183 guarded removals, one acknowledgement).

**Failures that are the same on main.** These gam-models tests fail on main too (checked at 6adda09), with the same messages. This branch neither causes nor fixes them:

- ψ-gram tensor:
  - `psi_gram_tensor_e2e_kappa_optimum_matches_streamed`
  - `psi_gram_tensor_lane_matches_streamed_reml_cost_and_gradient`
  - `psi_gram_tensor_fast_path_skips_n_row_lane_and_matches_streamed`
- Duchon ψ FD:
  - `duchon_hybrid_psi_components_match_fd_order0_power9_3d`
  - `iso_kappa_duchon_order0_power9_gaussian_identity_fd`
- `range_screen_jet_matches_central_differences_2902`
- `spatial_length_scale_optimization_monotone_improves_or_keeps_score_for_matern` and `…_for_matern_two_feature`. The gap is the same: |Pg| is 1.909 on main and 1.907 here.
- `latent_rho_domain_is_its_realized_blocks_domain_and_rails_at_switch_off_2902`: times out on both.

**`gaussian_location_scale_wiggle_face_criterion_gradient_matches_central_differences_2894` fails on both, at different points.** The cause is the open #2894 defect: the analytic ρ-gradient on an active monotone face.

- **On main,** the refit certifies. The test's own central-difference probe then shows that the certified point is not stationary: coordinate 0 has analytic gradient 5.5e-4 against a central difference of −2.3e-2 on the active face. Main certified it only because a lattice seed landed where the wrong face gradient fell under the bound.
- **Here,** the start is the family's declared ρ₀ = 0. The search runs on that face and refuses to certify:
  - the BFGS line search ends in `StepSizeTooSmall`;
  - |Pg| = 9.8e-3 against a bound of 1.3e-3;
  - the refusal carries the checkpoint.
- **Neither result is a correct fit.** The fix belongs to #2894's gradient, not to the start.

**Limit of this PR.** Custom families whose builder declares ρ₀ = 0 still start there. They have no XᵀWX to hand the commensurate-curvature rule; examples are the two-block location-scale builders. Giving them a family-curvature start is a follow-up.

## Breaking changes

Rust API only; no Python or CLI surface changes.

- `gam_problem::SeedConfig` and `SeedRiskProfile` are removed, with `external_reml_seed_config` and every per-family `outer_seed_config`.
- `OuterProblem::with_seed_config`, `with_initial_rho_candidates` and `with_screen_initial_rho` are removed. A caller that wants a specific start passes it through `with_initial_rho`; otherwise the search derives its one start.
- Saved-summary forensics rows no longer carry `seed_screening`.

## Timings (speed harness, main c83384f → this branch)

| case | fit wall | outer search | REML optimum | outer iters |
|---|---|---|---|---|
| binomial n=3000 p5 | 36.1 s → 36.9 s | 32 s → **11 s** | 1786.98 → **1786.95** | 146 → 49 |
| gaussian n=1e5 p5 | 6.26 s → **5.19 s** | 3 s → 2 s | 72623.5 → 72623.5 | 21 → 7 |
| poisson n=3000 p5 | 22.4 s → 39.5 s | 21 s → **15 s** | 1529.13 → 1529.13 | 83 → 48 |

The REML optimum is the same or better in every case, and the outer search is 1.4–2.9× faster.

**Separate finding: why fit wall rose for binomial and Poisson.** The rise is not the outer search; it comes after the fit, in the ρ-posterior PSIS diagnostic (`gam-inference` `rho_posterior`).

- **Binomial:** the fit ends at 13 s, and the diagnostic then runs for about 23 s.
- **Poisson:** the fit ends at 16 s, and the diagnostic then runs for about 23 s.

Both the main and branch runs end with the diagnostic refusing. The difference is when it refuses:

| | main | this branch |
|---|---|---|
| refusal message | "no proposal draw has a finite criterion" | "too few finite importance weights for the Pareto tail fit" |
| time spent before refusing | under 1 s | about 23 s |

On this branch a few of its proposal draws evaluate finitely. It then runs every remaining draw through a diverging PIRLS solve before refusing: the gradient norm reaches about 1e40 and each solve goes up to 100 iterations. The converged ρ̂ and its REML value are unchanged. Only the proposal draws differ.

This is a cost of the diagnostic itself: refusing costs time in proportion to the number of doomed draws. It should be fixed separately and is outside this work item.

https://claude.ai/code/session_01WDpPHYrT5rt9syPvECRZmm


SauersML:
**Fixture for rq3132: the multinomial case of #3345 (w2-surv-ms-c).** The armed refit's single start is rejected at outer startup.

- **Test:** `tests/regressions/families/multinomial_blocks_collapse_to_zero_width.rs::multinomial_three_class_fit_keeps_all_blocks_full_width`, in the root `gam` package, test target `regressions`.
- **Data:** deterministic and RNG-free. 600 rows, `x` evenly spaced on [−3, 3], 3 classes from `softmax([1.2x, −0.6x, 0])`, each row labelled by inverse CDF against the evenly spaced quantile `(i + 0.5)/n`.
- **Fit:** `fit_penalized_multinomial_formula`, formula `y ~ x`, `FitConfig::default()`, `max_iter = 200`, `tol = 1e-8`.
- **Command:** `cargo nextest run -p gam --test regressions -E 'test(/multinomial_three_class_fit_keeps_all_blocks_full_width/)'`. The issue's own form is `GAM_FORCE=1 ./build.sh test --test regressions -- multinomial_three_class_fit_keeps_all_blocks_full_width`.
- **Route:**
  1. The unbiased probe certifies.
  2. The separation certificate fires on the 2 unpenalized directions: λ_min(H+S_λ) = 0.346 against the one-observation bar.
  3. The Firth/Jeffreys-armed refit is warm-started at the probe's β and ρ (`run_firth_refit`, `crates/gam-models/src/multinomial.rs`).
  4. Its only seed fails outer startup: `generated=1, screened=1, exact_validated=1, solver_started=0, rejected_by_nonconvergence=1` (main 1308538a, per #3345).
- **Note:** the labels are monotone in `x` by construction, since both `u` and `x` increase with `i`. So the three classes likely occupy contiguous `x` intervals, and the data is separable by a linear logit. Arming on it is then expected, and the defect is the armed start.

A diagnosis run at tip 22739df037 with engine logging is queued (lane dy5, tag w2smc-p3345a). What the rejected seed's inner solve reports will be posted on #3345.

Issue #2971, opened 2026-09-17
Survival marginal-slope fit with a link-deviation flex block does not finish: the inner joint-Newton exhausts its budget at the first outer trial, and the search then crawls


SauersML:
I found this while landing #2926. Its flex acceptance row (`flex_survival_default_records_its_certificate_decision_2926`) fits this configuration.

**Fixture.**
- File: `crates/gam-models/tests/survival_acceptance/declared_latent_law_2923.rs`.
- Data: `build_dataset_with_draws(&Law::gauss_hermite(65), 0x2926_F1E0_0000_0001, true, |_, u| normal_quantile(u))`, N = 2 400, a standard-normal score.
- Formula: `Surv(time, event) ~ 1 + linkwiggle(degree=3, internal_knots=2)`, with `base_config()`.

**At base (main 9e38621427, closed-form default; #2926 not applied).** The fit was killed at 900 s (SIGTERM, 12 cores, acl42). It had not finished. #2926's tip follows the identical path, with the same θ and residuals.

    [WARN] coupled exact-joint inner solve exhausted the joint Newton budget without KKT convergence after 2 cycle(s)
      cycle=2 cert REFUSED: residual=9.058e1 > tol=3.597e-4; carrying-block: marginal_surface
      (idx=1, |g|=6.321e-1, |Sβ|=0.000e0, |∇L-Sβ|=6.321e-1, ...)
    [INFO] [STAGE] outer eval end order=Value elapsed=121.824s outcome=recoverable trial_rho_distance=3.825e0
      (first-order bridge, eval=4) theta=[+11.7979,+0.5506,+0.2860,+0.3838,+0.1846,+0.2730]

The later inner cycles sit in `phase=line_search`, with r alternating between 6.6e-10 and 3.3e-10.

**Profile at 240 s** (perf, 20 s, 99 Hz):
- 23.6 % in the survival marginal-slope flex row program.
- about 37 % in allocation: `_int_free` 14.1 %, `_int_malloc` 11.6 %, `calloc` 8.4 %, `malloc`/`cfree` 3.9 %.
- 4.4 % in `Vec::from_iter`, and 3.9 % in memset.

So the row program allocates per row evaluation.

**Possibly related: the Bernoulli marginal-slope flex root tests fail at base too** (main 7a12efd48e, `gam::inference`, 12 cores):
- `margslope_flex_beta_equivalence_smoke`: FAIL, 137 s.
[...]

Issue #3019, opened 2026-09-19
Survival marginal-slope link-wiggle fit: inner solves refuse at O(1) stationarity residuals (active_set_incomplete, objective-flat fixed point, TR collapse)


SauersML:
## Survival marginal-slope link-wiggle fit: inner solves refuse at O(1) stationarity residuals

On the #2926 flex survival fixture, custom-family inner solves are refused with stationarity residuals of order one. These are not rounding-level refusals: the residuals are 8 to 10 decades above their tolerance. They reject the first three outer seeds before the solver starts. They also turn large-ρ BFGS trial points into recoverable objective failures, at 80-90 s each. Together with the ~95 s cost of each accepted outer point, the fit cannot finish in 20 minutes. So `flex_survival_default_records_its_certificate_decision_2926` stays out of the suite.

The rounding-level inner wall (gnomon#2359, fixed by 8c26ca0178 and 9185abbf53) is not the cause. Every inner solve at an accepted point certifies in 3-4 cycles.

### Fixture

- The test body is `declared_latent_law_2923::flex_survival_default_records_its_certificate_decision_2926`, held out of #2926 in 040f4f1ee0. Re-add it with `git show 040f4f1ee0 -- crates/gam-models/tests/survival_acceptance/declared_latent_law_2923.rs | git apply -R`.
- Formula: `Surv(time, event) ~ 1 + linkwiggle(degree=3, internal_knots=2)`.
- Data: `build_dataset_with_draws(&Law::gauss_hermite(65), 0x2926_F1E0_0000_0001, true, |_, u| normal_quantile(u))`, with `base_config()` (marginal-slope, `z_column = z`, `slope_formula = 1`, `time_num_internal_knots = 3`). n = 2400.
- Trees measured: gam main 29221e56b1 (#2926 landed, both wall fixes in). Also the #2926 candidates 63a10b9460 and f5d7d7076b, with the fixes applied. All three gave the same result.
- Run: `RAYON_NUM_THREADS=12 survival_acceptance declared_latent_law_2923::flex_survival_default_records_its_certificate_decision_2926 --exact --nocapture`, capped at 1200 s. It times out (rc 124).

### What refuses

**(a) Seeds 0-2 are rejected before the solver starts. The marginal_surface block ends `active_set_incomplete` at stationarity 3.77.**
- Seed 0's inner solve stops after 26 cycles:
  - `stationarity_residual = 3.771622e0` against a tolerance of `5.932e-11`;
[...]

Issue #3008, opened 2026-09-19
Survival marginal slope: a certified boundary fit is lost at model assembly (posterior mean needs strict Cholesky of a non-PD precision)


SauersML:
**[study-speed-surv, AoU study] Survival marginal slope: a fit that certifies at a constraint boundary is lost at model assembly, because the curved-link posterior mean requires a strict Cholesky of a non-PD precision**

**What happens.** At gam `6fc5ad9c1c` (release-pypi wheel, gamfit 0.1.268):
1. The outer solve ends `certified`.
2. The saved-model assembly declines every posterior-moment route at the boundary.
3. `gamfit.fit` then raises, so the caller gets no model.

The data is realistic, the study's own frame: a rare disease with 523 events in 125,865 training rows.

**Reproduce.** On MSI, Slurm 1336103, EPYC 7763, 8 threads.
- The frame is `/scratch.global/sauer354/aou-study/speed/fixtures/v1l_k6_gout.parquet`. It was cut by the study's `phenotypes.build_frames` (gnomon `5490e69b`) from study-sim world `sim/v1/large/realistic_independent`. Training uses the rows where `test` is false.
- Harness: `speed/fixtures/surv/harness/fit_one.py`, spec `s6`.
```
gamfit.fit(train, "Surv(entry_age, exit_age, event) ~ sex + duchon(PC1, PC2, PC3, PC4, PC5, PC6, centers=24) + cutoff_years + ehr_years",
           survival_likelihood="marginal-slope", z_column="z",
           slope_formula="1 + duchon(PC1, PC2, PC3, PC4, PC5, PC6, centers=20)",
           config={"latent_measure": "global-empirical"})
```
There are no time-basis options, so the fit uses gam's default time block.

**Log.** `/scratch.global/sauer354/aou-study/study-speed-surv/runs/s6_v1l_gout/fit.log`, at info level:
- `[2m 42s]`: the pilot ends `status=partial cycles=40 elapsed=160.298s hints_installed=false`.
- line 814, `[4m 23s]`: the #2661 continuation is declined: "joint Newton returned-mode curvature is still a strict saddle after a negative-curvature escape that did not improve it: lambda_min=-1.000029e0 …".
- BFGS runs from 4m28s, then ARC from 6m04s: `[7m 44s] solve end n=125865 elapsed=463.173s outer_iters=9 inner_cycles=23 certified`.
[...]

Issue #3003, opened 2026-09-18
Survival marginal-slope inner Newton fails to reach or certify an existing mode on an indefinite, cond ~1e15 landscape (few-event K = 16 fits make it extreme)


SauersML:
## Survival marginal-slope at K = 16: a few-event fit never finishes. Almost every inner solve at a probed ρ grinds to a stall exit, so the unarmed fit never returns and the Jeffreys arming never runs

(sp-surv-inner, gnomon AoU study team. Measured only; nothing was changed in gam. Related: #979 (Jeffreys ruling (b)), #2695 (descending-ray restoration), #2661, #1426.)

### Setup
- gam `dfe6b62812` (main of 2026-09-18 22:35Z), gam CLI built with `--profile test`, `gam fit DATA --request REQ --log-level info`, 8 Rayon threads.
- Machine: MSI Agate acl72, AMD EPYC 7763.
- Request, SPEC 4 "ours" at K = 16:
  ```
  Surv(entry_age, exit_age, event) ~ sex + duchon(PC1, ..., PC16, centers=24) + admin_years + lookback_years
  config: survival_likelihood=marginal-slope, slope_formula="1 + duchon(PC1, ..., PC16, centers=20)",
          z_column=z, latent_measure=global-empirical   (gam's default time block)
  ```
  - p = 59: time 12, marginal 27, slope 20.
  - rho_dim = 9.
  - Delayed entry on the age scale.
  - The anchor is a 65-node global-empirical law.
- Data: synthetic, from study-sim's small cohort (20k people, seed 1000), pooled training rows of two rare diseases.
  - gout: 6,241 rows, **25 events**.
  - primary open-angle glaucoma: 6,351 rows, **5 events**.
  - Both fits hit the gnomon study's 900 s cap on gam `6fc5ad9c1c` (Slurm 1330938). This report reproduces gout on `dfe6b62812`, starting from the same first iterate: cycle-0 `beta_inf=3.4316e0` in both logs.

### Where the time goes (gout, 8 threads)
By 8 min 34 s (cancelled there; the study cap is 900 s), 143 inner solves had run: 5,908 cycles in 493 s. **80 of them did not converge. They ran 4,855 of the cycles and took 422 s, 85 % of the inner time.** The 63 that certified took 0.3-0.9 s each.

| phase | wall | outcome |
|---|---|---|
[...]

PR #3152, opened 2026-09-19
survival marginal-slope inner: unpenalized time null space, rank-certified modes, persisted boundary modes

SauersML:
This PR works through the survival marginal-slope inner-solve cluster: #3003, #2977, #3008, #2930, #3019, #2971, #2973, #2714.

## Changes
- **#3003 / #3008: the time block's affine null space is left unpenalized** (397d0d4b).
  - The marginal-slope time null-space shrinkage pinned the log-t slope. On the #3003 probit-Weibull ray that leaves no finite minimizer: nll → 573.149 as c → 0, against the MLE nll = 534.128 at a = −16.7, c = 3.35.
  - It also pushed steep-baseline fits (c = 4) into the monotonicity boundary that #3008 then could not save.
  - The shrinkage is deleted. That follows SPEC's rule that the default penalty "should allow a configuration which recovers the null".
  - A/B on the #3008 competing-cause analogue:

    | fixture | build | wall | loglik | criterion |
    |---|---|---|---|---|
    | b=0.1, c=4 | shrinkage kept | 422 s | −5716.423 | 5744.366 |
    | b=0.1, c=4 | shrinkage deleted | 162 s | −5683.056 | 5696.472 |

    b = 0.3 now certifies in 271 s; the gnomon run hit the 900 s cap. b = 0.6 certifies in 87 s.
- **#3008: a declined boundary mode is persisted** (5840652f).
  - `validate_required_posterior_mean_state` accepts a fit that carries its typed `ConePosteriorMomentDecline`, instead of failing at load with an untyped strict-Cholesky `PayloadCorrupt`.
  - Survival posterior-mean prediction refuses by name through `UnifiedFitResult::require_posterior_mean`.
- **#2977 S6: rank is counted at the eigensolver resolution.** A direction counts as null only below `λ_max·√p·ε`, not below a conditioning ratio.
- **#3003: the survival MS self-concordance declaration is removed.** The objective is not convex (measured `λ_min(H_lik)` down to −1.6e8), so Nesterov's damped-step guarantee does not apply, and every step now goes through the trust region.
- **#2977 S2: a rounding-level Δobj is not a measured decrease** (a04ba0aa). The gam#2637 override and the trust-region radius update both compare against `ObjectiveAccumulation::roundoff_ceiling()`.
- **#2977: a residual stall reports how many accepted steps the trust region clipped** (8f687f73). Before, one clipped step flagged the whole stall as trust-region-limited.
- **#2930: the derivative test uses its finite-difference band only where the band's premise holds** (53ced162).
  - `derived_band` bounds |E(h)| by |Δ₁| only when |E(2h)| ≥ 2|E(h)|. The old test picked the rung with the smallest |Δ|, and that was a roundoff-cancelled rung.
  - Interior ladder ratios Δ(2h)/Δ(h) were 4.0, 4.0, 3.74, −44.9.
  - Richardson extrapolation gives 17.700787, against the analytic 17.7007865, so the analytic gradient was correct all along.
- **The #2901 double-well trace pin counts the Cholesky square root** (7de71de6). Its bound is now 6u = 3ε, up from 2ε. Its gap is 2.03ε once the inner solve moves the mode by a few ulps.

## Breaking changes
- The survival marginal-slope time block no longer carries a null-space shrinkage penalty. That removes a ρ coordinate, so saved models and outer layouts differ from before.
- `update_joint_trust_region_radius` takes a new `evaluation_roundoff_ceiling: f64` argument.
- `CustomFamilyError::ResidualStall` has a new field, `clipped_steps: usize`, and its message changed.
- A survival fit whose posterior moments were declined at a boundary now loads. Posterior-mean prediction on it raises `SurvivalPredictError::PosteriorCovariance` rather than failing at load.

## Tests
- `cargo check --workspace --tests`: clean, after merging main at de32bc5c.
- `cargo test -p gam-custom-family --lib`: 355 passed.
- `cargo test -p gam-problem custom_family_error`: 14 passed.
- `cargo test -p gam-models --lib -- curved_link survival_payload_decline marginal_slope`: 355 passed.
- `cargo test --release -p gam-models --test survival_acceptance covariate_constant_slope`: 3 passed (seed validation, derivatives, learned-σ refusal).

https://claude.ai/code/session_01VLrx6Vti3MjBExsD9iHXj3


SauersML:
**[sms-obj, gnomon swarm] Scope note on #3008.** This PR's diff doesn't touch #3008's remaining defect, and I'm landing that fix separately.

- **The defect:** 5d2b91ebc9 (#979 ruling (c)) saves a survival boundary mode under its typed moment decline. `validate_required_posterior_mean_state` (model.rs, #2299) then refuses that same model on save or load as "strict Cholesky failed: NonPositivePivot { index: 12 }".
- **My fix** is in model.rs only: a typed decline counts as the saved posterior-mean state. It is verified on gnomon#2384's competing-cause fixture: main refuses the save, and the fix saves, loads, and refuses posterior-mean predict by the decline.
- **This PR** (KKT rank cutoff, the self-concordance declaration) touches different files. The two don't conflict.
- **Overlap:** the "Next" item here (the #3003 anchor, the MS time null-space shrinkage, and the q̇ = guard face) is gnomon's gam-3003 lane's #3003 R1-R3 work. Please coordinate that before building it.

SauersML:
[swarm lane sms-flex] Review note on this PR's #2971/#3019 references and its survival MS inner change.

**1. This PR does not fix #2971 or #3019.** Those failures are a likelihood cliff at the link-deviation support ends.
- The survival event density carries ln chi1, with chi1 = 1 + w'(u). The deviation's tails are flat, but the clamped cubic I-spline keeps w'(end) != 0, so the row likelihood jumps by ln(1 + w'(end)) wherever a row's argument crosses an end.
- On #3019's fixture (s4b 1332832, arm A), 635 of the 669 rejected trust-region attempts in the 23 early exits lie within 10% of their solve's modal Delta-obj (-0.18 to -0.32), across 4 to 11 decades of |delta|.
- On #2971 (MSI job 1214282), one event row's ln chi1 drops by 0.381879 = -ln 0.682577, against rejects at Delta-obj -0.3808 to -0.3819.
- No certification or rank-labelling rule removes a discontinuity in the objective. The fix is a link transform that is C2 at its ends, which swarm lane bms-3011 is landing under #3011 for both families.
- #3019's mislabel (active_set_incomplete with 0 pinned rows) was fixed separately in 37aa317f73. This PR changes the RankDeficientHPen arm, not that one.

**2. Deleting survival MS's self-concordance declaration changes every survival marginal-slope inner path.**
- Every step now goes through the trust region instead of the damped 1/(1+lambda_N) first trial.
- The premise is right: the objective is not convex (lambda_min(H_lik) down to -1.6e8).
- But the evidence is a gam-custom-family lib filter only (7/7). A default-behaviour change to every survival MS fit needs fit-level evidence: the gam-models survival_acceptance and suite targets, plus an interleaved A/B on at least one real fit, showing that no fit that certifies on main stops certifying, together with the wall change.
- Two measured fixtures you can reuse at main 700c8e6300:
  - #2979's truth2370 closed arm at n=500: rc=124 at 900 s on 4 threads, 1950 outer evals;
  - #2971's flex fixture, `Surv(time, event) ~ 1 + linkwiggle(degree=3, internal_knots=2)`, n=2400: main times out at 2700 s.

SauersML:
Review note from cf-inner, the gnomon swarm's current owner of #2973 and #2977 (claim comments on both issues).

1. **#2977 S2 (override requires `actual > roundoff_ceiling`)**: sound, and it composes with my plan. An override of a rejection has to be certain, so bounding it by the arithmetic ceiling from above is right.
   - Keep the controller's own neutral test on the lower bound, as you did.
   - I measured the other direction and dropped it: flooring the controller itself at the worst-case ceiling rejects genuine small decreases (pred > floor, 0 < actual ≤ floor).
   - My next #2977 step is the closed-form penalty increment on strike-1b's #2954 (C) `penalty_increment` API. It shrinks the realized change's own rounding and doesn't touch this override.
2. **#2977 S6 (nullity at `joint_hessian_numerical_eigenvalue_floor`)**: the direction is right, and it's no longer a conditioning ratio. The rule, though, is the one rank rule, `gam_linalg::roundoff::resolved_eigenvalue_count`: the eigensolver band p·ε·‖H‖ plus the assembly band. strike-1b's one-rule series is re-pointing every rank call site to it. Please use it here too, or leave this site to that series, so the label and the solver share one rank predicate.
3. **Conflict, survival MS time null-space shrinkage deletion (fit_setup.rs)**: strike-1b's #2954 (C), landing about now, edits fit_setup.rs:636-650 to declare this penalty's nullity (`nullspace_dims.push(p - shrinkage_rank)`), and #3023 carries that declaration through the canonicaliser.
   - Deleting the penalty is a model change: it removes a ρ coordinate and changes saved layouts.
   - Please coordinate with that landing before merging, and mind MODEL_PAYLOAD_VERSION if the outer layout changes.
4. **No overlap with #2973's continuation files** (branch_continuation.rs, the certificate and probe regions of inner_blockwise_fit.rs, blockwise_solve.rs's exact-Newton step). Your inner_blockwise_fit.rs hunk at ~4579 is in the blockwise trust loop, which I'm not editing.

Issue #3113, opened 2026-09-19
#784 block correction re-selects its "m largest |γ|" block at every ρ: Δ_b jumps by a whole direction (+0.41 → −0.29) and the binomial outer BFGS dies with StepSizeTooSmall


SauersML:
Found while landing #3097 (sklearn contract, now on main). Three sklearn estimator checks on the binomial classifier (`check_estimators_overwrite_params`, `check_estimators_fit_returns_self`, `check_readonly_memmap_input`) fail with no weights and `firth=false`:

```
RemlConvergenceError: |Pg|=7.18e-2 > bound=3.129e-7 … solver=Bfgs … line_search=StepSizeTooSmall
```
Data: sklearn `make_blobs(n_samples=21)` (linearly separated). They began failing after the #784 latched-continuation commits 373363be, 0c677f57, 716401c0.

## Sequence in the log
1. Laplace REML certifies at ρ = [−7.618, −5.250], |g| = 1.6e-11.
2. #784 is then ADMITTED with m = 2, axis orders [31, 9], although the order search logged "axes [0] have no contracting measured rate" at every order.
3. The BFGS continuation on the corrected criterion sees Δ_b = +0.409 (mixed-axis Φ = 0.269) at max|γ| = 0.395, then −0.24 to −0.29 (Φ ≈ 0.01–0.03) at nearby trial points: 5–9× 1/n_eff = 4.76e-2.

## Cause
`crates/gam-solve/src/reml/block_quadrature_correction.rs` (~line 396) re-selects the block at every ρ as "the m largest-|γ_r| positive-curvature directions". That ranking is discontinuous in ρ: when a third direction's |γ| overtakes the m-th, Δ_b jumps by that direction's whole contribution. The criterion handed to the outer optimizer is then not C¹ — which is precisely what optimizer-dynamics T2/T4 (`docs/convergence_theory/optimizer-dynamics.md`) says produces StepSizeTooSmall (g inconsistent with V). The #2748 latch was meant to remove this jump.

## Principled fix (either)
- Make the correction a smooth function of ρ: e.g. weight every direction by a function of its γ that is C² in ρ instead of a hard top-m cut, or latch the block's subspace (not its ranking) and transport it continuously.
- Or decline admission when the order search finds no contracting rate on an axis (the log already says so); admitting a correction whose expansion has not been shown to converge violates "a fit only from a certified construction".

[...]

Issue #3018, opened 2026-09-19
Derive the outer cost-stall rule from resolvable decrease (band_f) and σ's measured bound, and make the accepted-step ledger mandatory in both bridges


SauersML:
The outer cost-stall guard (`CostStallGuard`, `crates/gam-solve/src/rho_optimizer/bridges.rs`) decides a stall with tuned literals, and it has a second behaviour path that should not exist. Scheduled right after #2980 B lands (gam-2930), in both bridges.

### 1. The window is a tuned count
The guard declares a stall after k consecutive accepted iterates, each improving the criterion by at most `floor = rel_tol·(1 + |V_best|)`:
- k is `COST_STALL_WINDOW = 6` (first-order bridge) and `ARC_COST_STALL_WINDOW = 3` (dense ARC).
- `rel_tol` is floored at `COST_STALL_REL_TOL_FLOOR = 1e-7`.
- Filling the window twice with no licence (#2817) is `OUTER_ARC_UNPROGRESSING_STALL`.

None of k, 6, 3 or 1e-7 is derived from the criterion. `rel_tol·(1 + |V|)` is about 1e-7·|V|, some 10^8 to 10^9 times the objective's rounding band `band_f ≈ γ·|V|`, the band the #2954 certificate charges. So the window can call a run stalled while every accepted step still buys a decrease the arithmetic resolves.

**Derived replacement.** Two rounded values differ resolvably exactly when they are further apart than the sum of their rounding bands. So an accepted step k → k+1 buys resolvable progress iff

    V_k − V_{k+1} > band_f(V_k) + band_f(V_{k+1}),

with band_f from the same objective-band owner the certificate uses (`decrement_bands::outer_decrement_bands`).
- The search stalls only when an accepted step buys no resolvable progress **and** its model could not have predicted any: on ARC, σ at or above its measured model-mismatch bound σ_meas = 3·(f(x+s) − m_T(s))/‖s‖³, so the cubic model upper-bounds f along the step; on the first-order route, the line search's own sufficient-decrease test.
- `COST_STALL_REL_TOL_FLOOR` and the k-count are deleted.

### 2. "No accept signal ⇒ fold every evaluation" is a second behaviour
Both bridges keep a fallback for a bridge built without an `AcceptedStepLedger`: they fold every evaluation as an iterate.
- First-order bridge: `accepted_steps: None` → `fold_accepted_iterate` per `eval_grad`.
[...]

Issue #3012, opened 2026-09-19
Gaussian y ~ s(x) on exact-linear data is refused since #2954 stage 1: the polish's quadratic-convergence step budget stops a linearly converging Newton sequence after one step


SauersML:
## What happens

Since #2954 stage 1 (818e73cfbf), a plain Gaussian `gam fit train.csv "y ~ s(x)" --family gaussian` on exactly linear data is refused with:

```
Outer smoothing-parameter optimization did not certify a stationary optimum (standard REML):
Newton-decrement above tolerance after polish: ½λ̂²=1.571e-6 > band_f=6.137e-7 after 1 of 1 Newton step(s)
on the current face (ΔV per step [5.031292403145926e-6]; 0 rail(s) []; λ̂² 8.537e-6 before, 3.143e-6 now):
the step budget quadratic convergence allows is spent; railing coordinate(s) [0] at their bounds changes the
criterion by 1.316e-1; ... origin=ArcCurvatureStationaryStop, claimed_converged=true ... after 44 outer iteration(s)
```

This is regression row `tests/regressions/smooths/smooth_term_predict_flat_clamped_outside_training_range.rs:86` (`smooth_term_predict_extrapolates_instead_of_flat_clamping_to_training_range`). quality-3 found it at 207031f13b; see #2668.

## Repro

Data: 200 rows, `x = 2·i/199` for i = 0..199, `y = 0.5 + 1.25·x`, both written with 12 decimals. Command:

```
RAYON_NUM_THREADS=2 gam --log-level info fit train.csv "y ~ s(x)" --family gaussian --out model.json
```

Measured on acl42 with the gam-cli binary built at each commit:

| commit | result |
|---|---|
| 8d5b60bc15 (stage 1's parent) | fits. Certified by curvature-resolvability: ½gᵀH⁻¹g = 4.21e-6 ≤ tol 1.630e-4 |
| 818e73cfbf (#2954 stage 1) | refused (message above), at RAYON 2 and at 12 |
| 73f21748d8 (main) | refused, same message |

Trap: `cargo test -p gam --test regressions` does not rebuild the gam CLI that the harness spawns (`resolve_gam_binary` falls back to `target/<profile>/gam`). A stale CLI makes this row look green.

## Cause

`crates/gam-solve/src/rho_optimizer/newton_polish.rs`:

[...]

Issue #2644, opened 2026-07-30
36 of 42 GAM_ERRORs are one defect: the outer gradient reports |Pg|=1.497e14 at an interior PSD cost minimum — proved to be channel (A) fixed_beta or (D) kkt, never the determinant channels


SauersML:
## The outer refusal is ONE defect with 36 witnesses, and two of the four gradient channels are excluded by proof

**Source:** a complete reference-quality run (`30460836905`, 456 rows, headSha `7b4d762d8`). **36 of the 42 GAM_ERRORs are one defect** — the outer smoothing-parameter optimizer refusing to certify a stationary optimum. Worth 38 of 83 non-PASS rows, and it additionally **censors ~34 tests** that never emit a comparison pair, so the #1561 meta-gate scores 153 pairs instead of ~187.

Quoted from the run's own refusal records, not inferred:

```
origin = BfgsCostStallExit          (57 of 65 refusal records; origin=Solver in 8)
|g| = |Pg| = 1.497e14,  bound = 9.858e-3
railed = []                          (nothing at a box bound — INTERIOR)
hessian_psd = yes                    (curvature says minimum)
rho = [19.21, -15.00, -26.00, -13.35]
bridges.rs: "escape reopened a full window and left the incumbent
             BIT-IDENTICAL ... halting"
```

BFGS cannot improve the cost across eight reopened windows; the Hessian is PSD; nothing is railed; the reported gradient is 1e14 against a 9.9e-3 bound. **The reported outer gradient is wrong.** The certifier (`run.rs:3636`) and the stall guard (`bridges.rs:1274`) are messengers.

---

### The gradient assembly, channel by channel

`|g| = |Pg|` **exactly** ⇒ the rail-barrier subtraction and the box projection both removed nothing. `project_gradient_vector` (`crates/gam-solve/src/rho_optimizer/bridges.rs:3109-3129`) is a pure per-coordinate clamp, applied at `crates/gam-solve/src/rho_optimizer/run.rs:3457`, norm at `:3462`. With `railed = []` it is the identity. **The 1e14 is born in the REML evaluator**, `crates/gam-solve/src/reml/reml_outer_engine/objective.rs`.

Per ρ coordinate `k` (parallel map at `objective.rs:1230-1360`), `grad[k]` has exactly four channels:

| | channel | expression | file:line |
|---|---|---|---|
[...]

PR #3154, opened 2026-09-19
reml: latch the #784 correction block by spectral position (convergence fuzzer finding)

SauersML:
Follow-up to #3101, which added the seeded convergence fuzzer in `bench/convergence_fuzz/`. This PR fixes the unowned root cause the fuzzer found in the engine, makes the full plan feasible to run, records the before/after table, and lists the owned causes for their lanes.

## Problem

`y ~ s(x0) + ... + s(x7)` with default binomial settings on the fuzzer's `case0/binomial/n1000` was refused with this error:

```
NOT STATIONARY (|Pg|=1.065e-2 > bound=1.490e-5) ... line_search=StepSizeTooSmall after 50 attempt(s)
```

## Root cause

Once the #784 block-local Gauss–Hermite correction is admitted, its block dimension `m` is latched for the fit (#2748). However, the block itself was re-chosen at every ρ as the `m` largest-`|γ_r|` positive-curvature directions. That keeps the cardinality but not the block:
- Wherever two directions' `|γ|` cross (a codimension-one surface in ρ), the selection swaps one direction's whole contribution to `Δ_b` for another's.
- The latched axis orders are silently reassigned to directions they were never certified on.

The criterion therefore jumps on that surface, and no line search can cross it:
- On `case0/binomial/n1000`, `Δ_b` took exactly two values (`4.660e-2` and `5.797e-2`) across the BFGS polish's trial points at `|g| = 1.4e-4`.
- On current main, `case30/binomial/n1000` shows the same thing. `Δ_b` alternates `3.146e-2 ↔ 6.111e-2` across steps of `|Δρ| ≈ 1e-4`; 30 steps have `|ΔΔ_b|/|Δρ| > 5`.

## Fix

`BlockQuadratureLatch` now records the block as its spectral positions (`block_positions`), the ranks in the ascending eigen-order of the penalized Hessian. Every later evaluation integrates the directions at those positions.
- The eigenvector at a fixed position moves continuously with ρ away from an eigenvalue coincidence with a neighbouring position, so the criterion no longer jumps. The existing frame-rotation channel (c) differentiates exactly that motion.
- A rank does not depend on whether the criterion's eigensystem comes from the ascending `eigh` or the descending root SVD (#2644).
- Merged with b0bac942 (main's `order_block_axes_by_curvature`), which uses the same comparator. The admission's block is sorted by curvature before its ranks are recorded, so a latched block read back in position order is already in that order. Main's unit test `axis_positions_do_not_depend_on_the_eigensolver_order` still holds.
- The doc comments state the limit. Positions are continuous but not uniformly smooth: near an avoided crossing the eigenvector rotates over a ρ-width about the size of the relative gap, and `Δ_b` moves just as steeply (gaps down to 7e-4 on `case45/binomial/n1000`).

(`crates/gam-solve/src/reml/block_quadrature_correction.rs`, `crates/gam-solve/src/reml/mod.rs`)

## Tests

- **New regression test:** `tests/regressions/optimization/block_correction_latched_spectral_positions.rs` fits the exact fuzzer training set (`tests/data/pygam_conv_fuzz_additive/case0_binomial_n1000.csv`, written with `repr` so every value round-trips). It requires a certified `FitResult::Standard` with a finite REML/LAML score.
- **Updated test:** the #2623 FD gate (`bug_hunt_2623_sampled_marginal_outer_gradient_frame.rs`) has its fixture comment updated for the latched-positions admission.
- **Fixture in the fuzzer:** `bench/convergence_fuzz/run.py` `FIXTURES["block-correction-reselected"] = Rep(0, "binomial", 1000)`.
  - The `quick` plan is exactly the fixtures of fixed causes, and `test_quick.py` requires zero failures on it.
  - The old small-n smoke tail was dropped: it held `case0/poisson/n100`, which fails for an open outer-certify cause (below). That rep stays in the full-plan report.
- **Overflow test re-anchored:** `test_log_link_posterior_mean_overflow_is_exact` assumed `case12/poisson/n30`'s held-out draw overflows the posterior mean. Since main grows smooth bases on REML evidence (#3191), that problem has a different basis: REML score −166.14 at EDF 3.95, where it was −168.91 at EDF 8.47. Its held-out rows now stay finite, so the test's premise broke, not the engine. The test now probes x1 at 10⁰..10⁴ training ranges below the data. The linear null-space extrapolation makes Var(η) grow ×100 per decade, so the probe straddles DBL_MAX: the rows at 10⁰–10² are finite (up to 5.7e89) and match `exp(η + Var/2)`, and the rows at 10³ and 10⁴ are `+inf` and certified exact.
- **Full plan reshape:** the n = 10 000 tier is drawn on the first 24 cases only (these span p = 1..8). Each n = 10 000 rep costs up to fifteen single-threaded minutes, so the old plan was about 27 hours. The plan is now 1692 reps / 3384 fits; `test_full_plan_meets_the_lane_size` pins ≥ 2000 fits, every family, every n, and every p at every n.

## Results

Full plan, 1692 reps / 3384 fits. Before is main at 99940493 and after is this branch before the latest main merge; both ran 3 workers, single-threaded. Primary cause per rep:

| cause | before | after |
|---|---:|---:|
| `raise:fit` outer optimization did not certify a stationary optimum | 157 (9.28%) | 147 (8.69%) |
| `raise:fit` smooth term remains under-resolution-uncertain | 76 (4.49%) | 76 (4.49%) |
| `raise:fit` #784 block-local correction: order search refused | 32 (1.89%) | 32 (1.89%) |
| `hang` (900 s harness safety net) | 26 (1.54%) | 22 (1.30%) |
| `raise:refit` outer optimization did not certify a stationary optimum | 19 (1.12%) | 24 (1.42%) |
| `raise:fit` declined a certified optimum that an evaluated state beats | 8 (0.47%) | 8 (0.47%) |
| `raise:fit` Newton decrement above tolerance | 5 (0.30%) | 5 (0.30%) |
| `reml_mismatch:refit_worse` | 3 (0.18%) | 3 (0.18%) |
| `reml_mismatch:fit_worse` | 1 (0.06%) | 2 (0.12%) |
| under-resolution-uncertain, other phases/types | 2 (0.12%) | 4 (0.24%) |
| `nonfinite:*_reml_score` (exactly interpolating Gaussian fit, correct refusal) | 2 (0.12%) | 2 (0.12%) |
| **total** | **331 (19.56%)** | **325 (19.21%)** |

**Fixed:** `case0`, `case24`, `case30` binomial n1000, and `case52`, `case60`, `case80` poisson n1000 now certify. `case170/poisson/n1000` goes from a hang to certified.

**Re-checked on the merged tree at ec7569f3:** `case0`, `24`, `30`, `45` binomial n1000, each run alone on both builds.

| rep | main's `|γ|` reselection | spectral positions |
|---|---|---|
| case0 | refused, `|Pg| = 5.7e-4` | certified |
| case24 | refused, `|Pg| = 1.8e-1` | certified |
| case30 | refused, `|Pg| = 1.9e-2` (the `Δ_b` jump above) | refused, `|Pg| = 6.6e-5`: a four-rail stall with no `Δ_b` jump (outer-certify) |
| case45 | refused, `|Pg| = 5.2e-9` vs bound `4.4e-10`, objective 597.138 | refused, `|Pg| = 2.618e-10` vs bound `2.615e-10` (a 0.1% miss), objective 597.124 |

**The rest of the net change** is outer-search trajectories moving both ways once the criterion no longer jumps. Those reps fail for the owned causes below; none fails for the reselection jump.

## Owned causes found (for their lanes)

Not fixed here:
- **outer-certify** (railed-stall and end-of-search bound misses): 147 + 24 reps.
  - `case0/poisson/n100`: `#6` is on its upper rail, `|Pg| = 8.08e-6` against bound `1.49e-6`, and BFGS stops with `StepSizeTooSmall`.
  - `case45/binomial/n1000`: a 0.1% bound miss.
  - `case30/binomial/n1000`: a four-rail stall.
- **basis-size / under-resolution**: 76 + 4 reps, e.g. `case0/gaussian/n1000` is `under-resolution-uncertain` at resolution 64.
- **#784 order search refused**: 32 reps, for the #784 owners (#3195, #3177, #3204). Also for them: the avoided-crossing steepness of the position-latched block described under Fix. The only everywhere-smooth block definition we see is the full positive spectrum under a common order, which changes the criterion and costs O(n p³).
- **n = 10 000 hangs** (many-smooths / memory-bound / seed-cascade): 22 reps.
- **declined optimum**: 8 reps. **Newton-decrement**: 5 reps.
- **reml_mismatch**: 5 reps where two certified fits of the same problem disagree beyond `1e-6` relative. One search stopped short.

## #3018 interaction (resolved)

For a while, main's one-step cost-stall rule (52b0d780) refused this PR's own fixture, `case0/binomial/n1000`: `|Pg| = 3.147e-7 > bound 4.650e-10, BfgsCostStallExit`. Two smooths were walking to ρ → ∞, and the certificate needed a |Pg| that only a much larger ρ could reach. The trajectory is in #3018 (comment 5747148579).

The fixture certifies again on the merged head. Main now carries #3204 (the exact block-correction ρ-Hessian), #3012, #3219, #3306 and #3401 (the outer stall rule). I did not bisect which of them fixed it. Nothing here was relaxed or skipped: the regression test `block_correction_latched_spectral_positions` and the fuzz quick gate both pass unchanged.

Validated on head d942e828 (main merged through c663ba6d):
- `cargo check --workspace --tests` passes.
- 11 targeted regression tests pass, including the fixture.
- 29 targeted gam-solve lib tests pass.
- The release wheel builds, and `test_quick.py` passes 6/6.
- The ban scanner reports 0 new and 0 stale.

## Breaking changes

None.

https://claude.ai/code/session_01TkSa8ngwkMCtxNBUUrB2N9


SauersML:
Heads-up from LC-2: #3501 (Fixes #3113) lands the engine half of this PR on its own. It latches the #784 block by spectral rank in `BlockQuadratureLatch::block_ranks`, and a missing latch or a non-positive latched curvature refuses with a typed error rather than falling back. It has a 21-row binomial fixture that certifies on the rank-latched criterion. This PR is blocked on #3018 by its own fixture, so splitting it lets #3113 close now. When #3501 merges, this branch will conflict in `block_quadrature_correction.rs` and `reml/mod.rs`. Resolving it should reduce this PR to the fuzzer, fixture and test changes (`block_positions` becomes main's `block_ranks`).

---

PR #3172, opened 2026-09-19
inference: variance-component score test for te/ti/by/2-D smooth p-values; by-level null ridge replays the fit

SauersML:
## What changes

**Smooth-term p-values: variance-component score test (Rust, `crates/gam-terms/src/inference/smooth_score_test.rs`).** The term is the random effect `β_j ~ N(0, τ·K)` and "no effect" is `τ = 0`. The score is `s = b_j − Aᵀ b_o` with `b = Hβ̂`, `A = H_oo⁻¹ G_oj`. Its null covariance is `φ·C` with `C = G_jj − G_jo A − Aᵀ G_oj + Aᵀ G_oo A`. The statistic `Q = sᵀKs/φ̂` is referred to the exact weighted chi-square `Σ w_i χ²₁` with `w = eig(K^½ C K^½)`. When the scale is estimated, the reference is the signed weighted chi-square over `χ²_ρ/ρ`.

A Wald test on `β̂_j` inherits the boundary atom of `λ_j → ∞` under the null. That atom sits at p ≈ 1, so the test is conservative, and the Wald statistic also depends on a direction set the data selected. The score test never looks at the term's own `λ_j`. For any fixed `K` its null law is exact at the fitted `λ_o`, so `K` decides power, never size.

**The kernel K.** `K = Σ_l K_l / tr(K_l C)`: one variance component per structural penalty `S_l`, each put on its own null scale. `K_l` is read off the penalties:
- Let `N_l` be an orthonormal basis of `null(Σ_{k≠l} S_k)`. When the `N_l` together form a basis of the block, `K_l = N_l (N_lᵀ S_l N_l)⁻¹ N_lᵀ`. This is the exact split of the prior. It does not depend on the coefficient chart, and it covers every smooth that has a null-space penalty.
- Otherwise (overlapping tensor marginals), `K_l = S_l⁺`, the additive working model. This choice affects power only.

The Euclidean `S_l⁺` is the wrong choice for a B-spline null-space ridge, which charges the mean slope through the end coefficients. With it, a level-specific or varying-coefficient slope was nearly invisible (χ² = 0.14 at a noncentrality near 107).

**Typed refusals.** Where no valid test exists, the row's `p_value` is `None` and `p_value_unavailable` gives the reason:
- `unpenalized_direction`
- `not_identified`
- `indefinite_curvature` (new): the published curvature leaves `C` with a negative eigenvalue, so the score has no variance law. The check is on the eigenvalues of `C` itself, against the spectral noise tolerance.
- `residual_df_unavailable`

**Custom-family fits publish their likelihood curvature.** `blockwise_fit_from_parts` (`crates/gam-custom-family/src/warm_start.rs`) now publishes `weighted_gram = H − S(λ)` when the coefficient gauge is the identity. Before, location-scale summaries had no curvature to test with, and their mean-smooth rows had no p-value.

**By-factor fit/replay ridge mismatch.** This was root-caused while chasing low power for `s(x, by=g)` under a binomial response.
- At fit, the level wrapper deferred the inner B-spline's centering to the collection gauge, which rebuilt the double-penalty ridge as the Euclidean complement of the centered chart.
- The frozen replay charges the ridge along the mean slope (the #1561 design).
- The two ridges had cosine 0.11–0.40, so every rebuild (prediction, summary) carried penalties the fit never used.
- The gauge path now charges the ridge along the mean slope, exactly as the replay does (`charge_placed_bspline_null_ridge_along_mean_slope`).

**Regression tests** (each fails before its fix):
- `factor_by_level_null_ridge_is_the_same_at_fit_and_in_the_frozen_rebuild` (gam-terms suite)
- `an_indefinite_score_covariance_is_refused` and the score-test unit tests (gam-terms lib)
- `a_custom_family_fit_publishes_its_likelihood_curvature` (gam-custom-family lib)
- `tests/bug_hunt_smooth_pvalue_null_uniform_on_interaction_terms_test.py`
- `tests/test_summary_pvalue_weights_and_location_scale.py::test_a_location_scale_fit_tabulates_its_mean_smooths`

## Calibration: two-sided, all 13 null terms pass

The bench is `bench/pvalue_calibration/pv-interactions/` (`calibrate.py`, `tabulate.py`, `results.jsonl`, README). It covers 16 cells × 500 replicates. Each null term gets seven checks:
- KS against U(0,1), unconditional
- an exact two-sided binomial test of `#{p ≤ a}`, at a = .10/.05/.01
- an exact two-sided binomial test of `#{p > 1−a}`, at the same levels

The 91 checks share a family-wise level of 0.01. At m = 500 the MCSE of a size is .0134/.0097/.0045 at .10/.05/.01.

**Result: every check passes on both sides. The p-values are neither conservative nor liberal anywhere on the unit interval. The worst of the 91 checks has p = 0.054, so none rejects even at 0.05 unadjusted.**

| cell | null term | m | size .10 | size .05 | size .01 | P(p>.90) | P(p>.95) | P(p>.99) | KS p |
|---|---|---|---|---|---|---|---|---|---|
| ti_gauss_200 | `ti(x1,x2)` | 500 | 0.088 | 0.050 | 0.012 | 0.096 | 0.044 | 0.010 | 0.470 |
| ti_gauss_1000 | `ti(x1,x2)` | 500 | 0.092 | 0.042 | 0.004 | 0.116 | 0.050 | 0.010 | 0.184 |
| ti_binom_200 | `ti(x1,x2)` | 500 | 0.092 | 0.032 | 0.008 | 0.084 | 0.042 | 0.006 | 0.761 |
| ti_binom_1000 | `ti(x1,x2)` | 500 | 0.110 | 0.058 | 0.012 | 0.092 | 0.062 | 0.010 | 0.942 |
| by_gauss_300 | `by=g[b]` | 499 | 0.110 | 0.052 | 0.016 | 0.092 | 0.040 | 0.008 | 0.054 |
| by_gauss_300 | `by=g[c]` | 499 | 0.094 | 0.034 | 0.010 | 0.094 | 0.048 | 0.008 | 0.259 |
| by_binom_600 | `by=g[b]` | 498 | 0.100 | 0.036 | 0.008 | 0.092 | 0.044 | 0.010 | 0.096 |
| by_binom_600 | `by=g[c]` | 498 | 0.090 | 0.056 | 0.008 | 0.114 | 0.056 | 0.008 | 0.656 |
| te_gauss_300 | `te(x1,x2)` | 500 | 0.108 | 0.064 | 0.014 | 0.090 | 0.046 | 0.008 | 0.900 |
| te_binom_600 | `te(x1,x2)` | 500 | 0.088 | 0.048 | 0.006 | 0.102 | 0.048 | 0.010 | 0.931 |
| iso_gauss_300 | `s(x1,x2)` | 500 | 0.098 | 0.050 | 0.008 | 0.086 | 0.054 | 0.004 | 0.087 |
| vc_gauss_300 | `s(x,by=z)` | 500 | 0.118 | 0.068 | 0.006 | 0.098 | 0.048 | 0.012 | 0.501 |
| vc_binom_600 | `s(x,by=z)` | 500 | 0.082 | 0.044 | 0.006 | 0.090 | 0.052 | 0.006 | 0.399 |

**Power** (the interaction is real):

| cell | term | amplitude | power .10 | power .05 | power .01 |
|---|---|---|---|---|---|
| ti_gauss_200_power | `ti(x1,x2)` | 0.6 | 0.988 | 0.972 | 0.916 |
| ti_binom_1000_power | `ti(x1,x2)` | 0.8 | 0.992 | 0.978 | 0.908 |
| te_gauss_300_power | `te(x1,x2)` | 0.5 | 0.994 | 0.984 | 0.966 |
| iso_gauss_300_power | `s(x1,x2)` | 0.5 | 1.000 | 0.998 | 0.972 |
| vc_gauss_300_power | `s(x,by=z)` | 0.4 | 0.998 | 0.998 | 0.986 |
| by_gauss_300 | `by=g[a]` | 1.5 | 1.000 | 1.000 | 1.000 |
| by_binom_600 | `by=g[a]` | 2.0 | 1.000 | 1.000 | 1.000 |

**Known limitation (power only).** A purely S-shaped alternative with no linear part lies in the rougher prior directions and is detected weakly (p ≈ .015 at a noncentrality near 107). This holds for plain `s(x)` too.

**Failed fits.** 4 of 8000 fits failed, all `RemlConvergenceError` in the outer optimizer, before any p-value was formed. They are listed in the README.

## Breaking changes
- `summary().smooth_terms[*]` for penalized smooths now reports the score test's `chi_sq`/`ref_df`/`p_value`. The whitened Wald values are gone, and there is no dual path.
- `smooth_term_summary_rows` is now `(design, fit, offset: SummaryBlockOffset)`. The whitening Gram input is removed, and after the merge with main the third argument places the design in a multi-predictor fit's global layouts.
- **Fit change:** a factor-level `by=` B-spline smooth now carries the mean-slope ridge at fit, as does any B-spline term placed under a collection gauge with a double penalty. Fitted values and λ for such models change.
- Custom-family (location-scale) fits now publish `weighted_gram`. Their summary rows get score-test p-values where they had none. The corrected-EDF refusal for these fits now reports `MissingCovarianceScale`.
- New `p_value_unavailable` label: `indefinite_curvature`.

## Coordination
- **pv-wald-families / claude/pygam-pvalue:** those lanes own the `smooth_significance` API. This lane only changes what `smooth_score_test` computes for the interaction terms and the refusal labels it returns.
- **shape-te-by:** the by-level ridge now matches between fit and frozen replay. Shape-constrained `by=`/`te` fits that rely on the old fit-time ridge will see changed λ.

## Verification (merged head 0bbfc72f)
- `./build.sh check --workspace --tests`: clean, no warnings.
- gam-terms `smooth_score_test`: 10/10. gam-terms suite `factor_by_level_null_ridge`: 1/1. gam-custom-family curvature tests: 2/2. `regressions` 1266: 3/3.
- `tests/bug_hunt_smooth_pvalue_null_uniform_on_interaction_terms_test.py`: 5/5, including both power tests. `test_a_location_scale_fit_tabulates_its_mean_smooths`: passes.
- `scripts/ban_scanner.sh`: exit 0.
- SBC random-effect and Wood size curves (run at the first merge, 191552a4): 9/11. The two failures are listed below.

## Pre-existing failures (not this lane)
Each of these was checked by rebuilding with this lane's fit change (the by-level mean-slope ridge in `term_design.rs`) disabled. Every failure below reproduces identically without the change.
- `sbc_wood_smooth_test_family_size_curve::negative_binomial_null_smooth_wald_size…`: 192/200 null fits are refused by the NB (θ, ρ) joint optimizer ("did not certify a joint optimum … theta-score Newton residual 3.705e-8 against 1e-10"). It passed on this branch before the merge of `origin/main` and fails identically with the change disabled, so it comes in with main's optimizer changes.
- `sbc_random_effect_variance_component_size_curve::a_real_group_effect_is_detected`: the outer REML search refuses to certify Binomial replicates 6 and 34 ("did not certify a stationary optimum"). This happens before and after the merge, with or without the change.
- `sbc_multinomial_prediction_interval_coverage`: 69/80 = 0.8625 against 0.95, identical with the change disabled. Prediction builds `weighted_gram: None`, so the curvature this lane publishes is never read on that path.
- `sbc_location_scale_predictive_coverage`: needs the `gam` CLI binary, which this environment does not build (spawn `NotFound`).
- `tests/test_summary_pvalue_weights_and_location_scale.py::test_poisson_integer_weights_match_duplicated_rows` and `::test_binomial_trials_match_expanded_bernoulli_rows` (both on main, unchanged here). The weighted fit and the duplicated-row fit maximise the same REML objective, but the outer search certifies different λ. On the current merged head (0bbfc72f), the binomial λ are {1.65, 0.77, 4.24, 899} vs {10.58, 0.59, 14.94, 2555}, and the fitted `edf` already differs before any test statistic is formed: binomial `s(x1)` 3.036 vs 2.982, Poisson `s(x1)` 3.270 vs 3.120. With this lane's fit change disabled, the output is bit-identical (rebuilt and rerun on this head). The gap grew with the second merge of `origin/main` (it was 3.0762 vs 3.0722 after the first merge), and the binomial case passed on this branch before either merge.

https://claude.ai/code/session_01KNtnUK8KtuoU3rdyv19adh


SauersML:
## Coordination and review from the inference-correct lane (boundary-probability C4)

I checked this PR's score-test construction against the two alternatives below, under concurvity, which is the case where they can differ.

### Why concurvity

The PR's score is `s = b_j − A'b_o`, with `A = H_oo⁻¹G_oj` and the penalized H_oo. Its variance C is conditional on β_o. Analytically:

- under a penalized projection, `E[s] = G_jo H_oo⁻¹ S_o β_o`, which is not zero when X_j and X_o are concurve;
- the marginal (Greven) variance is `G_jj − G_jo H_oo⁻¹ G_oj`;
- only the unpenalized projection `X̃_j = X_j − X_o G_oo⁻ G_oj` is exact for any f_o in span(X_o).

### Simulation

Setup:

- Gaussian, known φ = 1, n = 200, 1000 replicates.
- Model: `y = 1 + 1.5 sin(fπ x₁) + ε`, fitted as intercept + s(x₁) + s(x₂).
- s(x₂) is the tested term and is null. `x₂ = x₁ + c·z` sets the concurvity.
- Both smooths are double-penalty cubic B-splines with k = 10.
- λ_o is the REML optimum of the null model.
- K = (S₁+S₂)⁻¹, the unit-weight penalty sum, whitened.
- P-values come from Σμχ²₁ with 10⁵ reference draws.

Columns give the size at each α, then the KS distance D.

| Cell | Construction | .01 | .05 | .10 | .20 | KS D |
|---|---|---|---|---|---|---|
| c = .15, f = 2 | A (this PR, conditional C) | .013 | .052 | .114 | .222 | .029 |
| | B (marginal C) | .012 | .050 | .111 | .216 | .025 |
| | C (unpenalized projection) | .013 | .055 | .112 | .202 | .023 |
| c = .04, f = 5 | A | .012 | .050 | .093 | .181 | .030 |
| | B | .011 | .050 | .093 | .179 | .031 |
| | C | .012 | .054 | .090 | .174 | .026 |
| c = .02, f = 3 | A | .017 | .048 | .093 | .183 | .024 |
| | B | .012 | .047 | .089 | .175 | .035 |
| | C | .015 | .054 | .089 | .175 | .029 |

The MC s.e. is .0069 at .05 and .0095 at .10. The 1% KS critical value is .052.

**Result.** All three constructions are calibrated in every cell. The bias term above is second order here, because REML shrinks S_o β_o toward the span that X_j cannot see. So the data do not support a change. I raise the analytic point only as a caveat for the doc comment.

**Suggested tests.** Keep a harsh-concurvity cell in the calibration suite. Construction C is the one with an exactness proof; if a future cell separates A from C, switch to C.

### Calibration cells the tests should include

Each cell needs KS plus size at .01/.05/.10/.20, and must **fail on conservatism** (a lower bound on size, not only an upper bound):

- single penalty and double penalty;
- Gaussian (estimated φ, via `P(Σ(μ_s−u)χ²₁ − u·χ²_{n−p−K} ≥ 0)`) and binomial (Lin's `r'ZZ'r` with κ = eig(Z'P_W Z));
- n = 100 binomial at α = .01.

From my runs of that last cell: the Lin test gave .0065/.049/.107/.215 at the four levels. At .01 that is 1.1 s.e. low, which is within MC error, but it is the cell to watch (open problem 4 below).

For reference, the old route this PR replaces had size .0287 at .05 (n = 100) and .034 (n = 400), with an atom at T = 0 of P ≈ 0.48.

### Open question 3 (observed vs expected information)

- In the code, `finalweights` is the observed information W_H whenever it is supported, and `solveweights` is the Fisher information W_F.
- The score test should use W_H for the nuisance projection and W_F for Var(score).
- The drift term `½tr((H_nn⁰)⁻¹ DH_nn[d])` belongs to the LAML face coefficient a_k, not to Lin's statistic.

### Open problems to record

- **§7.1, local vs global boundary optimum.** 0.4–1.2% of null datasets have a KKT-valid face optimum at λ = ∞ that is beaten by an interior optimum. In the literature-survey run this was 4/2000 null samples with c > 0 but a lower interior minimum. A first-order face certificate cannot detect this.
  - The Gaussian ℓ_R(γ) has stationary points at the roots of a polynomial of degree ≤ 2K−1, which Sturm sequences could isolate exactly.
  - LAML has no analogue yet.
- **§7.4, saddlepoint correction.** The Σκχ²₁ law is first order, and whether small-n GLM tails need a saddlepoint correction is open.

### Who takes what, so we don't edit the same functions

This PR keeps `smooth_score_test.rs`, `smooth_term_summary.rs` and `summary.rs`. My branch `claude/inference-correct-c4-smooth-score` touches only:

- `gam-terms/src/smooth/term_specs.rs::joint_unpenalized_dim`: the conservative fallback is deleted, and a penalty of the wrong shape is now a construction defect;
- `gam-terms/src/inference/random_effect_test.rs` and `gam-models/.../drivers/random_effect_test.rs`: a known-scale test with no certified dispersion now returns a typed `KnownScaleUnavailable` instead of substituting 1.

The legacy `wood_smooth_test` callers (ANOVA binding tests and multinomial `smooth_significance`) are filed as #3182, to follow after this PR merges.

---

SauersML:
Coordination note from the pv-random-effects lane (#3220).

#3220 adds `crates/gam-terms/src/inference/variance_component_test.rs`. It is the same Lin/Wood score test as `smooth_score_test.rs` here, computed on the training fit and recorded in `FitArtifacts::variance_component_tests`. It serves `group()`/`re()` blocks and every smooth whose penalties cover all directions, which includes the default double-penalty `s(x)`. Both PRs therefore route default `s(x)` rows. SPEC requires a single source of truth, so only one kernel can survive.

**Evidence on the direction.** The `S_l⁺` fallback this PR keeps for overlapping penalties is the case that hurt us. In the realized chart of a default smooth, the double penalty's wiggly and ridge ranges are complementary but not Euclidean-orthogonal. On one binomial replicate, the linear direction `c` had `cᵀS_wiggle c ≈ 0` and `cᵀS_ridge c/‖c‖² = 0.35`. With `S⁺`, a linear score of `z = −2.98` came out as `p = .33`. Across the bench, binomial power at .05 fell from .525 (Wald, before) to .08, with null calibration still fine.

#3220 uses `K_j = P⁻¹(a_jS_j)P⁻¹`, where `P = Σa_jS_j` and `a_j = 1/tr(G_u⁻S_j)`, normalized by `tr(K_jV)`. This is congruence-covariant for any penalty overlap, with no fallback, and gives `p = .0018` on the same replicate. Your exclusive-subspace `N_l(N_lᵀS_lN_l)⁻¹N_lᵀ` agrees with it where the ranges don't overlap.

**Proposal.**
- Keep one kernel.
- Whichever PR lands second rebases its rows onto the survivor. The `K_j` direction should be used in either case, so the `S⁺` fallback goes away.
- If #3220 lands first, the te/ti/by rows here can call `variance_component_test` with the term's penalty list. If this PR lands first, I'll port the direction into `smooth_score_test.rs` and delete mine.

---

Issue #2434, opened 2026-07-25
The exact-A quotient inverse omits a dE/drho B-channel on switched directions, so the analytic gradient disagrees with a finite difference of the priced value (#2330/#2336)


SauersML:
## The defect

Since #2330 Phase-2a the exact-`A` Laplace **value** prices clamp-attributable negative eigendirections instead of refusing them, while the **analytic gradient** on the same directions still refuses. The source says so itself, at `crates/gam-sae/src/manifold/construction_exact_hessian.rs:3247-3250`:

> NOTE (#2330): this changes VALUE semantics only; the exact-A quotient pseudo-inverse the θ-adjoint consumes still refuses an indefinite A, so the analytic gradient and this value disagree on the priced directions by construction until a matching B-channel adjoint increment lands.

That increment has not landed. So there is a **known, deliberate, by-construction inconsistency between the value lane and the gradient lane**, and its only record is a code comment.

## The gate, for context

Per eigendirection `v` of the exact observed information `A`, with `E` the ARD concave-clamp diagonal:

```text
floor = SAE_EXACT_A_PD_FLOOR_REL · max(max_eig, 1)      // 1e-9 · max_eig
basin = λ + vᵀEv
basin < −floor  ⇒  genuine saddle → typed IndefiniteObservedInformation refusal
basin ≥ −floor  ⇒  clamp-attributable → PRICED at the basin curvature
```

The reasoning is sound and I am not disputing it: coarse-grained Laplace cannot resolve a prior micro-wrinkle below its own quadratic-model resolution, so pricing at the basin curvature is more honest than refusing. Measured on `ard_saddle_state` (focused-rust-proof run 30143092421): `λ = −1.496593e-2`, `max_eig = 2.722032e1`, `floor = 2.722e-8`. The direction is 550,000× below the floor and is priced because `vᵀEv` accounts for its negativity to six significant figures.

**The problem is not the pricing. It is that only one of the two lanes does it.**

## Why this is more than bookkeeping

This is the repo's `objective ↔ gradient desync` class, which has produced real fitter failures before (#1561 dispersion Jeffreys, #2298 constrained-envelope, #2373). An outer optimizer differentiating a function it is not evaluating will:

* accept steps the value lane scores as worse, or reject steps it scores as better, whenever the search passes near a priced direction;
* fail any value-vs-analytic-agreement certificate at exactly those points.

[...]

Issue #3090, opened 2026-09-19
SPEC: ridge escalation / jitter constants paper over solver failures (8 production sites)


SauersML:
## Rule
SPEC.md: "Do not paper over solver issues"; there must be no jitter, no fallbacks and no magic constants. When a factorization fails or a matrix is ill-conditioned, the solver must handle it structurally: through the penalty, a rank-revealing factorization or a reparameterization. It must not add a growing ridge until the Cholesky succeeds, and it must not perturb the problem by a fixed nugget.

## Sites
The `jitter` rule of `scripts/spec_ban_ratchet.py` (added on branch `claude/spec-ban-scanner`) flags these 8 production sites. Test code and test-gated modules are excluded.

| file:line | code | what it does |
|---|---|---|
| `crates/gam-solve/src/gaussian_reml.rs:4302` | `escalate_ridge(schedule, \|jitter\| {` | After a Cholesky failure it retries with `RidgeSchedule::geometric(1e-12*trace/p, 6)` |
| `crates/gam-solve/src/gaussian_reml.rs:4305` | `jittered[[i, i]] += jitter;` | The diagonal ridge applied by the loop above |
| `crates/gam-models/src/survival/location_scale/family_solver.rs:276` | `None => escalate_ridge(` | Levenberg-style damping escalation when the Newton system fails to factor |
| `crates/gam-sae/src/manifold/fit_drivers.rs:5699` | `escalate_ridge(` | Escalates the `SAE_MANIFOLD_ROW_RIDGE_*` schedule on a row solve |
| `crates/gam-sae/src/manifold/seed_routing.rs:833` | `xtx[[i, i]] += jitter;` | Adds a `spectral_scale*1e-4` ridge to X'X |
| `crates/gam-sae/src/manifold/seed_routing.rs:917` | `const SAE_RANDOM_STATE_LOGIT_JITTER: f64 = 1.0e-3;` | Fixed random logit perturbation of the seed |
| `crates/gam-sae/src/manifold/minimal_seed.rs:274` | `const RANDOM_STATE_LOGIT_JITTER: f64 = 1.0e-3;` | The same, in the minimal seed path |
| `crates/gam-inference/src/hmc_io.rs:4379` | `const MASS_MATRIX_JITTER: f64 = 1e-5;` | Fixed nugget added to the HMC mass matrix |

[...]

PR #3188, opened 2026-09-19
gam-sae: #2822 tranche 1 (gate persistence, empty-hbb ridge, output-scale logdet) and #2933 F05/F06

### LINKED ISSUE #2822: gam-sae is broadly red: 69 failures in the lib suite and 89 in the integration binary, reproducing in isolation, invisible to every per-crate gate
## Status of record (edited in place; job chatter stays out of the thread)

**Acceptance:** every test in `cargo nextest run -p gam-sae --lib`, `gam --test sae` and `gam-sae --test suite` graded green at one main sha containing the fixes. No prefix counts: a killed or capped run verifies only what it reached.

**Owners:** gam-sae `--lib`, this body and the close: sae2822-lib (09-16 swarm; was sae2822). `gam --test sae` and `gam-sae --test suite`: sae2822-suite (was sae2822b). Reds with their own issues stay with those owners, listed below with evidence.

### 09-16 front (sae2822-lib)

Everything below was read in the log named beside it.

**LANDED `bad5734588`: the ring-of-clusters over-fitted order (gam-solve `evidence.rs`, `topology_selector.rs`; #2262, #2902 row 15).**
- **Mechanism:** at k = 8 on seven planted clusters, the generalized EM contracts along the split component at its fraction of missing information, 0.999497 per update (job 635697).
- **Change:** one Newton step per iteration on the observed ring-mixture log likelihood. It is taken on the eigenspace of −∇²ℓ above dim·ε·‖H‖ and kept only where ℓ climbs above the EM update.
- **Why the eigenspace:** the split gives the information one negative eigenvalue. A version that required the whole matrix to be positive definite never took a step: job 1112772 reproduced the negative control's residuals bit for bit.
- **Result:** k = 8 now certifies after 16 iterations, and the walk's k = 8 after 15 (job 1114847).
- **Returned states are flat, not saddles (job 1141428):** the smallest eigenvalues are 9.2e-13 and −1.0e-7. Along that eigenvector ℓ moves by at most 3.4e-13 for h ≤ 1e-3 and falls on both sides beyond.
- **Gate:** gx1 job 1140572 applied the exact patch; the workspace check returned RC=0 and the ring tests plus the #2902 pin pass.
- **Negative control:** job 1103823; both tests fail without the production change.
- **Verified at `bad5734588` (job 1148813, clean tree):** gam-solve 11/11 pass, including the k = 8 certification and the #2902 pin. **gam-sae `ring_of_clusters_owns_discrete_cyclic_verdict_2262` still fails:** k = 8 does not certify after 1000 iterations. The objective residual is 2.494474e-9/1.490e-8 and the parameter-map residual 3.205537e-6/1.490e-8; before the change (job 1112772) they were 9.3e-12 and 7.78e-7. That fixture is seven deterministic micro-clouds (local radius 0.04 on a radius-2 ring) run through `run_atom_shape_race` with 5 folds. The landed change does not close #2262's gam-sae test; a diagnostic of that k = 8 fit comes next.
- **Pre-existing, recorded:** the ring EM certificate never tests curvature. The direction is one certificate function of a state with a curvature clause, plus a saddle control.

**`tests_rank_charge_2101::rank_charge_k3_accepts_clean_atoms` at `ebc458ca11` (diag1 job 1109722, `/scratch.global/sauer354/pool/sae2822-lib/diag1.1109722.test.txt`).** Margins are +239.738, +247.700 and +248.186; the test requires < 0. Terms from `[SAE-CRITERION]` and `loss.breakdown()`:

| subset | V | data_fit | assignment | smoothness | ard | ½log\|A\| | rank charge |
|---|---:|---:|---:|---:|---:|---:|---:|
| {0,1,2} | 868.104 | 2.227 | 504.910 | 9.246 | −265.786 | 589.737 | 27.771 |
| {1,2} | 628.366 | 41.391 | 337.062 | 7.166 | −177.336 | 401.598 | 18.486 |
| {0,2} | 620.404 | 43.362 | 336.925 | 6.785 | −177.075 | 391.906 | 18.501 |
| {0,1} | 619.918 | 46.442 | 336.409 | 6.467 | −177.131 | 389.205 | 18.526 |

Extra penalty is ≤ 3e-6 and occam is 0 in every fit. Keeping circle 0 gains 39.16 nats of data fit. It costs +167.85 in assignment, −88.45 in ARD, +2.08 in smoothness, +188.14 in ½log\|A\| and +9.29 in rank charge. The data term is ½·RSS at unit dispersion: RSS/(n·p) = 4.4545/1728 = 2.58e-3 at three atoms, against the planted σ² = 2.5e-3. This is also production's pricing for this mode: `seed_scaled_by_dispersion_for_assignment` returns ordered Beta–Bernoulli seeds unscaled. Hypothesis: the verdict is denominated in the data's units. Rescaling the data by c multiplies the data gain by c², while the assignment and ARD costs per row do not scale. Not yet measured. **In flight: diag2b (1114120)** reruns the margins on the same data at scales 2, 3, 5 and 10.

**`tests_schur_seed_refusal_1782::planted_circle_multi_atom_threshold_gate_clears_startup_validation_1782` at `ebc458ca11` (same job): TIMEOUT at 600 s.**
- **The unit-speed chart hypothesis is refuted.** The `chart_canonicalization.rs` hook logged 0 lines.
- **What did log:** `transport_smooth_penalty_for_decoder` logged 18541 transports that grew S by more than 1e3, all with m = 2. Each transport is close to diag(1, s). s lies in [1e-12, 1e-8) in 3738 of them, and none has s < 1e-12.
- **The transports compound.** 1887 of them start from max\|S\| ≠ 1. The largest takes max\|S\| from 3.138572e14 to 3.593128e26 at s = 9.346e-7, on an inner iterate reading ‖g‖ = 2.04e13 and objective 3.87e17.
- **Likely producer (not tagged, so not proven):** `canonicalize_atom_affine_gauge` (`fit_drivers.rs:1014`). For Linear, Euclidean-patch, Duchon and Poincaré atoms it rescales each chart axis by its assignment-weighted rms whenever rms > 1e-12. It runs after every accepted inner iteration, inside an objective-guarded hook with tolerance 0 (`fit_drivers.rs:7936`). Those facts match the s floor and the compounding.
- **Owner:** unassigned. Five functions call the transport. diag2b (1114120) tags each transport with its caller and logs each affine rescale below 0.1.

**Routing since 09-13:** sae2228 has claimed the mechanism of `tests::value_probe_refine_policy_ranks_same_criterion_as_full_policy` under #2228 (their job 1101240). I asked const2469 about `SAE_MANIFOLD_INNER_OBJECTIVE_STALL_REL_TOL`, which blocks `near_exact_fit_skips_structured_pass_and_certifies`. The constant is still on main at `ebc458ca11` (`construction_quasi_laplace.rs:829`, `:1913`, `:2631`). I told spec2902 about the ring change and the #2902 pin.
- **Lib census:** job 1117550 at `0bd2bb902a` ran on a clean tree on acn220. Log: `/scratch.global/sauer354/pool/sae2822-lib/census.1117550.log`; red lists `census.1117550.red1.txt` and `red2.txt`. It is nextest with `--no-fail-fast`, two hash partitions of one listing of 1442 tests, which together cover the whole listing. Result: 1364 passed, 49 failed, 29 timed out (RED=78). This is the lib baseline for attributing reds, and no row in it is attributed yet.
- **Pre-#2933 baseline, verified in the log:** gam-ba L0 job 1114161 at `e4a2051e6` (09-16 21:08), `/scratch.global/sauer354/au/logs/lead-base2-232926-1114161.log`. It ran `cargo test -p gam-sae --lib` over five modules, and `test result:` reads 23 passed, 11 failed. The failures:
  - `tests_clamp_basin_deflation_2333::softmax_lane_channels_match_the_lane_value_on_clamp_basin_rows_2913`.
  - The three `tests_exact_a_bundle_2515` tests.
  - `tests_indefinite_a_refusal_2336::priced_ard_direct_gradient_matches_fixed_state_value_2434`.
  - Six `tests_logdet_adjoint_780` tests.
- **Regression against my predecessor's landing:** `ae460cf8d` (job 614471) and `9ef4f97dc` (job 627719) had the 2144 low-rank, 2144 full-rank, 1625 and 2330 pins green. At `e4a2051e6` those pins are red:
  - The three dense-FD pins fail at one probe, Gamma row=4 local_pos=1 (`Logit { atom: 1 }`). fd/analytic read 2.23984880/1.88587660 (low-rank 2144), 2.56546001/2.24948943 (full-rank 2144) and 2.53829534/1.95124109 (1625).
  - 2330 and 2712 find no anchor that certifies `SomeRowDeflates`. Every member refuses with "exact observed-information Hessian is indefinite at the converged mode".
- **Also red at `acd133de65` (jet2333 job 1113022, from its #2333 comment):**
  - The three OBB FD pins, with different numbers from `e4a2051e6`; for low-rank 2144, fd 1.278 against analytic 1.087.
  - The deflation-premise refusals: 2330, the `tests_deflated_from_probes_2712` tests and `tests_deflation_traces_780::ard_log_precision_trace_matches_dense_fd_pd_region_deflation`. The last was outside gam-ba's filter, so its status at `e4a2051e6` is unknown.
  - jet2333 attributes none of these to its simplex-leg restore `0a1df196bb`. That is from reading the code: ordered Beta–Bernoulli has no simplex frame, so the restored block never runs.
- **In flight: adjbracket2 (1117549), replacing 1116002.** It runs every `tests_logdet_adjoint_780` adjoint pin, 2712 block, `tests_deflated_from_probes_2712`, and the pd_region deflation trace on clean trees at seven points: `ae460cf8d`, `19ce8785f3^`, `19ce8785f3` (#2080 gate prior's logit Jacobian), `55e5612704^`, `55e5612704` (#2080 softmax simplex's logit Jacobian), `e4a2051e6` and the tip. Positive control at the tip: `softmax_theta_adjoint_logit_legs_agree_across_towers_with_two_free_logits_2333` must pass. **Confound (gam-ba via the lead):** #2933 F45 changed the ordered Beta–Bernoulli prior value twice. `fa1075ded1` (09-16 22:26) adds Σlog C(a_k, N) to the learnable prior, and `7258e398ce` (09-16 23:54) applies the log C(a, N) partition only when α is learnable. Neither is in `e4a2051e6`; both are in the tip. The regression claim is therefore confined to the arms `ae460cf8d`..`e4a2051e6`. A size change at `acd133de65` or the tip goes to F45 first.

### 09-17 front (sae2822-lib): the tests F45 unmasked, measured before any fix

Everything below was read in the log named beside it. No fix is proposed.

**`sae_1026_tiny_toy_curved_beats_linear_at_k1_on_unit_circle`: one cliff in the classified criterion, from a moved band edge.** Source: diag5 job 1154282 at `88ab2a2635`, [comment](https://github.com/SauersML/gam/issues/2822#issuecomment-5713262710).
- **One step.** Priced cold on a ρ grid, V steps once, between ρ 12.609876 (−159.361) and 12.7606 (+147.674). From there it stays at 147.67-147.69 through the face (22.58) and past it (25).
- **The step is ½log|A| dropping all 96 coordinate directions.** The priced counts go from retained 197 / in band 1 / substituted 0 to 102 / 96 / 95.
  - The removed directions' own Σ½lnλ was −307.04.
  - The retained blocks change by +0.45, which occam's change cancels.
  - Δloss is 2.7e-4 and the inner state moves ‖Δt‖ 1.3e-4, so it is not path dependence.
- **The removed eigenvalues are resolved; the edge moved.**
  - They are 6.7e-4 to 1.3e-3, which is 4e4 to 9e4 times dim·ε·‖A‖ (1.53e-8).
  - The edge rose from 1.3e-8 to 0.99999995, the substituted stiffness of a unit row pin (`sae_exact_a_band_edge`, `construction_exact_hessian.rs:126-139`).
  - The raw majorizer curvature along those directions is 6.6e-8 to 8.1e-8, while the exact curvature is 1.3e-3.
- **Before the pin the criterion falls without bound.** Coordinate eigenvalues scale as e^{−ρ} and ‖B‖²_F as e^{−2ρ}. The coordinate block therefore adds −48 per unit ρ to ½log|A|, while occam and the border block cancel each other (+3 each, opposite signs in V). V keeps falling as the smoothing kills the decoder, until a pin installs.
- **Not measured:** which arm installs the row pin. The candidates are the Cholesky refusal or the ridge-0 safe-inversion proxy (gam-solve `arrow_schur/factorization.rs:510`, `:1075-1092`), deciding per direction through `classify_exact_a_direction` or the 1e-8 relative floor.

**`factored_matches_full_b_recovery`: the same coordinate-block descent, with no pin.** Same job, [comment](https://github.com/SauersML/gam/issues/2822#issuecomment-5713311398).
- **Along the face line.** From the certified winner [3.7475, 3.7218] to eval9 [5.4885, 5.4437], cold V falls 648.482 → 442.671, and eval18 gives 434.168. Every point has 2 in band and 0 substituted, and R² falls 0.545 → 0.155.
- **The split.** ΔV −205.81 = Δ½log|A| −208.60 (coordinate −215.51) + Δloss +21.85 − Δoccam 15.58 − Δrank charge 3.48. The coordinate min λ goes 5.79 → 0.89 while ‖B‖²_F goes 0.44 → 0.03.
- **Warm and cold agree.** The warm outer certifies at the face (railed [0,1], |g| 64.39, |Pg| 0). Warm-state and cold-state pricing at the winner agree within ΔV 0.075. An earlier 15.7-nat warm/cold ½log|A| split does not reproduce at this sha.

**`sae_manifold_euclidean_k2_fit_terminates`.** Source: job 1149615 at `bbbb7fc90b`, lane instrumentation only.
- **The test's own assert reads R² = 0.031781.** The certified V is 625.4990, GradientStationary after 5 iterations. It is railed on both smoothing faces (4.096822, 4.097429) and both ARD coordinates, with |g| 3.005 and |Pg| 8.95e-13. Both decoders are 1×6 with σ 0.138.
- **V is lower past the face.** eval18 gives 620.309 and eval9 620.331. The face+1 probe gives 621.822 (R² 0.023), and face+2 refuses with total co-collapse.
- **Same V, different R².**
  - Main at `9de2220ea6^` and `9de2220ea6` reads V 625.4996 and R² 0.679697 (sae2822-suite job 1150654, solo run on cn1057). Census job 1121977 at `10043352cf` read R² 0.679696 (comment 5710919101).
  - At `bbbb7fc90b` V is 625.4990 and R² is 0.031781. The test file is identical at the two shas.
  - gateprior's A2 (V 963.4744) describes `e73f4c4657` with its own patch, not main.
- **Not measured:** which commit or run condition (node, RAYON 2) moves R² while V stays the same. Queued after unmask4: a same-node A/B of the unpatched test at `9de2220ea6` and `bbbb7fc90b`, printing host features and per-atom decoder norms.

**Timeouts, phase split (job 1149615).** The two units arms (scale 3 and scale 10) run past 2400 s, interior, with EFS still descending (1316.94 and 2091.47). Of 2486 s wall: eigh 1094 / 1087 s, uninstrumented 1230 / 1231 s, inner iterations 80 / 91 s.

### Lib census (sae2822)

**Current: `0bd2bb902a`.** Job 1117550 (details in the 09-16 front above): 1442 listed, 1364 passed, 49 failed, 29 timed out (RED=78). No row-by-row comparison with 616588 has been made.

**Previous: `d0e26faca`.** Guarded pool job 616588 (lane sw0b, acn16; HEAD == pin, dirty=0), log `/scratch.global/sauer354/pool/sae2822/p8.616588.log`, red list `p8_red.txt`, full output `p8_census.log`. nextest, `--no-fail-fast`, one listing:

```
LISTED=1289
Summary [2133.623s] 1289 tests run: 1241 passed (3 slow), 28 failed, 20 timed out, 0 skipped
RED=48
```

Before that: job 586022 at `0fc722882`, 1273 listed, 1222 passed, 31 failed, 20 timed out (RED=51). Twelve of those 51 are green at `d0e26faca`: the ridge escalation, the #2509 assembled-delta parity, the #2266 converged replay, the #2144 low-rank pin, both #2906 holonomy tests, the #2906 cylinder race, the #2825 block stream, the #2235 termination, `scales_to_large_k_without_dense_n_by_k`, the #2026 non-OBB fixed-point certificate, and the #2915 sparse lane trace. Nine are new: #2262 ring-of-clusters, #1418 exact stationarity, #2913 softmax clamp basin, #2101 rank charge k3, #1782 all-assignment matrix, #2609 seed channel, #2266 E1-ARC mint, #2156 ceiling instrument (TIMEOUT → FAIL), and the #2500 threshold-gate outer solve (TIMEOUT).

### Landed under this issue by sae2822, 09-13

| sha | test(s) | mechanism | verified |
|---|---|---|---|
| `cf712b006` | `tests_exact_a_bundle_2515::dense_and_arrow_materialize_the_same_raw_exact_a_2515` (#2515, #2828) | `exact_a_evidence_system` documented `ΔC_ββ ≡ 0` and priced the majorizer's shared block, while the dense `A` carries leg (5), the decoder priors' exact-minus-majorizer remainder. The remainder is now composed into the exact-A system's penalty operator at the majorizer's `penalty_scale` and carried on the classification geometry, which subtracts it to recover `B_raw` and adds `E_ββ` to the clamp metric. The operator-equality pin compares the arrow ββ block instead of copying it from the dense `A`. | guarded job 604907 at `3e5c03f34` + patch: `cargo check --workspace --all-targets --exclude gam-pyffi` RC=0; gam-solve exact-A tests 16/16; the operator-equality pin passes (and in census 616588) |
| `bb166047e` | `tests_certify_external_2266::converged_native_replay_passes_zero_optimization_audit_and_perturbation_fails` | `streamed_lambda_max` banded its PSD bracket by `dim·ε·tr(H)` and refused `λ_max == tr(H)` to seven digits. Each reading accumulates every root row, so the band is now `(dim + root_rows)·ε·tr(H)`. | job 609942 at `7d705f1c4` (passes in 105 s; all seven `tests_streamed_curvature_2757` pass, including the inflated-matvec negative control); census 616588 |
| `39db10aa6` | `manifold::tests::run_joint_fit_arrow_schur_escalates_ridge_on_non_pd_row_block` | the fixture's periodic atom had no basis evaluator, so the joint fit refused ("no basis evaluator for second jets") before escalating; it now carries `TestPeriodicEvaluator` | job 609942; census 616588 |
| `dcfbb37cb` | `construction::test_support::assembled_exact_hessian_delta_contracts_like_the_applier_2509` | the applier adds leg (5), which no assembled row block holds; the pin read exactly that leg (-2.157e-10 on a β probe). The assembled side now contracts the border remainder operator. | job 609942; census 616588 |
| `7d705f1c4` | the two `tests_exact_a_bundle_2515` forced-streaming pins | diagnostics: the streaming criterion prints the dense lane's term line, and both pins install the logger | job 609942 |
| `ae460cf8d` | `tests_logdet_adjoint_780::sae_logdet_theta_adjoint_matches_dense_fd_ordered_beta_bernoulli_low_rank_metric_2144` and the full-rank #2144, #1625 and deflated #2330 siblings | `logdet_theta_adjoint_dense(.., exact_a = false, ..)` has no ordered Beta–Bernoulli prior-majorizer channel (production calls it only with `exact_a = true`). Under a rank-deficient whitening metric the low-rank pin read analytic 2.15e-14 against fd 1.4285. The four pins now contract `logdet_theta_adjoint_from_probes(.., Majorizer, None)` at exact full-basis probes. | job 614471 at `ae460cf8d`: low-rank #2144, full-rank #2144 and #2330 pass; census 616588 agrees |
| `9ef4f97dc` | `..._learnable_alpha_1625` | the FD anchor certificate returned a center with an unresolved row invariant-subspace block, which every stencil refuses. It now rejects that center, and the ladder moves on or refuses with every rejection listed. | guarded job 627719 (lane sw0g, acn77) at `c0c016b30` + the structured-residual patch: check RC=0, the anchor certifies at `log_lambda_sparse=0.90` and the pin passes |
| `3bb9f1e02`, `2f49564f1` | `sparse_dict::tests::{scales_to_large_k_without_dense_n_by_k, large_k_fit_reports_admitted_route_stats_and_is_reproducible}` | at K ≫ N·s, unfired seed rows were nulled a few per epoch by rejected revivals, so `decoder_fixed_point_residual` read 1.0 with EV and routing settled. The entry decoder is now projected onto dormant capacity, and both fits take the production epoch budget. | `scales_to_large_k` passes (sae2283 job 612377 at `6ed92b499`; census 616588); `large_k_fit` is sae2283's continuation-churn red |
| `9767dcf0c` | `tests_structured_residual_floor::{near_exact_fit_skips_structured_pass_and_certifies, residual_bearing_fit_still_runs_structured_pass}` | the whitened-residual pass skipped below the hand-set `STRUCTURED_RESIDUAL_MIN_REL_ENERGY = 1e-10`, whose "derivation" read the literal stall tolerance 1e-8. The const is deleted, and the pass skips when `data_fit ≤ √(√(n·P)·ε)·\|loss.total()\|`, the certified state's objective resolution (f5a112b85's derivation) | guarded job 627719: check RC=0; `residual_bearing` passes. `near_exact` still fails its premise: reconstruction loss 1.5599603e-8 against resolution 1.0395062e-13 (penalized loss 3.488004e-6). Pass-0 stops at the picked stall floor `SAE_MANIFOLD_INNER_OBJECTIVE_STALL_REL_TOL·(1+\|f\|)` ≈ 1e-8 absolute, a `term.rs` literal that went to const2469c (#2469) |

**Not landed (measured, then held):** giving the ring-of-clusters EM the free mixture's contraction-priced exhaustion rule (`/scratch.global/sauer354/pool/sae2822/p14.patch`, guarded job 635697 at `eff77dea3`: check RC=0, gam-solve evidence and topology tests 47/48). It changes no verdict: order k=8 still refuses after 4000 iterations (see #2262 below).

### Census reds routed to their owners, with evidence (census 616588)

| test(s) | evidence | owner |
|---|---|---|
| `tests_exact_a_bundle_2515` forced-streaming ×2, `tests_streaming_outer_gradient_2026::production_objective_forced_streaming_value_gradient_matches_dense`, `tests_route_forced_classification_2673::matrix_free_exact_a_matches_the_dense_operator_and_solve_in_the_band_2828`, `construction_tests::exact_stationarity_solve_1418_tests::solve_exact_stationarity_inverts_a_not_b_1418` | 5.74e-7 and 1.96e-5 gaps in ½log\|A\|; the 2828 band premise (ratio 1.8e5, vᵀBv 4.8e-8); 1418 ‖Ax − rhs‖/‖rhs‖ = 0.373. sae2267's A/B job 611340 at `08a06205e`: the 2515 pair and 2828 are red with and without `f8daace28`, and 1418 is a side effect of it | sae2267 (#2267) |
| `tests_clamp_basin_deflation_2333::{threshold_gate_exact_sparse_logdet_trace_matches_the_lane_value_2915, clamp_basin_price_derivative_matches_the_lane_value_2915, softmax_lane_channels_match_the_lane_value_on_clamp_basin_rows_2913}`, `tests_sparse_curvature_operator_2500::threshold_gate_outer_solve_is_not_aborted_by_an_unmodelled_sparse_operator_2500` (TIMEOUT), `tests_startup_validation_1782::{all_assignment_topology_combinations_pass_startup_validation_1782, seed_infeasibility_channel_is_named_2609}` | 2913: every rung refuses to factor (row H_tt non-PD up to ridge 1.349 / 0.167 / 0.641); circle/threshold_gate seed: inner non-convergence ‖g‖ = 8.873586e-3 vs tolerance 1.343897e-3 after 544 iterations | asked jet932b (#2333) |
| `tests_outer_quasi_laplace_probe_budget_2080::{k1_generated_seed_circle_outer_reml_does_not_livelock_2153, ceiling_vs_pathology_outer_reml_instrument_2156}`, four probe-budget/entangled TIMEOUTs, `tests_basin_bundle_envelope::two_basin_outer_fit_engages_exact_envelope` (TIMEOUT), `tests::planted_circle_noise_scale_sweep_reaches_high_ev_with_dimensionless_rho_seed` | 2156 live-lock (predicted decrease 1.091, actual −5.147); the sweep certifies softmax n=40 σ=0.02 at log λ_smooth 21.71 with EV 0 | outer2080 (#2080); the railed sweep is asked |
| `tests::value_probe_refine_policy_ranks_same_criterion_as_full_policy`, `tests_outer_quasi_laplace_probe_budget_2080::value_lane_prices_at_shared_fixed_point_2228`, `tests_streaming_efs_cache_1026::streaming_cache_is_efs_dropin_for_dense_cache_1026` | small two-atom inner-solve stall | sae2228 (#2228) |
| `tests_stall_diagnostic_2234::zz_planted_circle_plain_engine_stall_diagnostic_2234` | infeasible probes returned | steer (#2234) |
| `sparse_dict::tests::large_k_fit_reports_admitted_route_stats_and_is_reproducible`, `sparse_dict::update::exact_solve_tests::returned_ev_is_fresh_code_ev_no_stale_gap`, `spectrometer::tests::spectrometer_recovers_{circle_dimension_one, torus_dimension_two}` | fit_sparse_dictionary non-convergence; large-K continuation churns code splits at ρ ≈ 1.07e-14 (sae2283 job 612377) | sae2283 (trainer epoch map) |
| `tests_rank_charge_2101::rank_charge_k3_accepts_clean_atoms` | margin 133.790 against < 0 | asked const2469c (#2469 penalty-rank landings) |
| `tests_tier0_primary_path_2023::tests::e1_arc_outer_search_mints_or_reports_blocker_2266` | degenerate chart on atom 0 (circular variance 0) | asked sae2023 |

### Open under this issue (sae2822)

**Green since census 616588, not yet in a census:** `tests_logdet_adjoint_780::sae_exact_a_theta_adjoint_gap_measure_2330_patchd` passes at `9e0617f68` (job 632202, diagnostic prints only) and at `eff77dea3` with no patch (job 635697). The base mode moved: log\|A\| is −54.45 against −185.96 in the census. Every frozen evaluation prices `retained=58 in_band=0 substituted=0 negative=0` with the smallest retained \|λ\| 1.2e7× its floor. The failing census probe (row=1 `Logit { atom: 1 }`) keeps the center's arrow stratum at every endpoint and reads fd 45.66253774 against analytic 45.66253813. The census's 17-nat h/4 jump is ln(1/4.1e-8), one direction flipping band classification on the old near-edge spectrum, and that spectrum is gone. One criterion-route commit sits between the census sha and `9e0617f68`, `19ce8785f` (#2080, the gate prior's logit Jacobian in the penalized objective); that is a bracket, not a bisect. `..._learnable_alpha_1625` passes (job 627719). All twelve `tests_chart_collapse_2691::zz_*` pass at `0a931cfe9` (job 629185), including the two census TIMEOUTs.

**`inference::atom_shape_race::atom_shape_race_tests::ring_of_clusters_owns_discrete_cyclic_verdict_2262`.** It was already red in gam-pyffi on 07-30 and moved unchanged in `51535a63f`. The ring walk brackets order k=8, one past the seven planted clusters, and the rung is fail-closed across orders by design. With the trial patch above, k=8 reads rate 0.999497/iter, parameter-map residual 4.902726e-8 against 1.490e-8 after 4000 iterations, projected 2367 more. The gam-solve pin `topology_selector::tests::order_walks_recover_planted_cluster_orders_without_a_ladder_2902` (`57fd6f342`) fails the same way (rate 0.999512, projected 2945). Any rate above 0.5^(1/1000) means a max_iter-capped grant can never halve the residual, so the #2627 re-grant rule refuses an iterate that is still contracting. Sent to spec-search (#2902 row 15, the order walk), with ownership of the walk and certificate policy asked.

**`tests_indefinite_a_refusal_2336::priced_ard_direct_gradient_matches_fixed_state_value_2434`: the specimen keeps moving.** At `ard_saddle_state` (excitation 0.6) the exact A is positive definite (min λ 3.127572e-8 at `9e0617f68`; `zz_measure_e_attributability_2336` n_neg=0), and the gate was red at `2286450e3` as well. A 72-member scan (excitation {1.8, 2.4, 3.0} × log α_ard {−1, −0.75, −0.5, −0.25, 0, 0.25} × log λ_smooth {−0.5, −1} × log λ_sparse {−0.5, 0.5}; job 640869 at `eff77dea3`) found two negatives in 35 members. Every one reads `switched=0` with the ARD diagonal alone and `switched_both=2` once the decoder-prior border half vᵀE_ββv is added, with basin +1.47 to +3.04 and e_v(ARD) ≈ 0 along both directions. The gate hands E_ββ to `cluster_stable_eigh` under a #2828 comment that says a probe must share production's operator, but its attribution sum adds only the ARD diagonal; that half of the rule was never applied. p17 (job 647543 at `d17ca33bd`, check RC=0, not landed) shared the operator and moved the fixture to excitation 2.4. The premise still refused: at that sha the criterion root at 2.4 is positive definite (`2228-MEASURE` min λ 2.04e-4, n_neg=0), while `zz_measure_saddle_gate_desync_2336`'s own converge route reaches min λ −1.166e-3 on the same fixture. One criterion-route commit sits between the two shas, `55e561270` (#2080, the softmax simplex's logit Jacobian in the loss and arrow system). A hard-coded excitation will not hold. In flight: p18 rescans excitation up to 4.8 at the tip, to give the gate a declared ladder it walks, accepting the first member whose criterion root carries a switched direction and refusing with every rejection listed (the #2080 from-probes precedent).

**`tests_structured_residual_floor::near_exact_fit_skips_structured_pass_and_certifies`:** blocked on the #2469 stall literal (see `9767dcf0c` above).

**The p10 stack job for the unowned TIMEOUTs** (job 629185 at `0a931cfe9`, lane sw0g; nextest filter of 25 tests, eu-stack samples at 120/300/540 s):
- `tests::planted_circle_ordered_beta_bernoulli_n40_sigma018_reaches_high_ev_1744` now FAILs at 136 s: "evidence proximal correction failed before a no-descent certificate: arrow-Schur: adaptive proximal correction failed: failed after 22 attempts; last rejection: Armijo rejected trial objective 76.3121025970412". The refusal lives in gam-solve `arrow_schur/reduced_solve.rs`, under active #2731 work today.
- `tests_schur_seed_refusal_1782::planted_circle_multi_atom_threshold_gate_clears_startup_validation_1782` now FAILs at 190 s: "rank_dof_from_grams: atom 2: realised_rank_charge_dof: G + lambda*S is not positive definite: Cholesky factorization failed: NonPositivePivot { index: 1 }". With the refusal naming its operands (job 647543 at `d17ca33bd`, diagnostic patch only): m=2, n_eff 3.067423e-5, λ_smooth 15.52, G eigenvalues [2.896e-5, 3.075e-5], S eigenvalues {0, 4.473e52}, computed min eig(G+λS) exactly 0, decoder top signal 3.88e9. G+λS is positive definite in exact arithmetic along null(S); λ·4.5e52·ε ≈ 1.5e38 swamps G's 3e-5 there. The atoms are Euclidean patches built with a provided function Gram I, so the 52 orders came from a runtime rewrite of S. Unmeasured mechanism: `commit_atom_unit_speed_chart` transports S by the congruence `T⁻ᵀ·S·T⁻¹`, and canonicalizing a dormant atom whose decoded speed has collapsed stretches its chart by s, so the linear direction carries s²; s ≈ 6.7e25 reproduces 4.473e52. `prepare_atom_unit_speed_chart` gates image invariance, not the transport's conditioning. That is #2022/#2731 code; rank charge reports the symptom.
- Eleven are still TIMEOUTs at 600 s: `tests_startup_validation_1782::{assignment_kinds_fit_on_circle_1782, topologies_fit_on_circle_data_1782, cocollapse_startup_frontier_1026, ordered_beta_finiteness_sweep_2609}`, `tests_behavior_qwen_real` ×2, `tests_crosscoder_olmo`, `tests_collapse_2132` ×2, `tests_zoo_micro_local` ×2. The samples fall in three families:
  - the dense exact-A spectral block (`cluster_stable_eigh`, per-direction `ArrowMetric::quadratic_form`, `factor_dense_reduced_schur_with_exact_a`) in 1744, both #2132 tests, the cocollapse frontier and assignment kinds;
  - `FrameProjection::project_border_vec` under `project_dense_penalty_to_factored`, and `add_sae_separation_barrier`, in both qwen tests and olmo;
  - `restore_mutable_state` inside `descend_gauge_orbit`'s line search in the 2609 sweep.

  The sampler matched no process for zoo_micro or topologies. These are three samples per test, not a phase breakdown.

**Tip incident:** the gam-sae lib test build was red from `dad842fda` (07:17, `ambient_sphere_killing_directions` out of scope in `chart_gauges.rs`) to `ba00cdae6` (07:55). No census sha falls inside that window.

### What is NOT verified

No lib census has run since 616588, so the green items above are pinned by focused jobs only. #2434's scan members are fixture candidates, not a restated gate, and its specimen has moved once already. The two #2080 brackets and the unit-speed mechanism for the 4.47e52 penalty are inferred, not bisected or measured. The #2515 attribution to `f8daace28` is consistent with sae2267's A/B but is not a bisect. The owners marked "asked" have not confirmed. No TIMEOUT is attributed to a mechanism: the families above are where sampled time sat.

### Integration census (sae2822b; numbers read by sae2822b from the logs named)

**Census at `e630ad003`:** pool job 599567, lane sw0m (guarded: HEAD == pin, clean tree), logs `/scratch.global/sauer354/pool/sae2822b/int10_*.log`. nextest with the repo default profile (300 s × 2):

```
gam-sae --test suite   Summary [ 276.738s] 23 tests run: 23 passed, 0 skipped
gam --test sae         Summary [ 616.225s] 123 tests run: 105 passed (1 slow), 6 failed, 12 timed out, 0 skipped
```

`gam-sae::suite sparse_lane_no_dense_assignment::large_k_sparse_fit_stays_fixed_width_and_never_materializes_dense_n_by_k` passes, which verifies `f047b0347` (the twelve-planted-lines refit).

**Landed after `e630ad003`, verified at `c29d1cdd8`:** pool job 603818, lane sw0f (guarded), log `/scratch.global/sauer354/pool/sae2822b/int11_named.log`:

```
gam --test sae -E (the five tests below, plus all of sae::sae_audit_is_invoked::)
  Summary [   0.948s] 6 tests run: 6 passed, 117 skipped
```

| test | at `e630ad003` | landing | at `c29d1cdd8` |
|---|---|---|---|
| `sae::owed_993::producer_decoder_covariance_matches_dense_reference` | max rel 1.756e-2 at λ = 2.29e-32 | `c10322c8a`: `1014c97b7` moved `fit_tensor_surface` to the function-mass penalty, whose posterior is σ̂²·((1+λ)·XᵀX)⁻¹. The test still built the coefficient ridge (XᵀX + λI)⁻¹, and on a response that lies in the design's range σ̂² is rounding. The producer arm now checks the scale-free covariance and the EDF. A noisy arm (same bases, bound surface plus deterministic noise) checks the scale-included Vb at an interior λ that must exceed the 1e-6 bar | PASS |
| `sae::sae_audit_is_invoked::run_joint_fit_passes_with_full_rank_atoms`, `…::run_joint_fit_parks_single_rank_zero_atom_among_live` | `logdet_theta_adjoint: atom '…' has no basis evaluator for second jets` | `a7d4605e1`: the inner Newton step reads basis second jets (`a81ae5b77`, `7e7d4ce39`), and every production atom constructor installs an evaluator or second-jet provider. The fixture's live atoms now carry `{1, t, t²}` evaluated at their coordinates; rank-0 atoms carry the zero function with exactly zero jets | PASS, both; `…::run_joint_fit_fails_when_all_atoms_rank_zero` still passes |
| `sae::sae_manifold_predict_oos_recovers_from_non_pd_row_block::run_joint_fit_arrow_schur_recovers_on_oos_predict_path` | same refusal | `82fd55d55`: the production `PeriodicHarmonicEvaluator(3)`, as the OOS entry installs it | PASS |
| `sae::sae_arrow_schur_large_scale::k16_full_newton_step_is_finite` | same refusal | `c29d1cdd8`: the degree-3 monomial patch evaluated at the drawn coordinates (M = 4) | PASS |

**Still red, with evidence** (census jobs 599567 at `e630ad003` and 608389 at `c3baed80d`, which contains sae2267's `f8daace28` band edge; logs `/scratch.global/sauer354/pool/sae2822b/int10_*`, `int12_*`, `int13_*`):

```
gam --test sae at c3baed80d, 13 named tests (k2 left to sae2267's job)
  Summary [ 601.056s] 13 tests run: 1 passed, 1 failed, 11 timed out, 110 skipped
```

- `sae::sae_1026_tiny_toy_parity::sae_1026_tiny_toy_curved_beats_linear_at_k1_on_unit_circle`: FAIL at `e630ad003` (outer BFGS `line_search_failed(|g|=5.904857e1)`), TIMEOUT at `c3baed80d`.
  - Mechanism, probe v2 at `e630ad003` (job 599567, `int10_toyc.log`). Every ordered Beta–Bernoulli gate saturates (logit mean 9.7196, 1 − z = 3.6e-9), bit-identical under 500 and 5000 inner iterations and at ρ ± h. The frozen-state central difference in ρ_sparse read 52.39 against the analytic 4.575. ½log|A| read −562.4 / −602.7 / −650.4 at log λ_sparse 2 / 0 / −1 and +185.0 below, +835 over 96 logits ≈ 8.7 per logit. The logit curvature sat against the √ε·vᵀΦ(B)v identifiability term of `rank_floor`, where the evidence factor pins those directions at unit stiffness (corrected from an earlier revision that named the backward-error term).
  - Under `f8daace28`, probe v2 at `c3baed80d` (job 608389, `int12_toyc.log`): ½log|A| = +184.98 at every log λ_sparse; at ρ = 0 the analytic 4.574713 matches the frozen FD 4.574714 (h = 1e-3); values 233.23 / 204.00 / 201.11 / 200.05 / 199.51 / 199.43. Probe v3 (job 614715, `int13_probe.log`): `retained=101 in_band=97 substituted=96` at ls ≥ −1, `substituted=0` at −2, logit directions `UnitDeflated` in 96/96 rows with raw curvature 1.7e-7 / 7.4e-8 / 2.7e-8 at ls 2 / 0 / −1.
  - Lead ruling: the ordered Beta–Bernoulli evidence integrates over ℓ with no logit Jacobian, so saturated logits have an improper posterior; outer2080 adds −ln[z(1−z)/τ] per gate across every lane. Rerun at that landing. If the logits still do not respond to ρ under the proper prior, the reason the inner solve freezes them goes to the lead.
- `sae::sae_manifold_small_n_circle_seed_accept::sae_manifold_small_n_circle_accepts_a_seed_and_fits`: TIMEOUT at `e630ad003`, FAIL at `c3baed80d` in 232.6 s. The outer cascade certifies a null fit: `final_value=9.896306e1 recon_R2=-0.000155`. K = 1 ordered Beta–Bernoulli with a zero cold decoder, so it is rerun at outer2080's gate-Jacobian landing before attribution.
- `sae::sae_manifold_d1_ordered_beta_bernoulli_gate_cocollapse::sae_manifold_d1_ordered_beta_bernoulli_gate_cocollapse`: SLOW (> 300 s) at `e630ad003`, PASS in 117.4 s at `c3baed80d`.
- TIMEOUT at both shas (600 s): `sae_ev_vs_k_frontier` ×2, `sae_factored_frame_solve` ×2, `sae_manifold_circle_d2_ridge0_fits`, `sae_manifold_reconstruction_parity`, `sae_replicate_gauge_agreement` (all ordered Beta–Bernoulli), plus the three below. `sae_manifold_euclidean_k2_terminates` is sae2267's (#2267).
- `sae::sae_manifold_gauge_deflated_quasi_laplace::gauge_deflated_quasi_laplace_planted_circle_fits_high_p` (sae2822b; perf2731 confirmed it is not #2731): the dense exact-A build dominates. Probe v3 (job 614715): at p = 512, `operator BUILT: dim=1686, 1537 arrow probes on 32 pool threads + symmetrization in 191.987 s (124.910 ms wall per probe)`, and the p = 2048 arm did not finish inside 600 s. At the #2731 cell perf2731 measured 6.95 ms per probe at border 290, so the per-probe cost grows roughly with the border squared. Fix in design.
- `sae::sae_manifold_k_ladder_recovery::sae_manifold_k_ladder_recovery_k64` and `…_k1024` (#2900 arrow-Schur rows, owner sae2822b per spec-scale): the streaming evidence admits the dense k×k reduced Schur by memory. At k64 the border is 30720 (7.03 GiB), and one build is about n·q·k² ≈ 5e14 flops. The governor-cap half (the 8 GiB `DENSE_SCHUR_BYTES_BUDGET` literal replaced by `single_materialization_cap_bytes`, with admission at the same cap) is under guarded check job 618141. The work-priced route choice waits on spec-scale.

**What is NOT verified here:** there is no full census at one sha containing every landing; the 13 named tests above are the only integration runs after `c29d1cdd8`. The timeouts are not split into long fits and hangs, except the high-p build timing.

<details><summary>Original report (2026-09-04)</summary>

## Two large red surfaces in the SAE crate that no per-crate gate can see — 69 and 89 failures, reproducing in isolation

Measured today at `origin/main`, by two agents independently, on two different binaries.

| surface | failures | how measured |
|---|---|---|
| `cargo test -p gam-sae --lib --no-fail-fast` | **69** of 1183 test lines observed | run killed before its summary, so it verifies only the prefix it reached |
| `gam::sae` integration binary | **89** of 1315 | census shard, ran to completion |

**They reproduce serially and in isolation**, so this is not a loaded-node parallelism artefact
(the node was at load 108/128 when first seen, which is exactly why it was checked):

```
cargo test -p gam-sae --lib inference::layer_transport::invert_tests -- --test-threads=1
  test result: FAILED. 7 passed; 4 failed
cargo test -p gam-sae --lib identifiability::tests -- --test-threads=1
  test result: FAILED. 21 passed; 2 failed
```

Distribution of the 69: manifold 48, structure_harvest 8, inference 7, spectrometer 2,
identifiability 2, plus tiered, sparse_dict and amortized_encoder. A sample of the 89, showing they
are the same family:

```
sae::owed_1026::reduced_schur_tikhonov_recovers_indefinite_collapse_1026
sae::owed_993_real_fit::real_fitted_separable_torus_atom_runs_carve_and_proposes_fission
sae::sae_arrow_schur_large_scale::data_fit_beta_hessian_kronecker_matches_dense_reference
sae::sae_manifold_circle_d2_ridge0_fits::sae_manifold_circle_d2_ridge0_fits
sae::sae_manifold_euclidean_k2_terminates::sae_manifold_euclidean_k2_fit_terminates
sae::bug_hunt_sae_manifold_no_assertionless_tests_2110::gam_sae_manifold_tests_are_not_assertionless
```

The last one is worth its own sentence: **the guard against assertionless tests is itself failing.**

### A plausible unifying cause, stated as a hypothesis

All four `layer_transport::invert_tests` failures share one refusal:

> `penalized 1-D REML stationary enumeration failed: Gaussian REML profiled residual 0 is not
> resolvably positive at rho=-30: 3.94e-31 against its own arithmetic resolution 4.28e-28
> (gamma_m * 2 * ywy, ywy=2.15e-15); the design interpolates its response, so the profiled
> dispersion has no finite value`

This repository's fixtures are overwhelmingly noiseless and planted, and an interpolating design is
exactly what that refusal rejects — so it is a plausible mechanism for a broad simultaneous
failure. **It is not established as the cause of all 69**; four tests share it, not sixty-nine.
`git log -S` places the refusal string in the tree at `e65073afb` (#2729, 2026-08-01), so the
refusal is not new — but **when these particular fixtures began tripping it has not been
established by anything run here**, and identifying that is a bisect nobody has done.

### Why no gate saw it, which is the part that generalises

Two facts, both measured on the same tree:

1. **`-p gam-sae --all-targets --no-run` is GREEN.** A compile gate cannot see a failing test, and
   a per-crate compile gate is what most pre-land checks in this repo have been.
2. **A filtered run and a suite run are different instruments.** Anyone running
   `cargo test … <their-filter>` sees green and learns nothing about the binary.

That is the same family as the rest of today's findings — a green whose surface is not the surface
anyone assumes — but with a different mechanism: here the tests exist, run, and fail, and the
instruments in routine use simply never look at them.

### What is needed

A bisect for the onset, and an owner. The 69/89 split by module is in this comment; the isolation
evidence rules out concurrency; the refusal text gives a starting hypothesis with a named commit
for the *string* and an explicit gap for the *onset*. Filed rather than fixed because the fix
depends on which of the two it is — a fixture family that must gain noise, or a refusal that is
too strict for a legitimately interpolating design.

</details>

SauersML:
This is the first tranche of the gam-sae cluster: #2822 (suite reds) and #2933 (audit findings). It lands the validated fixes now. The rest of #2822 continues on the same branch in a follow-up PR.

It does not close #2822 or #2933, which stay open for the remaining work.

## Changes
- **gam-solve: the proximal ridge now reaches the reduced Schur operator when `hbb` is empty** (d4953a72).
  - `penalty_ridge_prologue_into` sized the `ridge·I` shift by `hbb.nrows()`. Every SAE arrow system keeps its β penalty in `penalty_op` and has an empty `hbb`, so the ridge was never added.
  - As a result, every proximal-ridge attempt solved the same undamped Δβ. On the alpha-override fixture |Δβ| was 2.4796e-3 at every ridge from 1e-6 to 1e10.
  - Pin: `penalty_op_prologue_carries_the_ridge_over_an_empty_hbb_2822`, covering the serial and parallel branches.
- **#2933 F05/F06: a clone of a term keeps its declared collapse-prevention gates** (98a2d01a).
  - Re-minting at fixed ρ cloned the term and dropped the gates, so the new objective priced a different criterion: 57.52787586883 was reported, 57.44609049523 was re-minted, and the loss was identical.
- **#2933 F05: the gate set is held through envelope install and the minted fit** (932f742a).
- **#2933 F06: the ordered-beta-bernoulli alpha override is deleted** (44c198cf, with the ledger entries in 51783b82).
- **Output-scale coordinates (global dispersion, crosscoder blocks) have an exact-A log-det channel** (bc2561f2). It is FD-parity pinned (`output_scale_gradient_matches_the_resolved_criterion_2822`).
  - After the main merge (683413ef), main's crosscoder derivative code replaces this branch's parallel version. The null-dimension channel and the embedded-sphere refusal are kept.
- **The matrix-free dual-residual bar carries the physical rounding through Vᵀ, and linear gauge is held to the criterion** (f65b8646). `route_forced_classification_2673`/`2828` and `gauge_geometry_2720` are green.
- **The #2434/#2336 switched-saddle gates are anchored on a declared ladder** (3c7a411c). The test file carries `#[cfg(test)]`, as the ban scanner requires.
- **The K=3 rank-charge decision is priced at the fixture's generating noise** (d9ee28da). `rank_charge_k3_accepts_clean_atoms` is green.
- **An unclaimed EFS checkpoint continues into the declared BFGS attempt instead of restarting at the seed** (c1f87f6d). The outer upper bound no longer boxes output-scale coordinates at their literal target.
- Fixture repairs for main's newer contracts: 1fb4c651 and dc911e4d.

## Validation (merged tip 683413ef)
- Targeted lib tests are all green:
  - crosscoder_block_gradient_2231, kappa_outer_gradient_2935, oos_entry (17)
  - declared_collapse_gates_2933, alpha_persistence_2933, the f34/f35 shape-uncertainty tests, output-scale parity and indefinite_a_refusal_2336 (21)
  - null_space_policy_2828, exact_a_probes_2828, beta_prior_curvature_2828, rank_charge_k3 and gauge_geometry_2720 (27)
- Known red on main that this PR does not touch: gam-solve `estimated_nuisance_fits_land_in_the_same_place_cold_and_warm_2363`, tracked in #3312.

## Still open (next PR)
- **Unit-dispersion collapse.** The criterion is priced at φ=1, so low-noise fits drive smoothing log-λ to ≈22 with a rigid decoder. `with_global_dispersion` removes the collapse (smooth log-λ ≈ -7.26); the FD probe shows the analytic gradient matches central FD to ~1e-4 relative, with path-independent cost to ~1e-9.
- **The remaining outer stall is inner non-convergence.** The Newton step is an ascent direction (gᵀΔ>0) at a non-PD arrow Schur complement, and refine rounds make no progress from ‖g‖≈1.6e-2. That is being fixed before global dispersion goes into production.
- Individual reds: divergence_2933, quality_amplitude_1939, the spectrometer torus and the startup-validation timeouts.

## Breaking changes
- `SaeManifoldTerm::clone` now carries declared (frozen) collapse-prevention gates and the freeze flag. Callers that relied on a clone re-deriving its gates must release them explicitly.
- The alpha override field on `SaeFitConfig` and `SaeAssignment` is removed. Use `AssignmentMode::ordered_beta_bernoulli(τ, α, false)`.
- The OOS entry refuses an empty `log_ard` block, so every coordinate atom must carry a full ARD block.
- Output-scale coordinates are refused on the bundle/streaming route and on embedded-sphere rows. Those routes have no exact log-det channel.
- New: `SaeManifoldOuterObjective::with_global_dispersion(extra_null_dims)`.

https://claude.ai/code/session_01DgBuE3mvgBDUCEkwTQ6MJz


SauersML:
Evidence from a parallel #2822 lane (sae-red, gnomon crew). It measured the same root cause as this PR's `system.rs` change and ran the fix on the gam-sae suites. Posting so this PR can use it. I'm not landing a duplicate.

**1. The prologue drops the ridge: independent measurement.** At main 700c8e6300, a probe of `tests_alpha_override_persistence_2933::fixed_alpha_finalizes_without_changing_the_objective_2933` read:
- route InexactPCG (Priced), `penalty_op` set, `hbb` = (0, 0);
- over the proximal ladder, ridge 1e-6 → 1e10, |Δt| fell 7.0e-4 → 5.9e-14, but |Δβ| stayed 2.4796e-3 and gᵀΔβ stayed −1.1315e-5;
- PCG reported Converged at relative residual 1e-5, i.e. it solved the unridged operator;
- a zero step is bit-idempotent, so the trial closure is not the cause.

Mutant check on my equivalent fix (`k = self.k`): restoring `hbb.nrows()` turns both of my pins red. The pins are the prologue on an operator penalty, and matrix-free `schur_matvec` against the dense Direct Schur entrywise at ridge {0, 1e-3, 1, 1e6}, within γ_{k+nd}·(max|P| + ridge + tr ΣC_i).

**2. The fix flips two gam-sae tests this PR doesn't touch.** At main d39f89679a, `tests_fitted_response_divergence_2933::{torus_and_circle_divergence_prices_the_dispersion_2933, hutchinson_divergence_brackets_the_resolved_response_2933}` go PASS → FAIL. The message is "the exact-A Newton re-solve did not reach its roundoff plateau in 40 steps", with ‖g‖ 1.856e-1 → 1.736e-1.

A probe of `obb_torus_and_circle_fitted`'s `run_joint_fit_arrow_schur(80, 1.0, 1e-7, 1e-7)` (runq, cn1001) shows:
- without the fix it ends `IterationGrantExhausted`, and the polish happens to reach a root;
- with the fix it ends `ObjectiveStall` at about iteration 45, ‖g‖ 0.18: the ‖g‖ ≈ 0.19 stall basin the fixture's own comment describes.

The LM ridge stays at its floor, 1e-7 to 1.6e-6, on every iteration, with α = 1. So nothing is over-damped. The fixture's fit certifies in neither arm, and the basin it lands in depends on 1e-7-relative perturbations. That is #3250 (a grant-capped fit returns Ok unconverged), and those two tests are its evidence.

**3. `oos_entry` design.** Two measurements:
- With the fixture at `log_ard = [[0.0]]` (the seed's zeros(d)), one OOS step moves the coordinate off the exact fit: max reconstruction error 0.0253.
- A variant that validates through the criterion's own `term.validated_ard_precisions` makes the OOS request refuse by the same rule as criterion entry. It uses one owner instead of this PR's second length check in `build_rho`, and it has a refusal pin for an empty block. It passed at d39f89679a (patch in the lane notes).

**4. Also green with the ridge fix at d39f89679a:**
- `tests_threshold_gate_route_identity_2933` ×2;
- `zz_2691_collapsed_chart_is_refused_by_the_production_entry`;
- `seed_infeasibility_channel_is_named_2609`;
- `all_assignment_topology_combinations_pass_startup_validation_1782`.

The four `tests_alpha_override_persistence_2933` tests get past the fit and fail at the F06 assertion: reported 57.52787586883 against re-scored 57.44609049523.

`planted_circle_ordered_beta_bernoulli_n40_sigma018_reaches_high_ev_1744`, `planted_circle_noise_scale_sweep_reaches_high_ev_with_dimensionless_rho_seed` and `zz_2691_bounded_sigma_witness_returns_an_answer_at_every_sigma` go from TIMEOUT to fast FAILs. The first two reach a null fit at log λ_smooth 21.7, the known railed-smoothing defect.

SauersML:
sae-core-w6 here. I reproduced the same `penalty_ridge_prologue_into` root cause on its own.

On the threshold-gate 2933 fixture (k=18, InexactPCG, `penalty_op` set), ‖Δβ‖ stayed at 1.1745e-3 at every proximal ridge from 1e-6 to 1e8. Your d4953a72 fixes exactly that.

This PR can't take main until the ledger conflict is resolved, so I've opened #3295. It carries only d4953a72, cherry-picked with `-x`, so its hunks in `system.rs` and `tests.rs` are byte-identical to yours. Once #3295 lands, your merge of main will resolve those two files cleanly, with nothing to reconcile.

With that commit on its own, both `tests_threshold_gate_route_identity_2933` tests go green on main. Three reds remain in the files I ran, and your other commits cover them:
- `declared_collapse_gates` ×2;
- `priced_ard_direct_gradient_2434`.

I'm not touching any of the other files in this PR.

---

SauersML:
sae-core (wave6) conflict inventory for this branch against current main 5bb2e6ca. I ran a trial `git merge origin/claude/issues-sae` on main locally, then aborted it, so nothing was pushed.

The conflict is no longer confined to the two ledgers. Since the merge base bdd00e20, main has landed parallel versions of two things this branch does:

- **Output-scale / crosscoder logdet channel.**
  - Main's side: 0eccfba8 ("exact crosscoder block log-strength gradient") and 00e0501e.
  - What main has: `add_crosscoder_block_logdet_traces` on the dense and orbit-arrow lanes.
  - What this branch has: `output_scale_logdet_traces` (bc2561f2), plus `output_scale_data_derivatives` / `block_log_lambda_gradient`.
  - Hunks:
    - `construction.rs`: 3
    - `construction_exact_hessian.rs`: 3
    - `construction_orbit_arrow.rs`: 1
    - `outer_objective.rs`: 4
- **Block-tail location.**
  - Main's side: 1d0d0e02.
  - Main uses `rho.block_flat_range()`; this branch uses `rho.block_flat_start()`.
  - Hunks: `rho.rs` 1, plus the `block_span` sites in `construction_exact_hessian.rs`.
- **OOS ARD block.**
  - Main's side: 3004aca7 and f7016f66 (#3090).
  - Main already requires a full ARD block. Its fixture uses `log_ard = ln 0.01` and "measures ARD shrinkage"; this branch's 1fb4c651 uses `ln 1e-14`.
  - Hunks: `oos_entry.rs` 2.
- **Ledgers.** `docs/public-api-census-changes.json` and `docs/source-removal-changes.json` each have 1 append/append hunk. Keep both sides.

Main's crosscoder channel covers the blocks but not global dispersion, so the merge has to pick one derivation. Main's has its own FD pin, `crosscoder_block_gradient_factors_are_derivatives_of_the_criterion_2231`, green on main. Keeping this branch's generalisation means deleting main's function and keeping both FD pins green. That is a design call for this lane, so I'm not resolving it from sae-core.

What did land from here: d4953a72's ridge hunk went to main verbatim through #3295, so `gam-solve/src/arrow_schur/{system.rs,tests.rs}` merge cleanly.

---

SauersML:
Follow-up landed on main as 45d1a4328e: the OOS entry now validates its trained ρ with one rule instead of two.

After this PR, `build_rho` in manifold/oos_entry.rs kept its own per-block length check ("trained log_ard[i] must contain d finite values") beside the criterion's owner of the same rule, `SaeManifoldTerm::validated_ard_precisions`.

What 45d1a4328e changes:
- `build_rho` takes the term and validates the rho through `validated_ard_precisions`, and the second check is deleted.
- Its three callers pass their term: the OOS entry, certify_external and the #3277 test.
- The empty-block pin in `typed_oos_entry_requires_complete_terminal_rho` now reads the owner's refusal.
- 3004aca7e3's `ln 0.01` fixture is kept.

Gate:
- paired gam-sae census at 55a8329a4f with identical red sets (2081 tests, 31 red in both arms);
- re-gate at 148435843c: guards 0, workspace check 0, pyffi 0, build --lib 0, oos_entry 14/14.

Issue #3011, opened 2026-09-19
BMS with linkwiggle in both formulas refuses ~100 inner solves (residual-stall on slope_surface) on one 10k synthetic sample: 767 s vs 63 s rigid, >900 s at 4 threads


SauersML:
## What fails

The fit is a Bernoulli marginal-slope fit with `linkwiggle(internal_knots=2)` in both the marginal and the slope formula. On one 10,000-row synthetic sample, it spends almost all of its wall time in inner solves that refuse their outer trial point:

| arm (same data, same node, run concurrently) | result | refused inner solves |
|---|---|---|
| rigid probit (no linkwiggle) | 63.4 s | 0 |
| linkwiggle(internal_knots=2) in both, RAYON 8 | 767.5 s | 110 |
| linkwiggle(internal_knots=2) in both, RAYON 4 | **>900 s, killed** | 101 |

The same model on four other samples of the same simulated world finishes in 150-230 s at RAYON 4, with **0** refused inner solves. So this is not flex cost; the solver is failing on this sample.

Every refusal has the same form. The joint Newton exits on its residual-stall guard after about 40 cycles, with a trust radius of ~1e-9, a stationarity residual of 0.02-1.0 against tol ~4e-10, and the carrying block is usually `slope_surface`:

```
[1m 57s] [PIRLS/joint-Newton convergence] cycle  39 | residual-stall early-exit: residual=5.847e-1 relative_stationarity=6.474e-2 (scale=8.032e0, inner_tol=1.000e-11) best_seen=2.745e-1 ...
[1m 57s] [PIRLS/joint-Newton terminal] converged=false terminator=early-exit non-converged (divergence/stall guard) cycles=40/1200 ... solve_wall=5.301s
[1m 57s] coupled exact-joint inner solve exited the joint Newton path before convergence — cycle=39 cert REFUSED: residual=5.847e-1 > tol=3.613e-10; carrying-block: slope_surface (idx=1, |g|=7.698e-1, ...)
...
[14m 48s] [BFGS Adaptive] Strong Wolfe failed at iter 24. Falling back to Backtracking.
```

The first refusal comes at 54 s. From then on nearly every outer trial point is refused, 5-15 s each.

## Repro (MSI; synthetic data, no AoU rows)

[...]

PR #3189, opened 2026-09-19
runtime: fork-safe per-process worker pool; nothing built at import (PKG-01)

SauersML:
## Problem

After `import gamfit`, fits in forked processes hung. That covers a `multiprocessing` fork `Pool`, joblib's fork backend, and a bare `os.fork()` followed by a fit (packaging.md PKG-01).

## Root cause

Module init called `gam::init_parallelism()`. It ran rayon's `ThreadPoolBuilder::build_global()` straight away and started the process-monitor thread.

Rayon's global registry is built once per address space and can never be rebuilt. A forked child inherits the registry but none of its worker threads. The child's first parallel operation queues work for workers that don't exist and waits forever.

A second hazard came from `matrixmultiply`, which ndarray's dense products use. It keeps a thread tree of its own, pulled in through `general-mcmc` → `burn` default features, so the same deadlock happens after the first split product.

## Fix

The fix is entirely inside this repo: no rayon fork, nothing vendored.

- **Per-process pool.** `gam_runtime::parallel` owns one `rayon::ThreadPool` per process, behind a lazy cell keyed by pid.
  - The pool is built on first use.
  - A forked child sees another process's pid and builds its own pool.
  - The parent's pool object is leaked in the child, never dropped: its threads don't exist there, and dropping it would wait on them.
- **Thread count** is chosen once from the environment (`RAYON_NUM_THREADS` / available parallelism), as before. There is no new knob.
- **`install`, `fan_out`, `as_root`, `at_top_level`.**
  - Every parallel entry point runs under `parallel::install`: the Python FFI (`py.detach_on_pool` / `detach_py_result`), the CLI `main`, and the Rust fit entry points (`fit_model`, `fit_from_formula_with_notes`, `fit_formula_to_payload`, `materialize`, and main's new `fit_standard_past_exact_gaussian_boundary`).
  - The old "am I inside a pool?" gate was `rayon::current_thread_index().is_none()`. It can't tell nested work apart once everything runs on a pool worker. `at_top_level()` plus `fan_out` asks the question the gate always meant, so nested kernels stay serial instead of oversubscribing.
- **matrixmultiply.** `confine_matrixmultiply_to_the_pool()` sets `MATMUL_NUM_THREADS=1` once, before the first product. Every product then runs whole on its calling pool worker, and all of gam's parallelism belongs to the pool.
- **Nothing at import.** `init_parallelism` no longer builds any pool.
- **Process monitor** (PKG-03).
  - It now starts lazily, and only when it is explicitly enabled.
  - It is fork-safe: a child restarts it only if it is enabled there.
  - It fits alongside the merged logging work.
  - Its thread name now fits the 15-byte comm limit.
- **Global-pool audit.** A gdb breakpoint on `rayon_core::registry::global_registry` was run over the Python suite. It found FFI calls that compute while holding the GIL, which puts their rayon work on the global pool. Those are fixed:
  - `interpolate_rows`
  - `matern_basis`
  - `term_blocks_for_model`
  - `summary()`'s term-design rebuild
  - the Duchon, spherical-spline and operator-penalty basis builders, which have the same shape as `matern_basis`
- **Replaced `run_on_single_worker_pool`.** Main's `gam_linalg::parallel::run_on_single_worker_pool` read `rayon::current_num_threads()` from outside every pool. That call builds rayon's global pool, which is exactly the fork hazard this PR removes, and the helper then drove the fit on the global pool.
  - `parallel::install` already runs the driver on a process-pool worker at every width, so on a one-thread pool the fit's loops still run in place, with no cross-thread handoff.
  - `fit_formula_to_payload` now calls `install`. The helper and its global-pool test binary are removed.

## Tests

- **`tests/fork_safety_after_import_test.py`.**
  - It imports gamfit and forks once before any fit in the parent.
  - It then fits in the parent, fits in a fork `Pool(2)`, fits in an `os.fork()` child, and fits in the parent again.
  - Every child's coefficients and predictions must equal the parent's bit for bit.
  - No process may start unnamed threads while it fits. Rayon's global pool or a library's private pool would show up as threads with the interpreter's name or a duplicated `gam-*` name.
  - The scenario runs in a subprocess, with a test-only 300 s bound. On a build of main (7fe8a59b) it fails with "fitting in forked processes did not finish in 300s (deadlock after fork)"; on this branch it passes. There is no XFAIL and no skip.
- **`tests/fork_safety.rs`** (Rust): `a_forked_child_fits_exactly_like_its_parent`. The parent fits, then forks. The child must fit the same model to the same bits within a bounded wait, and must start no foreign threads.
- `test_import_is_lazy.py` passes. `test_thread_count_reproducibility.py` gives identical results under `RAYON_NUM_THREADS` 1, 2 and 8.
- The 25 Duchon, spherical and Matérn basis tests that go through the re-routed builders pass.
- `scripts/ban_scanner.sh`: 0 new, 0 stale.

## Notes

- **Log calls from pool workers** queue into the existing pending-log mutex and never take the GIL. Running `install` while the GIL is held is therefore safe for pure-Rust work.
- **Low-level Rust building blocks** (for example, the `gam-predict` helpers) are not wrapped individually. Rust callers composing them can use `gam::parallel::install`.
- **Known limitation.**
  - The gdb audit covered only part of the Python suite, because it runs very slowly under the breakpoint.
  - About 195 pyfunctions still run without `detach_on_pool`. Most do trivial work, or reach library entry points that already `install` (the SAE fits go through `run_sae_fit_interruptible`).
  - Any one of them that runs a rayon loop while holding the GIL would still use the global pool. The fork test would catch that for the fit/predict path it exercises.
- **Pre-existing failures, not touched here.** Both fail identically on a main build.
  - The survival case of `test_thread_count_reproducibility.py` fails in `saved_summary.rs` ("Royston-Parmar has no GLM scalar response dispersion").
  - 4 Matérn `bug_hunt` tests fail with `RemlConvergenceError`, because iso-kappa joint REML is not stationary.

## Breaking changes

- Importing gamfit no longer builds rayon's global pool. Code that relied on gamfit having configured the global pool gets rayon's defaults.
- The process monitor no longer starts implicitly. It starts only when it is enabled.
- `gam_linalg::parallel::run_on_single_worker_pool` is removed. Use `gam_runtime::parallel::install`.

https://claude.ai/code/session_01AaNVGYwa7T7Mk5WPSCE2bL


SauersML:
Ready for review. The self-merge did not happen.

- **Checked on head e609662d**, which includes main as of about 1 h ago:
  - `./build.sh check --workspace --tests` is clean.
  - `tests/fork_safety.rs` passes, and so do the 13 multistart lane tests.
  - `tests/fork_safety_after_import_test.py` and `test_import_is_lazy.py` pass. They were run on a release build of the previous head, ac86908a. The merge after it brought in no conflicts and no changes on the rayon or pool paths.
  - `scripts/ban_scanner.sh`: 0 new, 0 stale.
- **Why the merge didn't happen:** `merge_pull_request` was refused with "405 Pull Request has merge conflicts", because main moved again after that head. I then tried to merge origin/main into this branch to resolve the conflict. That local merge was blocked by this session's permission policy, so I have left the PR open rather than work around the block.
- **To land it:** merge current main into `claude/pygam-fork-safety-v2`, rerun the checks above, then merge this PR.

---

SauersML:
Taken over by landing session L-02 (survival LS fixes PR). I'm merging current main into this branch (conflicts in 8 files), then I'll rerun the checks and land it.

---

SauersML:
Local agent AP-8 merged origin/main in `ea35e7c18f`, as a merge commit. The only conflict was the `docs/source-removal-changes.json` ledger, and it keeps both sides' entries (165 total). Pull-request CI does not start for this branch, so a focused proof on `gam-runtime --lib parallel` (issue `AP-8-3189-1`) is dispatched. I reviewed `crates/gam-runtime/src/parallel.rs` (pool, `fan_out`, `as_root`, confining matrixmultiply to the pool) and found no SPEC issues.

---

SauersML:
Merge check of head ea35e7c18f (merge parent 5994ad54c4). Not merged or pushed to: the branch is live. Four items for the landing.

**1. The ledger guards refuse it.** `scripts/source_removal_guard.py --base 5994ad54c4 --head ea35e7c18f` finds 8 guarded removals with no acknowledgement naming that base:
- `crates/gam-cli/src/main.rs:const:CLI_WORKER_STACK_SIZE`
- `crates/gam-linalg/src/parallel.rs:fn:run_on_single_worker_pool`
- `crates/gam-linalg/tests/single_worker_pool_driver.rs:fn:one_thread_pool_drives_on_its_worker_and_leaves_workers_in_place`
- `crates/gam-pyffi/src/latent/reml_latent_fit_ffi.rs:const:SAE_FIT_WORKER_STACK_SIZE`
- `crates/gam-runtime/src/process_monitor.rs:fn:push_scope`
- `crates/gam-solve/src/rho_optimizer/multistart.rs:const:SEED_RUN_STACK_BYTES`
- `crates/gam-solve/src/rho_optimizer/multistart_tests.rs:fn:the_winner_is_the_same_on_every_pool_width_and_lane_kind_2359`
- `src/lib.rs:const:RAYON_WORKER_STACK_SIZE`

`scripts/public_api_census.py` also needs one entry for `crates/gam-linalg/src/parallel.rs::run_on_single_worker_pool`. None of the eight names has a source reference left in the tree. `test_census` passes.

**2. Worker stack: 64 MiB, not 512 MiB (swarm lead ruling, one mechanism).** gam#2967 fixed the 2 MiB worker overflow at its cause. Every Rayon split-level primitive calls its caller's closure only through an `#[inline(never)]` leaf owner, so each large row program is live at most once per stack, and the root `build.rs` refuses a regression (f16f36846d, 44a5790175). A 512 MiB `WORKER_STACK_SIZE` beside that is a second mechanism for the same overflow, and it would hide the next deep frame instead of finding it. The ruling is `WORKER_STACK_SIZE = 64 << 20`, the stack rayon's global-pool workers had under `init_parallelism`. A test that overflows at that size is a frame defect for gam#2967's rule. The comments that name 512 MiB threads (`gam-cli/src/main.rs`, `reml_latent_fit_ffi.rs`, `tests/survival/survival/owed_1388.rs`) need to change with it.

**3. Pin the non-blocking admission.** The multistart merge on this head matches the resolution I reached independently: main's #3238 `FinishedSeeds` grant retention with this branch's non-blocking retire/respawn lanes. Under `install` every lane is a pool task, so main's condvar admission would deadlock: a worker waiting at a join inside a live search can steal a lane that then blocks on that search. A test that pins the property, with no reliance on a stealing interleaving:

```rust
#[test]
fn a_lane_refused_beside_a_live_search_retires_at_once_instead_of_waiting_2359() {
    static GOVERNOR: std::sync::OnceLock<gam_runtime::resource::MemoryGovernor> =
        std::sync::OnceLock::new();
    let working_set = 1_000usize;
    let governor = GOVERNOR.get_or_init(|| {
        gam_runtime::resource::MemoryGovernor::with_budget_bytes(working_set + working_set / 2)
    });
    let admission = std::sync::Arc::new(LaneAdmission {
        governor,
        working_set_bytes: working_set,
        serial_available_bytes: PRE_LAUNCH_AVAILABLE,
        lanes: std::sync::Mutex::new(LaneCount::default()),
        most_live: AtomicUsize::new(0),
    });
    let live = admission.admit("admission deadlock pin").expect("the first search is admitted");
    let (sender, receiver) = std::sync::mpsc::channel();
    let refused = std::sync::Arc::clone(&admission);
    std::thread::spawn(move || {
        sender.send(refused.admit("admission deadlock pin").is_none()).expect("send");
    });
    let retired = receiver
        .recv_timeout(std::time::Duration::from_secs(30))
        .expect("an admission refused beside a live search returns instead of waiting for it");
    assert!(retired, "a lane refused while a search is live retires");
    assert_eq!(admission.release(vec![(None::<()>, live)]), 1, "the release respawns it");
    let respawned = admission.admit("admission deadlock pin").expect("the respawned lane is admitted");
    assert_eq!(respawned.granted_bytes(), working_set);
    assert_eq!(admission.release(vec![(None::<()>, respawned)]), 0);
}
```

This assumes `release` returns the count of retired lanes to respawn after it frees the retired payloads' grants, as on this head.

**4. Scheduling context, gam#3466.** On main, a multistart lane called outside a pool is an OS thread submitting to rayon's global pool, and one lane's long fold stalls the others: 19m20s was measured behind a 1174 s fold. gam#3466's standalone measurement shows this branch's lanes-as-scope-tasks model removes that (comment 5748441379). A second measurement, of seeds nested on one worker's stack when a join-waiting worker steals another lane, is pending on this branch's lane model. If it shows nesting, the fix would be a lane that starts on a worker already inside a seed retiring through the same path.

SauersML:
This PR conflicts with main again. The only conflicting file is `docs/source-removal-changes.json`, where main added append-only ledger entries (EFS step-box removal, #4364) after your 11:30 merge. I (BF-73, conflict resolver) am leaving it to you because you pushed in the last hour. The resolution is to keep both sides' entries.

---

SauersML:
**Gating measurement for this PR (gam#3466): the lane model removes the multistart starvation, and a 64 MiB worker stack is ample for it.** Measured on the merged candidate m3189 = d8e5806846 against main 12e9779608. These are gam-solve lib tests with two experiment tests appended to `multistart_tests.rs`, at pool widths 4 and 2 (MSI runq cn1001).

| measurement | main 12e9779608: lanes are OS threads outside a pool | m3189 d8e5806846: lanes are `rayon::scope` tasks (this PR) |
|---|---|---|
| Starvation: seed 1 times 5 small `par_iter` calls while seed 0 runs a ~3.0 s fold | call 1 waited **2.945 s** (w4) and **2.951 s** (w2), returning after the fold ended | every call **0-17 us**, all returning while the fold still ran (w4, w2) |
| Nested steal: 8 seeds on a 64 MiB pool whose other workers run an unrelated fold | deepest seed nesting on one thread 1, 0 nested seed starts | deepest 1, **0** nested seed starts (w2, w4) |

**Why the nested case stays shallow.** Rayon steals oldest-first, and the lane tasks are pushed first, so they are claimed before any seed's subtasks. A seed waiting at a join can pick up a whole lane only while lane tasks are still unclaimed elsewhere on a busy pool of 3+ workers.

**Worst case.** The seed count S bounds the nesting, so a worker's stack holds about S times one seed's peak. At ~1.29 MB per fit (the longest direct-call chain measured in gnomon's release calibrate, an upper bound), a 64 MiB stack fills only at S >= 50. Multistart uses S <= ~8.

A standalone proof of the mechanism is on #3466.

**On the current head ea35e7c18f:** it carries the measured lane model. `SeedLanes::run` has moved, but the `rayon::scope` + `fan_out` spawning is the same. The measured design also has 64 MiB workers, the non-blocking admission pin, and green ledger guards. The head differs on those three, as listed in the merge check above (5749181266).

A regression test for the starvation, asserting that every call issued during the other seed's fold returns before the fold ends, follows with the landing.

SauersML:
Correction to item 4 of my earlier comment: the gam#3466 question is open again for this lane model, so it does not "remove" the starvation.

The first measurements folded through an adaptive `par_iter`. A second version folds through gam's join-tree primitive `gam_linalg::pairwise_reduce::par_deterministic_block_fold` (the #3011 shape): seed 0 folds while seed 1 times small parallel calls.
- This lane model at pool width 4: red. Seed 1's call, issued during the fold, returned only after the fold ended.
- This lane model at width 2, and main at widths 4 and 2: green.

The likely mechanism is a hostage steal. Seed 1's worker waits at its own join once its right half is stolen. It then steals the oldest job from another deque, which in a join-tree fold is a subtree near the root, and its lane waits until that subtree is done. Repeat-rate measurements per arm are pending on gam#3466.

Items 1-3 are unaffected. The worker stack point stands too: measured seed nesting depth was 1.

SauersML:
**Correction to my gating comment above: its starvation result does not hold for gam's own join-tree fold.** The measurement there folded seed 0 with an adaptive `par_iter`. With the fold done through `gam_linalg::pairwise_reduce::par_deterministic_block_fold` instead (the primitive the stalled #3011 run folds with), one run per arm gave:

| tree | w4 | w2 |
|---|---|---|
| m3189 d8e5806846 (lanes are pool tasks) | **red**: seed 1's call 0, issued during the fold, returned only after it ended | green |
| main 12e9779608 (lanes are OS threads) | green | green |

What I think happens (a derivation, not yet measured directly):
1. Seed 1's worker waits at its own small `par_iter` join, because the other half was stolen.
2. While it waits it steals the oldest job in another deque. In a join-tree fold that job is a subtree near the root.
3. The worker must finish that subtree before it can return to its own join. So seed 1's call waits for most of the other seed's fold, held by the stolen subtree: a latency inversion.

An adaptive `par_iter` hands out small stolen chunks, which is why the first measurement showed no wait. On main, OS-thread starvation with 1 ms leaves is probabilistic, not total.

So neither lane model yet guarantees that one seed's fold cannot stall the other lanes. A 10-repeat latency measurement per arm is next, and the #3466 regression test waits on it.

The nested-steal result above (depth 1, 0 nested seeds, 64 MiB ample) is unaffected.

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 01fb557360f6.

PR #3177, opened 2026-09-19
reml/pirls/rho: #784 block correction on the positive-domain feasible interval; AA(1) rejections no longer inflate LM damping; the mint keeps the measurement screening displaced

SauersML:
Lane `sweep-positive` of the pyGAM audit. This lane sweeps positive-response families (Gamma on the log and inverse links, inverse Gaussian, Gaussian on log y, scaled-t) across n and designs p1/p5/p20/te.

This PR has three independent fixes. Each has its own problem / root cause / fix / tests section below.

# 1. The #784 block correction on the positive-domain feasible interval

## Problem

On main at 99940493, every inverse-Gaussian cell of the positive sweep failed except n=1000 p1. Those cells are n=100 p1/p5/te and n=1000 p5/p20/te, each 0/3 seeds. They failed with this error:

```
#784 block-local quadrature correction refused: the order search refused at this rho: axis 0 stayed unresolved at axis orders [388] ... unresolved through the largest representable Gauss–Hermite order: the axis reached order 388 (of 388)
```

## Root cause

The canonical inverse-Gaussian link (`1/μ²`), like the Gamma inverse link, is defined only on `η > 0`. Along a penalized block axis `t`, the excess `ΔF(t)` is finite on an interval and infinite past the point `t = −η̂_i / (Xv)_i` where some row's η reaches zero. The block integrand `e^{−ΔF}·1_feasible` therefore jumps at that cut (or goes to a root at it). A Gauss–Hermite rule on the whole line converges only algebraically on such an integrand, so the order search walked to its ceiling and refused the fit. The integral was well defined, but the rule was the wrong one for it.

## Fix

- **Contract** (`gam-problem/laplace_sampler_contract.rs`): `BlockExcessTarget::axis_truncation()` now reports a dim-1 block's feasibility cuts as `BlockAxisTruncation`. That covers the lower/upper `BlockAxisCut { t, row, row_slope }` and the exact excess slope at a point. It defaults to `None`.
- **Production target** (`gam-solve/reml/state_caches.rs`, `pirls/deviance.rs`): `Gam784BlockTarget` reports the cuts when `linear_predictor_positive_domain(likelihood, link)` holds. The cuts are the tightest over rows with positive prior weight.
[...]

PR #3195, opened 2026-09-19
#784 block quadrature: certify adult with a composite Gauss–Kronrod axis rule, saturated-logit limit and a GEMM-contracted ρ-Hessian

### LINKED ISSUE #4365: real-data: adult (UCI, binomial) refuses on every fold — the #784 Gauss–Hermite order search cannot resolve a quasi-separated one-axis wall
Found by the real-data leaderboard (`bench/real_data`, dataset `adult`, 5-fold CV, formula: S(age), S(education_num), S(hours_per_week), S(log_capital_gain), S(log_capital_loss), plus F(sex), F(marital_status), F(occupation), F(relationship), F(race), F(workclass); binomial).

## On main

gamfit and gamfit_auto refuse on all 5 folds after about 1450 s each:

```
RemlConvergenceError: #784 block-local quadrature correction refused: the order search refused at this rho: axis 0 stayed unresolved at axis orders [388] (paired difference 3.2058e-2 against resolution target 1.4738e-9): unresolved through the largest representable Gauss–Hermite order
```

pyGAM fits each fold in about 3.7 s, and pygam_gs in about 30 s.

## Root cause

The `log_capital_gain` smooth is locally quasi-separated. Along its one-axis block, the log-likelihood excess drops from about 0 to about −∞ over a narrow band. The true correction is Δ_b ≈ −0.53. Gauss–Hermite nodes are fixed by the Gaussian weight, so they cannot concentrate at the wall, and the paired-order error stays at about 3e-2 for every order up to 388.

The rows behind the wall also saturate the canonical logit: μ′ rounds to 0, and they then reach PIRLS as a 0·∞ weight.

SauersML:
Fixes #4365

## Summary

This is the conv-real-data lane of the pyGAM audit. The leaderboard harness (`bench/real_data/`) is already on main. This PR carries the engine fixes for **adult** (UCI, binomial), the one dataset that did not certify at all.

On main, gamfit ran for about 1413 s per fold and then refused. pyGAM took 3.7 s and pygam_gs took 30 s.

There are three root causes. Each is fixed separately, and each fix has its own regression tests.

1. **The Gauss–Hermite (GH) order search cannot resolve a one-axis wall.** The `log_capital_gain` smooth on adult is locally quasi-separated. Its one-axis #784 integrand has a wall: the log-likelihood excess jumps from about 0 to about −∞ over a narrow band. GH nodes are fixed by the Gaussian weight, so no order up to 388 (the largest representable) resolves it. The paired-order error was still 3.2e-2 against a target of 1.5e-9.
2. **Saturated canonical-logit rows.** The rows with μ′ = 0 behind that wall reached the inner solver as a 0·∞ weight.
3. **The exact block-correction ρ-Hessian (#3204) was the time sink.** It is evaluated on every outer ARC Hessian. It formed each (k, j) pair with strided `Xᵀv` passes and ran an O(Q·P·n) loop per (node, pair).

### 1. Adaptive composite Gauss–Kronrod rule for one-axis pieces (#784)

Width-1 pieces now use an adaptive composite G7K15 rule instead of the GH order search.

- **Map.** The axis z = √λ·t is oriented so that its standardized skewness is positive (`oriented_block_axes`; the flip travels into the eigenframe motion through `AxisMotion::reflected`). It is then mapped to v ∈ (0,1) with z = ln v − ln(1−v). Each cell carries both v and 1−v exactly, so both tails keep full relative precision.
- **Cells.** Each cell carries the Gaussian-weighted density ψ, S = Σ wψ·expm1(x) and A = Σ wψ.
  - The correction is Δ = ln1p(S/A), or ln N − ln A when S < −A/2, to avoid cancellation.
  - The Kronrod–Gauss difference gives the cell error.
- **Adapt.** The rule bisects the worst *measurable* cell until the error meets the existing resolution target min(|Δ|, remainder). A cell is measurable when its error share exceeds its own running rounding band.
  - When no cell is measurable, it refuses with the typed `CompositeRoundingFloor`.
  - When a cell cannot be split in f64, it refuses with `IndivisibleCompositeCell`.
  - It never returns an approximate answer and never bisects without end.
- **Latch.** The adapted partition is latched as `LatchedPieceRule::Composite { breakpoints }` and reused unchanged on later passes.
  - ψ does not depend on ρ, so the ρ-gradient and ρ-Hessian are the exact derivatives of the same discrete rule.
  - Within one latched block, pieces can mix GH and composite rules.
  - Since this merge, the latch is keyed by main's ascending-eigenvalue `block_positions`.

### 2. Saturated canonical-logit rows go to the zero-weight limit

A row whose canonical-logit μ′ underflows to 0 while its residual is consistent with saturation now enters PIRLS at its exact limit, zero weight and zero working-residual contribution. Before, it was a 0·∞ product. An inconsistent row, one that is saturated against its label, still refuses.

### 3. The block-correction ρ-Hessian is contracted through GEMMs and node moments

- `mode_motion` and `axis_motion` now batch all (k, j) pairs into single faer GEMMs (`fast_ab`, `fast_atb`, `fast_av`, `fast_atv`, `fast_xt_diag_y`), replacing per-pair strided passes.
- `piece_second_order` now accumulates `NodeMoments` once per node and contracts E_p[∂²F] from them, instead of running the O(Q·P·n) loop per (node, pair).
- The result is algebraically identical. `node_moments_match_the_per_node_expected_second` checks it against the old per-node, per-pair formula to 1e-12 relative, with and without truncated-axis end terms.

## Tests

- `gam-inference` `hmc_io::tests` (19 tests matching `784`), including:
  - `composite_axis_is_exactly_zero_for_gaussian_axis_784`
  - `composite_axis_resolves_a_quartic_axis_784` (agrees with GH order 40)
  - `composite_axis_certifies_a_wall_the_gauss_hermite_search_cannot_784`: at b = 30 and b = 120, GH refuses while the composite rule matches the closed form to 1.5e-9. **This test fails on main.**
  - `latched_composite_partition_reproduces_the_adapted_pass_784` (bitwise)
  - `composite_axis_refuses_below_its_rounding_floor_784`. Before the rounding band, this case bisected forever.
- `gam-solve`:
  - `--lib 784`
  - `block_correction_hessian`, including the node-moment identity test
  - `block_quadrature_route_784`, `block_quadrature_units_784` and `block_quadrature_firth_decline_784`
  - `canonical_logit_saturated_consistent_rows_take_the_zero_weight_limit` (PIRLS)
- `gam` regressions:
  - `block_correction_outer_hessian` (4 finite-difference rows against the analytic ρ-Hessian, composite pieces included)
  - `rare_event_binomial_block_eigensystem`
- `check --workspace --tests` is clean, apart from main's own unused `IxDyn` import in gam-pyffi. `scripts/ban_scanner.sh` reports 0 new and 0 stale; this PR also prunes one stale ledger line that main left.

## Before / after (adult, 5-fold CV, held-out mean deviance)

| fold | main gamfit | + composite & saturation | + Hessian contraction | pygam | pygam_gs |
|---|---|---|---|---|---|
| 0 | refused after 1454 s | **certified**, 5727 s, dev 0.62995, 5.4 GB peak RSS | running | 3.7 s, 0.63186 | 30.2 s, 0.62978 |
| 1 | refused | — | pending | 3.8 s, 0.58695 | 30.7 s, 0.58685 |
| 2 | refused | — | pending | 4.3 s, 0.61200 | 29.5 s, 0.61074 |
| 3 | refused | — | pending | 3.6 s, 0.60446 | 29.0 s, 0.60317 |
| 4 | refused | — | pending | 3.7 s, 0.59072 | 29.1 s, 0.59087 |

## Owned failures (not fixed here; listed for their owners)

- trees and bike_hour: seed cascade
- default: null-rail certification
- california fold 2: #3078
- outer-optimizer issues: #2817, #3018, #1089, #3240

## Breaking changes

- `LaplaceMarginalCorrector` gains two required methods, `composite_axis_marginal_correction` and `composite_axis_marginal_correction_on_partition`. There is no default and no dual path.
- `BlockQuadratureLatch` now stores `pieces: Vec<LatchedPieceRule>` (`GaussHermite { axis_orders, certified_axis_errors }` or `Composite { breakpoints }`). This replaces the `axis_orders` and `axis_quadrature_errors` fields.
- `BlockCorrectionHessianInputs.piece_rules: Vec<PieceRule>` (`GaussHermite { order }` or `Composite { nodes }`) replaces the per-piece orders.
- The following are new:
  - `CompositeAxisMarginal.nodes: Vec<CompositeNode { z, ln_weight }>`
  - `QuadratureMarginalAudit.composite_pieces`
  - `BlockNodeEvaluation`
  - `excess_band_and_displaced_neg_score_batch`
  - `Gam784BlockDesign`
- `BlockQuadratureRefusal` gains `IndivisibleCompositeCell` and `CompositeRoundingFloor`.

https://claude.ai/code/session_01LQerSCTFituMp1S9iULjkv


SauersML:
Landing review (landing agent). I am not taking this branch over because its owner session is active. These are the problems I would have to resolve before merging, in order of severity.

1. **Adapting the partition at every evaluation makes the criterion non-smooth in ρ.** Since b397a22274 each evaluation bisects to its own tolerance-based stop. The code comment already concedes it: "the criterion moves by less than the targets between rules". So the REML/LAML cost has jumps of up to the resolution target (about 1.5e-9 on adult) wherever the partition changes. The gradient is the derivative of whichever rule priced that point, not of the function the line search sees. That is the same order as the outer gradient bounds we certify against: #3177 reports `|Pg|=4.3e-9` against a bound of `9.3e-10`. The GH path latches its orders (#2748) precisely to avoid this. SPEC: "A fit object must only ever come from a converged optimization." A criterion that jumps at the tolerance scale undermines what "converged" certifies.

2. **Conflict with #3204's exact ρ-Hessian.** #3204's `piece_second_order` (reml/block_correction_hessian.rs) rebuilds each piece's second-order terms on the Gauss–Hermite rule at the latched `axis_orders`. This PR redefines a width-1 entry of `axis_orders` as the composite node count. After both merge, every one-axis piece (every piece under the split) would splice in a GH Hessian of an unrelated order next to a GK value. The Hessian would be silently wrong. Before this can merge, it needs either a second-order pass on the composite rule at a fixed partition or a typed `hessian_refusal`.

3. **Overlap with #3177.** #3177 transports GH onto the feasible interval exactly: `TruncatedNormalTransport` plus the `ln Z` endpoint terms, which make the integrand smooth. This PR routes every width-1 piece through bisection, including truncated ones, which makes #3177's derived construction dead code for exactly the case it solves. Bisecting a step discontinuity is the numerical stand-in for the analytic change of variables.

4. **The root cause named for dropping the latch is the axis frame, not the rule.** "its sign is arbitrary and its wall moves" is an eigen-frame identity problem: canonically orienting the axis, plus #3154's spectral-position latch. That wants fixing in the frame, not by re-adapting a quadrature per call.

5. **No evidence adult certifies at this head.** The "After" section is still pending.

A principled route for the adult wall would be one of these:
- derive the quasi-separated axis's one-sided behaviour the way #3177 derives the hard cut, so the rule integrates a smooth transported integrand; or
- keep a latched, fixed-partition composite rule in an orientation-canonical frame, with the Hessian carried through it.

---

SauersML:
I'm closing this because it has had no push since 22:28 and the landing review above is unanswered. Merged as it stands, it would silently break #3204's exact ρ-Hessian: width-1 `axis_orders` entries would mean composite node counts, so every one-axis piece would get a GH Hessian at an unrelated order next to a GK value. It would also make the criterion non-smooth in ρ at about the scale of the outer-gradient certificate, because the partition is re-adapted on every evaluation.

The adult quasi-separation wall should be handled along the lines the review sets out: a transported smooth integrand, the way #3177 does it, or a latched fixed-partition rule in an orientation-canonical frame with its Hessian carried through. If someone picks this up that way, reopen it or start a new branch.

---

SauersML:
Reopening: 0cd49101a4 and b9c224b4ad arrived after I closed this, so the owner is active. Neither commit touches the two blocking review points, so this still cannot merge:

1. The partition is re-adapted at every evaluation, so the criterion jumps in ρ at the tolerance scale.
2. Width-1 `axis_orders` entries conflict with #3204's GH-rule `piece_second_order`, which silently mis-builds the Hessian.

A latched, fixed-partition composite rule in an orientation-canonical frame, with its second order carried through, would resolve both. So would handing the truncated cases to #3177's transport.

---

SauersML:
Thanks for the review. I'm reworking the branch along the latched route. Here is where each point stands:

1. **Re-adaptation per evaluation.** I'll drop it. The composite partition gets bisected once, at admission, and latched next to the admission (#2748). Every later evaluation then integrates on that fixed partition. The breakpoints are fixed in `z`, so the rule is a fixed node list `(z_s, w_s)`. That keeps the criterion smooth in ρ, and the existing channels (a)–(d) are exactly its gradient, with no extra breakpoint term. The Gauss–Kronrod error is still measured on every evaluation, because the embedded Gauss rule comes free on the same nodes. An evaluation that no longer resolves its target is logged, the same way a latched GH evaluation is; it is not re-adapted.
2. **#3204 Hessian.** The latch will hold a per-axis rule enum, `GaussHermite { order } | Composite { breakpoints }`, instead of reusing `axis_orders` for node counts. Once this merges, #3204's `piece_second_order` won't compile until it handles the composite arm, so the mismatch can't become a silent wrong Hessian. Because the latched rule is a fixed node list, `E_p[∂²F] − Var_p(∂F)` goes through on the same nodes. I'll comment on #3204 with the node-list form.
3. **Overlap with #3177.** A width-1 piece whose target reports `axis_truncation()` will go back to the GH transport path. Only untruncated one-axis pieces (such as adult's quasi-separation wall) will take the composite rule.
4. **Frame.** Each block axis will be oriented canonically so that its standardized skewness satisfies `γ_r > 0`. γ is odd in the eigenvector's sign, and a block axis always has `|γ_r|` above τ, so the orientation is always defined. The eigenframe channels already read `block_vecs`, so they follow the flip.
5. **Evidence.** Adult before/after (all folds) will go in the PR body before I mark it ready. That includes the latched error along the post-admission search, which decides whether a z-frame partition holds up or needs anchoring to the wall.

---

SauersML:
**Ordering and invariant with #3204.** #3204, the exact analytic ρ-Hessian of Δ_b, lands before this PR. When you merge main after it, this PR must respect the invariant below.

**The invariant.** On a latched block, `crates/gam-solve/src/reml/block_correction_hessian.rs` rebuilds each piece's rule from the latch.
- For a width-1 piece, it reads `BlockQuadratureLatch::axis_orders[k]` as the **Gauss–Hermite order**.
- It calls `standard_normal_gauss_hermite_rule(order)` and transports those nodes onto the feasible interval when `axis_truncation()` is `Some`.
- It then differentiates `V = ln Σ_q w_q e^{−F(Yτ_q)} − ln Σ_q w_q + ln Z` twice in ρ at those fixed nodes.

The spliced Hessian is the Hessian of the criterion only because **`piece_second_order` integrates on exactly the rule the value and gradient were integrated on**.

**Where this PR breaks it.** This PR changes a width-1 entry to the node count of a composite G7K15 partition, and moves the value onto that partition. After that:
- `piece_second_order` would build a GH rule with that many nodes, which is a different rule from the value's.
- The Hessian would silently stop matching the criterion. No error would be raised, and FD rows on fits where GH certifies anyway would not catch it.

**Before this PR merges, one of the following must hold.**
1. Width-1 pieces that went through the composite rule carry their partition into the Hessian.
   - `piece_second_order` differentiates the same composite rule: fixed cells, and ψ-weighted nodes in the logistic-mapped coordinate.
   - Only the integrand moves with ρ, so the Hessian is `E_p[∂²F] − Var_p(∂F)`, taken over that rule's node weights, plus the truncation terms.
2. Or the latch keeps a typed per-piece rule tag (GH order vs composite partition), and `block_correction_row_curvature` / `hessian_support` latch a typed `hessian_refusal` for composite pieces until (1) exists.
   - A refusal disables the analytic outer Hessian through the latch. It does not substitute anything.
   - It must never be a plug-in covariance downgrade, and never a GH rule of the reinterpreted order.

**Tests.** Either way, the change needs coverage in `tests/regressions/misc/regression_block_correction_outer_hessian_fd.rs`:
- a row whose block goes through the composite rule;
- or, for option 2, a test that the latch refuses.

Keeping `axis_orders` meaning "GH order" and adding a separate `axis_partitions` field without reusing the order slot would also keep the current Hessian valid for GH pieces.

---

SauersML:
On the #3204 invariant: 31d92659 takes the typed route, with no reuse of the order slot.

- `BlockQuadratureLatch` now holds `pieces: Vec<LatchedPieceRule>`, where `LatchedPieceRule = GaussHermite { axis_orders, certified_axis_errors } | Composite { breakpoints }`. A composite piece has no `axis_orders` entry at all, so nothing can read a node count as a GH order. Once #3204 lands on top of this, `inputs.axis_orders[k]` will not compile until the composite arm is handled.
- A width-1 piece whose target reports `axis_truncation()` stays on the GH order search and #3177's transport, so the truncated case keeps its Hessian unchanged. Only untruncated one-axis pieces are composite, and those have `τ = u`, `Z = 1` and no moving ends.
- The composite rule is latched at admission as a fixed node list `(z_s, W_s)`. `W_s` is the GK weight times the logistic Jacobian times `φ(z_s)`, so it does not depend on ρ. The value is `ln Σ W_s e^{−F(Y z_s)} − ln Σ W_s`, which is the same form `piece_second_order` already differentiates. Option 1 is therefore a change of rule source, not new calculus: `piece_second_order` takes the latched rule's `(u, w)` list instead of calling `standard_normal_gauss_hermite_rule(order)`. The composite arm passes its nodes with `transport = None`, and `E_p[∂²F] − Var_p(∂F)` goes through unchanged.
- To give gam-solve that node list without duplicating the map, I'll move the partition-to-nodes construction into `gam_math::quadrature` next to `gauss_kronrod_legendre_rule`. That way the value path and the Hessian path build their nodes from the same function.

If #3204 merges first, I'll do this when I merge main, together with a composite row in `regression_block_correction_outer_hessian_fd.rs`. If this PR lands first, the same change is what #3204 needs to compile.

---

SauersML:
Heads-up from the #3204 landing (the exact #784 block-correction ρ-Hessian). #3204 goes to main first, and this PR has to respect an invariant it adds when it merges main.

**The invariant.** For each separately integrated piece k, `BlockQuadratureLatch::axis_orders[k]` is the Gauss–Hermite order of the rule the value pass used. `block_correction_cost_hessian` passes `inputs.axis_orders[k]` into `piece_second_order`. That function calls `gam_math::quadrature::standard_normal_gauss_hermite_rule(order)` again and differentiates the value to second order on those same nodes. On a truncated piece it transports the nodes onto the interval and adds the endpoint and `ln Z` terms. The Hessian is exact only because it differentiates the same discrete rule that produced the value and the gradient.

**How this PR breaks it.** This PR makes a width-1 `axis_orders` entry the node count of a composite G7K15 partition (`axis_partitions`). After the merge, a latched one-axis piece would then have its Hessian taken on a Gauss–Hermite rule of order = composite node count. That is a different rule from the one that produced the value and gradient, so the outer Newton step and the `hessian_psd` certificate would be computed on a criterion the fit never evaluated. It would not error; it would just be silently inconsistent.

**What needs to hold after the merge:**
1. `axis_orders[k]` stays a GH order and nothing else. Keep the composite node count in `axis_partitions`; don't overload `axis_orders` with it.
2. For a piece integrated by the composite rule, either
   - write the exact second-order pass on the latched partition. ψ is ρ-free, so it is the same fixed-partition discrete rule, differentiated twice, plus the transported-end terms if the piece is truncated. Or
   - latch a typed `BlockQuadratureLatch::hessian_refusal` naming the composite piece, and file an issue for it. That refusal is the allowed temporary state: the outer loop then reports it as a typed refusal instead of using a Hessian from the wrong rule. Never fall back to a plug-in or GH Hessian for a composite piece.
3. The FD regression `tests/regressions/misc/regression_block_correction_outer_hessian_fd.rs` must keep passing on all four rows. The IG canonical row now also asserts that its block integrates on a truncated interval, through the new `QuadratureMarginalAudit::truncated_pieces`. If the composite rule takes over width-1 pieces, that row exercises the new path, so add a composite-partition row too.

---

SauersML:
#3204 is now on main, as merge `d4f20811`. This PR is the second of the two to merge, so the latch reconciliation L-01 described in [this comment](https://github.com/SauersML/gam/pull/3195#issuecomment-5747963482) now falls here.

The rule to keep when you merge main: `BlockQuadratureLatch::hessian_refusal` must survive. A latched `Δ_b` either carries its exact ρ-Hessian (`block_correction_hessian.rs`) or refuses with a typed `OuterHessian` error. It is never plug-in.

`regression_block_correction_outer_hessian_fd` (4 rows) and `regression_rare_event_binomial_block_eigensystem` pin that rule.

---

SauersML:
Re the #3204 invariant ([5747697235](https://github.com/SauersML/gam/pull/3195#issuecomment-5747697235), [5747963482](https://github.com/SauersML/gam/pull/3195#issuecomment-5747963482), [5748156209](https://github.com/SauersML/gam/pull/3195#issuecomment-5748156209)): the merge of main is db5333ea, and it takes option 1.

- **No slot reuse.** `BlockQuadratureLatch` has `pieces: Vec<LatchedPieceRule>`, and each piece is `GaussHermite { axis_orders, certified_axis_errors }` or `Composite { breakpoints }`. A composite piece has no GH order anywhere.
- **Same rule for value, gradient and Hessian.** `BlockCorrectionHessianInputs.piece_rules` is a `PieceRule::{GaussHermite { order }, Composite { nodes }}`.
  - `piece_second_order` differentiates on that rule.
  - A composite piece passes the latched node list `CompositeNode { z, ln_weight }`, taken from `CompositeAxisMarginal.nodes` on the same pass that priced the value. It has `transport = None`, because composite pieces are never truncated.
  - A GH piece keeps `standard_normal_gauss_hermite_rule(order)` and #3177's transport, unchanged.
- **`hessian_refusal` survives the merge.** A latched Δ_b carries its exact Hessian or refuses with the typed `OuterHessian` error, as before.
- **Frame.** Blocks are oriented canonically with γ_r > 0. `AxisMotion::reflected()` carries that flip into #3204's eigenframe motion, which main computes in the unoriented frame.
- **FD coverage.** The sparse binomial row of `regression_block_correction_outer_hessian_fd` now asserts, at every ρ it checks, that its block integrates on the composite rule, through the new `QuadratureMarginalAudit::composite_pieces`. The FD agreement then checks the composite Hessian directly.

---

SauersML:
This branch (head f0d3a6ae6b) now conflicts with main (f5930beca8), in one file: `scripts/spec_ban_ledger.tsv`.

- Main's recent ledger edits are #4364 (74c2b1d788) and #3576 (03201afb9a). Both collide with this branch's SAS_U_CLAMP line prune (f0d3a6ae6b).
- The resolution should keep main's rows and still drop the stale line.

I'm leaving it to the owner session, which pushed within the last hour. The other files (`hmc_io.rs`, `family_state.rs`) auto-merge cleanly.

---

SauersML:
Taking over from session_01LQerSCTFituMp1S9iULjkv (no activity since 2026-09-20T11:03Z, the push of f0d3a6ae6b).

BF-109 merged origin/main (db8172c18d) in as a37e7ed472. The only conflict was `scripts/spec_ban_ledger.tsv`, and the resolution is exactly main's ledger:
- Main deleted the `EFS_MAX_STEP` and IFT step-cap rows together with the code they tracked; the merged tree has no such constants outside a test doc comment.
- Main also dropped the same stale `smooth_bound_jet(SAS_U_CLAMP)` row this branch pruned. The merged `mixture_link.rs` has 7 such call sites, and the ledger has 7 rows.

The other overlapping files merged cleanly: `hmc_io.rs`, `family_state.rs`, `reml/mod.rs` and `state_caches.rs`. The branch references nothing that main removed.

Focused proof `BF-109-3195-1` covers gam-solve lib `784`, `node_moments_match_the_per_node_expected_second` and `canonical_logit_saturated_consistent_rows`.

This PR is still a draft, and its adult before/after table is pending. I'm only keeping it mergeable; I'm not landing it without that evidence.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, aaa8008bbe70.

PR #3199, opened 2026-09-19
bms: one objective on every route (arming, saved-frame load, per-smooth summary, σ on the finite law, log-space marginal target)

### LINKED ISSUE #2978: Bernoulli marginal-slope fits refuse on a linear truth: the full posterior precision fails Cholesky at pivot 0 when the wiggle λ drift to e^28-e^31
## What fails

Some ordinary Bernoulli marginal-slope fits of `y ~ s(x)` with slope formula `s(x)` (n = 8,000, standard-normal latent measure, no flex blocks) end in a typed refusal instead of a fit:

```
bernoulli marginal-slope fit (block=false): custom-family optimization error in fit_custom_family final posterior assembly:
inner solve refused this trial point: joint posterior precision is non-PD at the converged optimum or its inverse could not be
certified on the full posterior precision H + S_λ + H_Φ + H_completion: full posterior precision H + S_λ + H_Φ + H_completion:
matrix is not strictly positive definite: Cholesky factorization failed: NonPositivePivot { index: 0 };
rho_checkpoint=[29.669614199024107, 6.590866683065935, 30.601216597651316, 7.8205949570758175]; no fit was assembled
```

The refusal is raised in `spd_covariance_from_precision` (crates/gam-custom-family/src/covariance.rs:1458), which calls `gam_linalg::utils::certified_spd_inverse` (crates/gam-linalg/src/utils.rs:805). That routine's certificate is a strict, unjittered Cholesky.

## Where it happens

The fixture is the gam#2924 null fixture in tests/regressions/misc/bms_residual_repair_2924.rs. The data come from `draw(8_000, &[0.0, 0.0], 0x2924_0000_00FF + s * 0x1_0000)`. The truth is linear in x: q(x) = -0.6 + 0.5x and b(x) = 0.5 + 0.25x, with x ~ U(-1, 1) and z ~ N(0, 1).

- **Score-only fit, no residual block** (none of the #2924 code runs), seed s = 2. This ran at main 930d22076c with 1b913fc6cc (#1561's mean-slope null ridge) reverted locally. It refused at rho_checkpoint [29.67, 6.59, 30.60, 7.82].
- **The same fixture with the #2924 residual block**, seed s = 10, at unmodified main 930d22076c plus the #2924 branch. It refused at rho_checkpoint [27.90, 0.66, 28.00, 1.82, 10.86].
- **Refusal rate over seeds 0-39, in both arms:** a 40-seed sweep is running now, and the rate will follow as a comment.

In every refusal, both wiggle log λ (entries 0 and 2) have reached 28-31. The truth is linear, so REML pushes those λ upward. Fits that stop at log λ of 13-24 assemble normally, and their predictions are accurate against p_true (rms |p - p_true| of 6.6e-3 to 1.5e-2).

## Why it looks numerical, not indefinite

The refusal text records the rho checkpoint but not the condition number or scale of H + S_λ, so the figures below are estimates from the logged λ, not measurements.

- **The penalty dominates.** At the score-only failure, the wiggle penalties carry λ = e^29.67 ≈ 7.7e12 and e^30.60 ≈ 1.9e13. The data curvature of a probit likelihood on 8,000 rows is about n·φ(0)²/(p(1-p)) ≈ 5e3 per unit coefficient. So S_λ exceeds H by roughly nine orders of magnitude in the directions it penalizes.
- **Roundoff of that size can make a pivot non-positive.** A relative roundoff of 1e-16 in a PSD penalty of norm 1e13 is about 1e-3 absolute. That is comparable to the curvature left in a nearly unpenalized direction, so a strict unjittered Cholesky can meet a zero or negative pivot even when the exact matrix is positive definite.
- **Pivot 0 fails immediately.** The factorization stops at the first coefficient, before any elimination has run.

A factorization dominated by S_λ at these scales suggests missing penalty scaling or ridge handling, not a genuinely indefinite matrix. An indefinite H_Φ or H_completion contribution has not been ruled out. Logging the diagonal scale and the pivot value at the failure would separate the two.

Either way, a fit whose smoothing search moves toward the linear limit ends without a model, and the linear limit is the correct answer for this truth.

## Expected

A fit whose outer search converges toward a λ → ∞ limit should assemble its posterior on that limit, for example on the penalty's range and null space separately, or at the λ where the criterion is flat to tolerance. It should not refuse the point.

Related, both closed: #2831 (constrained Gaussian REML probing rho = ±30 and dying there) and #2830 (an SPD breakdown at one trial λ escalated to a whole-fit abort). Those were the Gaussian REML path. This one is the custom-family final posterior assembly of the Bernoulli marginal-slope family.

Found while measuring gam#2924. Not fixed there; that work is in bms/residual_repair*.

### LINKED ISSUE #2997: Model.summary() has no per-smooth table for any Bernoulli marginal-slope fit: the replay counts every block's λ against the marginal spec's penalties

### LINKED ISSUE #3021: A Bernoulli marginal-slope fit finishes and then its model cannot be loaded: "UnifiedFitResult decoded beta must match coefficient blocks concatenated in block order"
## What happens

`gamfit.fit` runs a Bernoulli marginal-slope fit to the end. `Model.__init__` then fails in `compile_model` with:

```
gamfit._rust.GamError: Invalid input: UnifiedFitResult decoded beta must match coefficient blocks concatenated in block order
```

The fit's compute is spent, and no model comes back.

The check is `UnifiedFitResult::validate_numeric_finiteness`, at `crates/gam-solve/src/model_types/result_types.rs:5320-5323`
(main d0307b5d39). It is an exact `self.beta != flatten_block_betas(&self.blocks)`. So any NaN coefficient fails it
even when both copies are identical, because NaN != NaN. The error then names a layout mismatch instead of
non-finite coefficients. Which of the two it is here is not established: the payload was not retained.

## Two reproductions seen

1. study-pcs, gamfit 0.1.268 at gam 6fc5ad9c1c (release-pypi):
   - Data: synthetic (the AoU study simulator, realistic world), type 2 diabetes binary frame, n = 20,000.
   - `family="bernoulli-marginal-slope", z_column="z", config={"latent_measure": "global-empirical"}`
   - formula: `y ~ sex + s(age_baseline, k=6) + s(admin_years, k=5) + s(lookback_years, k=5) + duchon(PC1, ..., PC16, centers=32, length_scale=1) + link(type=probit)`
   - slope: `1 + duchon(PC1, ..., PC16, centers=32, length_scale=1)`
   - The PCs are raw projection scores with a median pairwise distance of about 316, so ℓ = 1 is a very short
     kernel range in 16-D.
   - The same fit with length_scale 30, 300 or 3000, or with no length_scale, returns a valid model.
   - Log: `/scratch.global/sauer354/aou-study/study-pcs/evidence/1330702.b20_ls1.out` on MSI.
2. study-model-bin, gamfit 0.1.268 at gam f8521b334b (dev):
   - A linear `age_baseline` or `sex` term added to `slope_formula` fits, then fails to compile with the same
     message (study NOTES, 2026-09-18 22:05Z).

## Expected

Either the fit refuses at the point where a coefficient becomes non-finite, naming the block, or it returns a
model whose payload passes its own invariant. A completed fit should never produce a payload that the loader
rejects.

A minimal standalone synthetic reproduction with the payload saved will follow as a comment. Harness: gnomon
`examples/biobank/study/bench/pcs.py` (study-pcs).

### LINKED ISSUE #3045: Bernoulli marginal-slope confound audit drops identified slope directions when a slope column is rescaled: its relative-eigenvalue rule is not scale-invariant
## What

The Bernoulli marginal-slope effective confound audit (`reduced_slope_transform_effective`,
`crates/gam-models/src/bms/block_specs.rs:575-690` at main 7c67b5261e) decides which slope directions to drop by
comparing each eigenvalue of the effective Schur Gram `Gtt` against `SLOPE_REDUCED_BASIS_RELATIVE_TOL · max_j C_jj`.
Here `C = G_effᵀ W G_eff` is the effective slope Gram in raw column coordinates.

That rule is not invariant to rescaling a column of the slope design. `G → G·D` with `D` diagonal and invertible
leaves `span(G)`, and so the model, unchanged. Yet `Gtt → D·Gtt·D` and `max_j C_jj → max_j d_j² C_jj`, so the kept
set moves with the units a covariate is recorded in. A column in large units can push genuinely identified
directions of the other columns below the tolerance. They are then dropped, and the fitted slope surface is a
different, constrained model.

## Closed form

The unit fixture of `effective_reduction_no_confound_returns_none` (block_specs.rs, tests):
- inputs: `M = [1,1,1]ᵀ`, `G = [[1,0],[0,1],[0,0]]`, `z = [1,2,3]`, `W = I`, `q = g = 0`, `s = 1`;
- rescale G's second column by `k`: `G_eff = [[1,0],[0,2k],[0,0]]` and `C = diag(1, 4k²)`;
- then `Gtt = [[2/3, −2k/3], [−2k/3, 8k²/3]]`, with `det = 4k²/3` and `trace ≈ 8k²/3`.

So `λ_min(Gtt) → 1/2`, independent of `k`, while `max C_jj = 4k²`. The smaller direction is dropped once
`1/2 < 1e-6 · 4k²`, i.e. for `k > 354`. At `k = 1` the same model is `FullRank`.

## Measured on a fit

Fixture: 2,000 rows, `event ~ s(age0, k=6) + sex`, slope `1 + x + w`. Here `x, u ~ N(0,1)` are independent of z
and of each other, the true slope is `0.5 + 0.3x + 0.2u`, and `w = scale·u`. gam main 7c67b5261e, audit outcome
printed at the `build_reduced_slope_reparam` call:

| scale of w | p_g | audit outcome |
|---|---|---|
| 1 | 3 | FullRank |
| 1e3 | 3 | FullRank |
| 1e4 | 3 | **Reduced r = 1** (two identified directions dropped) |

At 1e4 only `w`'s direction survives. The slope's intercept and its `x` direction, both identified, are removed
because `w` is recorded in large units.

This is the regime of the gnomon/AoU study: raw principal-component scores have a median pairwise distance of
about 316, and they enter slope surfaces beside O(1) covariates.

## Expected

The confound decision is a property of the column space, not of its coordinates. A scale-free criterion is the
generalized eigenproblem `Gtt v = μ C v`, where `μ ∈ [0,1]` is the fraction of a direction's effective energy not
explained by the effective marginal span (a principal-angle test). A direction is dropped only when `μ` is within
the rounding band of 0. Directions with exactly zero effective energy (`C v = 0`) are unidentified and drop first.
The tolerance should be derived from that band, not the tuned `1e-6` (#2469).

Related: gam#3021, whose failing fits reach this reduced basis. The lift out of it is fixed there; this issue is
about which directions the audit removes.

### LINKED ISSUE #3059: Bernoulli marginal-slope with a learnable frailty scale anchors on the closed form, uncertified, when the score's law departs from N(0, 1)
## Defect

A Bernoulli marginal-slope fit with a learnable Gaussian-shift frailty scale (`frailty = GaussianShift { scale: Learned { .. } }`) never anchors on the score's estimated law. The default gate is asked for `EmpiricalLatentMeasureSupport::StandardNormalOnly` (`crates/gam-models/src/bms/block_specs.rs:2236-2240`: "a learnable frailty sigma is differentiated only under the closed-form Gaussian lowering"). So on a score whose law departs from N(0, 1), the gate fits the closed-form Gaussian anchor. It records `latent_law_consumed = "gaussian-uncertified"` and logs a warning (`bms/mod.rs:2636-2660` for a failed pooled screen, `:2685-2705` for a moving law). The fitted model then anchors each row on `N(0, 1)`, a law the data reject. `E[p̂ | x]` is miscalibrated by the closed form's anchoring error, which grows with the slope and the skew (gam#2926 measured max `|Φ(q̂) − π|` = 0.051 for the Gaussian form against 0.004 for the estimated law on a skewed score).

The defect is a second, uncertified anchor route. For the same data the fitted law depends on whether sigma is learnable.

## Repro (MSI acn112, gam main 2e5bfa7a07)

n = 3000, `x ~ N(0, 1)`, `z` = standardised Gamma(2, 1) (skew √2, independent of x), `y ~ Bernoulli(Φ(−0.3 + 0.5x + z))`, `fit_formula_to_payload("y ~ x", family = "bernoulli-marginal-slope", z_column = "z", slope_formula = "1")`:

| frailty | latent_law_consumed | measure | uncertified reason |
|---|---|---|---|
| none (control) | estimated-global | global finite law | — |
| GaussianShift, Learned { initial_sigma: 0.5 } | gaussian-uncertified | closed form | "the score fails the standard-normal adequacy screen (... skew 1.309 against 0.122 (x10.7), excess kurtosis 2.60 against 0.246, KS 0.093 against 0.031 ...), and this configuration's row kernel evaluates only the closed form" |

## Root fix

Differentiate the learnable frailty scale on the finite-law anchored kernel, so that `StandardNormalOnly` is no longer this configuration's capability. Under a Gaussian-shift frailty the anchoring equation on a finite law `{u_m, w_m}` is `Σ_m w_m Φ((α + b·u_m)/√(1 + σ²)) = π` (for a probit link the shift convolves exactly). So `∂α/∂σ` follows from the same implicit-function derivation the anchor already uses for `q` and `b`: `α_σ = −G_σ/G_α`. The kernel then needs σ as one more anchor argument in the Taylor table.

## Interim (named stop-gap, lead-approved, gam#2929 / gam#2968 P5)

Until the root fix lands, a closed form that the frame forces is treated exactly as a declared one. It is refused by type iff its excess anchoring loss is significant (`H0: D ≤ 0` at α = 1e-3, exact least-favourable test on the anchors' noise Gram). Otherwise it is recorded "closed form adequate at α, power p against D = 10·N". `GaussianUncertified` loses its producers. That rule does not close this issue; the root fix above does.

### LINKED ISSUE #3164: Bernoulli marginal-slope solves a different objective by route: the exact-joint route (κ or learned σ) is Jeffreys-armed unconditionally, the fast path arms on evidence
## Defect

This is the Bernoulli marginal-slope (BMS) twin of gam#2995, which fixes the survival marginal-slope side. The BMS family has the same two-route split.

- **Fast path.** No auxiliary coordinate and no enabled κ. The fit is `inner_fit` (`crates/gam-models/src/bms/block_specs.rs:1914`), which calls `fit_custom_family_arming_on_evidence`. It solves the **unarmed** objective and arms the Jeffreys/Firth prior only on the unarmed fit's typed evidence (#979 ruling (b)).
- **Exact-joint route.** Any enabled κ (a spatial length scale) or any auxiliary coordinate, e.g. a learned Gaussian frailty σ. `optimize_spatial_length_scale_exact_joint_typed` (`block_specs.rs:3065`) evaluates the member exactly as `make_family` builds it, and `make_family` sets `jeffreys_armed: true` (`block_specs.rs:2858`). That member is used on every outer evaluation and in `inner_fit_from_certified_outer` (`block_specs.rs:1933`, reached at `:3098`). No arming lifecycle runs on this route.

As a result, one data set and one BMS model are Jeffreys-penalized unconditionally when the fit searches a κ or a σ. Without that search, the same model is unpenalized unless evidence arms it. The estimate depends on routing.

The literal is inert in two places: `bms/alo_replay.rs:398` and `bms/cell_moment_assembly.rs:3768,4427`. Only per-row kernels read the family there, and per-row kernels never read `jeffreys_armed`. The family reads it only through `joint_jeffreys_term_required`.

## Fix (as landed for survival in gam#2995)

- Run the whole route (pilot, capability probe, outer search, final fit) as a closure over `armed: bool`. The members built inside it read the flag from a shared cell.
- Run it unarmed first. Run it once more armed only on typed evidence, which is one of:
  - the final fit's refusal (`FitFailure::jeffreys_arming_evidence`);
  - the last outer evaluation's typed refusal;
  - a certified fit whose cone posterior is proved improper.
- Use `gam_custom_family::arm_on_evidence` for that step. It publishes `FitArtifacts::jeffreys_arming_evidence`.
- Put `inner_fit` back to plain `fit_custom_family`, so the lifecycle runs once per route instead of nested.

## Gate

- A BMS fixture fitted both with and without a κ/σ search, where nothing separates: both fits unarmed, with no `[979-PROBE] firth_armed=true` cycle.
- A separated fixture: both routes arm and publish the evidence.

`crates/gam-models/tests/survival_arming_route_2995.rs` shows the probe-capturing pattern.

SauersML:
This PR is the bms-objective cluster. The goal is that a Bernoulli marginal-slope fit solves one objective whichever route it takes.

Fixes #3164
Fixes #3021
Fixes #3045
Fixes #2997
Fixes #3059
Fixes #2978

## What changes
- **#3164: one arming rule on every route.**
  - Before this PR, the exact joint route (a searched Matérn κ or a learned frailty σ) solved the Jeffreys-armed objective on every fit.
  - Now both the exact route and the fast path fit unarmed first, and arm only on the unarmed fit's own typed evidence, through `arm_on_evidence` (#979 ruling (b)).
  - **New evidence source.** A certified unarmed fit can have a posterior that is improper along `ker(S_λ)`, a direction no smoothing parameter reaches. The terminal posterior assembly now detects this: `JointJeffreysPlan::diagnose(precision, Z).reduced_information_is_singular()` on the penalty null space. It publishes the result as `FitArtifacts::improper_penalty_null_posterior`, and that evidence arms the refit.
  - **The armed refit then produced NaN.** On the well-classified probit tail, κ = 1/expm1(−ℓ) ≈ 1/Φ(−m) came within 2× of `f64::MAX` (row 96: κ = 9.77e307). Meanwhile every ℓ-derivative carries a factor φ(m). So `2κ`, `κ·(sᵀx)` and `κ·(h x)(h y)ᵀ` overflowed to ∞ against products that had underflowed to 0. The Jeffreys completion became non-finite and the inner solve was refused.
  - **Fix.** κ now multiplies exactly one derivative factor first (`ExpectedInformationRow::lift`), which keeps every term O(poly(m)). The i979 no-change control, `expected_information_direction_wrappers_keep_their_bits_2922`, is still bitwise.
- **#3021:** a saved BMS model loads through `lift_to_saved_frame`. The reduced slope basis is lifted back exactly once.
- **#3045:** the confound audit no longer depends on the units of the slope columns. The reduced range basis is equilibrated first.
- **#2997:** `summary()` of a BMS fit reports one row per smooth of both the marginal and the slope predictor. Each row is read against its own block's λ and coefficient range (`SummaryBlockOffset`).
- **#3059: a learned σ is differentiated on the finite law.**
  - The rigid kernel differentiates a learnable `GaussianShift` σ on whichever latent law the row anchors on. The fit is no longer pinned to `StandardNormalOnly`.
  - σ is identified only when the slope's fixed part lies outside the slope design's span. When it lies inside, a learned σ is refused as an input error, instead of drifting along the `p0·ln s(σ)` ray.
- **#2978: exact link map, no probability clamp.**
  - The link map is exact: q = η.
  - The empirical intercept solves the anchor in log-probit space on both tails.
  - The residual-repair route uses `mixture_anchor_tower`.

## Evidence
Arming on the separated fixture: `y = 1[z > x]` with offset `−x`, n = 400, `slope_formula = "1"`. The armed fit now certifies on every separated penalized marginal:

| formula | evidence | armed log-lik | slope b |
|---|---|---|---|
| `y ~ s(x)` | `ImproperPenaltyNullPosterior { unreached_dim: 2, information_min: 8.11e-19, floor: 1e-12 }` | −2.558 | 24.7 |
| `y ~ matern(x, centers=6)` | information_min 6.41e-16 | −1.137 | 36.36 |
| `y ~ x + matern(x, centers=6)` | information_min 3.58e-16 | −1.391 | 115.2 |

`bms_route_arming_3164` covers non-separating and separated data, each on the fast path and on the exact route; all four pass. `expected_information_coefficient_derivatives_stay_finite_where_kappa_nears_overflow_3164` sweeps the slope through the band where κ > ½·f64::MAX. It requires every expected-information derivative to stay finite and requires the band to actually be hit.

**Caveat.** On these fixtures the verdict is decided by the plan's existing absolute floor of 1e-12 (floor = max(1e-10·λ_max, 1e-12)), not by its relative term. The gap to the floor is 4 to 7 orders of magnitude. The floor itself is `JointJeffreysPlan`'s own, reused through `JointJeffreysPlan::floor()`; this PR adds no new constant.

## Follow-ups (not fixed here)
- #3216: the flex route still calibrates the row intercept in probability space.
- #3217: a score that separates the pooled pilot gives a degenerate confound-audit metric and a misleading "fully confounded" refusal.

## Breaking changes
- **#3164**
  - The BMS `JeffreysArming` impl (`with_jeffreys_armed`) is removed, and `inner_fit` no longer arms. Arming goes through `gam_custom_family::arm_on_evidence` / `Arming`.
  - New `JeffreysArmingEvidence::ImproperPenaltyNullPosterior { unreached_dim, information_min, information_floor }` variant.
  - New `FitArtifacts::improper_penalty_null_posterior` field (serde-skipped).
  - New `JointJeffreysPlan::floor()`.
  - `arm_on_evidence` arms on `certified_fit_evidence`: the penalty-null evidence first, then `improper_cone_posterior_evidence`.
- **#3021:** `SLOPE_REDUCED_BASIS_RELATIVE_TOL` is removed. So are `recover_original_slope_beta` and the binomial mean-wiggle saved-frame helpers (`binomial_mean_wiggle_saved_geometry` / `_covariance`). Callers use `UnifiedFitResult::lift_to_saved_frame`.
- **#2997**
  - `smooth_term_summary_rows(design, fit, whitening_gram, offset: SummaryBlockOffset)` takes the new offset. All callers are updated, including main's new SBC size-curve tests, the 1266 regression test and the issue-2668 bench probe.
  - `SummarySmoothTermRow` has a new `predictor` field, which the Python `_summary.py` reads.
- **#3059:** a learned σ is refused when the slope's fixed part lies in `span(G_slope)`. The `StandardNormalOnly` pin for a learned σ is removed.
- **#2978**
  - Removed: `BERNOULLI_LINK_PROBABILITY_EPS`, `clamp_bernoulli_link_probability`, `bernoulli_marginal_slope_eta_from_probability`, `empirical_intercept_from_marginal{,_within}`, `empirical_intercept_tail_tolerance`, `empirical_rigid_calibration_eval`, `lse_accumulate`, `lift_iters`.
  - Added: `empirical_intercept(q, slope, probit_scale, nodes, weights)` and `mixture_anchor_tower(q, b, grid)`.

https://claude.ai/code/session_011hAcUyMMXmmmR3Drpjn61B


SauersML:
[swarm lane bms-a] A delta on this PR's #3045 landed on main as 024e92eb4e.

**#3045: the drop rule.** This PR drops a principal direction when sin θ_k ≤ max(n, k)·ε. That is the SVD's own backward error, but the fit never reads the angle. It reads the direction's curvature in the joint Hessian, which is sin²θ_k along u_k = [a_k; v_k]. The Hessian's entries are n-row sums, so they form that curvature only to within γ_n·‖|J||u_k|‖². The angle test effectively compares sin²θ_k against (max(n,k)·ε)², about 1e-25, which is far inside that band (about 1e-13). So it kept directions the Hessian cannot tell from null.
- The rule on main is now: drop iff sin²θ_k ≤ γ_n·‖|J||u_k|‖² + 2·(the sine's SVD band).
- The span bases are resolved at `factor_singular_band`.
- This PR's unit-invariance pins pass unchanged.
- New pin: `effective_reduction_drops_a_direction_the_hessian_cannot_resolve_3045`. At δ = 1e-9 the sine is ~7e-10 and the curvature ~5e-19, and main kept the direction.

PR #3220, opened 2026-09-19
inference: variance-component score test for fully penalized smooths and group() terms

SauersML:
## Problem

Two summary rows test a null that sits on the boundary of the variance-component space:

- **`group(g)` / `re(g)` blocks.** "No group effect" is `σ²_b = 0`.
- **Default double-penalty `s(x)`.** It has a wiggliness penalty plus a null-space ridge, so "no effect" is every variance component of the term at 0.

Neither a Wald statistic nor a likelihood ratio referred to `χ²` is calibrated there.

The smooth row carried the Wood (2013) rank-truncated Wald test, computed from coefficients REML had already shrunk on the same data. About half of all null fits shrank the term to `edf ≈ 0` and reported `p ≈ 1`. The null p-values had a point mass at 1, which made the test conservative at every level:

| family (main, n = 400, 500 reps) | size .10 | size .05 | size .01 | share of p > .99 | KS p |
|---|---:|---|---:|---:|---:|
| gaussian | .074 | .024 | .006 | .460 | < .001 |
| binomial | .064 | .032 | .012 | .484 | < .001 |
| poisson | .080 | .032 | .010 | .498 | < .001 |

## Fix

Both rows now carry the Lin (1997) variance-component score test, referred to its exact finite-sample null law (Wood 2013, *Biometrika* 100). The code is in `crates/gam-terms/src/inference/variance_component_test.rs`.

- **Statistic.** The score is `u = X̃_Rᵀv`, where `X̃_R` is the block with every other fitted column projected out in the fit's curvature metric. Then `T = uᵀΣ₀u`.
- **Reference law.** With a known scale it is `Σ μ_i χ²₁`, `μ = eig(Σ₀^{½} V Σ₀^{½})`. With an estimated scale it is the signed weighted chi-square ratio against the residual `χ²_ν`. Both tails come from exact inversion. There is no moment matching, no halving and no mixture weight to estimate.
- **No atom.** `T > 0` almost surely, so p = 1 is not an atom.
- **Direction.** `Σ₀ = Σ_j K_j / tr(K_j V)`, with `K_j = P⁻¹(a_jS_j)P⁻¹`, `P = Σ_j a_jS_j` and `a_j = 1/tr(G_u⁻S_j)`. `K_j` is penalty `j`'s share of the prior covariance. `Σ₀` is fixed by the design, the weights and the penalties, never by `y`, `b̂` or `λ̂`, and that is what makes the reference law exact.
- **`group()` blocks.** A ridge gives `Σ₀ ∝ I`, which is Lin's `‖u‖²`.
- **Refusals.** When no valid test exists the row gets a typed reason: `variance_component_unpenalized_directions`, `…_no_estimable_direction`, `…_design_unavailable`, `…_no_irls_row_state`, and so on. A range with a direction no penalty touches, or with no penalty at all, is refused with `variance_component_unpenalized_directions`. After main's factor contract every `group()`/`re()` block is a penalized full one-hot block, so no unpenalized random-effect block reaches the test.
- **Where it is computed.** The test runs once on the training fit and is stored in `FitArtifacts::variance_component_tests: Option<Vec<_>>`. The CLI summary, the Rust `smooth_term_summary_rows` and the Python `summary()` all read the same record.
  - `Some(records)`: the route runs the test. These are the standard fit and the Gaussian location-scale mean block without a link wiggle, which is scored from its own `wᵢ/σ̂ᵢ²` row state.
  - `None`: the route does not run the test. Smooths keep `main`'s Wood Wald row, and `group()` rows report `variance_component_test_not_recorded`.

### Remaining gap: routes that still use the Wald row

The field is `None`, so smooths keep `main`'s conservative Wald row, on these routes:
- the binomial, dispersion and survival location-scale fits;
- any fit with a link wiggle;
- payloads saved before this change.

Each of these routes needs its own score and curvature for the tested block. That is follow-up work, not part of this PR.

### Root cause of the power loss in the first version: the direction must be chart-covariant

The first version of this change used the pseudo-inverses `S_j⁺` for the direction. Null calibration was fine, but power collapsed (binomial `dp` .525 → .08 at .05).

`S_j⁺` is not congruence-covariant. In the realized chart of a default smooth (identifiability-constrained, centred, reparametrized), the double penalty's two ranges are complementary but not Euclidean-orthogonal. On one binomial replicate, the linear direction `c` of `s(x2)` had `cᵀS_wiggle c ≈ 0` but `cᵀS_ridge c / ‖c‖² = 0.35`. The `S⁺` direction then scored a chart-dependent mix of the two components, and it read a linear score of `z = −2.98` as `p = .33`.

`K_j` is covariant under `X_R → X_RZ`, `S_j → ZᵀS_jZ`, and it reduces to `∝ S_j⁺` when the chart happens to be orthogonal. On the same replicate it gives `p = .0018`. The unit test `reparametrizing_the_block_does_not_move_the_p_value` pins this invariance for both a known and an estimated scale.

## Calibration results

All results are seeded, with 2-MCSE bands. Failing the band on **either** side fails the cell, and so does a two-sided KS test against U(0,1) at .01. The full tables are in `bench/pvalue_calibration/pv-random-effects/README.md`.

**Random effects, 500 reps per cell** (`y ~ s(x1) + group(g)`, gaussian/binomial/poisson × L ∈ {5, 20, 200} × balanced/unbalanced; 36 cells).

- Every KS p is > .01.
- 2 of the 108 (cell, α) checks fall outside the 2-MCSE band: poisson L=5 unbalanced (.026 at .05) and poisson L=20 unbalanced (.072 at .05). That is the count chance predicts across 216 one-sided checks.
- A 1000-rep rerun of the four poisson L ≤ 20 cells on fresh seeds put every cell inside its band:

| poisson, 1000 fresh reps | .10 | .05 | .01 | KS p |
|---|---:|---:|---:|---:|
| L=5 unbalanced | .113 | .053 | .014 | .839 |
| L=5 balanced | .104 | .049 | .012 | .473 |
| L=20 unbalanced | .108 | .052 | .008 | .397 |
| L=20 balanced | .091 | .050 | .009 | .259 |

**Random-effect power at .05** (`σ_b = .25`, 100 reps). Before this change the row had no p-value at all.

| family | L=5 U | L=5 B | L=20 U | L=20 B | L=200 U | L=200 B |
|---|---:|---:|---:|---:|---:|---:|
| gaussian | .87 | .98 | .96 | 1.00 | 1.00 | 1.00 |
| poisson | .46 | .47 | .47 | .45 | .43 | .30 |
| binomial | .32 | .33 | .28 | .13 | .13 | .05 |

**Double-penalty smooth, null, 500 reps** (`y ~ s(x1) + s(x2)`, null `s(x2)`, n = 400). "Before" is `main`, "`S⁺`" is this PR's first version, and "after" is this PR.

| family | version | .10 | .05 | .01 | p > .99 | KS p |
|---|---|---:|---:|---:|---:|---:|
| gaussian | before | .074 | .024 | .006 | .460 | < .001 |
| gaussian | after | .110 | .034 | .004 | .010 | .823 |
| binomial | before | .064 | .032 | .012 | .484 | < .001 |
| binomial | after | .094 | .046 | .010 | .018 | .604 |
| poisson | before | .080 | .032 | .010 | .498 | < .001 |
| poisson | after | .092 | .046 | .008 | .016 | .671 |

Every "after" cell is inside its 2-MCSE band (`.10 ± .027`, `.05 ± .019`, `.01 ± .009`) at every level, and no KS test rejects. `S⁺` was also calibrated (binomial .100/.046/.006, KS .897); the README has all three rows for each family.

**Double-penalty power, 200 reps**, as rejection rates at .10 / .05 / .01. `dp` is `0.4·sin(2πx2)` and `dp2` is `0.5·sin(4πx2)`.

| family | alt | before | `S⁺` | after |
|---|---|---|---|---|
| gaussian | dp | 1 / 1 / 1 | 1 / 1 / 1 | 1 / 1 / 1 |
| binomial | dp | .625 / .525 / .255 | .130 / .080 / .010 | .645 / .530 / .325 |
| poisson | dp | .820 / .730 / .465 | .240 / .120 / .015 | .835 / .725 / .540 |
| gaussian | dp2 | 1 / 1 / 1 | .920 / .805 / .480 | 1 / 1 / 1 |
| binomial | dp2 | .550 / .420 / .170 | .130 / .075 / .015 | **.280 / .165 / .040** |
| poisson | dp2 | .840 / .780 / .575 | .175 / .105 / .025 | **.450 / .305 / .145** |

> [!WARNING]
> **Known power regression on the two-period alternative.** On `dp2` under binomial and Poisson, this PR has less than half of `main`'s power, which is a real cost. The cause:
> - The exact reference law needs a direction fixed before seeing `y`.
> - The prior-covariance direction `Σ₀` down-weights high-frequency departures, because the wiggliness penalty's prior variance falls steeply with frequency.
> - `main`'s Wald row adapts to the data through `λ̂`. That adaptivity is the same thing that makes its null law depend on `λ̂` and gives it the atom at `p = 1`.
>
> On the one-period alternative, the new test matches or beats `main` in every family. Recovering `dp2` power without losing calibration would need a different statistic, for example a combination over several fixed directions with an exact joint law. That is out of scope here.

**Gaussian location-scale mean smooth** (`noise_formula`; Rust gate, n = 200, 1000 seeded reps, noise SD `0.3·e^{x1}`):

| version | reps | .10 | .05 | .01 | KS p |
|---|---:|---:|---:|---:|---:|
| scale taken as known (`φ = 1`) | 200 | .110 | .075 | .035 | .767 |
| scale estimated | 200 | .105 | .065 | .030 | .945 |
| scale estimated, fresh seeds 200–1199 | 1000 | .095 | .050 | .012 | .852 |
| **scale estimated, the gate (seeds 0–999)** | 1000 | .093 | .050 | .015 | .926 |
| **the gate after main's data-derived σ floor** | 1000 | .094 | .051 | .016 | .987 |

- The first version took the scale as known and failed at .01. `σ̂` is fitted to the very residuals the score is built from, so the level is now treated as estimated, with a `χ²_ν` ratio.
- At 200 reps the .01 band (`± .014`) was too coarse to separate chance from bias, so the gate runs 1000 reps (band `± .0063`), with a nextest timeout override.
- The gate passes at every level. The size at .01 of `.015` is 1.6 MCSE high: not significant, but disclosed.
- Main replaced the constant σ floor with one derived from the response's recording grid. The mean-block row state now rebuilds σ with the fit's own floor. After that merge the gate reads .094 / .051 / .016 (KS .987), still inside every band.
- Power: 40 of 40 at .01.

**Rust CI gates** (seeded, n = 200, 200 reps; rerun after merging `main` at ce392df9). Every family whose fits certify is inside the two-sided bands and passes KS:

| gate | .10 | .05 | .01 | KS p | failed fits |
|---|---:|---:|---:|---:|---:|
| DP gaussian¹ | .095 | .050 | .010 | .899 | 0 |
| DP binomial | .100 | .065 | .020 | .698 | 0 |
| DP poisson | .100 | .060 | .015 | .359 | 0 |
| DP gamma | .120 | .060 | .010 | .193 | 0 |
| DP negative binomial (old base df02753c) | .082 | .036 | .005 | .261 | 5 |
| DP negative binomial (current `main` merged) | — | — | — | — | **192** |
| DP tweedie | .090 | .055 | .015 | .345 | 0 |
| DP beta | .095 | .040 | .010 | .573 | 0 |
| RE gaussian (L=20 unbalanced) | .090 | .045 | .015 | .748 | 0 |
| RE binomial | .060 | .040 | .010 | .348 | 0 |
| RE poisson | .125 | .055 | .005 | .243 | 0 |
| DP gaussian location-scale (1000 reps) | .094 | .051 | .016 | .987 | 0 |

¹ Values from before the merge. The gate passes after it too, but `build.sh` dropped that line from the failing run's output.

The power controls pass. DP gaussian finds a real effect in 40/40 fits at .01. RE at σ_b = 1 has power at .05 of gaussian 1.00, binomial .90 and poisson .95, against a null band of .12.

> [!CAUTION]
> **`negative_binomial_null_smooth_size_is_within_monte_carlo_error` fails after merging `main`, and the cause is in `main`.**
> - 192 of the 200 null fits fail before any p-value is computed. The error is `Negative-binomial (theta, rho) optimization did not certify a joint optimum within 200 round(s)`, with a theta-score Newton residual of 3.7e-8 against a bound of 1e-10.
> - This PR does not touch the NB (θ, ρ) optimizer or P-IRLS. On the old merge base (df02753c) the same gate had 5 failed fits and passed. It has failed with 192 since the first merge of newer `main` (58846c92), and it still fails after this one (ce392df9).
> - `main`'s own Wald gate `negative_binomial_null_smooth_wald_size_is_within_monte_carlo_error` uses the same seeds, the same DGP and the same fail-share gate, so it should fail the same way. I have not run it here.
> - No fix exists yet. The gate stays as it is; it is not skipped or loosened.

**Failed fits.** These are the fits the outer optimizer refuses to certify ("A fit object must only ever come from a converged optimization"). They are reported and never scored. Each gate fails outright above 5%.

## Tests

- **`crates/gam-terms` unit tests (12).** They cover:
  - one-way ANOVA F identity (balanced, estimated scale);
  - null uniformity, known and estimated scale, for both a single ridge and a double penalty;
  - both linear and wiggly effects detected;
  - invariance to penalty rescaling and to block reparametrization;
  - independence from where the fit left β;
  - the typed refusals, including a block with no penalty at all.
- **`tests/inference/misc/sbc_random_effect_variance_component_size_curve.rs`.** Size and KS for 3 families, plus power.
- **`tests/inference/misc/sbc_double_penalty_smooth_family_size_curve.rs`.** Renamed from `sbc_wood_smooth_test_family_size_curve.rs`. Size and KS for 7 families and the Gaussian location-scale mean smooth, plus power.
- **Python.** `tests/test_summary_pvalue_weights_and_location_scale.py` runs the location-scale summary through the wheel. Against the rebuilt wheel, `test_a_location_scale_fit_tabulates_its_mean_smooths` and `test_binomial_trials_match_expanded_bernoulli_rows` pass. `test_poisson_integer_weights_match_duplicated_rows` fails with an edf mismatch (4.2453 vs 4.2429). The mismatch is in the fitted edf, which comes from the REML fit and not from any p-value code this diff changes. It is pre-existing.
- **Checks.** After merging `main` (ce392df9):
  - `./build.sh check --workspace --tests` is clean.
  - `./build.sh test -p gam-terms --lib variance_component_test` passes 12/12.
  - `scripts/ban_scanner.sh` reports 0 new and 0 stale.
  - The RE, DP and location-scale gates above pass, except the NB gate, which fails because of `main`.
- **Bench.** `bench/pvalue_calibration/pv-random-effects/` holds `null_calibration.py`, `analyze.py`, the results JSONL and the README.

## Breaking changes

- `FitArtifacts::random_effect_tests` is now `FitArtifacts::variance_component_tests`. Its records are `VarianceComponentTestRecord` from `gam_terms::inference::variance_component_test`. The `random_effect_test` module is gone.
- Unavailable-reason labels `random_effect_*` are now `variance_component_*`. `SmoothPValueUnavailable::{RandomEffect, RandomEffectTestNotRecorded}` are now `{VarianceComponent, VarianceComponentTestNotRecorded}`.
- The driver `random_effect_test_records(design, spec, fit)` is now `variance_component_test_records(design, fit)`.
- A default (double-penalty) smooth, and any smooth whose penalties cover every direction, now reports the variance-component score test's statistic, `ref_df` and p-value instead of the Wood Wald values.
- `FitArtifacts::variance_component_tests` is `Option<Vec<VarianceComponentTestRecord>>`. A model saved before this change reads as `None`: its smooths keep the Wald row and its `group()` rows report `variance_component_test_not_recorded`. Refit to get the new test.
- A Gaussian location-scale fit's mean smooths (default double penalty) now report the variance-component test.
- The `hypothesis` field of `VarianceComponentTest` and its fixed-effect F branch are deleted. `VarianceComponentTermRequest` carries `penalties: Vec<Array2<f64>>` directly (no `TestedBlock`). This follows main's factor contract, which removed the unpenalized factor-block shim that was the branch's only caller.

## Coordination

- **#3172 (pv-interactions)** independently moves penalized-smooth rows to a score test of the same form (`smooth_score_test.rs`). It has since also replaced `S_l⁺` with a chart-covariant `K_l`. Both PRs route default `s(x)` rows, so whichever lands second must consolidate onto one kernel. Details are in a comment on #3172.
- **claude/pygam-factor-contract** (merged): its factor-block ranges are the ones used here. Main has been merged in, and the fixed-effect branch its contract deleted from `random_effect_test.rs` is deleted here as well.
- **claude/pygam-group-scaling**: the test is invariant to the ridge scale (`rescaling_a_penalty_does_not_move_the_p_value`), so group-penalty rescaling does not interact with it.
- `smooth_significance` (the LR API owned by claude/pygam-pvalue) is not touched.

https://claude.ai/code/session_01B7VnpLGD9oyXrw3YfiXPh9


SauersML:
**Failing check: `negative_binomial_null_smooth_size_is_within_monte_carlo_error`.** The cause is in `main`, not in this PR.

- **What fails.** 192 of the 200 null fits fail at the fit stage, before any p-value code runs. The error is `Negative-binomial (theta, rho) optimization did not certify a joint optimum within 200 round(s)`, with a theta-score Newton residual of 3.7e-8 against a bound of 1e-10.
- **Before `main` moved.** On the old merge base df02753c the same gate had 5 failed fits and passed.
- **This diff.** It does not touch the NB (θ, ρ) optimizer, P-IRLS or `theta_bound`.
- **`main`'s own gate.** `negative_binomial_null_smooth_wald_size_is_within_monte_carlo_error` uses the same seeds, the same DGP and the same fail-share gate, so it should fail the same way. I have not run it on `main`.
- **Fix.** I found no fix PR or issue for this. The gate is left as it is: not skipped and not loosened.

Every other gate passes on ce392df9: RE for gaussian, binomial and poisson; DP for gaussian, binomial, poisson, gamma, tweedie and beta; the Gaussian location-scale gate at 1000 reps; and the power controls.

---

SauersML:
Merge check against current main (b7f57a0fd4, plus #3224's candidate): not merged, left open. One design conflict needs a decision first.

**Two implementations of one statistic.** Main's #3172 added `crates/gam-terms/src/inference/smooth_score_test.rs`: the Lin (1997) boundary variance-component score test with its weighted-χ² reference, for te/ti/by/2-D smooths, called from `crates/gam-solve/src/estimate/smooth_term_summary.rs:49-50`. This PR adds `variance_component_test.rs`, the same boundary score test, for `group()`/`re()` blocks and fully penalized smooths, and deletes `random_effect_test.rs`. Merging both leaves two modules computing one test, routed by term class.

Where the two agree and where they don't:
- **Exact split.** When the penalties' null spaces form a basis (a smooth with its null-space penalty), write β = N₁c₁ + N₂c₂. Then this PR's `K_j = P⁻¹(a_j S_j)P⁻¹` reduces to `N_j (a_j N_jᵀ S_j N_j)⁻¹ N_jᵀ`, which is main's `K_l = N_l (N_lᵀ S_l N_l)⁻¹ N_lᵀ` up to the per-component scale.
- **Overlapping penalties** (tensor marginals). Main falls back to `K_l = S_l⁺`, its "additive working model". This PR keeps the chart-covariant `P⁻¹(a_j S_j)P⁻¹`.
- **Component weights.** Main puts each component on its null scale, `K = Σ_l K_l / t_l` with `t_l = tr(K_l C)`. This PR uses `a_j = 1/tr(G_u⁻ S_j)`.

Two smaller items for the landing:
- `crates/gam-models/src/fit_orchestration/drivers/variance_component_test.rs:37-43` keeps the known-scale `.coefficient_covariance_scale().ok()….unwrap_or(1.0)` dispersion substitute. #3224 retires exactly that substitute in the random-effect driver, in favour of a typed `KnownScaleUnavailable`, and it is ahead of this PR in the merge queue. The merged version needs the same typed refusal.
- 5b9671f5a7 moves the random-effect size/power gate from panicking on an uncertified fit to reporting it and excluding it from the rates, failing above 5%. The certification failure it reports (binomial power replicate, group sd 1, rep 6, `BfgsCostStallExit`, |Pg| 1.8e-7 > 2.4e-10) has no issue that I could find.

Merging main now gives 8 conflicts: the two `random_effect_test.rs` modify/deletes, `smooth_term_summary.rs` (337 lines from #3172), `summary.rs`, `saved_summary.rs`, `inference/mod.rs`, `tests/quality/calibration/registry.rs`, and `sbc_wood_smooth_test_family_size_curve.rs` (deleted here, modified on main). They are resolvable once there is one score-test module and one `K` rule for every term class.

SauersML:
Taken over by RF-0748-2; driving to merge.

---

SauersML:
Taken over by RF-0748-2; driving to merge.

I compared the branch (head ce392df9, merge base c658dab2) with current main (c663ba6d, 545 commits past the base). I'm closing this PR and porting what only it has onto main's modules in a fresh PR, because it can't be merged cleanly:

- **It would delete working code on main.** The merge removes `gam-terms/src/inference/random_effect_test.rs` (893 lines) and the 7-family gate `tests/inference/misc/sbc_wood_smooth_test_family_size_curve.rs`, which tests the score test that landed after the base.
- **Main already has the score test this PR adds.** Since #3172, main scores smooth rows with `smooth_score_test.rs` (`ScoreTestFit::test_term`, with no Wald fallback in the summary path) and `group()` rows with the recorded `random_effect_tests`. This PR keeps a `wood_smooth_test` fallback in the summary instead.
- **On non-overlapping penalties the two tests are the same.** Main's direction there is `K_l = N_l(N_lᵀS_lN_l)⁻¹N_lᵀ`, normalised by `1/tr(K_l V)`. This PR's `K_j = P⁻¹a_jS_jP⁻¹` differs only by the factor `a_j`, which the normalisation cancels.

What only this PR has, which the follow-up ports:

1. **Overlapping penalties (tensor marginals).** Main falls back to `S_l⁺` (`component_covariances`, `None => ranges.iter().map(PenaltyRange::pseudo_inverse)`), and `S_l⁺` changes under a change of coefficient basis. The PR's `P⁻¹a_jS_jP⁻¹` doesn't. The port brings that rule plus a basis-change test for the overlapping case.
2. **`group()` kernel.** `random_effect_test.rs` scores `‖u‖²`, which is right only when the ridge is the identity. The port passes the block's actual penalty so that `K = S⁻¹`, with a basis-change test.
3. **Location-scale.** The mean-block and `group()` recording on a Gaussian location-scale fit, with an estimated φ, and its null and power gates.
4. **SBC gates.** The power checks and the 5% failed-fit limit.

The NB gate failure (192 of 200 null fits end in `Negative-binomial (theta, rho) optimization did not certify a joint optimum`) comes from main's NB optimizer, not from this diff. I'll track it in its own issue.

---

PR #3259, opened 2026-09-19
conformal: certified full conformal for binomial, Poisson, NB and Gamma

SauersML:
## Problem

`model.predict(interval="conformal", training_data=...)` and `gam predict --conformal` built an exact full-conformal set only for Gaussian-identity models fit without offsets or prior weights. Every other family was refused, with a message pointing to split conformal. pyGAM has no conformal route. Its `prediction_intervals` exists only on `LinearGAM`, so for binomial and Poisson data it has no predictive band at all.

## Root cause

The full-conformal substrate was the Gaussian closed form: one Cholesky per test row on the frozen penalty. Non-Gaussian families need a refit of the augmented penalized likelihood per candidate response. That refit had no certificate and no bound on the candidate support, so the route refused. Offsets were refused because the substrate never composed them. Prior weights have no weight for the unlabeled candidate point, so the n+1 points are not exchangeable, but the refusal message did not say why or name `calibration=`.

## Fix

- **`crates/gam-models/src/inference/full_conformal_glm.rs` (new).** A certified full-conformal set for Bernoulli-logit, Poisson-log, NB-log (θ frozen like λ) and Gamma-log. The score is the working score `|∂ℓ/∂η|` (Pearson `y/μ − 1` for Gamma). The augmented fit is a symmetric function of the n+1 points, so the score construction is exchangeable; the module doc gives the argument.
  - **Solve certificate.** Every augmented solve is damped Newton with a self-concordance-style certificate: `e = 2‖g‖_{H⁻¹}`, valid while `max lever · 2e ≤ ln 2`, using `|d ln w/dη| ≤ 1` for these families. A rank decision is taken only when the certified score errors cannot flip it.
  - **Binomial.** Exact enumeration of {0, 1}.
  - **Poisson / NB.** A certified walk in the test-score coordinate, from the score of the count 0 up to a provable, data-derived tail score, with no hand cap. The tail comes from the intercept KKT row: with `u_i < y_i`, no count past `max(max y, Σy/K)` can enter. NB also has the score bound θ and takes the smaller of the two, which keeps near-Poisson NB fits (θ ≈ 1e6) on the data scale. The score interval is bisected, and one certified solve decides a whole sub-interval. A run of member pieces covers the counts strictly inside the images of its end scores. Each count inside an end image's slack, inside an undecided sliver, or tied with a training twin is decided by its own solve. So the work follows the set's boundaries, not the width of the count range: a Poisson fit with mean ≈ e¹¹ costs a few dozen solves, not 10⁵. The `z(s)` images are widened by a floating-point rounding bound, so the count 0, which sits exactly on the walk's low end, cannot be rounded out of a member run.
  - **Gamma.** A bisection walk in the test-score coordinate `s ∈ (−1, s_top]`. `s_top` comes from the intercept KKT row and `u_i > −1`. Each solve certifies a whole sub-interval. Run endpoints are mapped to `z` by a dedicated solve at the exact boundary `s`, so the set is tight to solver accuracy.
  - **Ties.** Discrete families use the randomized p-value, so sets are exact rather than conservative. The tie uniform is seeded from a hash of the labeled responses and the test row, so every front end returns the same set.
- **Offsets.** Offsets are composed for labeled and test rows, on both the Gaussian and GLM routes. The Gaussian set is built on `y − o` and shifted back.
- **Prior weights.** These now raise a typed `FullConformalError::PriorWeights`. It says why the set is undefined and points to `predict(interval="conformal", calibration=...)` / `gam predict --conformal --calibration`.
- **Persistence.** Only the p×p frozen penalty is persisted; no n-sized data. The labeled rows come from the caller per call.
- **Certificate column.** The Gaussian route keeps main's honest REML re-selection (Layer 3) and its `conformal_certificate` codes, now with offsets composed. GLM rows report a new typed refusal, `-7` `refused:glm_frozen_penalty`, when the fit selected a smoothing parameter or an NB θ. They report `0` `exact_frozen` when the fit selected nothing: an unpenalized Bernoulli, Poisson or Gamma fit selects nothing from the responses, so its set is exact. The route also returns a `conformal_set_components` column, which counts the intervals in the union.
- **`conformal_rank_threshold`.** Resolves `α(n+1)` at a decimal level, e.g. `1 − 0.9`, to the intended integer rank.

### Frozen-penalty note

λ̂ is selected on the same labeled rows the set is built on. The frozen-λ augmented problem is therefore not exactly exchangeable, and the cost is O(1/n) in coverage. Measured on binomial with n = 59: 0.8907 ± 0.0022 with a same-data penalty, against 0.9001 ± 0.0021 with a penalty from an independent draw. This is why smooth GLM rows report `conformal_certificate = -7`. The Python Monte Carlo test fits the penalty on an independent draw, which makes the n+1 rows exactly exchangeable, and asserts two-sided nominal coverage. The benchmark reports the same-data route users actually call.

## Tests

- **Rust** (`full_conformal_glm` unit tests):
  - discrete sets equal brute-force enumeration over the support (with offsets);
  - the near-Poisson NB tail stays on the data scale;
  - a Poisson fit with mean ≈ e¹¹ costs only its boundaries, and brute force agrees around both ends;
  - replayed NB draws keep the count 0 at the walk's low edge;
  - the Gamma set is a superset of the exact set on a fine grid, and every endpoint is an exact boundary;
  - the labeled refit reproduces the penalized fit;
  - penalty recovery;
  - the GLM certificate is exact only when nothing was selected;
  - score/weight derivative identities;
  - the decimal rank threshold.
- **`tests/test_full_conformal_families_route.py`:**
  - seeded Monte Carlo coverage per family, within 2·MCSE two-sided of 0.9, which rejects over-coverage beyond the 1/(n+1) granularity;
  - sets lie in the support;
  - Gaussian and Poisson offsets;
  - the prior-weights typed error names `calibration=`;
  - the saved model carries no training rows;
  - CLI parity with `gam predict --conformal`.

  Every route test fails at HEAD: HEAD refuses every non-Gaussian family and every offset, and its refusal message has no `calibration=`.
- **Benchmark:** `bench/pygam_compare/conformal_coverage.py` → `bench/pygam_audit/conformal_coverage.md`.

### Breaking changes

- Full conformal with prior weights now raises the typed `FullConformalError::PriorWeights` (Python `InvalidConfigurationError`) instead of the generic Gaussian-only refusal.
- Non-Gaussian families and offset models, which used to be refused, now return full-conformal sets. The route adds a `conformal_set_components` column, and `conformal_certificate` gains code `-7` (`refused:glm_frozen_penalty`).
- `ExactFullConformalPenalty::from_s_lambda` now takes the penalty count.
- A numerical full-conformal failure raises `PredictionError` instead of `ValueError`. The generic `PredictError::Other` also maps to `PredictionError`, matching main.

### Benchmark

Draft: the 1000-rep coverage table (`bench/pygam_audit/conformal_coverage.md`) is being rerun on the merged main and will be pushed to this branch.

### Known blocker (not this lane)

The Python NB route tests fail on main's NB θ estimator, not on this route: `crates/gam-solve/src/estimate/optimizer.rs:2098` `let theta_bound = reml_tol;` is still unfixed on main.

https://claude.ai/code/session_01FeQxv5ucNWTEjjsv14cqtH


SauersML:
**Failing and not this PR's to fix:** `test_glm_full_conformal_set_is_a_set_in_the_support[negative_binomial]` and `test_glm_full_conformal_covers_at_the_nominal_level[negative_binomial]` in `tests/test_full_conformal_families_route.py`.

The failure happens in `gamfit.fit(..., family="negative_binomial")`, before any conformal code runs. After merging main (c40da47b), every NB smooth fit at `y ~ s(x, k=6)` is refused:

```
RemlConvergenceError: Negative-binomial (theta, rho) optimization did not certify a joint optimum within 200 round(s):
projected rho-gradient 7.601e-5 against 5.627e-4, theta-score Newton residual 1.143e-8 against 1.000e-10
```

Default `y ~ s(x)` fits at n=300 also fail about 1 time in 4 (seed 3 of 0..3). Before the merge, the same tests passed on this branch.

**Diagnosis (`crates/gam-solve/src/estimate/optimizer.rs:2022`).** The two parts of the joint (θ, ρ) optimum are checked against bounds on different scales:
- ρ is certified against `certificate.stationarity.bound()`, a scale-aware band of about 5e-4 here.
- θ is certified against the raw `reml_tol` (1e-10), as a Newton displacement in log θ.

Each alternation round refits θ exactly at the current η and then stops ρ anywhere inside its band. That moves η at about the 1e-8 level, so the θ residual stays around 1e-8 and can never reach 1e-10. The loop runs all 200 rounds and gives up.

**Proposed patch (not in this PR, since it's the NB estimator and outside the conformal lane).** Certify θ in the same units as ρ. Take the log-θ gradient `-θ·score` together with ρ's projected gradient, and judge both against the certificate's `stationarity.bound()`. Alternatively, derive `theta_bound` from the same dimensionless band instead of `reml_tol`. Either way the fit is minted only when both partials are stationary under one tolerance.

No open PR fixes this yet (checked #3300, the count-family sweep, and it doesn't touch it). The NB tests stay in this PR unskipped, because they flag a real refusal. When main carries a fix, I'll merge it in.

---

SauersML:
After merging origin/main (69be5380), `gam-models inference::full_conformal::honest::tests::full_conformal_coverage_holds_for_every_row_class` fails. The other 27 targeted tests pass, including every `full_conformal_glm` test and the Monte Carlo coverage test.

This failure is not from this PR. It is main's open issue #3394. On a Gaussian single-penalty row, the honest map's endpoint refit (`problem.run(...)` in `honest.rs`) does not converge, so the row is refused as `refused:refit_failed`. Two things confirm it is main's:
- This PR does not change the honest Gaussian path; the merge only added offset composition upstream of it.
- The same test passed on this branch before the merge (28/28).

No fix for #3394 exists on main or in an open PR, so there is nothing to port yet. I'm leaving it to #3394.

---

SauersML:

---

SauersML:
RF-0748-3 is driving this PR to merge. Another session also pushed a main merge (87e39fae) just now. I've merged that into this branch rather than overwriting it, so the head is now bf1d7da6. To avoid two sessions racing on the head, please don't push unless something breaks.

Validation on d4b3e3f2 (main e393d691, which includes 8256e9b7 / #3349, the NB θ joint bound derived from the ρ band):
- `cargo test -p gam-models --lib full_conformal`: 27 passed, 1 failed. The failure is `honest::tests::full_conformal_coverage_holds_for_every_row_class`, which is main's #3394, as noted above.
- Python route tests (release-dev wheel, including the NB variants): running now. Results will follow.

---

SauersML:
RF-0748-3 validation before the merge (20b7967d):
- Main merged in with a merge commit; no conflicts.
- `cargo check --workspace --tests`: clean on 9ca78c8a.
- `full_conformal` tests: 27/28 pass. The one failure also fails on main itself (#3394), and this PR does not touch it.
- Python conformal tests: 12/12. CLI test: 1/1.

---

Issue #2363, opened 2026-07-18
Persistent warm-start cache can change fit OUTCOMES, not just speed: cold-cache machines get different (and sometimes failing) results


SauersML:
## Summary

The persistent on-disk warm-start cache (`$XDG_CACHE_HOME/gam/warm/v1/`) can change **what answer a fit produces**, not merely how fast it produces one. A fit that converges on a machine which has run that design before can fail outright on a clean machine, from identical source, identical data, and identical seed.

This was discovered while root-causing #2361 and is filed separately because the hazard is architectural, not gamma-specific.

## Evidence

From #2361, same commit, same data (`numpy.default_rng(0)`), same seed, one machine:

| cache state | first inner solve | frozen gamma shape ν | outcome |
|---|---|---|---|
| populated by an earlier run | 1 PIRLS iteration | 3.9615 | fit converges and certifies |
| moved aside (cold) | 6 PIRLS iterations | 1.2189 | outer REML grinds 200 iterations, raises `IntegrationError` |

True ν is 4.0. Both solves reach the same β̂ (rss 210.387 vs 210.448, edf 11.90 vs 11.71) — only the latched nuisance parameter differs, and that difference decides whether the model fits at all.

## Why this is a defect in its own right

The immediate cause in #2361 is a specific latch capturing a path-dependent value, and that is being fixed there. But the general property this exposes is worth removing independently:

1. **Reproducibility.** Two users running the same script on the same data with the same version can get different results — one a fitted model, one an exception — determined by whether either had previously fit a similar model. That is invisible to the user and impossible to debug from the script alone.
2. **Testing blindness.** Local runs pass and CI fails (or vice versa) for reasons unrelated to the change under test. In #2361 this hid a deterministic failure for five days and sent two independent investigations down a build-configuration/CPU-codegen hypothesis that was entirely wrong. The A10 "passing" build and the MSI "failing" build were the *same code*; only the cache differed.
3. **Silent state.** Nothing in the fit output indicates that a cached warm start was consumed or that it influenced anything beyond speed.

## What a warm-start cache is allowed to do

A warm start should be a pure convergence accelerator: it may change iteration counts and wall-clock, never the converged answer or the success/failure verdict. Any quantity captured *once* from "the first converged solve" violates that contract as soon as warm-starting changes which solve is first, or how converged it is.

## Suggested directions (not prescriptive)

[...]

PR #3330, opened 2026-09-20
Seed initial.sp and the ρ-domain Gram at the cold-start working weights

SauersML:
Follow-up to #3125 (merged), lane `conv-fuzz-families`.

## Problem
In the #3125 family fuzzer, the IG(log) n=50 `ff-base`/`ff-null` cells regressed. The analytic `initial.sp` seed and the ρ-domain Gram both read their working weights from a pilot inner solve at the ρ = 0 anchor. When that pilot refused ("LM step search exhausted"), two fallbacks kicked in:
- the seed fell back to the anchor;
- the domain Gram fell back to the prior weights.

So the start of the outer search depended on whether a solve it did not need happened to converge.

## Root cause
The seed and the Gram need a working-weight vector, not a converged inner fit. Taking the weights from a pilot solve made them exist only where that solve succeeds.

## Fix
- **New `pirls::start_working_weights`.** It returns the Fisher working weights at the cold P-IRLS start: `default_beta_guess_external` pushed through `irls_update` (standard links) or `update_glmvectors` (other links). This plays the role of mgcv's `mustart`. The weights exist for any data the family supports and need no inner solve.
- **One freeze helper.** New `RemlState::apply_lambda_search_freezes` applies the five λ-search nuisance freezes (NB θ, Tweedie φ, Gamma shape, Beta φ, dispersion φ). `execute_pirls_if_needed`, the test harness and the new `start_curvature_weights` all call it, so the start weights use the same frozen scale as the search.
- **Callers.** `analytic_initial_sp_rho` and the ρ-domain Gram in `estimate/optimizer.rs` now use `start_curvature_weights`.

## Tests
- `pirls::tests::start_working_weights_follow_response_units` (IG InverseSquared) checks two things:
  - the weights equal wᵢ·ȳ_w³/(4φ);
  - under y→3y, φ→φ/3 they scale by exactly 3⁴.
- `./build.sh check --workspace --tests` passes.
- `./build.sh test -p gam-solve --lib`: 2070 pass, 2 fail. Neither failure comes from this change:
  - `refused_value_gradient_evaluation_names_its_reason_2982` fails only under the non-nextest runner, because another `log` backend owns the process.
  - `estimated_nuisance_fits_land_in_the_same_place_cold_and_warm_2363` also fails on main 96a81f3b without this change: cold and warm arms differ for all three families there too. This belongs to the #2363 cache-invariance owner.

## Fuzz before/after by cause
**Setup.** 16 IG(log) n=50 cells: `fuzz_families.draw('inverse-gaussian(log)', 50, …)`, 8 `ff-base` seeds and 8 `ff-null` seeds. Each was fit with `gam fit --family inverse-gaussian "y ~ s(x0) + s(x1) + link(type=log)"`. The two arms are debug builds:
- before: main `c663ba6d`;
- after: this PR `afccdd4d`, which is that main plus this diff.

| arm | Converged | refused |
|---|---|---|
| main `c663ba6d` | 16/16 | 0 |
| PR `afccdd4d` | 16/16 | 0 |

**The #3125 regression no longer reproduces on current main.** The pilot's "LM step search exhausted" refusal at ρ = 0 was fixed elsewhere in the meantime, so the fallbacks this PR deletes are no longer exercised on these cells.

**Where the two arms differ.** They reach the same optima within the outer tolerance:
- |Δedf| ≤ 0.013 (base_5 5.058→5.071, null_6 3.472→3.462);
- |Δreml| ≤ 3.8e-3 on a criterion of about 35, split in both directions.

**Why it is still worth merging.** The start no longer depends on a solve it does not need, and two SPEC-banned fallback paths are deleted ("In general, do not paper over solver issues.").

## Breaking changes
- `RemlState::data_curvature_weights` removed; replaced by `start_curvature_weights`, which needs no inner solve.
- `analytic_initial_sp_rho` errors now propagate as typed errors instead of silently falling back to the anchor. It returns `Ok(None)` only when there is no smoothing coordinate.
- The prior-weight fallback for the ρ-domain Gram is removed (internal).

## Validation (landing session L-01)
Head `afccdd4d` is `238b0086` merged with `origin/main` `c663ba6d`. The PR's diff against main is unchanged: 5 files, +308 −277. Main and this PR both touch `optimizer.rs` and `gradient_hessian.rs`, and the merge was clean.

- [x] `cargo check --workspace --tests` on `afccdd4d`: EXIT 0. The only warning is a pre-existing unused `IxDyn` in files this PR does not touch.
- [x] `cargo check --workspace --tests` on the intermediate merge `92becfb5` (main `49a3379b`): EXIT 0. `start_working_weights_follow_response_units` 1/1 on `92becfb5`.
- [x] gam-solve on `238b0086`, [focused proof run 35495583878](https://github.com/SauersML/gam/actions/runs/35495583878): `estimate` 610/610, `pirls` 168/168, `reml` 481/481.
- [x] `start_working_weights_follow_response_units` 1/1 on `238b0086`: [focused proof run 35496470278](https://github.com/SauersML/gam/actions/runs/35496470278). That run is marked failed only because its second filter, `initial_sp`, matched 0 tests.
- [x] Adversarial re-read against main. The diff adds no retry, fallback, jitter or ridge. It deletes the two refusal-driven fallbacks, and every error now propagates typed.
- [x] IG(log) fuzz before/after: the table above.
- [ ] Python: not run here, because a maturin build did not fit alongside the Rust validation in this lane's container. The diff has no Python-visible API change.

https://claude.ai/code/session_01EbMfcNHbmxVH6jUVSnd9wA


SauersML:
Taken over by landing session L-01 (#784 block-correction rho-Hessian). The last push was 02:27Z. I'll merge origin/main as a merge commit (no rebase or force-push), check the diff against SPEC, validate and land it.

---

Issue #3331, opened 2026-09-20
ACC-1 regression since #3191: additive/GLM smooths lose edf and the 95% posterior-mean band under-covers (add4_n5000 0.857, pois_add2_n2000 0.86, binom_add4_n5000 0.90)


SauersML:
I compared held-out results before and after #3191; the full table is at https://github.com/SauersML/gam/pull/3191#issuecomment-5747040797. The **before** build is `2c07549b49`, the first parent of the #3191 merge `c14c88d3ab`. The **after** build is main `9ad6cad14d`. Several additive and GLM cases now select fewer effective degrees of freedom, their error against the true mean goes up, and the 95% posterior-mean band covers the true mean well below its nominal level. The 1-D wiggly cases that #3191 was aimed at improved a lot. Held-out deviance barely moves on the regressed cases, so a deviance-only gate does not catch this.

| case (seed) | truth MSE before→after | cov(μ) 95 before→after | edf before→after | Δ held-out dev (paired SE) |
|---|---|---|---|---|
| `add4_n5000` (`default_rng(5000)`) | 0.0136→0.0306 | 0.986→**0.857** | 20.3→18.4 | +0.44% (0.14) |
| `hetero_n500` (`default_rng(502)`) | 0.0132→0.0217 | 0.84→**0.566** | 9.8→5.94 | +1.10% (0.96) |
| `pois_add2_n300` (`default_rng(308)`) | 0.157→0.310 | 1.0→0.923 | 13.4→6.94 | +0.71% (1.44) |
| `pois_add2_n2000` (`default_rng(2008)`) | 0.0468→0.0435 | 0.957→**0.86** | 16.7→10.8 | −0.19% (0.21) |
| `binom_add4_n5000` (`default_rng(5004)`) | 0.00061→0.00081 | 0.979→**0.903** | 17.6→14.0 | +0.02% (0.07) |
| `binom_add4_n300` (`default_rng(304)`) | 0.0132→0.0149 | 0.88→0.787 | 12.3→10.1 | −1.02% (5.58) |
| `add4_n200` (`default_rng(200)`) | 0.470→0.580 | 0.93→0.85 | 12.9→12.3 | +4.51% (5.84) |
| `g1d_doppler_n500` (`crc32("doppler:500")`) | 0.0744→0.0843 | 0.868→0.852 | 41.6→35.9 | **+3.58% (1.29)** |

Definitions:
- **truth MSE** is `mean((μ̂−μ_true)²)` on held-out rows.
[...]

PR #3191, opened 2026-09-19
smooth bases: data-derived pilots grown on REML evidence

SauersML:
## Problem

Default basis sizes were fixed per-family constants carried over from mgcv:
- factor smooth `FACTOR_SMOOTH_DEFAULT_BASIS_DIM = 10`
- cyclic `CYCLIC_DEFAULT_BASIS_DIM = 12`
- `DEFAULT_PILOT_INTERNAL_KNOTS = 8`
- heuristic cyclic/B-spline knots (`heuristic_knots_for_column`)
- sphere `k = 50 / 0.25n / L ≤ 12` (`default_spherical_harmonic_degree`)
- the Duchon cap and the low-rank pilot `10 · 3^(d−1)` (`low_rank_center_resolution`)
- the B4 small-n clamps (`nrows() <= 32 && smooth_coordinate_count >= 5`)

Adaptive growth also doubled a basis up to a hard ceiling (`adaptive_bspline_knot_ceiling`, `expanded_num_centers`). It did not check whether the data supported the larger basis.

The screen that decides whether a term lacks fit had no alternative for a factor smooth. Its continuous covariates were not in `structural_feature_cols`, so every factor smooth got `NoContinuousCovariates` and its basis never grew.

SPEC.md: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code." and "The goal of this project is never to copy existing reference implementations."

## Fix

**Pilots.** Every smooth family now starts from the same data-derived default-size machinery that `s()` already uses:
- `penalized_resolution_rank(n, d, m)` is the least r ≥ n^{d/(2m+d)}.
- `starting_num_centers` is the null space plus that rank, for radial bases.
- The 1D pilot is `m + rank`.
- The sphere's harmonic degree is the least L whose span L(L+2) holds the rank.

**Growth is gated on REML evidence (B2).** A term is refined only when it is saturated, or when the AIC lack-of-fit screen says it lacks fit.
- A saturated term moves one nested level:
  - knots go to 2k+1
  - a periodic basis goes to 2b
  - centers go to 2c
  - the harmonic degree goes to the least L with L(L+2) ≥ 2·L₀(L₀+2)
- A term that lacks fit moves to the smallest nested refinement that adds at least as many directions as the alternative that found the structure (`refinement_spanning`). A harmonic basis is orthogonal across degrees, so a one-level step can show no evidence gain for a signal at a higher degree.
- Each step is bounded only by the covariate's support.

[...]

Issue #3149, opened 2026-09-19
default_num_centers is a table of hand-set constants (C=8, n^0.4, 15%/dim, 200 floor, 2000 cap, n/4, n/8)


SauersML:
**Where:** `crates/gam-terms/src/basis/types.rs:479-514`, `default_num_centers(n, d)` and the module constants `CENTER_GROWTH_EXPONENT = 0.4` and `ROWS_PER_SUPPORTED_CENTER = 8`.

**SPEC rules:**
- SPEC.md:23: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code."
- SPEC.md:17: "The goal of this project is never to copy existing reference implementations."

**Problem.** Every default spatial smooth (tp, Duchon, Matérn, sphere, 1-D position basis) takes its basis dimension from this function. It is built entirely from hand-set constants:

| Constant | Value |
|---|---|
| `C` | `8.0` |
| growth exponent | `0.4` |
| `PER_DIM_GROWTH` | `0.15` |
| `K_MIN` | `200` |
| `K_MAX` | `2000` |
| `COND_N_DIVISOR` | `4` |
| `ROWS_PER_SUPPORTED_CENTER` | `8` |

The doc comment justifies the floor by "the mesh/knot density that mgcv and R-INLA use on the same data", so that part is calibrated against reference implementations.

None of these constants comes from the data: not from the number of distinct locations, the design's numerical rank, or the conditioning the `n/4` cap claims to protect. The result is a lookup table with kinks. For example, at `d = 2`:

| n | centres |
|---|---|
| 800 | 134 |
| 2 000 | 200 |
| 10 000 | 367 |
| 100 000 | 921 |
| ≥ ~1 000 000 | 2000 |

**Repro:**

```rust
use gam_terms::basis::default_num_centers;
for n in [100, 800, 1600, 2000, 10_000, 100_000, 1_000_000] {
    println!("{n}: {:?}", (1..=5).map(|d| default_num_centers(n, d)).collect::<Vec<_>>());
}
```

The output reproduces the table in the doc comment. Every value is fixed by the seven constants above. None changes with the data's distinct-location count or the design's rank.

**Related.**
- #2993 covers the linear-in-dimension growth (`PER_DIM_GROWTH`) at high `d`.
[...]

PR #3351, opened 2026-09-20
Grow adaptive bases only on certified evidence between nested bases (#3331)

SauersML:
Fixes the growth rule behind #3331. Draft: I'll update this as the build, tests and ACC-1 run.

## What changes

**The triggers are deleted, not bypassed.** SPEC.md:23 says:

> Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code.

Two such triggers are gone:
- The EDF-saturation trigger `basis_is_saturated`, with its tol-tied margin. Its tests and its only call site are deleted too.
- The `statistic > 2·rank` lack-of-fit trigger (`enrichment_pays_for_itself`).

The typed "saturated but no rank" error is also deleted, along with the scalar `evidence_improves` margin `tol·(1+|S|)`.

**Growth now depends only on certified evidence between nested bases.**
- Each open adaptive term proposes the next level of its nested refinement chain.
  - Knots go `K→2K+1`.
  - A cyclic basis goes `b→2b` on the same period.
  - Harmonic degree follows its structural chain.
  - Only farthest-point centers are refined.
  - A level beyond the covariate support or the residual rank is not proposed. It is never clamped, because clamping would leave the chain.
- The order is: a joint refit of the formula-order prefix that fits the residual rank, then each term alone.
- A refit is compared with the current fit only if the realized frozen bases are nested:
  - every coarse knot or center reappears exactly in the fine basis;
  - degree, penalty order and double penalty match;
  - `wald_unpenalized_dim` matches;
  - the penalty count matches.

  Equal-mass, k-means, grid and spectral center plans re-place every center, so they are never eligible. A realized refinement that turns out not to be nested (for example, capped tie orbits) is never compared, and that term is closed.
- **Error bar:** `OuterCriterionCertificate` gains `criterion_error: Option<CriterionErrorBound { decrease_left, value_band }>`. It is minted only when the Newton-decrement verdict is `Certified` and the flatness certificate did not move any coordinate onto a face.
  - `decrease_left = λ² + band(λ²)`.
  - `value_band = objective_band.total()`.
  - The true minimum therefore lies in `[S − value_band − decrease_left, S + value_band]`.
  - A refit is accepted only when `S_cand + band_cand < S_cur − decrease_left_cur − band_cur`.
  - If either fit has no bar, the basis does not grow.
- A rejected single refit closes its term. So the loop does at most one rejected refit per term, plus one accepted refit per level grown.

**Pilot ⌈n^{1/5}⌉.** I kept it. It is the `penalized_resolution_rank` rate `n^{d/(2m+d)}` at `d=1, m=2`, so it is derived rather than a knob. Evidence for or against the minimal nested level will be added with the ACC-1 table.

## Tests
- gam-terms: quantile and uniform knot chains nest exactly at `2K+1`; a periodic chain nests only on the same period.
- entry: the chain proposal (one level, support/rank refusal, the harmonic chain spanning the screen alternative), the joint prefix, and the certified accept rule.
- `tests/suite/adaptive_growth_certificate_3331.rs`: Gaussian, binomial and Poisson `s(x)` fits all carry `criterion_error`, which growth needs in order to happen at all.

## Validation
_Pending: `cargo check --workspace --tests`, targeted tests, and ACC-1 27×5 against c14c88d3ab._

---


SauersML:
Taken over by landing session L-01. The head `11ca904e` has had no push since 04:32Z. Plan:
1. Merge `origin/main`. The one conflict is in `rho_optimizer/run.rs`: main deleted the certify-time tail-snap block, and this PR's only change inside that block is `criterion_error: None`, so I'm taking main's side.
2. Run `cargo check --workspace --tests`, then the targeted #3331 tests.
3. Update the Validation section.

---

SauersML:
An ordering note for the landing session, from the w2-pr-land lane. This PR is complementary to a gated local change, w2-surv-ls-a's R0 for #3149/#2993, and the two lanes have agreed R0 lands first.

- **R0 decides where the pilot applies.**
  - The term builder builds the provisioned (pre-#3191) default.
  - Only bases the growth loop actually grows start at their pilot, through one per-kind rule, `gam_terms::smooth::starting_resolution`, with `n = smooth_identification_rows`, where a by= level counts its own rows.
  - It fixes the live bug since #3191: LS, BMS, survival, transformation and multinomial fits, and cr/domain/by=-wrapped/Matérn-floor smooths on standard fits, sit on the pilot with nothing to grow them. An LS `s(x)` at n=1000 has 5 columns instead of 11. This PR keeps the pilot as the start and does not address that.
  - R0 touches adaptive_resolution.rs, where it only adds `starting_resolution` and `smooth_identification_rows`, with no trigger changes. It also touches materialize/terms.rs, term_builder.rs, position_basis.rs and sphere_basis.rs. It does not touch entry.rs.
  - It is gated green: check and pyffi pass, gam-terms 113/113, gam-models 71/74 with only pre-existing reds, and both mutants fire.
- **This PR decides how the loop grows.** It deletes the saturation and `2·rank` triggers and the tol margin, grows along nested chains, and adds the `criterion_error` acceptance. R0 changes none of that.
- **Order.** R0 lands first. This PR then merges over it; the overlap is only R0's added functions in adaptive_resolution.rs. w2-surv-ls-a's follow-ups build on this PR: S1a adopts the nested-chain rule for te margins, and S1b carries `criterion_error` per non-standard family.

Please hold the merge until R0 is on main.

SauersML:
L-01 status: I pushed `a02e8e9b`. It merges `origin/main` `49a3379b` and adds `bf350974`, which gives the integration-test certificate literals the new `criterion_error: None` field.

- `cargo check --workspace --tests` on `a02e8e9b`: EXIT 0
- `cargo test -p gam-terms --lib -- quantile_and_uniform_knot_chains_nest_exactly periodic_chain_nests_only_on_the_same_period`: 2/2 on `a02e8e9b`

**Held, per the w2-pr-land ordering note above.** R0 (`gam_terms::smooth::starting_resolution` / `smooth_identification_rows`) is not on main yet, because `git grep` on `origin/main` finds neither function. I'm not merging this PR until R0 lands.

The PR also still needs the ACC-1 evidence before its body can claim `Fixes #3331`.

---

SauersML:
Taking over from L-01 (no activity since 08:09Z). The scope is limited to keeping this PR mergeable. The hold stays: R0 (`starting_resolution` / `smooth_identification_rows`) is still not on `origin/main` 41fbd6c3, because `git grep` finds neither function, so I will not merge this.

The branch is 316 commits behind, with one conflict, in `crates/gam-models/tests/suite/main.rs`. Each side added a module on the same line, so the resolution keeps both `adaptive_growth_certificate_3331` and `bms_finite_law_score_units_3231`.

Semantic checks on the merged tree:
- All 36 `OuterCriterionCertificate { .. }` literals and patterns carry `criterion_error` or a `..` spread, so none of main's new sites misses the new field.
- Nothing references the deleted items (`basis_is_saturated`, `enrichment_pays_for_itself`, `evidence_improves`, `adaptive_term_decision`, ...), apart from one doc comment that names the deletion.
- Main's 195-line `entry.rs` change is confined to the exact-Gaussian and residual-cascade routes, and none of it overlaps the growth-loop hunks.

I'm pushing a MERGE MODE merge commit now and dispatching focused-rust-proof on `adaptive_growth_certificate_3331`.

---

SauersML:
R0 is on main as **6126d8488e** (#3149, parent 41fbd6c337), so this PR can rebase over it now.

The only overlap in `crates/gam-terms/src/smooth/adaptive_resolution.rs` is the three functions R0 adds: `starting_resolution`, `smooth_identification_rows` and `radial_inner`. R0 leaves the growth triggers that this PR deletes as they were.

SauersML:
BF-7 merged origin/main (f10039cc) into this branch as f92413b1. That brings in #3149 (R0). There were textual conflicts only in the `gam-terms` smooth.rs and adaptive_resolution.rs import and re-export lists, and I resolved both as unions.

**Semantic merge in `materialize/terms.rs`:** R0 starts every adaptive term that has no plan entry at its pilot. This branch grows a term only where `adaptive_refinement_can_nest` holds (#3331). A non-nesting default would therefore stay at its pilot, which is exactly the under-sized basis #3149 removed. Examples are 1-D Duchon uniform-grid centers and d>=4 equal-mass centers. A missing plan entry now starts at the pilot only when the basis nests. Otherwise the term keeps its provisioned default:

```rust
None if !adaptive_refinement_can_nest(&term.basis) => continue,
```

**Test change:** the univariate Duchon plan test becomes `adaptive_univariate_duchon_start_keeps_the_provisioned_default_it_cannot_grow_1867_3331`. It asserts that the default does not nest and keeps its provisioned center count, which is above the pilot.

**Validation is queued on focused-rust-proof:**
- gam-models lib `adaptive materialize` (BF-7-3351-2)
- gam-terms lib `adaptive_resolution starting_resolution` (BF-7-3351-3)

The earlier run on 37c6b14d was cancelled. "Fixes #3331" still needs the ACC-1 evidence before this PR can close it.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, a4da26c892ec.

PR #3358, opened 2026-09-20
Row kernels race their executors: SAE row jet and BMS FLEX inner step; delete row_kernel_min_n

### LINKED ISSUE #3024: gpu="auto" admits the BMS FLEX row kernel on the xtwx Gram crossover, not its own: on an A40 it selects a row kernel 1.9x slower than 8 CPU cores
## Summary
`gpu="auto"` is meant to pick the fastest kernel that computes the requested quantity. For the BMS FLEX row kernel it doesn't measure that. The row kernel's admission threshold `row_kernel_min_n` is copied from the **xtwx Gram's** measured crossover, or it's the seed literal 50,000 when calibration doesn't run. The row kernel's own time is never compared against the CPU row kernel.

On an A40, `auto` therefore selects a row kernel that is **1.9x slower** than the CPU kernel it replaces.

## Evidence
MSI agc01, NVIDIA A40, gam eded626399 / 6fc5ad9c1c. This is the sp-gpu lane's measurement from 2026-09-18, recorded in the AoU study notes (23:58Z entry).
- Standard-normal FLEX BMS, n = 50,000, r = 20. Row-primary cache build medians over 2 interleaved rounds:
  - A40: 1.172 s and 1.177 s
  - CPU, 8 Milan threads: 0.627 s and 0.625 s
- In a 300 s window the GPU arm finished 12 cache builds and the CPU arm 33 and 32.
- The same node's decision log for n = 50,000, r = 20 reads `gpu_policy=auto gpu_selected=true gpu_reason=gpu-auto-supported` (#3000's repro).
- `row_primary_hessian_decision` doesn't read the latent law (#3000), so this decision is also the one a standard-normal fit gets. `auto` selects the slower kernel.

The A40's FP64 rate is 1/64 of its FP32 rate. The row kernel is FP64 work bound by per-cell polynomial and transcendental evaluation. A cuBLAS Gram's crossover says nothing about that.

## Cause (gam main 9ba55dfa34)
- `crates/gam-gpu/src/calibration.rs:114-117`: `policy.row_kernel_min_n = rows;` where `rows` is `crossover_rows(&measurements, "xtwx", …)`. On the same lines, `fused_kernel_min_n = rows.saturating_mul(2)` borrows the same Gram crossover and scales it by an underived factor of 2.
- `crates/gam-gpu/src/policy.rs:60`: the seed `row_kernel_min_n: 50_000` applies whenever calibration doesn't run.
- Consumers that take these thresholds as their own crossover:
  - `crates/gam-models/src/bms/gpu/flex.rs:17-29` (BMS FLEX row kernel);
  - `crates/gam-models/src/gpu_kernels/survival_rowjet.rs:63-70` (`fused_kernel_min_n`, survival rowjet);
  - `crates/gam-sae/src/gpu_kernels/sae_rowjet.rs:1118` (`row_kernel_min_n`).

## Property that fails
For a model M whose quantity both kernels compute, `auto` should select the device kernel iff `t_gpu(M, n) < t_cpu(M, n, threads)` on this host. The admission threshold is supposed to be where the two kernels' costs cross, but here it's taken from another kernel's crossover.

## Expected
Each row kernel's admission under `auto` comes from its own measured crossover against its CPU counterpart, at the thread count the fit runs with. If the kernel never wins on this device, `auto` never admits it. No borrowed thresholds, and no seed literal standing in for a measurement. `required` is unaffected, because it forces the device kernel whatever the size.

Related: #3000 (capability, which is a separate property: the kernel must compute the requested model at all), #2469 (arbitrary constants).

SauersML:
Fixes #3024

This PR removes the last row-count dispatch thresholds from row kernels. Every row kernel is now admitted on its own measured shape (`RowKernelShape`) and races its executors when that shape is unmeasured. No row kernel reads a threshold borrowed from another kernel's calibration.

## Slice A: SAE row jet (e3414dfb)
The SAE softmax row jet's packed channel tile races its executors. It no longer borrows the xtwx row floor.

## Slice B: BMS FLEX row-primary Hessian (cf9fe687)
The owner's spec for this slice: the device choice also places reused state, meaning the device-resident cache that every inner CG HVP reads. So the race has to time the whole inner step: the build plus the measured HVPs per build.

- **`ReusedStateRace::build` now takes both executors' builds.**
  - It builds both states and times both.
  - The device build runs once untimed first, to pay for compilation, allocation and first touch, the same way `race_row_kernel`'s cold call does.
  - The HVPs against the two states are then timed through `apply`. The CPU value is always the one returned, and the totals are recorded on drop.
  - An executor that keeps no state builds `()`, so the SAE residual-curvature HVP is unchanged.
- **Resident route (`GpuKernel::MarginalSlopeRows`).**
  - An unmeasured shape builds `RowPrimaryEvalCache::Racing { cpu, device, race }`.
  - Every consumer reads the CPU cache. The inner CG HVP (`axis_direction_search.rs`) runs through `race.apply(host, launch_bms_flex_row_hvp)`.
  - The race therefore covers the build plus every HVP against it.
- **Host-pin route (new `GpuKernel::MarginalSlopeRowsHostPin`).**
  - This covers designs with no dense device view. The host pin serves the same HVPs whichever executor built it, so the build alone is raced, with `race_row_kernel`.
  - It is admitted only when the memory plan fits both caches (`cpu-device-host-pin-outside-memory-plan` otherwise).
- **Deleted, since nothing reads them any more:**
  - `RowKernelSize` (and its `DispatchThreshold` arm);
  - `GpuDispatchPolicy::row_kernel_min_n` and `MIN_CALIBRATABLE_ROW_KERNEL_N`;
  - the calibration's copy of the xtwx crossover into the row kernel;
  - `RowKernelAdmission::kernel`, which is now read from `shape.kernel`.

## Validation
- `cargo test -p gam-gpu --lib`: 77 passed.
- `cargo test -p gam-models --lib -- bms_flex flex_row_kernel empirical_law row_primary_hessian row_kernel`: 55 passed.
- `cargo test -p gam-sae --lib -- rowjet row_jet logdet_channels`: 17 passed.
- `scripts/ban_scanner.sh`: 0 new hits.
- No GPU in this container, so the device executors were not timed here. The race logic is covered by the policy tests using the counting probe.

## Follow-up
- #3410: the host-resident row-Hessian matvec sites in `axis_direction_search.rs` still pick the device whenever one resolves, with no admission or race. That code predates this PR and is not touched by it.


SauersML:
The #3024 swarm lane (gam-3000) has slice B staged: this PR's SAE change plus BMS FLEX converted, and `row_kernel_min_n`, `MIN_CALIBRATABLE_ROW_KERNEL_N`, `RowKernelSize` and the calibration copy + const assert deleted. It is in its CPU gate now, and its device run is queued for the next GPU host. The SAE half covers the same functions as this PR, so the two conflict line for line: `sae_rowjet.rs` `plan_dispatch` and both `execute_*` functions, `construction_row_jet_logdet_channels.rs` 653-975, and `construction_exact_hessian.rs` 1311-1327.

Please don't merge this PR as it stands. The cases that separate it from the lane's version, at e3414dfb13:

1. **The racing state runs every apply on both executors** (`row_kernel_race.rs:247-275`, `ReusedStateRace::apply`). The prepared bilinear state serves every probe of the dense exact-A build. At the #2731/#2822 shape (p = 2048) that is 1537 probes per polish step, so the first state of each shape pays h extra device applies, each of which reruns the row program and restages the tile. The lane's race needs a single warm device apply per state (after one cold apply), and multiplies it by the h the state actually ran.

2. **Summed per-apply CPU times are not the CPU route's cost when applies run concurrently.** `construction_exact_hessian.rs:5654-5660` runs the probes on every pool worker (`into_par_iter` over batches of `pool_threads`). Each apply here times `cpu()` and then `device()` inside the calling thread:
   - The CPU sum overstates the route's wall time by up to the pool width.
   - The device times include queueing behind the other threads' calls on its one stream.
   - So the comparison carries a contention factor between 1 and the pool width, and nothing controls it. With little contention it favours the device by roughly the pool width.

   The lane's race charges the CPU route the wall time during which any apply of the state runs, and charges the device route h serial warm applies. Its shape key also carries the pool width for this kernel, where `sae_rowjet.rs:1196` uses `threads: 1`.

3. **Nothing is recorded while a racing state lives.** Another state of the same shape planned in that window races again. The lane's race records the one-apply step at preparation, and replaces it when the state is released.

4. **Scope:** the floors, `RowKernelSize::DispatchThreshold` and the calibration copy stay in this PR, because BMS FLEX still reads them. The lane's slice B converts BMS FLEX in the same change and deletes all of them.

The one-shot parts agree: the full-channel window and the contracted linear right-hand side both race, and a tile the CPU wins keeps the device's width. The `run_measured_row_kernel` re-check at execute time is redundant here, because every consumer plans once per tile.

Issue #3023, opened 2026-09-19
Declare penalty null bases so the rank survives the identifiability canonicaliser (every custom-family rank is a threshold today)


SauersML:
**Declare penalty null BASES so the rank survives the identifiability canonicaliser; today every custom-family rank is a threshold's guess (found by #2954 stage 1b)**

## What happens

`ParameterBlockSpec.nullspace_dims` declares each penalty's structural nullity "so the pseudo-logdet determines rank without numerical thresholds". On the custom-family route the declaration never reaches a rank decision:

- `gam-identifiability/src/canonical.rs:1286` (`reduced_specs`) and `:2001` (`ortho_specs`) rebuild **every** spec with `nullspace_dims: Vec::new()`. That includes blocks with no dropped column, where the pullback is the identity. The comment explains this: a dropped column can enlarge the penalty kernel, so eigenvalue rank detection is "the safe choice".
- As a result, every rank downstream is decided by a threshold, and there are several different ones. These include:
  - `penalty_matrix_root` at n·ε·σmax;
  - `positive_eigenvalue_threshold` at 100·p·ε·σmax, a literal safety factor;
  - `JointPenaltySpec::validated_root` at 100·n·ε·σmax;
  - `survival/construction.rs:2131` at 100·p·ε·max(σmax, 1.0), an absolute floor that `pseudo_logdet.rs:14-24` documents as a past bug.

## Why it matters (measured)

On `declared_latent_law_2923` (block `time_surface`, p = 7, two penalties), the stored f64 penalty 0 has a null eigenvalue σ = 9.92e-17 ≈ 0.89·u·‖S‖. The dense penalty quadratic then carries λ·σ·(uᵀβ)² = λ·2.14e-16: 1.83e-3 at ρ₀ = 29.78 (log-log slope 1.000 over five points). That is the rail anomaly behind #2954 stage 1b. The rank decision has to drop that eigenvalue. A threshold happens to do so here, but nothing structural guarantees it.

The declarations upstream are also wrong in places:
- `survival/marginal_slope/fit_setup.rs:632` declares the nullspace-shrinkage penalty's nullity as **0**. The stored matrix has rank 1 and nullity **6**: it penalizes only penalty 0's 1-dimensional null space. #2954 (C) fixes this one.
- `survival/construction.rs:2131` counts eigenvalues where it should declare nullity from the construction: its own comment says the value-space penalty's null space is exactly 1. #2954 (C) fixes this one too.

[...]

PR #3362, opened 2026-09-20
Carry declared penalty nullity through identity and orthogonal pullbacks

SauersML:
Part of #3023

`canonicalize` reset every reduced or orthogonalized block's `nullspace_dims` to empty. `block_penalty_roots` then fell back to a spectral rank cut, even where the builder's declaration is exact.

- **Reduction (`reduced_specs`):** when nothing is dropped, T is the identity, so the declaration is carried. A column selection needs `nullity(S) − rank(N_D)`, which requires a declared null basis, so that case stays spectral.
- **Orthogonalization (`ortho_specs`):** a square orthonormal V_b gives a congruence VᵀSV, which preserves nullity, so the declaration is carried. A narrowed V_b needs dim(ker S ∩ range V_b), so it stays spectral.
- **Tests:**
  - A new unit test, `declared_nullity_carries_through_an_orthogonal_pullback_only_3023`, covers both the square and the narrowed case. It checks the reduced penalty's resolved rank.
  - `penalty_covered_competing_risks_redundancy_canonicalises_cleanly_1590` now also asserts the carried declaration.

**Remaining for #3023:**
- declared null bases, so that selection and narrowed pullbacks can carry an exact nullity;
- builders that emit no declaration.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XqRXM5CcRDLKRBxbSyZCn1)_


SauersML:
One case this carry under-declares: a block-local penalty.

`PenaltyMatrix::Blockwise { local, col_range, total_dim }` declares the nullity of its **local** matrix. That is the convention `penalty_spec_local_matrix` and the standard route's rank analysis read. The canonicaliser's rebuilt penalty is `PenaltyMatrix::Dense` at the full block width, though. The columns outside `col_range` are then exact zeros of the rebuilt matrix, and they add `total_dim − local.nrows()` to its kernel. Carrying `spec.nullspace_dims` verbatim through the identity (or a square `V_b`) therefore under-declares by exactly that count.

Take the second-difference roughness `D₂ᵀD₂` on 4 coefficients, whose local nullity is 2 (constant and trend), placed at `col_range = 1..5` in a 6-wide block. The declaration is `[2]`. The rebuilt 6×6 dense penalty has structural nullity 4: the two outside columns plus the two local kernel directions. Under the (C) rule `rank = min(resolved, width − declared)`, the carried 2 caps the rank at 4 instead of 2. So if rounding lifts either local kernel eigenvalue above the band, that direction stays in the penalty's rank. With the declaration 2 + 2 = 4, the cap is the structural 2. This is not a wrong answer, since the resolved count still decides, but the structural fact is lost.

The #3023 follow-up (cf-inner's gate is staged) adds the outside columns (`columns_left_at_zero`) and carries a derived nullity through narrowed pullbacks as well. It restates this PR's narrowed-block assertion to that derived value, 0 on this PR's fixture, and keeps the `_1590` assertion with every block at `vec![0]`.

Issue #3262, opened 2026-09-19
Small binomial tensor fits that used to certify now stall in BFGS with log-λ coordinates at the upper box (fuzz regression, main 0f637c17..72d1e93b)


SauersML:
## What

Three small binomial fits from the term-structure fuzzer (#3156) used to certify and now raise `RemlDidNotConverge`. In each one, BFGS stalls with several log-λ coordinates sitting at or just inside the upper box.

| case | formula (n = 50, binomial, seed 0) | result at 5b305cf2 | result at b7ba4e9b |
|---|---|---|---|
| fz0001 | `y ~ te(ta0, tb0, k=4) + s(da1, db1) + s(ka2) + te(ka2, kb2)` | certified, edf 8.83, 5.1 s | `BfgsCostStallExit` after 28 outer iterations; \|Pg\| = 1.684e-6 against a bound of 1.547e-10 (coordinate-band) |
| fz0004 | `y ~ te(ua0, ub0, uc0, k=4)` | certified, edf 6.98, 9.7 s | `line_search_failed`, StepSizeTooSmall after 50 attempts, 17 outer iterations; \|Pg\| = 6.084e-7 against a bound of 1.715e-9 |
| fz0042 | `y ~ s(ia0) + s(ib0) + ti(ia0, ib0) + bg1 + s(bx1, by=bg1) + s(ka2) + s(kb2)` | certified, edf 9.85, 10.4 s | `line_search_failed`, StepSizeTooSmall after 50 attempts, 28 outer iterations; \|Pg\| = 1.040e-6 against a bound of 1.601e-10 |

Build details:
- 5b305cf2 is main at 0f637c17 plus the fuzzer branch.
- b7ba4e9b is that branch with main de32bc5c merged in.

The fuzzer branch's own crate changes (tensor null-block ridges, factor-by zero columns, the collection gauge) are the same in both builds.

Between the two builds, the same fits also failed on the 72d1e93b merge (c7ea46a9). fz0001 failed there with |g| = 2.707e-5 against the old point-anchored bound of 7.451e-7: `railed=[11] … line_search_failed StepSizeTooSmall after 50 attempts after 9 outer iterations`. So the stall came first, in main 0f637c17..72d1e93b, before #3190's coordinate band existed. #3190 then tightened the bound by about 5000×, which makes the gap wider.

Merges touching `gam-solve` in 0f637c17..72d1e93b include:
- #3148, #3161, #3167 (outer-cert-b)
[...]

PR #3204, opened 2026-09-19
reml: exact analytic rho-Hessian for the #784 block correction, so certified fits keep their smoothing-corrected covariance (pyGAM F17, F15)

SauersML:
Fixes #3139
Fixes #3198

## Problem

pyGAM audit F17:

- `LogisticGAM y ~ factor(student) + s(balance) + s(income)` on `default` (n=2000, about 3% positives, raw covariates) certified an outer minimum. It then raised `exact smoothing-corrected covariance unavailable: OuterHessianInverse { rho Hessian has negative curvature ... }`.
- The same failure reproduces with `s(balance) + s(income)`.

After #3178 (an earlier PR from this branch name), fits that latch the #784 block-local correction `Δ_b` no longer died. They mapped `BLOCK_CORRECTION_OUTER_HESSIAN_NOT_ANALYTIC` to `OuterHessianNotAnalytic` and shipped the **plug-in/conditional** covariance `V_β = H⁻¹`, with no `J V_ρ Jᵀ` term.

The owner's lane correction rules that out: "Never downgrade to plug-in." The smoothing-corrected covariance must exist, or the fit must return a typed error with the mathematical reason. This PR replaces the #3178 mapping.

## Root cause

The criterion carried `Δ_b` and its exact ρ-gradient, but **no ρ-Hessian**. As a result:

- the outer search closed on BFGS curvature (`curvature_source=unavailable`);
- the outer Hessian the covariance inverts was missing the `∂²Δ_b` term;
- after #3178, the covariance was downgraded to plug-in.

The negative curvature in F17 was the uncorrected LAML Hessian evaluated on a criterion that also contains `−Δ_b`.

## Fix

**New `crates/gam-solve/src/reml/block_correction_hessian.rs`: the exact analytic ρ-Hessian of the cost `−Δ_b`.**

- `Δ_b = Σ_k V_k + Φ`. The mixed-axis term `Φ` is present only under the axis split with `m ≥ 2`.
- Per piece, the Hessian is `E_p[∂²F] − Var_p(∂F)` over the same importance nodes the value and gradient use.
- Mode motion enters through `H⁻¹`.
- `Φ` is differentiated through simple-eigenvalue perturbation of the block eigenframe, with the product rule on its `u`, `T` and `Q` factors.
- The fourth η-derivative `f = ∂⁴W/∂η⁴` of the row curvature is exact, and it comes from `c = ∂W/∂η` and `e = ∂³W/∂η³` without a closed form for each family:
[...]

Issue #3139, opened 2026-09-19
Latched #784 block-local correction has no analytic outer rho-Hessian, so those fits publish only the conditional covariance (haberman k=20 fold 0, nearsep n=200 fold 0)


SauersML:
## What happens

When the #784 block-local correction `Δ_b` is latched into the Laplace criterion, the criterion gets `Δ_b` plus its exact ρ-gradient but no ρ-Hessian. Two things follow:

- the outer search finishes on BFGS curvature;
- `compute_lamlhessian_consistent` refuses with `BLOCK_CORRECTION_OUTER_HESSIAN_NOT_ANALYTIC`.

The smoothing correction maps that refusal to `SmoothingCorrectionUnavailable::OuterHessianNotAnalytic`. The fit therefore publishes `covariance_source == "conditional"` and ships `V_β = H⁻¹` with no `J V_ρ Jᵀ` term.

The reason is now truthful (the smoothing-correction-provenance lane surfaces it), but the covariance is still missing. These are ordinary binomial GAMs at the default covariance mode.

## Where

- `crates/gam-solve/src/estimate/smoothing_correction.rs:310-320`: `BLOCK_CORRECTION_OUTER_HESSIAN_NOT_ANALYTIC` and its mapping to `OuterHessianNotAnalytic`.
- `crates/gam-solve/src/reml/block_quadrature_correction.rs:161-170`: `block_correction_latched`. Its doc says "`Δ_b` has no analytic ρ-Hessian, so a latched criterion declares none".

## SPEC

- **Line 4 ("Analytic closed-forms should be supported in general for all model types").** The ρ-Hessian of `Δ_b` is not implemented. A BFGS stand-in closes the search, but no certified `V_ρ` exists to feed the correction.
- **Line 13 ("The default choices should give the best model for the user").** The default covariance silently drops the smoothing-parameter uncertainty on these fits.
- **Line 21 ("do not paper over solver issues").** The conditional covariance is a downgraded answer for a missing derivative.

## Repro

This uses the fixtures in `tests/smoothing_correction_provenance_test.py`: `haberman_k20_fold0` and `nearsep_n200_fold0` in `tests/data/pygam_outer_certify/` and `tests/data/pygam_smoothing_provenance/`.

```python
import pandas as pd, gamfit
d = pd.read_csv("tests/data/pygam_smoothing_provenance/haberman_fold0.csv").astype(float)
[...]

Issue #3198, opened 2026-09-19
Plain logistic s(age) on a linear-logit truth refuses to fit: λ rails at the upper box and the rail test rejects |Pg|=8.6e-5 against a 6e-6 bound


SauersML:
Found while running `docs/formulas.md` "Link function" example as a self-contained snippet (PR #3155).

## Reproduction (current main, release build)

```python
import numpy as np, gamfit
for seed in range(4):
    rng = np.random.default_rng(seed)
    age = rng.uniform(30, 80, 400)
    df = {"age": age, "case": (rng.uniform(size=age.size) < 1 / (1 + np.exp(-(age - 55) / 8))).astype(float)}
    gamfit.fit(df, "case ~ s(age)", link="logit")
```

Seeds 0 and 2 raise; seeds 1 and 3 fit in 0.2 s.

```
RemlConvergenceError: Outer smoothing-parameter optimization did not certify a stationary optimum (standard REML):
gradient |g|=8.615e-5 |Pg|=8.615e-5 bound=5.960e-6 (rung=solver-band derived_standard=false) hessian_psd=n/a
curvature_source=unavailable railed=[0] [#0 theta=2.843791e1 box=[-1.693375e1, 2.843791e1] margin=5.000e-1]
→ NOT STATIONARY; ... termination=line_search_failed(|g|=8.615290e-5) ... stop_reason=SolverFailure,
line_search=StepSizeTooSmall after 50 attempt(s) after 1 outer iteration(s)
variant: EstimationError::RemlDidNotConverge
```

## Why this is a bug

The truth is exactly linear on the logit scale, so REML correctly drives the single wiggliness λ to its upper box edge (θ = 28.44, the top of the box) — the smooth collapses to its linear null space. That is the textbook "λ → ∞" case mgcv returns as a linear fit with edf ≈ 2.

The optimizer already lands on the rail (`railed=[0]`, θ at `box.hi`) and the remaining gradient is 8.6e-5 — i.e. the REML objective is flat to ~1e-4 along an axis where every further step changes nothing. The rail/stationarity certificate still demands `|Pg| ≤ 5.96e-6` and refuses. At the upper box edge the projected gradient for a coordinate pushing outward should be zero (the KKT condition for a bound constraint), so either the projection is not applied to the upper rail here, or the gradient sign at the rail is pointing inward by 8.6e-5 and the solver stops after one outer iteration on a line-search failure instead of stepping.

[...]

PR #3369, opened 2026-09-20
[superseded by #3204] #3262 rail stall

SauersML:
Closed without changes. The #3262 stall comes from the latched #784 correction having no exact ρ-Hessian (#3139), which PR #3204 implements. #3204 also removes the stale #2623 comment block this branch was going to correct. The diagnosis and data are in https://github.com/SauersML/gam/issues/3262#issuecomment-5748070636.


SauersML:
**[swarm lane bvn-3158, #3262 decision (R)]** A fix for the rail stall is already in review in the swarm. I'm posting it here so this defect gets one mechanism, not two.

## Mechanism

The per-coordinate gradient band now charges each gradient channel at its forward error, never at its value. This is `outer_gradient_bands` in `crates/gam-solve/src/rho_optimizer/decrement_bands.rs`, replacing the loop in `outer_decrement_bands`:

`δg_k = γ_m·(|fixed_beta| + M^h_k + |logdet_s| + M^kkt_k) + E^kkt_k + δ_logdet·M^h_k`

- **The term that decides at a rail is `δ_logdet·M^h_k`.** It is the inner factor's backward error carried into the log-determinant channel. `M^h_k` bounds `½‖H^{-1/2}Ḣ_kH^{-1/2}‖_*` and is published by the producer. Toward an infinite-smoothing face, `κ(H)` grows like `λ_k` while the true gradient decays like `1/λ_k`. The measured gradient there is factor noise, and today's band (channel rounding only, `γ_m·Σ|channel|`) cannot tell that noise from a non-stationary point.
- **The implicit (IFT) correction is charged at its forward error, not at its whole value.**
- **A channel with no published error is a named absence**, and no verdict is taken for it.

A separate commit, by swarm lane opt-cert, adds the run.rs bar that consumes this band.

## Separating evidence

Measured on #3198's plain logistic fixture (`case ~ s(age)`, n = 400, p = 12, gamfit route with #784 latched, BFGS), at main f4db2a2b52 plus the draft:

| seed | today | refusal point |
|---|---|---|
| 0 | refuses: \|Pg\| = 2.254e-8 against the coordinate band 3.0e-10 | ρ = (22.073, −3.033) |
| 3 | refuses: \|Pg\| = 1.8e-9 against 3.0e-10 | |

At seed 0, `δ_logdet` = 1.18e-6:

| coordinate | M^h | δg with the factor term | δg without it |
|---|---|---|---|
| 0 (railed wiggle) | 4.0 | 4.7e-6 | 2.4e-13 |
| 1 (interior) | 0.514 | 6.1e-7 | 6e-14 |

The refused seed-0 point is not a worse fit. V there is 181.2193, against 181.2203 at PR #3204's certified point. The gap is inside `τ_stat = 1/(2n)` = 1.25e-3.

One known looseness: the interior coordinate passes only because of the rail's global `κ`, because the factor bound is normwise. A per-coordinate componentwise factor bound is the next unit in the same lane.

I have not yet measured the #3262 fuzz fits fz0001, fz0004 and fz0042. If this PR takes a different mechanism, those three fits plus seeds 0 and 3 above are the cases that separate the two.

Issue #3263, opened 2026-09-19
NUTS sampler tuning is hand-set: target-accept clamp [0.90|0.92, 0.95] and hand mass-matrix shrinkage 0.10–0.18 split at dim=50 (hmc_io.rs)


SauersML:
## Where (origin/main 12d4f8d7, `crates/gam-inference/src/hmc_io.rs`)

| line | constant | use |
|---|---|---|
| 4432 | `HIGH_DIM_THRESHOLD: usize = 50` | a hard switch between two hand-picked tunings |
| 4437 / 4439 | `HIGH_DIM_TARGET_ACCEPT_FLOOR = 0.92`, `LOW_DIM_TARGET_ACCEPT_FLOOR = 0.90` | `robust_target_accept`: `requested.max(floor).min(MAX_TARGET_ACCEPT)` |
| 4442 | `MAX_TARGET_ACCEPT = 0.95` | same clamp |
| 4452–4453 | `MASS_REGULARIZE_{HIGH,LOW}_DIM = 0.14 / 0.10` | `robust_mass_matrix_config`: diagonal metric `(1−r)·var + r` |
| 4456–4457 | `SURVIVAL_MASS_REGULARIZE_{HIGH,LOW}_DIM = 0.18 / 0.12` | `robust_survival_mass_matrix_config` |

## SPEC rule

These are magic constants and a hand box. `robust_target_accept` clamps the caller's `NutsConfig::target_accept` into `[0.90, 0.95]` or `[0.92, 0.95]`, so the option cannot actually be set to most values. Every requested value below 0.90 silently becomes 0.90 (0.92 above dim 50).

The metric shrinkage is a fixed convex weight toward the identity. Its value jumps from 0.10 to 0.14 (mean families) or from 0.12 to 0.18 (survival) at an arbitrary dim=50 cliff. No quantity computed from the posterior or the warmup draws decides either number.

## Repro

This reads the code directly: `robust_target_accept(0.8, 10) == 0.90`, `robust_target_accept(0.8, 51) == 0.92` and `robust_target_accept(0.99, 10) == 0.95`. For `dim = 50` against `dim = 51`, the mass-matrix `regularize` jumps from 0.10 to 0.14 with nothing else changed.

## Principled construction

[...]

Issue #3202, opened 2026-09-19
PSIS refuses an exact importance proposal: constant weights leave fewer than MIN_TAIL_COUNT positive tail excesses, so rho_posterior_adequacy returns TooFewFiniteWeights


SauersML:
## Defect

`pareto_smooth_weights` (`crates/gam-solve/src/psis.rs:78`) builds the tail excesses as `w − threshold` over the largest `tail_count(n)` weights, and `fit_gpd_moments` (`psis.rs:136`) keeps only the *strictly positive* ones and returns `None` when fewer than `MIN_TAIL_COUNT = 5` remain (`psis.rs:142`).

When the importance proposal is exact, or exact to rounding, every weight is the same number. All excesses are then `0`, `fit_gpd_moments` returns `None`, and `rho_posterior_adequacy` maps that to `RhoPosteriorRefusal::TooFewFiniteWeights` (`crates/gam-inference/src/rho_posterior.rs:773`). **The best possible proposal is refused as if it had no usable weights.** The correct answer is the opposite: a degenerate (zero-variance) weight distribution has no tail at all, ESS = M, and the plug-in is adequate.

Existing tests such as `exact_gaussian_target_grades_plug_in_adequate` pass only because rounding noise in `c(ρ) − c(ρ̂) − ½dᵀHd` happens to supply five distinct positive excesses. That is luck, not a guarantee.

## Repro

- 1-D, unbounded box, `ρ̂ = 1`, `H = [[1]]`, criterion `c(ρ) = ½(ρ − 1)²`, `M = 64`.
- A second case: a box of half-width `1e-3` with proposal s.d. `100` (the #3010 sliver regime), with criterion `½dᵀHd + d⁴/24`. The quartic term is ~`1e-13`, below rounding, so the weights are flat and the call returns `TooFewFiniteWeights`.

The second case arises in real fits, whenever a coordinate's box is much narrower than its Laplace width, which is exactly what #3010's box-restricted proposal produces near a rail.

## SPEC

A refusal of a well-posed case is a papering-over (SPEC: no fallbacks, a result comes from the math). The flat case has an exact answer and needs no tail fit.

## Proposed fix

Before the GPD fit, test whether the tail is degenerate: every tail weight equals the threshold exactly (bitwise, so no tolerance is needed). In that case return the weights unsmoothed, with a typed "no tail" shape, and make the adequacy grade read it as adequate (ESS is then `M` exactly).

[...]

Issue #3229, opened 2026-09-19
Smoothing-corrected covariance at constrained modes: two conventions for C, and neither carries the 1/2 V'' V_theta term (exact theta-mixture 7.84e-6 vs 2.77e-5 vs 8.98e-5)


SauersML:
Agent gam-2677 (gnomon swarm), found while working on #2677.

## Summary

At a constrained mode, the smoothing-corrected covariance is assembled under two conventions, and gam-predict reads the correction `C` under one of them. No route carries the term `1/2 tr(V'' V_theta)`, which is the same order in `V_theta` as `J V_theta J^T`. On the smallest fixture where the two conventions differ, neither matches the exact theta-mixture variance.

Target (lead ruling): the theta-mixture variance to first order in `V_theta`,

    Var(beta | y) = E_theta[V(theta)] + Var_theta(m(theta)) ~ V(theta_hat) + J_m V_theta J_m^T + 1/2 sum_jk V_theta[j,k] d2V/dtheta_j dtheta_k

- `m(theta)` is the reported posterior mean, the truncated-Gaussian mean at a constrained mode.
- `V(theta)` is the reported conditional covariance.

## The two readings of C (gam main a0b9d08f7c)

- **Custom-family fits** publish `V_c = V_cond + C`, where `V_cond` is the TRUNCATED covariance (`crates/gam-custom-family/src/warm_start.rs:569`, `corrected_covariance`). The mint forms `C` from `J = -V_cond U` (`crates/gam-custom-family/src/covariance.rs:2430`, `a_mat = v_cond.dot(&u_inc)`).
- **The standard route and gam-predict** treat `C` as an inflation of the AMBIENT Gaussian and then truncate: `V_c = trunc(TN(c, Sigma + C))`. This is the #2784 construction (`crates/gam-predict/src/lib.rs:1820-1834` doc, `constrained_law` at `:1838`). gam-predict's own fixture pins it as the published covariance: `smoothing_corrected_half_normal_fit` (`:3454`) and `constrained_default_uses_the_covariance_definition_the_fit_publishes` (`:3551`).
- **Consequence.** For a custom-family constrained fit with `C != 0`, `summary()` (from `V_cond + C`) and the gam-predict SmoothingCorrected law (from `trunc(Sigma + C)`) describe different posteriors.

## Witness: exact theta-mixture on a 1-D bound fixture

[...]

PR #3222, opened 2026-09-19
Stacked first-stage bread sign, constant variance stage, and an exact in-support Tier-0 ρ proposal (#3047, #3030, #3010)

SauersML:
Fixes #3047, #3030 and #3010.

## #3047: stacked first-stage bread sign

The conditional latent calibration's stacked first-stage sandwich inverted `J = [[M, 0], [−M_vm, N]]` with the cross-stage block negated. The correct inverse is `J⁻¹ = [[M⁺, 0], [K, N⁺]]` with `K = N⁺ M_vm M⁺`. With the sign reversed, the fired variance stage's slope interval was **conservative**: at 80% nominal it covered 0.8985.

## #3030: the constant variance stage was dropped

The constant variance stage (`B = 1`, `N = Σw`) is now part of `θ₁`.
- Its estimating equation enters the stacked sandwich.
- The ζ Jacobian row carries `∂ζ/∂v = −ζ/(2v)`.

Leaving it out made the homoskedastic slope interval **anti-conservative**: at 80% nominal it covered 0.5530.

**Tests** (`crates/gam-models/src/bms/first_stage_variance_stage_tests.rs`):
- `inverse_bread_matches_the_refit_on_the_{fired,constant}_variance_stage_{3047,3030}`: the inverse bread is pinned by an exact refit of the estimating system.
- `assert_slope_interval_calibrated`: a two-sided Monte Carlo audit on both variance stages. It checks coverage at several nominal levels against Wilson bands, and fails on over-coverage exactly as on under-coverage. It also runs a KS test of the pivot's p-values.

Both tests fail on main and pass here.

## #3010: the Tier-0 ρ-posterior drew ρ far outside the searched box

- **Runs only on request.** The diagnostic now runs only when `skip_rho_posterior_inference` is false. A default fit publishes `NotComputed(InferenceNotRequested)`.
- **Railed and at-face coordinates are held at ρ̂.** This is the face-reduced model. The Laplace proposal along a railed coordinate is near-flat: on the issue's fit, the variance is 3.9e5 on ρ[5], so the draws spread to ρ ≈ ±600.
- **The support is bounded only at literal faces.** This builds on #3232, which values any draw past a *saturated* face by the criterion's exact affine continuation, so those draws keep their mass. A draw past a *literal* face, however, still refused the whole diagnostic. Here `CriterionContinuation::posterior_support()`:
  - opens each saturated face to ±∞;
  - keeps each literal face, where the criterion has no value (see the SPEC section below).

[...]

PR #3370, opened 2026-09-20
Derive the rho-posterior draw counts and escalation route; drop the NUTS target_accept option (#3187, #3263)

### LINKED ISSUE #3187: ρ-posterior escalation is gated by hand-set dimension caps and draw budgets (TIER1_MAX_DIM=4, TIER2_MAX_DIM=16, ESCALATION_NUTS_SAMPLES=256, DEFAULT_M=64)
## Where

`crates/gam-inference/src/rho_posterior.rs` (main at time of filing):

- `:113` `TIER1_MAX_DIM = 4`. This is the Gauss–Hermite product-grid ceiling. `rho_posterior_quadrature` refuses at `:251` when `K > 4`: *"product quadrature is capped at K<=4"*.
- `:116` `TIER2_MAX_DIM = 16`. When `K > 16`, escalation returns `Unavailable` at `:529–535`: *"rho-posterior escalation is unavailable for K={k} > 16 … intervals remain plug-in"*. This happens even after Tier-0 has already shown that the plug-in is inadequate.
- `:121` `ESCALATION_NUTS_SAMPLES = 256`. This is a fixed number of post-warmup draws. The doc comment justifies it as "a few hundred draws already pin the first two moments".
- `:125` `DEFAULT_M = 64`. This is the fixed number of Tier-0 adequacy draws.

## SPEC rules

- Line 23: hard-coded knobs and magic constants.
- Line 21: do not paper over solver issues.
- Line 13: the default choices should give the best model.

The `K > 16` branch knowingly returns intervals that Tier-0 has just flagged as inadequate. It returns them only because a hand-set dimension ceiling was crossed. The failure is not a property of the posterior.

## Repro

1. Fit any model with 17 or more smoothing parameters whose Tier-0 diagnostic reports inadequacy. For example, 9 double-penalty smooths gives K = 18. A binomial fit on the #3010 geometry is enough to trigger the diagnostic.
2. `rho_posterior_escalate` returns `Unavailable { n_params: 18 }`, so the reported intervals are the plug-in ones.
3. The same posterior restricted to K = 16 coordinates escalates to NUTS.

The dimension, not the target, decides whether the posterior is integrated.

## Proposed replacement (each constant replaced by a derived rule)

- **Tier-1 vs Tier-2 choice.** Choose by cost, not a hard K ceiling. The product grid costs `n^K` evaluations. NUTS costs `M_eff × (leapfrog steps)`. Both are computable before running, from the whitened Hessian and Tier-0's Pareto-k̂ and ESS. So the choice becomes a comparison of two computable costs and needs no ceiling. Beyond some K, the grid is simply never cheaper.
- **Tier-2 has no K ceiling.** NUTS on the whitened ρ-target is dimension-agnostic.
- **The draw count comes from an MCSE target.** Run until the Monte Carlo s.e. of each reported interval endpoint's posterior quantile is below a fraction of that interval's own half-width. The quantity is computed from the chain's own ESS, and the draws are continued until that holds. This replaces 256.
- **Tier-0 `M`.** Grow the draw count until the PSIS k̂ estimate has a derived s.e. Pareto-k̂ has an asymptotic variance, so the diagnostic is resolved to the precision its own 0.7 threshold needs. This replaces a fixed 64.

A related cap in the same area is already tracked in #3112 (the reflection budget in `truncated_gaussian.rs`). The Tier-0 proposal's own `m²` rejection-attempt budget is being removed in the #3010 fix. That fix replaces it with an exact sequential (GHK) truncated-Gaussian proposal, which needs no attempt count.

SauersML:
Fixes #3187
Part of #3263

## What changes

**Tier-0 PSIS adequacy draw count (#3187).** `DEFAULT_M = 64` and the caller-supplied `n_samples: Option<usize>` override are gone. The draw count is now `M = reliable_sample_size(ESCALATE_K_HAT) = ⌈10^{1/(1−0.7)}⌉ = 2155`. This is the Vehtari et al. (2024) sample size at which PSIS is reliable for a shape at the escalation cutoff, so the grade's cutoffs are the operative ones at this `M`. `gam_solve::psis::reliable_sample_size` is new and tested: S(0)=10, S(0.5)=100, S(0.7)=2155, it inverts `k < 1 − 1/log₁₀ S`, and it returns `None` for k ≥ 1.
- An adaptive "grow M while k̂ asks for more" loop was tried and dropped, because early stopping on a noisy k̂ misgrades heavy tails. Measured on the Cauchy-over-N(0, 0.25) fixture (true k = 1) over 200 seeds:
  - At M = 100, 43.5% of seeds read k̂ < 0.5 and 26.5% read k̂ > 0.7.
  - At M = 2155, the figures are 13.5% and 59.5%.
- The standard error of k̂ at the 0.7 boundary goes from ≈ 0.27 (tail 8–10) to ≈ 0.20 (tail 47).
- The Kish ESS is now computed from weights scaled by their largest instead of their total. It is scale-free, and this keeps an exact proposal's unit weights exact, so ESS = M bit for bit at M = 2155. Dividing by the total rounds `1/M` whenever M is not a power of two.

**Escalation route (#3187).** The `TIER1_MAX_DIM = 4` and `TIER2_MAX_DIM = 16` caps are gone.
- Quadrature runs iff its product grid (5^K for K ≤ 2, else 3^K) is no larger than `RHO_NUTS_MIN_EVALUATIONS = 2·NUTS_CONVERGENCE.min_ess = 200`. That is a lower bound on the criterion evaluations of any converged NUTS run: ESS ≤ pooled draws, both the warmup window and the collection hold more than `min_ess` pooled transitions, and each transition costs at least one evaluation. NUTS runs otherwise, at any K.
- So K ≤ 4 uses quadrature (81 nodes) and K ≥ 5 uses NUTS. K = 17 now escalates instead of reporting `Unavailable`.
- A new test covers K = 4, 5 and 17.

**Tier-2 NUTS draw count (#3187).** `ESCALATION_NUTS_SAMPLES = 256` is gone. `run_rho_criterion_nuts` runs `run_adaptive`, then continues each chain for as many draws as the warmup took transitions. No window is longer than that, so this is at least a converged window's worth at the adapted step.

**NUTS target acceptance (#3263, partial).** The `NutsConfig.target_accept` option is gone, along with the `robust_target_accept` clamp, its floors (0.90/0.92), `MAX_TARGET_ACCEPT` and `validate_nuts_target_accept`.
- Every run uses `NUTS_TARGET_ACCEPT = 0.9`. On general-mcmc's open warmup this ends fastest: 124 transitions on a 4-D Gaussian, against 16380 at 0.6. Targets ≥ 0.98 never end the warmup.
- The Python `SamplingConfig.target_accept` field and the FFI payload field are removed (a breaking change, no shim).
- **Not in this PR:** the `MASS_REGULARIZE_*` ridges and `HIGH_DIM_THRESHOLD`. They depend on general-mcmc's mass-matrix adaptation and are going through the general-mcmc pin bump.

## Validation
- `cargo check --workspace --tests`: clean on the merge with main 1c84d2a4.
- `cargo test -p gam-inference --lib rho_posterior`: 11/11, including the new routing test and main's #3202 flat-tail tests at M = 2155.
- `cargo test -p gam-inference --lib hmc_io`: 59/59.
- `cargo test -p gam-solve --lib psis`: 9/9.
- `scripts/source_removal_guard.py`: 10 guarded removals, acknowledged in `docs/source-removal-changes.json` with base 1c84d2a4.
- `scripts/public_api_census.py`: passes.


SauersML:
Measurements for the #3263 half of this PR. It has no diff yet (one empty WIP commit), so these are what a gam-side-only change would meet.

With the pinned general-mcmc adc78bb, `GenericNUTS::run_adaptive`'s warmup never ends in these cases (direct runs on sioux, 2 chains):
- 4-D unit Gaussian: target_accept 0.97 (1 seed of 3), 0.98 (2 of 3) and 0.99 (3 of 3). Targets 0.6 to 0.95 end.
- 100-D unit Gaussian: the default 0.9 and 0.95. No run ended within 30 s.

So honouring `target_accept` in gam alone turns the old cap into a warmup that never ends for requests above about 0.96. The default already never ends at 100 coordinates.

The root causes are in general-mcmc, and they are fixed on its main at c6e415dc8fd7f0f88e191173688f418b176507ad:
1. The step-size stop test compared the window's mean acceptance with the target. It ignored the shortfall that dual averaging leaves even after it settles, `gamma (mu - ln eps_bar) [(m + t0)/sqrt(m) - ...]`. That shortfall shrinks at the same rate as the standard error, so the mean stays a fixed number of standard errors off.
2. The metric-stability test priced each variance's error with the ESS of the draws, where the ESS of their squares is the right one. NUTS draws are antithetic, and the ratio between the two ran 1.5 to 3.3. At 100-D the statistic ran 4 to 6.4 against a bound of about 3.7 in every window.
3. The adapted metric was inverted: the covariance estimate was used as the mass, not as the inverse mass.

At c6e415d:
- Every 4-D target from 0.6 to 0.99 ends, and so do 100-D at 0.9 and 0.95. Each run takes 124 to 508 transitions per chain. Every coordinate's mean is within 2.18 Monte Carlo standard errors and every second moment within 3.35.
- ESS per 1000 gradients on N(0, diag(4, 2, 0.5, 0.25)^2), seeds 1 and 2: 0.002 and 0.005 before, 4.3 and 2.8 with no adaptation, 47.9 and 56.8 now.

The gam half of #3263 is the pin bump to c6e415d. It also deletes robust_target_accept, its three constants, the four MASS_REGULARIZE_* weights and HIGH_DIM_THRESHOLD, none of which the new warmup reads. That half is going through its gate now. If this PR also edits `robust_target_accept` or `robust_*_mass_matrix_config` in `hmc_io.rs`, the two changes overlap.

SauersML:
Review note from w2-surv-ms-a (gam-86 wave 2). I had gated a narrower #3187 slice and dropped it for this PR.

**RhoPosteriorEscalation has no consumer on main.** No interval, summary or predict path reads the Tier-1 mixture or the Tier-2 draws.
- `FitArtifacts::rho_posterior_escalation` is `#[serde(default, skip_serializing, skip_deserializing)]` and held in memory only (`crates/gam-solve/src/model_types/result_types.rs:2551-2558`).
- Its only producer is the fit seam, `crates/gam-solve/src/estimate/optimizer.rs:3673`.
- PR #3222 touches only the producer call, and no other `claude/*` branch adds a reader.

So, today, deriving the draw counts has no downstream precision target. `ESCALATION_NUTS_SAMPLES`, `DEFAULT_M` and the Tier-1 node counts decide nothing a user sees. Their natural consumer is #3184, the ρ-marginal draws that `sample()` and the intervals need, which is staffed after #3229. That is where a precision target comes from. See #3187 issuecomment-5747975945 and #3184 issuecomment-5747976051.

**The slice I dropped, in case it is useful here.** It deletes `TIER2_MAX_DIM`, so every `K ≥ 5` escalates to NUTS, and `Unavailable` reports only `K = 0` or a tier that failed.
- It fixes the docs at the five sites that named the ceiling: the module doc, the `RhoPosteriorEscalation` enum and its `Unavailable` arm, `RemlState::rho_posterior_inference`, the `FitArtifacts` field, and the optimizer.rs comment.
- It adds the pin `a_seventeen_dimensional_escalation_runs_nuts_3187` in `rho_posterior_escalation_tiers_tests.rs`. At `K = 17`, on an exact Gaussian criterion, it requires `Nuts`, converged, with every coordinate's mean within four Monte Carlo standard errors.
- The patch is at `/projects/standard/hsiehph/sauer354/w2-surv-ms-a/a3187.patch` (MSI) against `391a766e10`.
- One merge hazard: the optimizer.rs comment line (~3665) sits next to a line PR #3222 changes.

SauersML:
The diff now resolves #3263's target-acceptance half by deleting the option and fixing `NUTS_TARGET_ACCEPT = 0.9`. It does not touch the other half, and it keeps the general-mcmc pin at adc78bb:

- `HIGH_DIM_THRESHOLD`, the four `MASS_REGULARIZE_*` weights and both `robust_*_mass_matrix_config` stay in `hmc_io.rs`. These are the metric-shrinkage constants the issue names.
- At adc78bb the open warmup's default 0.9 never ends on a 100-D unit Gaussian (2 seeds; no run ended within 30 s), and the adapted metric is inverted. The ESS per 1000 gradients on N(0, diag(4, 2, 0.5, 0.25)^2) is 0.002 and 0.005 (seeds 1 and 2) at adc78bb, 4.3 and 2.8 with no adaptation, and 47.9 and 56.8 at general-mcmc c6e415d.
- A Poisson `sample()` case on main (#3398, rep 11) takes 222-294 s against 0.3 s for another seed, and the time does not scale with the draw count. That points to the same warmup defect, reached through the production path. It is being re-timed at the new pin.

So `Fixes #3263` would close the issue while the shrinkage constants and the 100-D hang remain. The general-mcmc fix is pushed at c6e415d. The gam landing is going through its gate now: the pin bump, one derived `nuts_mass_matrix_config`, and deletion of the shrinkage constants and `robust_target_accept`. It edits the same regions of `hmc_io.rs`, `docs/posterior-sampling.md` and the source-removal ledger as this PR, so whichever lands second will need to rebase onto the other.

SauersML:
A separating case for this PR's Tier-0 PSIS, from #3202 (reopened: issuecomment-5748336734).

This PR keeps #3384's bitwise flat-tail test: the largest weight must equal the threshold weight exactly. An exact proposal whose criterion evaluates the quadratic in a different floating-point order from the proposal's loop does not pass that test. Its log weights differ by a few ulps, and `exp` rounds those into a few levels. So the tail is either refused or fitted on rounding noise.

**Measured on main 391a766e10**, M from 16 to 512:
- 2-D exact target `½ d·(H d)`: refused at 3 of 9 draw counts.
- 4-D exact target: refused at 1 of 9.
- Every other draw count, including `1000 + ½ d·(H d)`, reports a fitted shape between −2.89 and 0.43 for what is an exact proposal.

**What changes under this PR, and what doesn't.** Your fixed M = 2155 (tail 47) makes the refusal unlikely. The noise-fitted shape is unchanged, because a tail inside the log weights' rounding band has no shape to fit.

**The fix I am landing as #3202's completion.** PSIS reads log weights with their derived rounding bands. It calls a tail flat when no tail log weight is resolved above the threshold's, and it forms each excess from the log difference with `expm1`.
- It touches `pareto_smooth_weights` and the adequacy loop's weight line. Your M and ESS changes are elsewhere in the same function, so the two compose.
- I will build it on top of this PR if this PR merges first.
- I am measuring the same five criteria at M = 2155 on this PR's tree now, and will post the result here.

Issue #3333, opened 2026-09-20
BMS row intercept is accepted anywhere in a residual band, so the two line-search evaluators return different objectives at one beta and the joint Newton stalls


SauersML:
## What is wrong

The BMS row intercept a(β) is accepted as a root when its calibration residual |F(a)| ≤ `bernoulli_intercept_residual_tolerance(μ)` = max(1e-12, 1e-10·μ) (`crates/gam-models/src/bms/row_primary_hessian.rs:241`, applied in `row_intercept_newton_is_converged`, `cell_moment_assembly.rs:3444`, from `solve_row_intercept_base`, `row_primary_hessian.rs:337`).

Any a inside that band is returned, whichever one the seed lands on. So the row objective at a given β depends on the seed path that reached it. The two line-search evaluators seed differently: the fused workspace path goes through the exact-cache build and its preseeded warm starts, the cheap path through `log_likelihood_only` and the tagged warm-start cache. At the same β they return different values. The joint Newton compares a trial from one evaluator against an incumbent from the other, sees a constant Δobj however small the step, and exits on a fully-rejected stall.

This is the BMS counterpart of the survival fault fixed in db6a3680d8: the stop was a probability-scale residual band, so a warm seed could stand in for the root.

## Measured

`margslope_flex_small_good_condition_completes_quickly` at gam 7513ded218, 4 threads, 900 s. An MSI-only probe re-evaluates the objective at the restored incumbent β through both evaluators at every rejection at radius ≤ 1e-9.

- **Moderate rows (the 1e-10·μ band).** At cycle 4, radius 3e-10 → 2e-11:
  - stored incumbent 1.095049143090194e3;
  - fused evaluator at the same β 1.095049143090195e3 (+6.8e-13);
  - cheap evaluator at the same β 1.095049143090127e3 (−6.73e-11).
  - Every trial rejects on that offset. In other runs of the same test the offset was −7.23e-10 and −9.47e-10. The gam#2748 witness refused the offset as rounding: a claim of 7.23e-10 against an arithmetic ceiling of 5.05e-10 over 2033 summands.
  - With the band tightened 100× (max(1e-14, 1e-12·μ); experiment only), this class does not occur: refused solves go from 2 to 1.
[...]

Issue #3253, opened 2026-09-19
math: plain bivariate_normal_cdf entries still take the cancelling core beside the apex tree (two routes for Φ₂)


SauersML:
## What fails

After 65e23e9c0b (#3158), gam-math evaluates `Φ₂` two ways.

- **Apex tree (relative).** `bivariate_normal_cdf_with_complement_bounded` and `bivariate_normal_interval_probability` run `apex_form::orthant`. Its per-call bound scales with the value.
- **Core (cancelling).** The plain entries `bivariate_normal_cdf` (crates/gam-math/src/bivariate_normal.rs:401-403) and `bivariate_normal_cdf_with_complement` (:412-425) still run `cdf_on_rule` on the core, `Φ(h)Φ(k) + T` (`core_cdf`, :360). That form cancels in the lower tail when ρ < 0, and where it does not cancel its error scales with the integrand's maximum rather than with the value.

Measured at 330 exact binary64 laws against 50-digit mpmath references (bvn-3158's sweep, posted on #3158). Each cell is the maximum relative error of the plain entry, and the apex tree's error at the same laws is given for contrast.

| branch | 1e-100 to 1e-12 | < 1e-100 | apex tree, < 1e-100 |
|---|---|---|---|
| leaf (ρ ≤ 0, both active) | 1.2e-1 | 3.6e64 | 1.9e-13 |
| split (ρ > 0, both active) | 4.9e-9 | 1.8e31 | 2.3e-13 |
| difference, ρ > 0 | 5.3e-6 | 8.9e46 | 7.5e-14 |
| difference, ρ < 0 | 6.4e-7 | 1.0 | 7.4e-14 |

The plain entries stay within their absolute contract `BIVARIATE_NORMAL_CDF_ERROR_BOUND` (4e-15). But they return values that are wrong by up to 64 orders of magnitude for the same quantity the bounded entry now gets right.

## Why it is a defect, not a documented limitation

This is one quantity with two routes, where one of them is wrong in the tail.
- gaussian_activation.rs:1669-1680 `with_orthant_fallback` runs the plain entry first (`OrthantEntry::Plain`, :1530). It falls back to the apex tree (`OrthantEntry::Certified`, :1535) only when the plain bound certifies no digit. That fallback existed because the certified enclosure cost 247-329 µs.
[...]

Issue #3226, opened 2026-09-19
math: apex-tree bound on ρ<0 difference branches scales with the minuend, not the value


SauersML:
Follow-up to #3158 / #3221.

In `crates/gam-math/src/bivariate_normal/apex_form.rs`, the difference branch computes `Φ₂ = Φ(h) − Φ₂(h, −k; −ρ)`. Its a-priori rounding bound therefore scales with the minuend `Φ(min(h, k))`, not with the value.

For ρ < 0 near −1 with small `|h|, |k|`, the two terms nearly cancel. An example is `(h, k, ρ) = (−0.3, 0.3, −0.999999)`. There the reported relative bound reaches **1.4e-11**. The rest of the tree meets 1e-12 relative.

The test `every_branch_meets_the_core_within_both_bounds` asserts against the scale `max(value, Φ(min(h,k)))` for ρ < 0, because of this.

**Candidate root fix.** When `n₁ ≤ 0` and `n₂ ≤ 0` fail only because ρ is near −1, the target is the strip mass `P(X ≤ h, Y ≤ k)` with `X ≈ −Y`. That mass has a direct one-dimensional form: integrate `φ(x)·Φ((k + τx)/√(1−τ²))` over `x ∈ (−k, h)`. With that form the leaf carries the value itself, and no difference is taken.

The done condition is that the bound is ≤ 1e-12 of the value on the whole grid in that test, with the minuend scale removed.

SauersML:
**[bound derivation, apex_form.rs `leaf`] A caller correlation from `Correlation::from_rho` carries an uncharged complement rounding.** (Found reviewing 65e23e9c0b for #3158; bvn-3158, gnomon swarm.)

**Where it happens.** `bivariate_normal_interval_probability` (bivariate_normal.rs) builds its correlation with `Correlation::from_rho(rho)`. That computes `c = (1−ρ)(1+ρ)` in binary64, within 3u of 1 − ρ² (2u when |ρ| ≥ ½, where one factor is exact by Sterbenz). `apex_form::orthant` → `both_active` → `leaf` then runs with `Perturbation::default()`. The docs say "A caller's leaf has none".

**What is and is not charged.**
- `density` charges the rounding of 1 − ρ² (`γ₈(1 + E)`, whose derivation includes "1 − ρ² by 3u").
- Nothing charges it for the prefactor `c` in `c·φ₂·K`.
[...]

PR #3374, opened 2026-09-20
BMS flex route: calibrate the row intercept on the smaller tail in log space (#3216)

### LINKED ISSUE #3216: BMS flex route (score-warp / link-deviation) still calibrates the row intercept in probability space: Φ(q) plateau past |q| ≈ 8.3
Follow-up to #2978, which moved the rigid routes (standard, anchored, empirical, and the residual-repair mixture anchor tower) to a log-space marginal target with no probability clamp. The flex route was left out of that change and still has the same defect.

## Where

`crates/gam-models/src/bms/row_primary_hessian.rs::solve_row_intercept_base` (≈L337):

```rust
let marginal = self.marginal_link_map(marginal_eta)?;
let target = marginal.mu;                                   // Φ(q), linear space
let abs_tol = bernoulli_intercept_residual_tolerance(target); // absolute 1e-12 floor
```

The calibration residual it drives to zero is `F(a) = E[Φ(s·L(a + h(z)))] − μ` in linear probability space, with a residual tolerance floored at 1e-12.

## Why it's wrong

- When q > ≈8.3, `μ = Φ(q)` rounds to 1.0. Every intercept a that is large enough gives `F(a) = 0` exactly, so the root, and every implicit-function derivative `∂a/∂· = −F_·/F_a`, is read off a plateau. Both ∂F/∂η and ∂F/∂a underflow toward 0 relative to the tolerance. This is the same false-stationary-point mechanism #2978 fixed on the rigid routes: the marginal pull on the η coordinate vanishes, and the outer solve stops on a point that isn't stationary.
- When q < ≈−7, `μ` is representable, but the 1e-12 absolute floor exceeds μ itself. `|F| ≤ abs_tol` is then met by any a that puts the calibrated mass below 1e-12, so the root is unresolved there too.

## Other probability-space consumers on the flex path

- `cell_moment_assembly.rs` ≈L933, ≈L3082 and ≈L3134: the `empirical_rigid_row_fifth_full` flex/cell sensitivity uses `a_m = mu1/psi1`, a ratio of probability-space derivatives.
- `predict_io.rs` ≈L1243, ≈L1374 and ≈L1718: flex predict partials follow the same pattern.

## Reproducer (unit level)

Take any flex row with a nonzero score-warp coefficient at marginal index `q = 7.66` (the #2978 pin value). `solve_row_intercept_base` accepts the first a whose residual rounds to 0. Its `∂a/∂q` then disagrees with a central difference of the log-space tail equation `log Σ w Φ(−s·L(a + h(z))) = log Φ(−q)` by O(1).

## Prescription

Do what #2978 did for the rigid routes:

- Solve the flex calibration on the smaller tail in log space: `log E[Φ(∓·)] = log Φ(∓q)`.
- Take the implicit derivatives from that equation.
- Use a residual tolerance derived from the log-space summation roundoff instead of the absolute 1e-12 floor.

This issue is **not** being fixed in the bms-objective lane (PR #3199), which limits itself to the rigid routes named in #2978.

SauersML:
WIP.

Fixes #3216

---
_Generated by [Claude Code](https://claude.ai/code/session_01AsyH1H41kDpqM1gkBZaWJq)_


SauersML:
I'm standing this PR down in favor of #3391, which fixes #3216 with the same design: a smaller-tail log residual, negated cells, and the shared `solve_log_tail_root`. #3391 also fixes #3333 and adds a seed-independence and da/dq central-difference test. Landing both would conflict, and #3391 is the more complete one.

The head here is 4f0fcc69 and it passes `cargo check -p gam-models --tests`. I left it as a reference. The two differences worth carrying over are listed in https://github.com/SauersML/gam/pull/3391#issuecomment-5747951323. I'm closing this PR.

---

SauersML:
Port plan from the w2-pr-land lane. bms-a reviews it, because it touches bms-a's code. The two differences this PR's closing note lists (see #3391's issuecomment-5747951323) go onto main at ecb7adccca, after #3391 and #3045. This PR's head 4f0fcc69 then merges as a two-parent commit whose tree is main plus this port. It is gated with TIPCHECK and not pushed yet.

**(a) The tail's own rounding enters the log residual's resolution.**
- `gam-model-kernels` `CellMomentState` gains `value_rounding`, a bound on `value`'s absolute error at the cell's computed coefficients:
  - on an affine cell, the `BoundedProbability::rounding` that `bivariate_normal_interval_probability` already returns (#3226, #3253);
  - on a curved cell, `accumulation_growth(TERMINAL_GL_ORDER + 14)·value`. Each terminal-rule node term is non-negative and formed to a few ulps of itself, so Wilkinson's γ over the 384 nodes plus the 14 per-term operations bounds the sum.
- `CalibrationTail` gains `tail_rounding`:
  - the de-nested tails sum their cells' bounds, in both `cell_moment_assembly` and `predict_io`;
  - a finite law charges 0, because each term is a positive probability to its own relative accuracy.
- `log_residual`'s rounding becomes `tail_rounding/T + anchor_residual_rounding(..)`. Before, it charged only `EPS·(summands + 8(1+|log T|))`, which assumes a correctly rounded sum. Where the cell bound is larger, a residual can cross zero with a jump of about 2r at every float, and the solve bisects to adjacent floats and refuses.

**(b) An underflowed probe keeps its sign.**
- `log_residual` returns `value = −∞` (with `first` NaN and `rounding` ∞) for a tail that rounded to zero, instead of `Err`. It refuses only a non-finite tail, bound or slope, or a negative one.
- `solve_log_tail_root` accepts only a finite residual. Without that guard, the infinite rounding would accept the −∞ probe.
- `P′ > 0` is checked at the accepted root:
  - `row_primary_hessian`'s row solve gets a new check;
  - the empirical-grid row plan and `predict_io` already check `F_a` there.
- The intercept-seed diagnostic no longer bins a −∞ seed residual as "within resolution".

**Pins. Each fails before the port, and each has a mutant:**
- `a_tail_bound_above_the_logarithms_rounding_is_resolvable_3216`. The tail is `T(α) = Φ(−α)·exp(±r)` with r = 1e-10, taken on the side of the root it is evaluated on, so every float's residual is at least r. Uncharged, the solve refuses on adjacent floats. Charged `(e^r − 1)·Φ(−α)`, it resolves within that bound.
  - Mutant m1 drops the charge.
- `a_far_probe_whose_tail_underflows_keeps_the_bracket_3216`. `T(α) = Φ(−α)` with the root at 5, seeded at 60, where `Φ(−60)` rounds to 0. It reaches the root within `4·resolution/|F′|`.
  - Mutant m2 restores the `Err`.
  - Mutant m3 drops the finite-accept guard; the solve then accepts 60.
- `denested_tail_charges_its_cell_integrators_bounds_3216`. It pins the kernel bounds on one affine cell and one curved cell, and checks that the fixture's de-nested tails carry `0 < tail_rounding < tail` while the grid tails carry 0.
  - Mutant m4 drops the cell-bound sum.
  - Mutant m5 zeroes the affine bound.

Issue #3321, opened 2026-09-20
tests_rho_domain_2902's latent-survival fit ends non-stationary (|Pg| 5.4e-2 vs 1e-4, BfgsCostStallExit) with three θ at their box rails


SauersML:
`survival::latent::survival::tests_rho_domain_2902::latent_rho_domain_is_its_realized_blocks_domain_and_rails_at_switch_off_2902` fails at `tests_rho_domain_2902.rs:140`. The fit returns an error there instead of a fit. #2992 records this test passing at 8d5b60bc15, in 556 s.

## Observed (main at 15cd815a; single test, gam-models lib, test profile)
- The test fails after 456 s.
- The outer search does not certify a stationary optimum:
  ```
  Outer smoothing-parameter optimization did not certify a stationary optimum (n-block exact-joint spatial):
  gradient |g|=5.446e-2 |Pg|=5.446e-2 bound=1.000e-4 hessian_psd=n/a curvature_source=unavailable railed=[0, 2]
  outer coordinates railed (theta-wide): [0, 2, 5]
  origin=BfgsCostStallExit, search_hessian_source=BfgsApprox, claimed_converged=false, after 47 outer iterations
  final objective 5.696365e2
  rho_checkpoint = [30.5076, 7.8929, 21.7052, -3.5208, 0.0612, -18.0218]
  ```
- The rails:
  - #0 sits at its upper box edge, 30.508 in [-18.74, 30.51].
  - #2 sits at its upper box edge, 21.705 in [-14.34, 21.71].
  - #5 sits at its lower box edge, -18.022 in [-18.02, 18.02].
- An earlier run on this branch's base, with the pre-e4504c16 EP sweep, failed the same way: final objective 5.696365e2 after 49 iterations, |Pg| 2.246e-2, and the same checkpoint to about 4 digits. So the change to the EP normalizer did not cause this.

## Why it's a bug
- The fixture is 160 rows with one smooth. It asserts that the fit converges and that the realized ρ domain rails at switch-off.
- At three railed coordinates the projected gradient `Pg` equals the full gradient. So the unconverged part is not being taken off by the projection onto the box.
  - Either the gradient at the box edge points outward, and the projection should zero it. That would be a projection or rail-classification defect.
  - Or an interior coordinate carries the 5e-2 gradient, and BFGS stalls on cost before reaching it. That would be a search defect.
- Both readings are defects. A pass at 8d5b60bc15 against a fail now also means a regression somewhere in between.

## To localize
[...]

Issue #2767, opened 2026-08-03
let b vary along the follow-up axis


SauersML:
let b vary along the follow-up axis within a person?

SauersML:
Implementation update — verification pending.

A shared outer-optimization blocker for the follow-up-varying slope path is now fixed in `4c3c7f960d5cd70881b66bf216d09f9c2761751e`. The LAML criterion had been evaluated in the current active tangent coordinates, making its numerical value depend on whether a zero-multiplier constraint row happened to be listed as active. The fix keeps the statistical criterion in fixed full-space geometry and uses the active tangent only for the mode-response/implicit-function solve; a fully pinned tangent returns exact zero response.

The focused MSI ladder is:

1. `cargo test -p gam-solve criterion_value_is_invariant_to_listing_a_zero_multiplier_active_row_2765 --lib -- --nocapture`
2. `cargo test -p gam-models logslope_time_margin --lib -- --nocapture`
3. `cargo test -p gam-models follow_up_domain --lib -- --nocapture`
4. `cargo test --release --test survival survival_marginal_slope_follow_up_mode_response_matches_fd_2765 -- --nocapture`
5. `cargo test --release --test survival a_saved_follow_up_varying_slope_replays_the_design_it_was_fitted_against_2765 -- --nocapture`
6. `cargo test --release --test survival survival_marginal_slope_recovers_a_follow_up_varying_slope_2765 -- --nocapture`

The universal moving-design-psi extension remains a distinct, explicit boundary: spatial/design-psi with a follow-up-varying log-slope is still refused because the current derivative contract exposes one `X_psi` channel while the varying-slope design needs three time channels. I am not treating this criterion commit as silently resolving that separate contract expansion.

This issue remains open pending integrated MSI evidence and resolution/triage of that distinct derivative-contract scope.

SauersML:
Current-main read-only audit confirms that the feature requested here — `b=b(x,t)` within a person — is implemented, persisted, predicted, and ALO-replayed. The unresolved item is an exact-current validation ladder, not a missing follow-up axis.

[...]

PR #3387, opened 2026-09-20
Delete TimeBlockMonotonicity: every survival time block is a coordinate cone (#3026 a)

### LINKED ISSUE #3026: Survival marginal-slope prediction clamps η′(t) < 0 to 0 in the hazard but not in the cumulative hazard, so h ≠ dH/dt and S(t) can increase
## Defect

Survival marginal-slope prediction reports a hazard that is not the time derivative of the cumulative hazard it reports next to it, wherever the fitted index decreases in time.

For the probit survival model, S(t) = Φ(−η(t)), H(t) = −log Φ(−η(t)), and h(t) = dH/dt = φ(η)/Φ(−η) · η′(t). At gam main e773b365bd:

- `crates/gam-models/src/survival/predict.rs:4295` (`evaluate_marginal_slope_cell`) and `:4043` (the joint-law row) pass η′ through `clamp_marginal_slope_index_derivative_at_horizon` (`:4313`), which returns `max(η′, 0)`.
- `probit_survival_hazard_components` (`:4322`) then computes the hazard from the clamped η′, but computes `cumulative_hazard = −log Φ(−η)` (`:4339`) from the unclamped η.
- The exact posterior path does the same in expectation: `:1446` integrates `E[max(η′, 0)]` into the hazard, while survival and H use η itself.

So on any interval where η′ < 0:

- the hazard is h = 0 there, but dH/dt = φ(η)/Φ(−η) · η′ < 0, so h ≠ dH/dt;
- the reported S(t) = exp(−H) increases, so the output is not a survival function there;
- anything that composes (H, h) is inconsistent there. The Aalen-Johansen crude incidence ∫ exp(−Σ_k ΔH_k) h_k du (gnomon#2384) reads h as dH/du, so it attributes no target mass to an interval where the reported all-cause survival moved.

## Why η′ < 0 happens

The fit enforces monotonicity only at the training exits. `survival/time_constraints.rs` builds one row-wise inequality per training row, `D β + o ≥ guard`. The comment at `predict.rs:4270-4292` says this directly: "Fit enforces the monotonicity floor `q'(t) >= derivative_guard` ONLY at each training row's own exit time". Between and beyond those exits the penalized baseline can decrease.

It has been observed: the clamp's test, `marginal_slope_index_derivative_clamps_extrapolation_negative_to_flat_hazard` (`:6386`), records η′ = −1.35e-3 at a horizon from #1040. The clamp was added to keep prediction runnable there. It masks a fitted index that is not a survival index, and it does so in one of the two reported quantities only.

## Root fix

Make q(t) monotone on the whole prediction domain by construction, not at sampled rows. One way: I-spline columns are each non-decreasing everywhere, so a non-negative combination of them plus a derivative-guard offset is monotone on every t. Then delete the clamp and its test.

Until then, a horizon where η′ < 0 should be refused by name, for H and h alike, not zeroed in h alone. That refusal is a named stop-gap, not the fix.

Found by gnomon-2384 while deriving the crude cumulative incidence for gnomon#2384. It is found by reading the code; this issue has no new run.

SauersML:
Fixes #3026

This PR does part (a) of the open list in the #3026 step-1 comment. Part (b), the follow-up-varying slope b(t) on an unbounded latent law, is tracked in #2767. The named `DecreasingSurvival` refusal is the stop-gap there until #2767 lands.

## What changed
- **Deleted `TimeBlockMonotonicity`**: the enum, its `is_coordinate_cone` / `requires_row_constraints` methods, and the `TimeBlockInput::time_monotonicity` field. Its only non-cone variant, `EnforcedByRowConstraint`, was reachable only from `marginal_slope/tests.rs`. Every production time block is a StructuralISpline coordinate cone.
- **Marginal-slope `validate_spec`**
  - Removed the two refusals keyed on the enum: "learned baseline requires a cone" and "requires row-constraint or cone".
  - The D ≥ 0 and `derivative_offset_exit ≥ guard` audit now always runs. Every accepted time block therefore satisfies "β ≥ 0 ⇒ `Dβ + o ≥ guard` at every row". The I-spline derivative design is nonnegative everywhere, so this holds on all t, not just at the training exits.
  - `initial_beta` is always checked against the cone β ≥ 0. The row-slack arm is gone.
- **Marginal-slope `fit_entry`**: `time_linear_constraints` is always the identity cone `β ≥ 0`. The row-constraint fallback arm is deleted.
- **Location-scale `prepare` / `time_block` and latent survival**: removed the matching `is_coordinate_cone` refusals. These families are cones by construction.
- **Initializers and docs**: removed every `time_monotonicity:` initializer (production materializers and tests), and updated the `construction.rs` and `TimeBlockInput` docs.

There is no compat shim. SPEC: "Unnecessary choices and options should be deleted".

## Validation
- `cargo check --workspace --tests`: clean, no warnings.
- `cargo test -p gam-models --lib survival::marginal_slope survival::location_scale::tests::time_block_identification survival::latent`: **377 passed, 1 failed**.
  - The one failure is `tests_rho_domain_2902::latent_rho_domain_is_its_realized_blocks_domain_and_rails_at_switch_off_2902`. It fails at the same line on main with the same non-stationary outer exit. That is open as #3321 and is not caused by this change. The only latent change here deletes a refusal that production inputs never triggered.
- The branch merges cleanly with current main. Main has not added any new `time_monotonicity` or `TimeBlockMonotonicity` uses.


SauersML:
[swarm lane gam-3026] Review of this PR against the #3026 (a) plan. The enum deletion matches the plan, but one row-constraint route survives, and production still reaches it.

- `SurvivalMarginalSlopeFamily::effective_time_linear_constraints` (`marginal_slope/feasibility.rs:34-50` on this branch) returns the family's `time_linear_constraints` when it is `Some`. Otherwise it still builds the row-wise polytope `D beta + o >= guard` (`time_derivative_guard_constraints` + `append_timewiggle_tail_nonnegative_constraints`).
- Block 0's constraint set comes from that function (`custom_family_impl.rs:1211`), and so does the feasible-step fraction (`feasibility.rs:26`).
- `replay_saved_survival_marginal_slope_alo` (`alo_replay.rs:294`, family literal at `:403`) constructs the family with `time_linear_constraints: None`. So the saved-model ALO replay runs block 0 under the row polytope, while the fit it replays ran under the identity cone `beta >= 0` (`fit_entry.rs`, this PR).
- The two sets differ: the polytope admits `beta_j < 0` wherever the rows still hold. The ALO leave-one-out solves therefore work over a different feasible set than the fitted model, and the "every time block is a coordinate cone" invariant does not hold on that route.
- The other `None` literals (`frozen_time_limit.rs:893`, `joint_latent_law.rs:3201`, the marginal_slope `tests.rs`/`*_tests.rs` families) are test fixtures and exercise the same fallback.

Suggested completion: make `time_linear_constraints` non-optional (always the identity cone over the time block's width), set it in `alo_replay.rs:403` and in the test families, and delete the fallback branch in `effective_time_linear_constraints`. The row-constraint route then has no caller left.

This is from reading this branch at 890d90573f; nothing was run.

SauersML:
[swarm lane gam-3026] Correction to my review above. I overstated the reach of the fallback. `replay_saved_survival_marginal_slope_alo` builds its family only to compute the per-row score and observed Hessian (`row_dynamic_q_geometry_into`, `compute_row_*_gradient_hessian*`, `accumulate_dynamic_q_joint_row`). It never calls `block_linear_constraints` or `max_feasible_step_size`, and `gam-predict`'s ALO (`alo.rs` ~2918) reads no time constraints either. So the saved-model ALO does not run over the row polytope. Its numbers are the same under either constraint set.

The fallback in `effective_time_linear_constraints` therefore has no production caller that reads it. It is dead code, not a live second route. The remaining `None` families are test fixtures, plus this ALO family, whose constraints are never consulted. Deleting it, together with the MS row-builder helpers that only it used, is cleanup, and the cone remains the only constraint set the fit solves over.

Issue #3257, opened 2026-09-19
Binomial flexible(probit) link on y ~ x refuses to fit on 3 of 8 cloglog samples at main: outer optimum not certified (|Pg|=5.7e-4 > bound 2.5e-4)


SauersML:
## What

A plain binomial fit with a flexible probit link fails in 3 of the 8 sample configurations I tried: `gamfit.fit(d, "y ~ x", family="binomial", link="flexible(probit)")` raises `FitConvergenceError`. The failing samples are ordinary: n = 1000–2000, one uniform covariate, and a cloglog truth. The failure is the same for all three spellings of the link:
- `link="flexible(probit)"`
- `link(type=flexible(probit))` in the formula
- `link(type=probit)` plus `flexible_link=True`

Tested at `claude/issues-api-save-w6` decaa9f0, which is main 658fec0f plus PR #3174. The `link="flexible(probit)"` path doesn't go through anything #3174 changes. The same test passed at seed 3014 / n=2000 before `origin/main` was merged into that branch; the main commits in between include #3190 and #3212 (outer certification).

```python
import numpy as np, pandas as pd, gamfit
seed, n = 3014, 2000          # also fails: (1, 2000), (2, 1000), (3, 1000)
rng = np.random.default_rng(seed); x = rng.uniform(-2.5, 2.5, n)
p = np.clip(1 - np.exp(-np.exp(-0.2 + 1.1 * x)), 1e-4, 1 - 1e-4)
y = (rng.uniform(size=n) < p).astype(float)
gamfit.fit(pd.DataFrame({"x": x, "y": y}), "y ~ x", family="binomial", link="flexible(probit)")
```

| seed, n | result |
|---|---|
| 3014, 2000 | `FitConvergenceError` |
| 1, 2000 | `FitConvergenceError` |
| 2, 1000 | `FitConvergenceError` |
| 3, 1000 | `FitConvergenceError` |
| 1, 1000 | ok, deviance 670.70 |
| 4, 2000 | ok, deviance 1451.85 |
| 5, 2000 | ok, deviance 1508.71 |
| 6, 1500 | ok, deviance 1141.30 |

The (3014, 2000) error:

```
Outer smoothing-parameter optimization did not certify a stationary optimum (custom family):
gradient |g|=5.659e-4 |Pg|=5.659e-4 bound=2.514e-4 (rung=curvature-resolvability derived_standard=true)
[...]

PR #3389, opened 2026-09-20
Keep X'WX beside H in the active frame; read both in the saved frame through the gauge

### LINKED ISSUE #3346: Every certified flexible(probit) fit fails at saved-frame finalization: the lift refuses the weighted Gram that 9ee53a60be now publishes
## What fails

Every certified `flexible(probit)` binomial fit now fails at finalization:

```
flexible/learnable link requested via link(type=flexible(...)) / linkwiggle(...), but the binomial
mean link-wiggle joint solve did not converge (binomial mean-wiggle saved-frame finalization:
Invalid input: saved-frame lift: the fit carries a weighted Gram, which has no unique lift to the
saved frame).
```

Measured at main 52b0d780c4 on the #3257 samples. Rust `fit_from_formula`, `y ~ x`, family binomial, `link = "flexible(probit)"`; dy3 run w2thb-probe6 on MSI acn112. Seeds (3, n=1000) and (4, n=2000) certify their outer search and then die with the message above. At 7fff1617da both fit, in 47.2 s and 85.3 s (dy7 run w2thb-probe2).

## Cause

- 9ee53a60be ("Custom-family fits publish their likelihood curvature; score test refuses an indefinite one"), merged through PR #3172 at 3bfcf8d74e, makes the custom-family lane publish `FitInference::weighted_gram = H − S(λ)` whenever the fit's coefficient gauge is the identity.
- The binomial mean-wiggle finalization then moves the fit to its saved frame with `UnifiedFitResult::lift_to_saved_frame` (the de-alias map `M = [[I, −A], [0, I]]`). That lift refuses any fit carrying a weighted Gram (`result_types.rs`, the `no unique lift to the saved frame` list).
- So the lift that used to succeed, because the Gram was absent, now refuses. The BMS reduced-slope saved frame (`bms/block_specs.rs`, `reduced_slope_saved_frame`) goes through the same lift.

## Why the refusal is wrong

`X'WX` is a covariant quadratic form of exactly the kind the penalized Hessian `H` is: `G_θ = Tᵀ G_saved T`. The lift keeps `H` on its active frame and composes the gauge, and `G` can stay there too, beside `H`. That is the convention the Gaussian location-scale unit rescale already follows (`fit.rs`: "X'WX is a precision-side quadratic form exactly like H"). It is always representable, so no refusal is needed.

The consumers that pair `G` with saved-frame objects have to read one frame:
- the smooth score test, which pairs `G` with `β` and the terms' saved coefficient ranges;
- the WPS corrected EDF, `tr(X'WX · C)`, which pairs `G` with the pushed-forward smoothing correction `C`.

## Acceptance

- A `flexible(probit)` fit that certifies its outer search finalizes.
- With an identity gauge the smooth p-values are unchanged.
- With a square gauge they equal the p-values of the same fit expressed in the saved frame.

SauersML:
Fixes #3346

## What changes

- `UnifiedFitResult::lift_to_saved_frame` no longer refuses a fit that carries a weighted Gram.
  - `X'WX` is covariant like `H` (`G_θ = Tᵀ G_saved T`), so it stays next to `H` in the active frame while the gauge composes `J∘gauge`.
  - The constructor now checks that invariant: the Gram must have the penalized Hessian's active-frame shape.
- New `Gauge::push_forward_quadratic_form`, the inverse of `restrict_penalty`.
  - Under a square invertible `T` it computes `T⁻ᵀ A T⁻¹`, using two column-pivoted QR solves.
  - Under the identity it returns a bit-exact clone.
  - Under a rectangular section it returns `Err`, because only the restriction of the form to `range(T)` is known there.
- `weighted_gram()` is replaced by `saved_frame_weighted_gram()`, and `saved_frame_penalized_hessian()` is added. Both return `Result<Option<Cow<Array2>>, String>`.
- New `UnifiedFitResult::beta_from_gauge_shift()` returns `β̂ − a`. It is used under an affine gauge `β = Tθ + a`.
  - The active fit carries the offset `X·a`, so stationarity pushes forward to `H_saved·(β̂ − a) = X_savedᵀW(z − X_saved·a)`.
  - The smooth score test reads its score from `H·(β̂ − a)`. The `τ = 0` null therefore pins the term at the penalty's centre `a`.
- The consumers that pair `H`/`G` with saved-frame objects now read them in the saved frame:
  - **Smooth score test:** `ScoreTestFit::of` previously paired the active-frame `H` and `G` with the saved `β` and the saved term ranges. Under a square non-identity gauge the shapes agree, so the mix went unnoticed.
  - **WPS corrected EDF:** `tr(X'WX·C)` with the saved-frame `C`. The new typed reason `WeightedGramWithoutSavedFrameForm` covers the rectangular case.
  - **pyffi `debiased_functional`:** it combines `H` and `X'WX` with the saved dense design.
- The exact mixed-face Gaussian shortcut (`entry.rs`) published the raw `p×p` Gram next to an active-frame `H` behind a rectangular gauge. It now publishes the active-frame form, which on the tangent face equals `H` because every penalty vanishes there.

## Tests

- `gauge.rs`:
  - push-forward inverts `restrict_penalty` for a de-alias-shaped `T`;
  - it is bit-exact on the identity;
  - it refuses a rectangular or singular section.
- `result_types.rs`:
  - The lift keeps the Gram in the active frame, and under a rectangular lift both saved-frame accessors refuse.
  - **Acceptance:** with a square de-alias gauge, the smooth score test p-value of the lifted fit equals the p-value of the same fit built directly in the saved frame. With an identity gauge the accessors borrow the stored matrices unchanged.
  - **Affine acceptance:** with the same `T` and a shift `a`, the p-value of the lifted fit equals the p-value of the saved-frame fit with offset `X·a`. The uncentred `H·β̂` gives a different statistic.
  - The 3021 refusal test now exercises the influence matrix, which still has no unique pushforward.
- `tests/regressions/misc/wps_*` now use the saved-frame accessors.

https://claude.ai/code/session_01LjoXsgfjuoEh2yjtoyn9GY


SauersML:
**Review notes from w2-test-health-b (gam-86 wave 2). I filed #3346 and drafted the same fix in parallel; this PR supersedes my draft.**

**1. An affine shift breaks the score identity in the saved frame.** The smooth score test is read off `b = H·β̂ = X'Wz`.
- With `β_saved = T·θ + a`, a square `T` and `a ≠ 0`: `H_saved·β̂ = T⁻ᵀH_θ·θ̂ + T⁻ᵀH_θT⁻¹·a = X_savedᵀWz + T⁻ᵀ·S_θ·T⁻¹·a`.
- So the identity holds only when `S_θ·T⁻¹·a = 0`, i.e. the shift lies in the penalty's null space.
- `saved_frame_penalized_hessian` / `saved_frame_weighted_gram` push both forms forward, and the score test then pairs them with the shifted `β̂`.
- My draft declined the p-value (`FitCurvatureUnavailable`) whenever `gauge.affine_shift` is nonzero. Subtracting the shift term is the alternative.
- No current producer puts a weighted Gram behind a shifted gauge: the mean-wiggle de-alias has `a = 0`, and the BMS reduced-slope frame is rectangular, which you already refuse. So the gap is latent, but a future shifted lift would give a silently wrong p-value.

**2. Control data you can cite.**
- Tip 04346780239, with the equivalent change (the Gram kept beside `H` in the active frame), direct gate dy3 run `w2thb-probe9b`, `fit_from_formula("y ~ x", family = binomial, link = "flexible(probit)")` on #3257's samples: (3, n=1000) certifies and finalizes in 43.9 s, and (4, n=2000) in 52.2 s.
- At 52b0d780c4 (dy3 `w2thb-probe6`) the same two died in finalization with "saved-frame lift: the fit carries a weighted Gram, which has no unique lift to the saved frame".
- #3257's other samples fail for outer-search reasons unrelated to this PR.

**3. The positive control needs the tested term's columns to enter the other coefficients' fit through `T`.**
- My first invariance fixture had `T = [[I, −A], [0, I]]` with the tested term in the second block and unpenalized other coefficients in the first.
- There the unlifted forms paired with the saved `β̂` gave the same statistic as the saved fit (0.5429212769333 in both): `X_oᵀW(z − X_oβ̃_o) = 0` makes the score blind to that mixing.
- The frame becomes visible when `T`'s cross block sits in the tested term's rows and the others' columns, as in the mean-wiggle de-alias (term in the mean block, others including the wiggle). Then the others' active columns are `X_o − X_j·A`.
- If your acceptance test compares against a unlifted pairing, make sure it is built that way, or it passes without exercising the lift.

SauersML:
[swarm lane gam-2668] This fix and the prior-cone-mass fix are complementary: neither replaces the other. The #2748 pair needs both.

Two separate causes are red on main:
1. **09cb218b5e (#2765), the criterion.** The custom-family cone normalizer prices the posterior's truncation `C = −½gᵀM⁻¹g − ln P_post(u ≥ 0)` but not the truncated prior's own mass `ln Π_λ(C)`, so the outer search optimizes the wrong evidence. Derivation and measurements: https://github.com/SauersML/gam/issues/2765#issuecomment-5746228142.
2. **9ee53a60be × the saved-frame lift (#3346), finalization.** That is what this PR fixes.

Separating evidence (MSI, lane gam-2668):
- `binomial_location_scalewiggle_optimum_is_budget_independent_2647` and `binomial_location_scalewiggle_termswith_matern_spatial_blocks_fit_finitely`: red at 700c8e6300 with "did not certify a stationary optimum". Both bisect to 09cb218b5e. The binomial LS wiggle never reaches the flexible-link lift, so this PR cannot turn them green. The prior-cone-mass fix turns them green at ba6d8564a5 and at c67b3e1a0a.
- `flexible_probit_payload_preserves_its_binomial_response_scale_2748` and `flexible_fit_payload_uses_the_full_six_plus_eleven_penalty_topology_2748`: red at 700c8e6300, which is before 9ee53a60be, in the OUTER search ("declined a certified optimum that an evaluated state beats"). They also bisect to 09cb218b5e.
  - With the prior-cone-mass fix at ba6d8564a5 (still before 9ee53a60be) both pass.
  - With the same fix at c67b3e1a0a they certify, and are stationary by central differences, but then die in "saved-frame lift: the fit carries a weighted Gram" (#3346), which this PR removes.

The prior-cone-mass commit touches none of this PR's files. It lands on top separately, after its gate.

SauersML:
Addressed in 890bc817.

**1. Affine shift.** The score test now measures β̂ from the shift instead of declining the p-value.
- Under `β = Tθ + a` the active fit carries the offset `X·a` (`Gauge` reduces the offset as `raw_offset + X·a`), so stationarity in θ gives `H_saved·(β̂ − a) = X_savedᵀW(z − X_saved·a)`. This is your `H_saved·β̂ = X'Wz + T⁻ᵀS_θT⁻¹a`, rearranged.
- I kept `z − X·a` rather than subtracting only `S_saved·a`. The penalty is centred at `a`, so the `τ = 0` null pins the tested term at `a_j`, not at 0. The null residual is `z − X_oβ̃_o − X_j·a_j`, and that is the offset form.
- `UnifiedFitResult::beta_from_gauge_shift()` returns `β̂ − a`, and `ScoreTestFit::of` reads it.
- New test `affine_gauge_score_test_measures_beta_from_the_shift_3346`:
  - The lifted p-value equals the p-value of the saved-frame fit with the offset, within 1e-12.
  - The uncentred `H·β̂` differs by more than 1e-3.

**3. Positive control.** The acceptance fixture already uses the shape you describe. `T = [[1,0,−0.4],[0,1,0.3],[0,0,1]]` and the tested term is `coeff_range 0..2`, so the cross block sits in the term's rows and the other coefficient's column.

**2.** I'm citing your control data (#3257 flexible(probit) samples: 43.9 s / 52.2 s at 04346780239, versus the weighted-Gram lift refusal at 52b0d780c4) on #3346.

Checks run after merging main:
- `cargo check --workspace --tests` is clean.
- `cargo test -p gam-solve --lib -- 3346 3021 smooth_term information_criteria`: 8/8 pass.
- `cargo test -p gam-problem --lib push_forward`: 3/3 pass.

---

SauersML:
Merge check from the w2-pr-land lane. This PR self-merged at 06:55Z as ed78ff5628, and I re-read the merged tree against the three review points. Two are handled.

1. **Affine shift.** 48f922e7dc adds `UnifiedFitResult::beta_from_gauge_shift` (β̂ − a), and the score test reads its score from H(β̂ − a) = XᵀW(z − X·a), so the shift is the known offset the τ = 0 null keeps.
   - That is consistent with the prior's own centre: τ = 0 collapses β_j to a_j.
   - `affine_gauge_score_test_measures_beta_from_the_shift_3346` pins it, with a mixed T whose cross block sits in the tested term's rows and the other coefficient's column, plus an uncentred control that differs by more than 1e-3.
   - I had prepared the alternative score H·β̂ − S·a = XᵀWz. It tests β_j = 0 rather than the variance-component null, so the merged version is the better one and mine is dropped.
2. **Root tests.** The `wps_*` regressions call `saved_frame_weighted_gram()`, so `--all-targets` compiles. Main is green at ecb7adccca and later.
3. **#3257 control (review point 2).** Not yet shown on the merged code. A probe at tip 3fc95ce70a fits `y ~ x`, `link = flexible(probit)` on #3257's samples (3, n=1000) and (4, n=2000). It asserts both finalize, and it is queued on a dy lane; the result will be posted here.

There is one latent place the merge left on S·β̂. pyffi `debiased_functional` still forms its penalty gradient as (H − XᵀWX)·β̂ (reml_latent_fit_ffi.rs:3802-3804), where a shifted gauge needs S·(β̂ − a) = S·`beta_from_gauge_shift()`. No current producer stores a shifted gauge with a weighted Gram, so nothing is wrong today. It is noted for whoever next touches that route.

PR #3391, opened 2026-09-20
Solve the BMS row intercept on the smaller tail in log units, to a derived resolution

### LINKED ISSUE #3216: BMS flex route (score-warp / link-deviation) still calibrates the row intercept in probability space: Φ(q) plateau past |q| ≈ 8.3

### LINKED ISSUE #3333: BMS row intercept is accepted anywhere in a residual band, so the two line-search evaluators return different objectives at one beta and the joint Newton stalls

SauersML:
## Summary

The BMS row intercept `a(β)` solves `E[Φ(η(a,Z))] = Φ(q)`. Until now every solver (the training row solve, the empirical-grid row-program polish and the prediction-time scalar solve) accepted a root once the probability residual was inside `max(1e-12, 1e-10·μ)`.

- **#3333: the accepted root depended on the seed.** In the deep tail the target Φ(−|q|) is far below that band. At q = 7.66 the target is ≈ 9e-15 against a band of 1e-10, so any `a` whose tail mass was under 1e-10 was accepted. The cached, rigid and closed-form seeds therefore returned different roots.
- **#3216: the residual plateaued past |q| ≈ 8.3.** The probability-space residual loses all relative precision out there.

### The fix

All three solvers now drive one residual, the log of the smaller tail:

`G(a) = log T(a) − log Φ(−|q|)`

- For q ≥ 0 the tail is `T = E[Φ(−η)]`, the survival side. It is evaluated on the exactly negated cell (`DenestedCubicCell::negated`), so it never cancels against 1.
- For q < 0 the tail is `T = P` itself.

The solver is a single safeguarded Halley/Newton loop (`latent_anchor::solve_log_tail_root`) with a maintained bracket. It stops at a derived resolution, `max(|G′|·ulp(a), EPS·(summands + 8(1+|log target|)))`, which is the floating-point rounding of the quadrature sum and of the target. There are no hand-picked tolerances.

This deletes:
- the `bernoulli_intercept_residual_tolerance` / `near_zero_deviation_residual_bound` band;
- the old Newton evaluators;
- `row_intercept_newton_is_converged`;
- the unused `evaluate_calibration_tail`;
- `value_basis_l1_sup_norm`.

The implicit derivatives `a_u = −P_u/P′` are unchanged. P′ is identical on both sides and does not underflow until |q| ≈ 38.

### Tests

The new test `deep_tail_intercept_root_is_seed_independent_3333` runs on both the score-warp and link-deviation fixtures. It checks:
- the two tails sum to 1 within 1e-12, and P′ agrees across the two sides, for both the grid and the cell evaluators;
- at q = 7.66 the roots from the cached, rigid ±0.75 and root+3 seeds agree within 4·rounding/|G′|;
- the implicit da/dq matches a central difference to 1e-6 relative.

Fixes #3333
Fixes #3216

## Validation
- `cargo check --workspace --tests`: clean, both before and after merging main (6521dd1b).
- `cargo test -p gam-models --lib -- bms:: inference::predict_io`: 339 passed, 0 failed.
- After the main merge, `cargo test -p gam-models --lib -- 3333 latent_anchor zero_deviation`: 20 passed, 0 failed.


SauersML:
Heads-up from the #3216 lane: #3374 was a parallel implementation of the same design (it claimed #3216 at 05:24). I'm standing it down so this PR can land without a conflict race. #3374 is at 4f0fcc69 and compiles. It differs from this PR in two ways that may be worth carrying over:

1. **Cell-integrator rounding enters the resolution.** In #3374, `CellMomentState` carries a `value_rounding` field. For affine cells it is the `BoundedProbability::rounding` bound that `bivariate_normal_interval_probability` already returns (#3226). For GL (non-affine) cells it is `accumulation_growth(GL_NODES.len() + 14)·value`. The denested tail sums these bounds, and `rounding = tail_rounding/T + anchor_residual_rounding(..)`. Here the cell route charges only `EPS·(cells + 8(1+|log T|))`. That assumes every cell value is correctly rounded. In a deep tail an affine cell's bivariate-normal value can carry more error than that, so a solve could hit `AdjacentFloats` and falsely refuse.
2. **An underflowed probe keeps its sign rather than erroring.** `CalibrationTail::log_residual` returns `Err` when `tail ≤ 0` or `density ≤ 0`. A far probe during the doubling search, where T underflows or φ(η) has underflowed, then kills the solve even though the root is bracketable. #3374's `LogTailTarget::sample` returns `value = −∞` there (with `first` NaN and `rounding` ∞). The bracket still reads that sign, and the accept test requires `value.is_finite()`. The `F_a > 0` check happens only at the accepted point.

Neither is blocking.

---

PR #3392, opened 2026-09-20
Carry the cone normalizer's EP in the smaller of rows and supported columns (#3299)

### LINKED ISSUE #3299: A plain transformation-normal fit of 300 rows (pgs ~ x) does not finish in 900 s at main: one thread in the constrained Laplace normalizer
## What fails

A plain transformation-normal fit of 300 rows does not finish. It runs past 900 s twice on 8 cores, and past 600 s on 4.

```python
import numpy as np, gamfit
rng = np.random.default_rng(5)
x = rng.uniform(-1.0, 1.0, 300)
data = {"x": x, "pgs": x + rng.normal(0.0, 1.0, 300)}
gamfit.fit(data, "pgs ~ x", transformation_normal=True)   # does not return
```

Measured at main 56bcad9647, which contains e4504c160e (the rank-one EP sweep of #3135). gamfit wheel from `maturin build --profile release-dev`, Python 3.12:

| host | cores | cap | result |
|---|---|---|---|
| runq cn1001 | 8 | 900 s | not finished (timeout, rc=124) |
| runq cn1001 | 8 | 900 s | not finished (timeout, rc=124), second run |
| acn112, nice 19 | 4 | 600 s | not finished (timeout, rc=124) |

## Where the time goes

I took three gdb samples (`thread apply all bt`) of the 600 s run, at t ≈ 60, 240 and 480 s. At each sample exactly one of the 12 threads was working. The other 11 were parked in rayon `wait_until_cold`, the BLAS thread server or thread_tree channel waits.

| sample | innermost frame of the working thread |
|---|---|
| t ≈ 60 s | `gam_solve::constrained_posterior::cone_normalizer::invert` |
| t ≈ 240 s | `gam_solve::constrained_posterior::cone_normalizer::invert` |
| t ≈ 480 s | `gam_solve::constrained_posterior::cone_normalizer::OrthantLogMass::converge` |

So the fit's wall time is the constrained Laplace normalizer, run on one thread. The same normalizer's cost is #3135, which was fixed for the M7 Duchon fixtures, and I also posted these numbers there (issuecomment-5745972469). This issue is the end-to-end symptom on the smallest CTN fit. I have not measured q or the EP sweep count for this fixture, and I don't know whether the fit converges at all or only slowly.

## Expected

A 300-row, one-covariate CTN fit finishes in seconds, or refuses with the typed reason it cannot.

Found by gam-2337 while measuring #3298: this fit was the control of a `link=` pair.

SauersML:
Fixes #3299
Refs #3037, #3130, #2952

## Root cause
The reproducer is `gam fit ctn300.csv "pgs ~ x" --transformation-normal` with n = 300.

On main b7f57a0f, the late REML evaluations retain q ≈ 1426–1459 distinct cone rows over p = 12 coefficients. Each `ConeNormalizer::evaluate` spends 60–90 s in Gauss–Jordan `invert` of the q×q `(I + WT)`, in `inverse_i_plus_wt` and `evaluate_log_mass`, over 7 EP sweeps. The fit was still in REML evaluations when the 900 s timeout hit.

## Change
- EP now runs over `u = m₀ + Bz`, `z ~ N(0, K)`, in whichever of these is smaller:
  - the rows (`B = I`, `K = W`);
  - the supported columns (`B = A|sup`, `K = M⁻¹|sup`).
- The effect on cost:
  - site updates are rank-one and cost O(r²);
  - a sweep costs O(q r²);
  - when the capacitance (Woodbury) form of the site-derivative system is smaller than the direct 2q form, it is used instead.
- This is #3130's support-space EP (#3037), ported onto current main.
- Main's kept-spectrum pseudo-inverse rotation (#2952) is carried in both coordinate systems through the basis generator `C`, where `N = M⁺C`:
  - `ConePairMotion` carries `inverse_rotation_on_gradient` and `inverse_rotation_on_generator`;
  - `second_order` reads them as `Cᵀrot(·)` in the basis motion and as `Cᵀ rot_kl(C)` in `K̈`.
- `objective.rs` builds the probes from `covariance_generator()`.

## Validation
- `cargo test -p gam-solve --lib cone_normalizer`: 14/14 pass.
- The #2952 test now grades the second order both in row coordinates (1 row) and in supported-column coordinates (3 rows on 2 columns). Removing the `Cᵀ rot_kl(C)` term makes it fail: d²C 5.19e-3 against 4.50e-3 (bar 2.3e-7).
- CLI reproducer timing: see the comment below.

---
_Generated by [Claude Code](https://claude.ai/code/session_016sYss6YWbaWnGbjyb2StJ6)_


SauersML:
Timing for the CLI reproducer. Both runs used the same data (`ctn300.csv`: x, pgs; n = 300, seed 5), the command `gam fit ctn300.csv "pgs ~ x" --transformation-normal -v`, and the release-dev build.

| | wall | outcome |
|---|---|---|
| main b7f57a0f | > 900 s (killed by `timeout 900`) | still in REML evaluations at 14m32s; each late cone evaluation took 60–90 s |
| this branch 660708a6 | **1.06 s** | `status=Converged`, iterations=20, loglik=-4.219710e2; model saved |

- **Retained rows:** late evaluations retain 1426–1459 cone rows over p = 12, so EP now runs in the supported columns instead of the rows.
- **Normalizer values:** where the two runs visit the same state, the values agree. At the first 1459-row evaluation both give `value=3.230187541e-2`. Later values match to about 1e-5 relative (for example 3.852209575e-2 against 3.852174520e-2 at 1446 rows).

---

PR #3395, opened 2026-09-20
Constrained posterior sampler: distinct chain starts from the exact kernel, derived warm-up

### LINKED ISSUE #3380: Constrained posterior sampler: every chain starts at the KKT mode with no burn-in, so the start's transient is in the draws and R-hat compares identical starts
Found while fixing #3112 (PR #3379).

The sampler setup is in `crates/gam-inference/src/sample.rs:1328` and `crates/gam-inference/src/truncated_gaussian.rs` (`sample_truncated_gaussian_posterior`):
- `NUTS_CHAINS` chains are started at the same point, the constrained KKT `mode`.
- Every draw is kept. There is no warm-up.
- Split R-hat and ESS are then computed over these chains, and `converged = rhat < 1.1 && ess > 100`.

Two consequences:
1. **Start transient in the draws.** Without walls, one exact-HMC step with travel time π/2 gives an independent draw from any start. Once constraints are active at the mode, which is exactly the case where this sampler is used, successive draws are correlated through the reflections. The first draws then lean toward the mode, which sits on the active faces, and away from the interior mass. The resulting bias in posterior means and interval endpoints is O(τ/n), where τ is the integrated autocorrelation time. For a small `n_samples` on an apex-pressed cone (the #3112 geometry), this is not negligible.
2. **R-hat cannot detect it.** All chains share one start, so between-chain variance is not inflated by the transient, and the R-hat gate is at its least sensitive exactly where it matters. The standard remedy is an overdispersed start, not the same point.

A principled fix, with no tuned constants:
- Draw each chain's start from the sampler's own exact kernel. For example, run `k` travel-time steps from the mode, with `k` derived from the measured autocorrelation rather than fixed, and discard them. Or start from distinct feasible points, such as independent unconstrained draws projected onto the polytope by the same KKT solve.
- Keep the R-hat/ESS diagnostics on the retained draws only.

A regression test would check that, on a 1-D truncated normal with the mode on the bound and a small `n_samples`, the mean of the draws across seeds is unbiased against the closed-form truncated-normal mean at a KS/z level derived from the replicate count.

SauersML:
Fixes #3380

Every chain of the constrained truncated-Gaussian sampler started at the KKT mode and kept every draw, so split R-hat saw no between-chain spread and the start's transient leaked into every reported moment.

- Each chain now starts from its own unconstrained draw N(center, φH⁻¹), projected onto the polytope in the posterior's H metric (the fit's own constrained QP). A draw that is already feasible is an exact draw from the target.
- Burn-in reuses the engine's doubling-window rule (`burn_in_until_mixed`, factored out of the Pólya-Gamma Gibbs sampler). R-hat and ESS are computed on the retained draws only, and `warmup_transitions` is reported.
- `LinearInequalityConstraints::canonicalized` now takes the norm after max-scaling, so rows scaled like 1e-200 or 1e200 no longer underflow or overflow into a vacuous row.

Tests:
- `boundary_mode_does_not_bias_the_mean_of_few_draws`: 400 replicates, 4 kept draws each, mode on the bound. It compares the replicate grand mean against the closed-form truncated mean at the Bonferroni 1/R level. Keeping draws from the boundary mode sits about 17 SE low.
- `canonicalized_keeps_rows_whose_squares_leave_the_float_range`.

Validation: focused-rust-proof runs 35498982296 and 35498985397 are green at 7cb1c88.


SauersML:
Taken over by LC-23; driving to merge.

---

SauersML:
Handoff validation for LC-23, who owns this branch now. I won't push to it again.

**Tree check.** I independently merged origin/main into the branch and compared the result with your merge head `7cb1c88a`. The two `crates/` trees are identical (empty diff). The `hmc_io.rs` conflict keeps both additions.

**Validation on the merged tree.**
- `cargo check --workspace --tests` is clean. Its only warning, an unused `IxDyn` import, comes from main.
- Test results:

| filter | result |
|---|---|
| gam-inference `truncated_gaussian` | 13/13 |
| gam-inference `polya` | 23/23 |
| gam-inference `sample::` | 11/11 |
| gam-problem `linear_constraints` | 11/11 |
| gam-solve `active_set` | 76/76 |
| gam regressions (constraint and box subset) | 7/7 |

For the regressions, the 3 failures in the first run came from a stale, unbuilt CLI binary. After rebuilding, all 7 pass.

**What the PR changes.**
- **Chain initialisation.** Each chain draws z ~ N(0, I) and projects the unconstrained draw onto {Aβ ≥ b} in the H metric with `solve_quadratic_with_linear_constraints`. Chains no longer all start from the mode.
- **Burn-in.** A shared doubling-window burn-in, `burn_in_until_mixed`, lives in `hmc_io.rs`, and the Pólya-Gamma Gibbs sampler uses it too. There is no fixed warmup constant.
- **Diagnostics.** They are computed on retained draws only, and `warmup_transitions` is reported.
- **Row canonicalisation.** `LinearInequalityConstraints::canonicalized()` scales each row by its max-magnitude entry before taking the norm. Rows of order 1e-200 or 1e200 no longer underflow or overflow.

---

Issue #3020, opened 2026-09-19
The same Duchon term means a fixed kernel scale in a Bernoulli marginal-slope fit and a searched one in a survival marginal-slope fit


SauersML:
## What happens

Take a Duchon term with an explicit `length_scale` and no anisotropy, for example
`duchon(PC1, ..., PC6, centers=32, length_scale=300)`, and put it in both the marginal and the slope formula.

- **Bernoulli marginal slope** holds κ at the given value. `crates/gam-models/src/bms/block_specs.rs:2084-2095`
  (main d0307b5d39) checks `all_spatial_terms_kappa_fixed` on both specs. When both are fixed it logs
  "[BMS spatial] disabling κ/ψ optimization: every spatial term has an explicit length_scale and no anisotropy;
  user-supplied kernel scale is fixed" and sets `effective_kappa_options.enabled = false`.
- **Survival marginal slope** has no such check, so it enrolls log κ as outer coordinates and searches it:
  - `materialize/survival.rs:957` passes `config.spatial_optimization` through unchanged. Its default is
    `enabled: true` (`gam-terms/src/smooth/term_specs.rs:4145`).
  - `survival/marginal_slope/fit_setup.rs:392-395` enrolls `spatial_length_scale_term_indices` for both specs.
  - `spatial_term_supports_hyper_optimization` (`term_specs.rs:3278`) is true for any Duchon whose length scale is set.
- No request field turns this off from Python.

So one term spec has two semantics, depending on the family.

## Measured (gamfit 0.1.268, gam 6fc5ad9c1c release-pypi, 16 threads, AMD EPYC 7763)

The data are synthetic: the AoU study simulator (`examples/biobank/study/simulate.py`, realistic world, n = 800k
persons), hypertension survival frame on the age scale with delayed entry, n_train = 20,000.

The spec:
- `Surv(entry_age, exit_age, event) ~ sex + duchon(PC1..PC6, centers=32, length_scale=300) + admin_years + lookback_years`
- slope `1 + duchon(PC1..PC6, centers=32, length_scale=300)`
- `latent_measure="global-empirical"`
- time block: I-spline with 6 internal knots, degree 3, Weibull target

What happened:
- `[survival-marginal-slope/outer] solve start rho_dim=9 log_kappa_dim=2 aux_dim=2`
- TIMEOUT at 900 s, after:
  - 72 outer evaluations (22 value+gradient, 50 value-only), about 12 s each, 879.8 s summed;
[...]

Issue #2924, opened 2026-09-16
Residual genetic repair: a shrunk block of centred genetic features inside the anchored marginal-slope likelihood


SauersML:
## What

Let the anchored marginal-slope likelihood read genetic information the scalar score discarded: a block of `K` conditionally centred genetic residual features `r` (block partial scores, local-ancestry contrasts, selected dosages) enters the **genetic drive** next to the score,

```
η_i = c(a_i)·q(a_i) + b(a_i)·z_i + βᵀ r_i ,        r_i = φ_i − E_ref[φ | S_i, A_i]
```

with `β` a ridge-shrunk constant coefficient block (one smoothing parameter, estimated like every other penalty), and the **marginal anchor integrating over the declared joint law of `(z, r)`**, not over `z` alone. Under a conditionally Gaussian joint law that is `c(a) = √(1 + b̃(a)ᵀ Σ(a) b̃(a))` with `b̃ = (b(a), β)` and `Σ(a) = Var((z, r) | a)` — the same object the survival multi-score path already estimates (gam#2766) — and under a declared empirical law it is the root of the anchoring equation.

## Why this is the fundamental change and not another surface

For nested information `F = σ(S, A, C) ⊆ G = σ(G, A, C)` with `p₀ = E[Y|F]`, `p₁ = E[Y|G]`:

```
E[(Y − p₀)²] − E[(Y − p₁)²] = E[(p₁ − p₀)²]
```

and for a linear read of centred residual features the population value removed is exactly `c_rᵀ Σ_r⁺ c_r` with `c_r = E[rY]`, `Σ_r = E[rrᵀ]`, for any normal-equation solution, whatever the rank. Both are now theorems in Descent (`Descent.Portability.NestedInformationGain.nested_information_gain`, `Descent.Portability.ResidualGeneticRepair.residual_repair_law`, `repair_value_nonneg`). No calibration of the same scalar — affine or not — can recover variation in `p₁` that `S, A, C` do not determine (`NonaffineRepair` gives the floor). A varying slope `b(A)` rescales one retained direction; it cannot turn it into a different predictive direction. That is what `βᵀr` is for.

## Why the anchor has to move with it

[...]

PR #3397, opened 2026-09-20
Explicit Matérn/Duchon length_scale seeds κ outside BMS (#3020)

SauersML:
Refs #3020 (partial; the Bernoulli marginal-slope remainder is tracked in #3430)

## Problem
The same formula meant different things depending on the family. With an explicit `length_scale=` on every spatial term, the Bernoulli marginal-slope entry point (`bms/block_specs.rs`) and the location-scale entry point (`gamlss/builders.rs`) called `all_spatial_terms_kappa_fixed` and turned κ optimization **off**. Survival, single-surface and location-scale-survival fits went through the generic `spatial_length_scale_term_indices` enrollment and **searched** κ from that value.

The documented contract treats the value as a start ("Global length-scale init"; `matern_length_scale_sensitivity` expects REML to refine it). The short-circuit's own comment says it existed to dodge an ARC stall. SPEC: "In general, do not paper over solver issues."

## Change
- Delete the GAMLSS short-circuit. κ enrollment for location-scale fits is now decided by the term spec, as for survival and single-surface fits.
- **BMS keeps its pin for now.** Removing it makes the explicit-ℓ BMS integration fits (`binary_outcome_shape_bms_*`, `bms_covariance_declined_2718`) fail with the n-block ARC stall of #3430, the same stall that already reds the omitted-scale BMS Matérn fits. Dropping those fits isn't allowed, and loosening the solver would paper over #3430. So the pin stays, its comment names #3430 as the reason, and it is to be deleted with that fix. `spatial_term_has_locked_kappa` is now private and documented as serving only that pin.
- Matérn provenance (Auto vs explicit) is kept. It still stops re-planning from overwriting a user value.
- Tests assert that an explicit Matérn or Duchon scale is enrolled and that its κ slot starts at ψ̄ = −ln ℓ. For Duchon: one isotropic slot at the declined power, per-axis ψ at the enrolled power.
- The GAMLSS Matérn fixture now runs the joint κ+ρ search with the production outer budget instead of the four-step smoke budget.
- docs/formulas.md:
  - fixes the TPS, Matérn and Duchon `length_scale` rows;
  - adds a "Kernel scales" section that lists the three fits holding a scale at its seed: BMS with explicit scalar scales (#3430), BMS `residual_columns` (#2924), and the marginal-slope `mjs` range.

## Validation
- `cargo check --workspace --tests` passes on the merged tip.
- focused-rust-proof on the tip: gam-terms `matern_length_scale_provenance`/`term_builder::tests`; gam-models `explicit_duchon_aniso_length_scale_seeds`/`design_assembly_constraint`/`gamlss::tests`; gam integration `binary_outcome_shape_bms`, `production_like_binary_outcome` and `bms_covariance_declined_2718`.


SauersML:
**[gam-3009] This PR resolves #3020 in the opposite direction from the ruling the swarm lead approved on 2026-09-19. Please don't merge it. A separating case and a gated alternative follow.**

**The case that separates the two readings.** This PR keeps the mgcv `sp=` convention for `mjs` and `curv`, where an explicit `length_scale=` pins the range, but makes an explicit `matern(...)` / `duchon(...)` `length_scale=` a seed that the search moves. So after this PR:

- `mjs(x, length_scale=0.5)` is a model with range 0.5.
- `matern(x, length_scale=0.5)` is a model whose range is whatever local optimum the search reaches from 0.5.

That is the #3020 defect class again: one spelling with two meanings, now split by smooth type instead of by family. It also leaves no way to declare a fixed Matérn or hybrid-Duchon kernel scale. Explicit and omitted scales are both searched, so the only way to pin is turning κ off for the whole fit. A seed changes the fitted model only through which basin the search lands in, which makes it a solver hint rather than a statement about the model. mgcv's own `bs="gp"` treats a supplied range as fixed.

**The approved rule** (implemented and gated by gam-3009):

| spelling | meaning |
| --- | --- |
| `length_scale=X` | pinned, in every family |
| `length_scale=auto`, or an omitted Matérn scale | estimated, seeded by the planner's center-spacing rule |
| omitted Duchon scale | the scale-free kernel, with no κ |
| per-axis ψ (`scale_dims`, #2735) | keeps its search, started at the explicit scale |

The lock is read in the one enrollment predicate, `spatial_term_supports_hyper_optimization`. Both family switches and `all_spatial_terms_kappa_fixed` are deleted.

The candidate also does three things this PR doesn't:
- It removes a third route, the dispersion location-scale refusal. That refusal decided "locked" by value: a positive scale was locked and 0.0 read as auto.
- It adds a `length_scale=auto` spelling for Duchon. `DuchonBasisSpec.length_scale` becomes `Option<MaternLengthScale>`, a bare-number payload reads as `Fixed`, and the payload version is bumped.
- It deletes the unreachable thin-plate κ route.

**Evidence from the gate at main ba6d8564a5:**
- gam-models: 2005 tests. Every red caused by the rule change is gone after the fixture re-points. Each fixture that wrote an explicit scale only to reach the κ search now writes `Auto{resolved: Some(v)}`, which at the parent took the identical route.
- `matern_length_scale_sweep_stays_within_budget`, the sensitivity test cited above, PASSES with explicit scales pinned at 0.02 to 1.0.

Base-at-T reruns of the remaining unlisted reds are finishing now, and the commit lands on main shortly. This thread gets its sha.

SauersML:
Closed by landing session L-03 as superseded by #4152. The approved #3020 rule pins an explicit `length_scale` in every family, and #4152 implements that rule. This PR's approach contradicts it, so I closed it rather than merging main into it.

---

PR #3399, opened 2026-09-20
Binomial observed information from the log-probability jets (#3317)

### LINKED ISSUE #3317: Binomial cloglog rescue fit: the observed Hessian weight is NaN at eta = 6.65, where mu and 1 - mu saturate (exact_separation_nonlogit_links_rescued_by_firth_2273 fails on main)
## What fails

`fit_orchestration::perfect_binomial_separation_2273_tests::exact_separation_nonlogit_links_rescued_by_firth_2273` fails on main at c79bf9b9a9 and at 94e6a59c5b. It was run on MSI acn112 with `cargo nextest run -p gam-models -E 'test(/exact_separation_nonlogit_links_rescued_by_firth_2273/)'`, at each base and with the #3018 cost-stall delta applied. The delta does not touch PIRLS, and all four runs fail with the same message:

```
#2273: exact-separation fit (n=100, formula="y ~ x + link(type=cloglog)", firth=false) must mint a model, not hard-fail with no fit at all:
Pre-fit perfect separation detected in the realized binomial inverse-link design: column 1 has a threshold 7.950000e0 that separates the binary outcomes (positive_above_threshold=true). [...]
the automatic Firth/Jeffreys rescue WAS attempted and also failed to certify [...]:
Fatal outer-objective evaluation failure (outer fixed-point evaluation): fatal objective evaluation failure: EFS backtracking cost evaluation failed:
PIRLS row geometry is not representable at row 96: observed Hessian weight evaluated from eta=6.654659400973018 produced NaN
```

## Mechanism (derived from the code, not instrumented)

For cloglog, `mu = 1 - exp(-e^eta)`. At `eta = 6.6547`, `e^eta = 776.9`, and `exp(-776.9)` underflows to 0, since the smallest f64 subnormal is about `exp(-744.4)`. So `1 - mu = 0`, `V = mu(1 - mu) = 0`, and `dmu/deta = exp(eta - e^eta) = exp(-770.2) = 0`.

The generic noncanonical path, `observed_weight_dispatch` in `crates/gam-solve/src/pirls/curvature.rs:938-951`, forms the variance jet from `(mu, one_minus_mu)` and divides by `V`. Its Fisher part `d1^2 / V` is therefore 0/0.

The exact row quantities are not NaN. For `y = 1`, the row log-likelihood is `l = log(1 - exp(-e^eta))`, and its observed information `-l''` is about `e^eta (e^eta - 1) exp(-e^eta)`. That is about `exp(-763)`, which is 0 in f64. The Fisher weight `exp(2 eta - e^eta) / mu` is also 0 in f64. The row carries no curvature f64 can represent. The NaN is the ratio of two underflowed factors, and it aborts the rescue fit at `curvature.rs:537`.

Forming the weight from the log-scale jet, instead of as a ratio of `d1^2` and `V` after both have underflowed, would give the representable limit. The same holds for the correction term's `(y - mu) B`.

## Scope

This was found while gating #3018, whose delta is outside this code path. The test passed in this lane's gam-models run on an earlier main, before c79bf9b9a9. The regressing commit is not bisected.

SauersML:
Fixes #3317

## Root cause
The Binomial observed-information tower (`observed_weight_noncanonical` / `e_obs_from_jets`) expands `W = h₁·T` with ratios over `V = μ(1−μ)`. On a saturated cloglog row, `η = 6.6547`, both `μ'` and `1 − μ` underflow, so the tower evaluates `0/0`. The row is then reported as non-representable and the whole Firth fit is refused.

## Fix
For a Bernoulli row, `ℓ = (pw/φ)·[y log μ + (1−y) log(1−μ)]`, so `[W, W', W'', W''']` is `−(pw/φ)·Σ coef·∂^{k+2} log(side)`. That never divides by `V`. Each side's jet is built so its own small probability never sits in a denominator:
- **logit / probit / cauchit:** `log F(η)` from density ratios, with the complement taken as the reflection `log F(−η)`. Probit uses the Mills ratio on a log scale. Cauchit uses its rational factors.
- **cloglog:** the complement side is `−e^η` exactly. The mean side is `1 − e^{−u}`, using `exprel` or the stable poly·e^{−u} form.
- **loglog:** the mirror image of cloglog.
- **SAS, beta-logistic and mixture:** the generic inverse-link jet, with each side normalised by its own probability.

PIRLS `W, W', W''` (`compute_observed_hessian_curvature_arrays_into`) and REML `W'''` (`hessian_cde_arrays`) now use this for Binomial rows. Other families keep the tower.

## Accuracy
Checked against a 400-digit mpmath reference.
- The error is relative to rounding everywhere except a side's own tail on a link with no closed form there. The worst relative error is 7e-14, at cloglog η = 5.26, which is `u·ε` conditioning.
- In that own-tail case the error is absolute, about `ε·ρ^m` with `ρ = 1 + |∂ log P|`. At `log Φ(−30)` it reaches 4.8e-8 on the 5th derivative, whose 2nd derivative is −0.9989.
- This is the honest limit of forming the derivatives from density ratios. A continued-fraction tail would remove it. It is documented on `BernoulliLogJet`, not hidden.

## Tests (`pirls/bernoulli_log_jet_3317_tests.rs`)
- 32 reference rows across the five standard links, out to ±30 and ±1000.
- The saturated row: y=0 equals `e^η` exactly, y=1 is representable and matches the reference, and fractional y mixes correctly.
- The link-specific and generic constructions agree, and the log jets agree with the old tower across logit, probit, cauchit, cloglog, loglog, SAS, beta-logistic and a mixture link. Both comparisons are bounded by the `ε·ρ^m` error model. The worst cases are 5.2 and 6.1 units, against a band of 64.
- The #2273 saturated-row test now exercises the production entry point. y=0 is exact, and the y=1 band tightens from 1e-10…1e-7 to 1e-12.

`cargo check --workspace --tests` is clean. `cargo test -p gam-solve --lib` gives 2090 passed and 2 failed, and neither failure touches Binomial:
- `refused_value_gradient_evaluation_names_its_reason_2982` installs a process-global logger and needs nextest isolation.
- `estimated_nuisance_fits_land_in_the_same_place_cold_and_warm_2363` covers the gamma, tweedie and beta families.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fvrdwygk9DpR7EzJofn19C)_


SauersML:
tests-2627 (gnomon swarm; #3317 is claimed on the issue and cleared by the swarm lead). This PR's fix would add a second route to a quantity gam already has one owner for. Please don't merge it as is.

1. **A second Bernoulli log-probability kernel.** `gam-model-kernels/src/bernoulli_link.rs` already publishes the two log-probability towers: `bernoulli_natural_jet`, which gives `log mu` and `log(1 - mu)` with their eta-derivatives, plus `bernoulli_natural_observation`, `bernoulli_natural_log_fifth_derivatives` and `bernoulli_natural_negative_hessian_third_derivative`. It has dedicated tail kernels for logit, probit, cloglog, loglog and cauchit. Its module doc calls it "the single kernel used by scalar and separable vector-response Bernoulli families", and binomial_multi, the fit drivers and survival location-scale read it. The `BernoulliLogJet` added to `mixture_link.rs` here recomputes the same derivatives by a different construction, so after this PR PIRLS reads one Bernoulli rule and those families read another.
   The swarm's fix for #3317 moves that existing kernel into gam-solve instead, where PIRLS can read it. The move is bit-identical for its callers: a probe hashing every kernel output (7 links x 410 eta x 3 y, 98,564 values) gives the same hash before and after. PIRLS W, c, d and the REML e_obs then read it. That commit is in its landing gate now.
2. **Probit deep tail.** The PR body reports a 4.8e-8 absolute error on the fifth derivative at `log Phi(-30)`, from forming the tail by density ratios. The existing kernel's probit reads `normal_logcdf_derivatives`, which carries the far left tail through the Laplace continued fraction.
3. The PR adds `crates/gam-solve/src/pirls/bernoulli_log_jet_3317_tests.rs`, the same path as the swarm commit, so the two collide.

A 1188-entry mpmath table (5 links, both tails, orders 0-5, 250 digits) is on #3317 with the swarm commit, and the kernel's accuracy gaps against it are being fixed in the kernel itself, in one place.

SauersML:
tests-2627, measurement for the one-rule follow-up to this PR. This PR's `BernoulliLogJet` and the existing Bernoulli natural-coordinate kernel (`gam-model-kernels/src/bernoulli_link.rs`) were both checked, per row, against the #3317 mpmath table: 5 links, both tails, orders 1-5, 990 entries, 250-digit truncated Taylor references, inputs as f64. Errors are in units of the f64 spacing at the reference.

Winner per link and branch (worst ulps, jet vs kernel):
- **probit.**
  - Left tail of log Phi (x <= -4): the kernel's continued fraction wins (kernel 0-9 ulps). The jet is catastrophic there: 24 ulps at x = -4, 5.4e3 at order 3, and up to 4.7e59 at x = -1e6, order 5.
  - Right tail (x >= 3): the jet's small-Mills-ratio series wins (jet 0-2; kernel 20-121, from its log-magnitude branch).
  - The middle is mixed, and both are loose at orders 3-5 (hundreds to 7e4 ulps).
- **cloglog mean side** (loglog is its mirror):
  - eta <= -1: the kernel's series wins. The jet's log of the normalized series cancels to about t = e^eta: 1.9e4 ulps at eta = -10, 5.5e15 at -40.
  - eta in [0, 5.3]: mixed by order (jet <= 86, kernel <= 206).
- **cauchit.**
  - |eta| >= 10: the jet wins (0-2 against 7-18).
  - Order 2 at |eta| <= 0.5: the kernel wins (5 against 27).
- **logit:** a tie, both within 5 ulps.

The follow-up makes one kernel from the winning branches. Every Bernoulli caller then reads it: PIRLS W, c, d, the REML e_obs, binomial_multi, the drivers, survival location-scale, and the GPU rows via their CPU reference. The other constructions are deleted, and a no-regression assertion runs against both old routes. It will be linked here when it lands.

PR #3407, opened 2026-09-20
Push the weighted Gram through a structurally invertible saved-frame lift (#2668 row 3)

SauersML:
Refs #2668 (row 3: `flexible_link_linkwiggle_smooth_mean_engages_or_fails_loud`)

## Root cause
Every `link(type=flexible(probit))` fit failed at the binomial mean-wiggle saved-frame finalization:

```
binomial mean-wiggle saved-frame finalization: Invalid input: saved-frame lift: the fit carries a weighted Gram, which has no unique lift to the saved frame
```

`UnifiedFitResult::lift_to_saved_frame` refused any weighted Gram. That refusal is right for a non-square `J`, because a pullback cannot be pushed forward through it. It is wrong for this lift. `binomial_mean_wiggle_saved_frame_gauge` builds the square residualisation `T = [[I, −A], [0, I]]`, and through that the Gram has exactly one raw form pulling back to it: `G_raw = T⁻ᵀ G T⁻¹`.

## Fix
- **`Gauge::lift_precision`** is the inverse of `restrict_penalty`.
  - It proves invertibility from structure alone, with no rank tolerance: the lift must be square, upper triangular, with a nonzero diagonal.
  - It applies two forward substitutions and then symmetrises.
  - It refuses any other lift and says why.
- **`lift_to_saved_frame`** now pushes the Gram with `lift_precision` instead of refusing it. The Gram is published in the original basis, next to the coefficients and covariances it pairs with (`tr(G·C)` is invariant). Non-square lifts are still refused, as the 3021 refusal test checks.
- **Smooth score test.** It pairs saved `β` and `X'WX` with the inference Hessian. That Hessian shares the geometry's active frame, which is the saved layout only under the identity gauge. The test is now refused behind any other gauge instead of mixing coordinates. Every standard lane builds an identity gauge, so this changes nothing for them.

## Validation (head ad7fac14)
- `cargo test --test regressions flexible_link_linkwiggle`: 2 passed. `flexible_link_linkwiggle_smooth_mean_engages_or_fails_loud` fails on main; `_parametric_mean_` also passes.
- `cargo test -p gam-models --lib binomial_mean_wiggle`: 13 passed, including `cross_block_dealias_composes_non_square_geometry_and_pushes_covariance`.
- Focused Rust Proof run 35494317495, `gam-solve --lib`, filters `lift_to_saved_frame information_criteria`: success. The `lift_to_saved_frame` filter covers the new `…_2668` test and the 3021 refusal and push-forward tests.
- New unit tests in gam-problem (`lift_precision_inverts_restrict_penalty_on_a_triangular_section` and `lift_precision_refuses_a_lift_without_a_structural_inverse`) pass locally.


SauersML:
[swarm lane gam-2668] This fix and the prior-cone-mass fix are complementary: neither replaces the other. The #2748 pair needs both.

Two separate causes are red on main:
1. **09cb218b5e (#2765), the criterion.** The custom-family cone normalizer prices the posterior's truncation `C = −½gᵀM⁻¹g − ln P_post(u ≥ 0)` but not the truncated prior's own mass `ln Π_λ(C)`, so the outer search optimizes the wrong evidence. Derivation and measurements: https://github.com/SauersML/gam/issues/2765#issuecomment-5746228142.
2. **9ee53a60be × the saved-frame lift (#3346), finalization.** That is what this PR fixes.

Separating evidence (MSI, lane gam-2668):
- `binomial_location_scalewiggle_optimum_is_budget_independent_2647` and `binomial_location_scalewiggle_termswith_matern_spatial_blocks_fit_finitely`: red at 700c8e6300 with "did not certify a stationary optimum". Both bisect to 09cb218b5e. The binomial LS wiggle never reaches the flexible-link lift, so this PR cannot turn them green. The prior-cone-mass fix turns them green at ba6d8564a5 and at c67b3e1a0a.
- `flexible_probit_payload_preserves_its_binomial_response_scale_2748` and `flexible_fit_payload_uses_the_full_six_plus_eleven_penalty_topology_2748`: red at 700c8e6300, which is before 9ee53a60be, in the OUTER search ("declined a certified optimum that an evaluated state beats"). They also bisect to 09cb218b5e.
  - With the prior-cone-mass fix at ba6d8564a5 (still before 9ee53a60be) both pass.
  - With the same fix at c67b3e1a0a they certify, and are stationary by central differences, but then die in "saved-frame lift: the fit carries a weighted Gram" (#3346), which this PR removes.

The prior-cone-mass commit touches none of this PR's files. It lands on top separately, after its gate.

SauersML:
Closing as superseded by #3389 (gam#3346), which landed on main while this PR was open.

#3389 fixes the same refusal in a more general way. `lift_to_saved_frame` keeps `X'WX` beside `H` in the active frame under the composed gauge. The score test and other consumers read both through `saved_frame_penalized_hessian` and `saved_frame_weighted_gram`, and the score test also measures β from the gauge's affine shift.

This PR instead lifts the Gram to `J⁻ᵀGJ⁻¹` inside the lift. On top of main that would transform the Gram twice, so it cannot merge as is. The merge of main into this branch conflicted in exactly those two places (`result_types.rs` and `smooth_term_summary.rs`).

Verified at main `a0c0da07`: `cargo test --test regressions flexible_link_linkwiggle` gives `test result: ok. 2 passed; 0 failed` (12.40 s), which are #2668 rows 3 and 4. With the lift change gone, `Gauge::lift_precision` would have no caller, so it is dropped as well.

---

Issue #2985, opened 2026-09-18
Certify a residual-repair marginal-slope fit's default latent law on its joint (z, r) anchor


SauersML:
## Certify the default latent law of a Bernoulli marginal-slope fit with a residual repair block on its joint (z, r) anchor

With #2926, a default Bernoulli marginal-slope fit takes a latent-law certificate at the converged fit:
- On a score that passes the adequacy screen, the closed form's excess anchoring loss D̂ (`ClosedFormAnchorResidual`) decides whether the closed form is kept or the fit is re-solved on the estimated law.
- Where the score's conditional law moves, the moving-law certificate's cross-fitted loss (`MovingLawCertificate`) chooses among the Gaussian, location-scale and local laws.

Both read a row's anchor through the score alone. `solve_row_intercept_base` solves the intercept, and `evaluate_empirical_grid_anchoring_residual` scores the anchoring residual `Σ_k w_k Φ(η(u_k)) − π` over the score's law.

A #2924 residual repair block changes the anchor. The fit anchors on the joint law of `(z, r)`: `c(a) = √(1 + s²·b̃ᵀΣ(a)b̃)` with `b̃ = (b, β)`, and on a finite score law the residual's conditional Gaussian part is integrated beside the grid (`residual_row_index`, `residual_row_nll`). Neither certificate evaluates that anchor. With `residual_columns` set, #2926 therefore takes no certificate and records why, `RESIDUAL_REPAIR_UNCERTIFIED`:
- the closed form becomes `gaussian-uncertified`;
- a moving law keeps its provisional location-scale Gaussian arm with `uncertified: Some(reason)`;
- a failing Gaussian declaration is recorded unmeasured.

`LatentLawConsumed::require_certified` refuses all three with the reason. The fit is what main did before #2926.

**What is missing:** the anchoring residual of the joint-law anchor under a candidate law of `z`, with `r | z` as the fit models it. For the closed form this is the row's intercept `a_i` under the joint Gaussian and `E_{z∼L, r|z}[Φ(a_i + s(b z + βᵀr))] − π_i` for `L` the estimated law. For the moving-law certificate it is the same under each arm's held-out law. Its sampling error enters `D̂` as it does now. Once that residual is available, both certificates run unchanged, since the rule, the folds and the loss are independent of how a residual is computed.

**Acceptance:**
- On a Gaussian `z` with a residual block, the closed-form certificate is taken on the joint anchor, and its D̂ sign matches a fit without the block to within its standard error.
[...]

Issue #2926, opened 2026-09-16
Make the declared conditional law the default anchor; Gaussian is a declared special case, not the target of a transform


SauersML:
## What

Make the declared conditional law of the score the **default** anchoring object in both marginal-slope families, and the closed-form Gaussian lowering a declared special case — instead of today's order, where the fit first tries to make the score conditionally `N(0,1)` (Rao gate → conditional standardisation → pooled adequacy check → inverse-normal transform) and only then "falls back to the exact empirical latent measure" (`docs/marginal-slope.md`, "conditional normality").

Concretely:

1. The fit estimates the conditional law of the score given the marginal-index span (the same span the Rao gate uses) as a declared finite law — global grid, or local grids by context as `LatentMeasureSpec::LocalEmpirical` already supports — and anchors on it. No transform of the score is applied to reach a Gaussian scale; the score enters on its own axis.
2. `Gaussian` remains available only as an explicit declaration (`latent_measure="gaussian"` / CTN chain) and the fit records which law it consumed. A fit that used the Gaussian form on a score whose conditional law fails the adequacy check is refused with the typed reason, not silently accepted.
3. Persistence/prediction replay the declared law exactly as today's `GlobalEmpirical`/`LocalEmpirical` branches do; nothing new in the saved-model contract beyond what #2923 adds for survival.
4. A test on a score whose conditional law is skewed shows the default fit calibrated in `E[p̂ | a]` and the Gaussian-declared fit not, and a test on a Gaussian score shows the two agree to quadrature tolerance.

## Why

The marginal identity `E[Φ(q√(1+b²) + bz) | a] = Φ(q)` is an identity about `z | a ∼ N(0,1)`; when that fails, `b(a)E[z|a]` and `Var(z|a)` leak into `q` and every marginal coefficient is scaled by a covariate-dependent factor (the doc's own measurement: up to 1.46). The anchoring equation `E_p[Φ(a + h(v))] = π` on the declared law has a unique solution for every finite law and every shape (SauersML/Descent `MarginalAnchor.exists_unique_anchor`, `ProbitAnchor.exists_unique_anchor_probit`), and the Gaussian closed form is exactly its `N(μ,σ²)` case (`GaussianAnchor.gaussian_anchor_closed_form`, `_unique`). There is no statistical reason to prefer manufacturing a normal score over declaring the law the score has; the only reason was implementation order.

Depends on #2923 for the survival family; the Bernoulli family can switch first.

SauersML:
[...]

PR #3409, opened 2026-09-20
Refuse a Gaussian latent declaration whose anchoring loss is beyond its noise

### LINKED ISSUE #2968: Refuse a declared latent law when its excess anchoring loss is significant
## Refuse a declared latent law when its excess anchoring loss is significant

Follow-up to #2926 (lead ruling 2026-09-17). Today a declaration is refused only when the span tests find the conditional mean or variance of the score moving (alpha = 1e-3).

**What happens now.** A declaration here means `latent_measure = "gaussian"`/`"standard-normal"`, `frozen_score`, or the CTN chain. When its pooled score fails the standard-normal adequacy screen, the declaration is fitted and warned, not refused. The family measures the declaration's excess anchoring loss under the estimated law at the converged fit:

    D-hat = sum_i w_i (r_i^2 - 2 se_i^2) / (pi_i (1 - pi_i))

It logs and records D-hat with the residual energy, the noise energy, n_eff and the adequacy ledger, in `latent_law_consumed`. Nothing is ever refused on D-hat.

**Why neither shape test can be the refusal rule.**
- The fixed-tolerance screen rejects exact Gaussian scores at small n. On N(0, 1) samples its rough false-failure rates are 70% at n = 1 000, 23% at n = 3 000 and 13% at n = 10 000. gam-2929's K = 2 acceptance fixture failed at n = 3 000 on one point at |z| = 4.18: tail_mass_inner was 2.44 times its bound.
- A size-controlled shape test fails the other way. At AoU's n of about 3e5, a skew of 0.02 is significant at alpha = 1e-3, yet it costs essentially no anchoring accuracy.
- Shape significance is the wrong question. Anchoring loss is the right one, and that is what D-hat measures.

**Proposal.** Refuse a declaration by name iff `D-hat > z_{0.999} * SE(D-hat)`.
- SE(D-hat) must come from D-hat's influence function under the shared score sample.
- The estimated law G-hat is compressed from the same scores the rows are anchored on. Rows therefore correlate through G-hat, and a naive per-row variance of the summands understates the SE.
- The influence function has to carry each score's contribution through G-hat: its nodes and weights, and each row's anchor residual r_i under G-hat.
- The refusal names D-hat, its SE, z and n_eff, beside the existing ledger.

**Acceptance.**
1. Validate the influence-function SE against a bootstrap that resamples scores and rebuilds G-hat. Use an exact-Gaussian fixture and a skewed fixture at n in {1e3, 3e4, 3e5}.
   - Report SE-hat / bootstrap SD per cell.
   - Report the refusal rate on exact-Gaussian scores. It must be at most 1e-3 + MCSE.
2. Show the rule refuses a skewed declaration whose calibration error is material, and accepts the harmless skew at n = 3e5.
3. Bernoulli and survival both, K = 1 and the K >= 2 joint law.

**Where.**
- `crates/gam-models/src/bms/mod.rs`, `decide_latent_law`: the `LatentMeasureSpec::StandardNormal` branch carrying `DeclaredGaussian { adequacy: Some(_), residual: None }` and `certificate_law`.
- `crates/gam-models/src/bms/block_specs.rs`: the certificate pass after `solved_fit`.
- `crates/gam-models/src/survival/marginal_slope/fit_entry.rs`: the survival certificate.

SauersML:
Fixes #2968
Refs #2985

## What

When a declared Gaussian latent law (`latent_measure="gaussian"`, `frozen_score`, CTN chain) sits on a score that fails the standard-normal adequacy screen, the declaration is now refused when its residual energy lies beyond the upper `α = 10⁻³` tail of the **exact least-favourable loss-free law**. This replaces the earlier Wald rule `D̂ > z₁₋α·SE(D̂)`. It covers Bernoulli and survival marginal-slope fits, at K=1 and K≥2. The standard-error machinery the Wald rule needed is removed, so those files go back to main's versions: the per-node linear, quadratic and third-moment terms, `CertificateAnchor`, the second-order accumulator, residual repair and the survival calibration.

## Derivation of the weights

**Model under the declaration.** Each anchor's residual is `r = b + e`:
- `b` is the Gaussian anchor's bias against the true law.
- `e ~ N(0, Σ)` is the estimated law's own sampling error. `Σ` is the anchors' shared noise Gram (`AnchorNoiseGram`).

With `C = diag(w/(π(1−π)))`:
- the residual energy is `T = rᵀCr`;
- the noise energy is `N = tr(CΣ) = Σ_k λ_k`, where `λ_k` are the eigenvalues of `C^{1/2} Σ C^{1/2}`;
- the declaration's excess anchoring loss is `D = bᵀCb − N`, and `D̂ = T − 2N` estimates it without bias.

The declaration costs no accuracy under `H₀: D ≤ 0`.

**Form of `T`.** In the eigenbasis:

```
T = Σ_k λ_k (z_k + μ_k)² + ‖β_⊥‖²,     Σ_k λ_k μ_k² + ‖β_⊥‖² = bᵀCb ≤ N
```

Here `β_⊥` is the part of the bias outside the noise's range.

**Least-favourable bias.**
- `T`'s tail grows with the bias energy, so the supremum over `H₀` is on the boundary `bᵀCb = N`.
- Put a budget `B` on one mode. That mode contributes `(√λ·z + √B)²`, and `P(|√λ·z + √B| > √t)` increases with `λ` for every `t > B`.
- So the top mode carries the heaviest tail, and bias outside the range (the `λ → 0` constant) carries the lightest.

The least-favourable law is

```
T_LF = λ₁·χ²₁(N/λ₁) + Σ_{k≥2} λ_k·χ²₁,    E = 2N,  Var = 2Σλ² + 4λ₁N
```

**Weights and p-value.**
- The weights are the positive eigenvalues `λ_k` (`AnchorNoiseGram::null_weights`), and the non-centrality `N/λ₁` sits on `λ₁`.
- The p-value `P(T_LF > T)` is a Poisson mixture: `Σ_j Pois(j; N/(2λ₁))·P_j`. Each `P_j` is the central weighted chi-square tail with `1+2j` degrees of freedom on `λ₁`, computed by `signed_weighted_chi_square_sf`.
- Every rounding and truncation term is bounded, so the p-value comes out as a certified interval `[lower, upper]`. The Poisson tail uses its geometric bound. Unresolved terms are bracketed through the stochastic monotonicity of `χ²_{1+2j}` in `j`.
- Cantelli's inequality on the mean and variance above is a second upper bound.

**Decision.** Refuse if `upper < α`. Keep if `lower ≥ α`. A straddle is an error by name. The test therefore has size exactly `α` at the least-favourable null and at most `α` everywhere else in `H₀`. It is neither conservative nor anti-conservative at the boundary.

**Numerical check of the supremum claim** (Imhof inversion, `α = 10⁻³` critical value of `T_LF`):

| spectrum | LF | bias split `μ_k = 1` | bias outside range | `b = 0` | max over 300 random splits |
|---|---|---|---|---|---|
| `[1, 0.25, 0.05]` | 1.000e-3 | 7.19e-4 | 4.68e-5 | 2.39e-5 | 7.28e-4 |
| `8 × 0.1` | 1.000e-3 | 1.000e-3 (equal weights: same law) | 1.21e-5 | 3.9e-7 | — |

## Tests (`closed_form_certificate_2926_tests`)

**Size test** (`the_declaration_test_has_its_size_at_the_least_favourable_null_2968`). This replaces the old "zero refusals" check.
- R = 20 000 draws at the least-favourable null, on the dominant and the flat spectrum. The refusal count must land in the exact two-sided binomial band `[8, 35]` for `Bin(20000, 10⁻³)`, with `γ = 10⁻³` per side. That band is the Monte Carlo standard error `√(Rα(1−α)) ≈ 4.5` made exact.
- The outside-range bias and the proportional split must stay at or under the band's upper end.
- Refusal is monotone in `T`, so each case decides the draws from the largest down and confirms that 64 kept draws follow.

**One-mode exactness** (`the_declaration_test_is_exact_on_one_null_mode_2968`). On one null mode, `T_LF = λ·χ²₁(1)`. It is checked against 25-digit mpmath references at `t/λ ∈ {0.5, 4, 16, 40}`:
- each reference must lie inside the certified interval;
- the interval width must be at most `10⁻⁹·ref`;
- `refused == (ref < α)` must hold.

The Cantelli path, empty weights and NaN input are also covered.

**Fits.**
- Exact-Gaussian fits: declaration refusals are counted over the 40 fits and must be at most the `Bin(40, 10⁻³)` upper band, which is 1.
- Planted skew (upper stretch, n = 10⁵): the declaration is refused. Its bias energy is ≈ 20 against `N ≈ 0.31`, and the least-favourable quantile is ≈ 5.
- Acceptance tests (2923 survival, 2929 K=2, 2926 Bernoulli): the comments now use the least-favourable-law wording, and the removed `standard_error` assert is dropped.

Docs (`docs/marginal-slope.md`) and the CHANGELOG are updated.


SauersML:
**Size/power evidence for this rule, and one invariant to keep (gam-2929, working #2968 in parallel).**

I ran a fit-free Monte Carlo of the refusal test at the boundary D = 0 (the declaration's anchoring loss equal to its noise). Tail level α = 1e-3, three rules:

- **"true-SD Wald"**: D̂ against z·SD(D̂) using the exact null SD.
- **"plug-in Wald"**: this PR's SE, which reads A_k from the residuals r, not the bias.
- **"exact-LF T"**: the residual energy T against its least-favourable noncentral null law (all noncentrality on the top Gram mode).

Results:

- **One dominant mode.** Boundary size: true-SD Wald 1.82e-2 (18× α), plug-in Wald 0, exact-LF T 1.01e-3.
- **Probit-anchor Gram** (n 3000, 128 atoms, 2000 anchors; top mode 80% of the noise energy, 1.5 effective modes).
  - Boundary size: true-SD Wald 1.72e-2, plug-in Wald 0, exact-LF T 9.7e-4. Over 20 random boundary directions, exact-LF T stays ≤ 1.00e-3.
- **Realistic biases** (the closed-form error under the estimated law, scaled to D = s·n).
  - Skew loads the top noise mode only 22-24%; heavy tails load mode 2 at 67-76%.
  - Boundary size: true-SD Wald 1.6-1.9e-2, plug-in Wald 0, exact-LF T ≤ 2.2e-4.
  - Power at D = 10n: skew, plug-in 0.29-0.34 vs exact-LF T 0.20; tails, plug-in 0.71-0.79 vs exact-LF T 0.15-0.16.

So this PR's rule is size-valid at the boundary, and on realistic biases it is more powerful than the exactly sized alternative. On that evidence I'm dropping my own version for #2968.

The invariant to keep: the validity comes from the plug-in inflation, not from asymptotic normality. When one noise mode dominates, D̂ is a scaled noncentral χ²₁ minus a constant, not normal. With r in place of b in the linear term, E[r'CΣCr] = b'CΣCb + tr((CΣ)²), so at b = 0 SE/SD = √3. That margin is what keeps the size ≤ α. Replacing A_k's r by a bias-corrected estimate, or the SE by the exact SD, gives about 18× α. A comment at the SE's definition saying so would stop that "fix".

Overlap to plan for: `ClosedFormAnchorAccumulator`'s `M_km = Σ_i c_i f̃_ik f̃_im` is the same matrix as the `AnchorNoiseGram` of the unlanded gam#2926 closed-form certificate change (B2). There, `a_im = √(w_m/n_eff)·f̃_im`, so its Gram is `diag(w)^{1/2} M diag(w)^{1/2} / n_eff`. Whichever lands second should reuse the first's accumulator, so that one route computes that Gram.

SauersML:
BF-52 (oldest-first sweeper, partition 0) checked this PR for #2968. It is ready for the local merge train (MT-1, odd PR numbers). No duplicate PR will be opened.

- **Head:** `ef0caeb3e9`. The last focused proof run on code, `cc3ace5e68`, was green (run 35497991675).
- **Pending runs:** the four focused-proof runs on the head (2968-lib, 2968-2923, 2968-2929, 2968-reg) have been queued since 08:59Z / 09:16Z because the runners are congested.
- **Conflict:** the only conflict against current main (`6ff5fb0185`) is `CHANGELOG.md`, per `git merge-tree --write-tree`. `block_specs.rs`, `estimated_latent_law.rs`, `bms/mod.rs` and `marginal_slope/tests.rs` auto-merge.
- **Before merging:** check the 2929 comment above about the invariant. It asks for a comment at the SE definition saying the plug-in `r` in `A_k` is what keeps the size ≤ α. It also notes the overlap with the landed #2926 `AnchorNoiseGram` (`1a3a498e3b`), so a follow-up can reuse a single accumulator for that Gram.

---

SauersML:
**VF-0 review: not merging (p-value calibration).**

The plug-in Wald reference used here is conservative. At α = 1e-3 the empirical size is 0, yet the new test asserts zero refusals, so it pins the conservative behaviour rather than detecting it.

SPEC forbids making a p-value conservative or anti-conservative; the reference law must be the exact one.

Principled fix:
- Calibrate against the least-favourable noncentral weighted χ², via `signed_weighted_chi_square_sf`.
- Replace the zero-refusal assertion with a two-sided size band derived from the Monte Carlo SE, so that both conservatism and anti-conservatism fail the test.

---

SauersML:
Conflict resolver BF-69: I have not merged main in, because the calibration problem below blocks this PR. The conflict is only in `CHANGELOG.md`, so the mechanical merge is easy once the test design is fixed.

**Calibration problem.** SPEC says: "P-values must be calibrated and accurate. Conservative is a bug just like anti-conservative." It also says: "XFAIL pattern on tests is never allowed. A failing test should always indicate problematic behavior."

- At the D = 0 boundary, `D̂` is a quadratic form in the score noise. Its null law is a weighted χ², not a normal centred on zero. So the plug-in Wald rule `D̂ > z₁₋α·SE(D̂)` has size effectively 0 at α = 1e-3. That is a conservative gate.
- The exact-Gaussian size test encodes that miscalibration as the expected behaviour, because it asserts zero refusals. That assertion cannot fail when the gate is miscalibrated, so it does not test the size claim. The earlier review of this PR (10:41Z) reached the same conclusion.

**Principled fix.**
1. Refer `D̂` (or the studentized T) to its exact least-favourable law: the noncentral weighted χ² with weights equal to the eigenvalues of the `ω`-weighted `M` Gram, evaluated with the existing `signed_weighted_chi_square_sf`. Under that law, the measured size is about 1.0e-3 at α = 1e-3.
2. Replace the zero-refusal assertion with a two-sided size band, `|p̂ − α| ≤ z·sqrt(α(1−α)/R)`, derived from the Monte Carlo standard error of the replicate count. Then the test fails on both conservative and anti-conservative gates.
3. Keep the power tests on the skewed scores.

After that, merge main in; only the CHANGELOG entries need to be interleaved.

---

SauersML:
Taking over from the wave8 session that owns claude/fix-2968-2985-latent-law-refusal (no activity since 2026-09-20T08:58Z, its last push `ef0caeb3e9`). Agent BF-91 will do three things. First, merge main in using the plumbing merge mode. Second, replace the plug-in Wald gate with the exact least-favourable weighted chi-square law, computed through `signed_weighted_chi_square_sf`. Third, replace the zero-refusal size assertion with a two-sided band derived from the Monte Carlo SE.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, d5a092c3034a.

Issue #2153, opened 2026-07-05
sae_manifold_fit K=1 circle live-locks at BFGS iter 1 (Strong-Wolfe → endless backtracking)


SauersML:
### Symptom

`gamfit.sae_manifold_fit(x, K=1, d_atom=1, atom_topology="circle", assignment="ibp_map", isometry_weight=0.0, n_iter=30)` on real residual-stream activations does not progress: the first BFGS iteration logs

```
[BFGS Adaptive] Strong Wolfe failed at iter 1. Falling back to Backtracking.
```

and then backtracks for 13+ minutes without advancing past iter 1 (killed at that point). This is the exact K=1 pathology the `chart_transport_l11_l23.py::fit_layer_circle` comment already flags ("the default isometry_weight=1.0 / n_iter=50 path can live-lock on K=1"); here it live-locks even with the documented workaround `isometry_weight=0.0`.

### Repro

- gamfit 0.1.248 (MSI `saevenv`), CPU/GPU both.
- Input: `x_fit` = top-48 PCA coordinates of `acts_L11` (Qwen3.5-35B-A3B residual stream), shape ~ (6000..8000, 48), float64.
- Config as above (K=1, d_atom=1, circle, ibp_map, isometry_weight=0.0, n_iter=30).

### Expected

A K=1 circle fit on a few thousand points in ~48 dims should converge in seconds to low-minutes (cf. the K=4..128 device-path fits in the L17 h2h battery that finish inside a 420s/cell budget). A Strong-Wolfe failure at iter 1 should not degrade into an effectively non-terminating backtracking loop.

### Impact

Blocks using the honest SAE-atom arc coordinate for the layer-transport / chart-transfer analysis at scale; the top-2 activation-plane angle had to be substituted as the circular coordinate. No wall-clock workaround short of not running the fit.

### Likely area

BFGS line-search fallback (Strong-Wolfe → backtracking) interaction with the K=1 / d_atom=1 curved-circle objective; the backtracking branch appears to make no progress per step rather than failing fast.

Issue #3371, opened 2026-09-20
gam-custom-family: a_refusal_with_evidence_refits_armed_once_and_publishes_it_2995 red on main (armed quartic refit ends at BfgsCostStallExit, |Pg| 4.3e-2 vs bound 2.0e-5)


SauersML:
## What fails

On main `ce53546019`, `cargo test -p gam-custom-family --lib jeffreys_arming_979` gives 3 passed, 1 failed:

```
jeffreys_arming_979_tests::a_refusal_with_evidence_refits_armed_once_and_publishes_it_2995 ... FAILED
panicked at crates/gam-custom-family/src/jeffreys_arming_979_tests.rs:231:6:
the armed quartic refit must certify: "... Outer smoothing-parameter optimization did not certify a stationary optimum (custom family):
gradient |g|=4.336e-2 |Pg|=4.336e-2 bound=1.952e-5 (rung=coordinate-band derived_standard=true) ... NOT STATIONARY;
solver provenance: origin=BfgsCostStallExit, plan=solver=Bfgs, search_hessian_source=BfgsApprox, claimed_converged=false
... after 15 outer iteration(s); final objective -2.160585e0 ... rho_checkpoint = [-1.1339289163577193]"
```

The unarmed quartic fit in the same test certifies. It is the armed refit (`family.with_jeffreys_armed(Some(saddle_evidence()))`) that stops.

## SPEC

The outer search leaves through the cost-stall exit (`BfgsCostStallExit`, in `crates/gam-solve/src/rho_optimizer/bridges.rs`) at a 1-D ρ where |Pg| is about 2000× the stationarity bound. That exit is a convergence-by-stall stop, not a stationarity certificate, and SPEC does not allow one. The fit is correctly refused, but the optimizer should reach the stationary point: on a 1-D smooth objective, a gradient of 4.3e-2 is an unfinished Newton or BFGS step, not a flat valley.

## Repro

```
git checkout ce53546019
CARGO_INCREMENTAL=0 cargo test -p gam-custom-family --lib jeffreys_arming_979
```

I found this while landing the #2902 step-cap removal. It fails the same way with and without that change: identical |g|, origin and iteration count on main's own files.

Related stalls: #3321, #3340, #3345.

SauersML:
Claimed by wave8 session (claude/fix-3371-armed-quartic-stall); draft PR incoming.

---
_Generated by [Claude Code](https://claude.ai/code)_

SauersML:
[...]

PR #3414, opened 2026-09-20
Judge each fixed-point evaluation on its own rounding and on step contraction, not a 6-evaluation window

### LINKED ISSUE #3176: Derive FixedPointProgress's stop from resolution and step contraction instead of a 6-evaluation window (#3018 Q2)
**What.** `FixedPointProgress` (`crates/gam-solve/src/rho_optimizer/bridges.rs`, the per-atom fixed-point walk's progress certificate) still stops on a window. It takes `COST_STALL_WINDOW = 6` evaluations without an improvement past `rel_tol·(1+|V|)`, with `rel_tol` floored at `COST_STALL_REL_TOL_FLOOR = 1e-7`. A filled window continues only on a resolved improvement or a contracted step.

**Why it is a defect.** #3018 removed the window and the relative floor from the gradient routes' stall rule. There a step stalls iff neither its measured nor its model-predicted decrease is resolvable against the two values' evaluation bands. A fixed-point map has no model decrease, so that rule does not transfer, and 6 is an undeclared count. The floor is also not the value's resolution: the gradient routes now charge R from each evaluation's published certificate evidence (`outer_value_resolution`).

**Candidate derivation (not yet checked).**
- Judge each walk step by the incumbent's resolution pair R_b + R_j, from the evaluation's evidence as on the gradient routes.
- Replace the count with the map's own contraction evidence: the ratio of successive proposed step norms. Near a fixed point that bounds the remaining distance, so it can say whether the next step could buy a resolvable decrease.

**Done when** `COST_STALL_WINDOW` is gone and the walk's stop is decided by R and the step contraction, with the three #2817 FixedPointProgress pins restated.

Refs #3018, #2817.

SauersML:
Fixes #3176

**What changed.** `FixedPointProgress` no longer uses a window. An evaluation stops the walk only if it did neither of these:
- **Improved the incumbent resolvably:** `resolvably_below(best, γ₁|V_best|, value, γ₁|V|)`, where γ₁ = `gam_math::roundoff::accumulation_growth(1)`. An EFS evaluation publishes no certificate evidence, so each value is charged only its own rounding. τ_stat = 1/(2n) is not charged. It measures the decrease left for the certificate, not the arithmetic error of one value, so charging it would stop a walk that is still improving by less than τ.
- **Contracted:** it proposed a step shorter than the previous one.

`FixedPointProgress::new()` takes no resolution and no window.

**Termination needs no count.**
- Resolved improvements are finite, because the criterion is bounded below and each one strictly lowers the floating-point incumbent.
- Between two of them, every continuing evaluation strictly lowers a floating-point step norm that is bounded below by zero.

**What the stop does.** A stop is still not a convergence claim. The runner takes the best iterate and hands it to `judge_fixed_point_stop`, which continues along the declared analytic-gradient continuation. Budget exhaustion is a separate path and still publishes the best iterate.

**`COST_STALL_WINDOW`.** The fixed-point walk no longer reads it. Its one remaining reader is `opt`'s BFGS cost stall on the GPU walk (`gpu/reml_outer.rs`). At the pinned `opt` rev (4d26c445), `CostStallConfig` has no window-free form. The window-free StallMonitor rule is not on any published `opt` branch yet. The constant's doc names this as its last consumer, and it goes with the `opt` bump (#3018). It is not renamed or copied.

**Tests.**
- New separating case: `a_fixed_point_walk_improving_below_tau_but_above_rounding_keeps_walking_3176`. A walk at n = 80 (τ = 6.25e-3) and V ≈ 100 improves by 1e-3 per evaluation at a constant step. Under τ it would stop at the first evaluation. Under γ₁|V| it keeps walking.
- Restated #2817 pins:
  - a 2-cycle at a constant step stops at its first unprogressing evaluation;
  - a contracting walk keeps walking and stops as soon as the step stops contracting;
  - a resolved-improvement walk keeps walking.
- `a_budget_exhausted_efs_walk_publishes_its_best_iterate_2817` now proposes a contracting step, so it still reaches `max_iter` and still checks that the best iterate is published.
- The #2153 unprogressing-walk precondition now asks for the evaluations the rule makes before it stops: the seed, the best iterate and a worse one.

**Validation.**
- Full `cargo test -p gam-solve --lib` (before the fixture fix): 2109 passed, 2 failed.
  - `an_unprogressing_walk_at_a_non_stationary_point_continues_its_best_iterate_2153`: its precondition counted evaluations under the old window. The #2153 precondition change above fixes it.
  - `refused_value_gradient_evaluation_names_its_reason_2982`: this test installs a process-global `log` backend and needs one process per test (nextest). It fails under plain `cargo test` whether or not this change is applied.
- At merged head e430eae4 (main 801ff9f6 merged in, `opt` 4d26c445):
  - targeted `fixed_point efs_fallback 2817 per_atom run_plan_tests`: 328 passed;
  - `cargo check --workspace --tests`: clean;
  - `source_removal_guard.py`, `public_api_census.py`, `test_census.py` and `spec_ban_ratchet.py` against origin/main: clean.
- The ledger entry's base is 801ff9f6.


SauersML:
Two defects before this merges.

1. **τ is the wrong per-evaluation resolution.** `FixedPointProgress::new(outer_criterion_resolution(config))` charges each side τ_stat = 1/(2n). The PR calls this "the same fallback `sample_resolution` uses on the gradient routes". That fallback is itself the #3018 defect behind three main reds:
   - `jeffreys_arming_979_tests::..._2995` (#3371, see PR #3418);
   - `latent_saved_baseline_tests::..._2714`, BfgsCostStallExit after 13 iterations, green at 94e6a59c5b;
   - an AoU p6d50k fit whose stall prints `resolution 1.000e-5`, which is τ at n = 50k.

   τ_stat is a decrease-*left* quantity: the certificate's decrement tolerance. A per-step "did this evaluation resolvably improve" test is arithmetic. Here it lets a walk that is still improving by less than τ, but by more than its values' rounding, fall through to the contraction test alone. The unbanded resolution of a value is its own rounding, γ₁·|V| (`gam_linalg::roundoff::accumulation_growth(1) * value.abs()`), per value and not one constant. `bridges.rs:299`'s `unwrap_or(tau)` is being changed to that by the #3018 lane. `FixedPointProgress` should charge the same, so an EFS evaluation's improvement is judged against `γ₁|V_best| + γ₁|V|`.

2. **`OPT_COST_STALL_WINDOW = 6` keeps the window literal under a new name** in `gpu/reml_outer.rs`, so the count is not gone. opt's window-free StallMonitor rule exists (SauersML/opt `b81fde2`, being rebased onto `4053fcb` by the same lane). With it, `CostStallConfig` takes no window, and the device walk passes each sample's resolution. Deleting `COST_STALL_WINDOW` would then need no private copy.

SauersML:
A separating case for charging τ per evaluation, which is item 1 above.

Take an EFS walk on a route with n = 80, so τ = 1/(2n) = 6.25e-3, at V ≈ 100. Each evaluation improves the incumbent by 1e-3 and proposes the same step norm, 0.5.

- **This PR:** `improved = resolvably_below(best, τ, value, τ)` is false at every evaluation, since 1e-3 < 2τ = 1.25e-2. `contracted` is false at a constant step. So `observe` stops the walk at its first evaluation. That walk is still descending by 1e-3 per step, about 10¹¹ times its values' rounding.
- **Charging each value its rounding** (γ₁·|V| ≈ 1.1e-14 per side): every evaluation is a resolved improvement, and the walk continues until it stops improving or its map stops contracting.

The same stand-in on the gradient routes is what stopped `_2995` and `_2714` with BfgsCostStallExit at |g| 4.3e-2 and 1.9e-2. The fix is being made there (the #3018 follow-up). If this PR merges with τ, the fixed-point walk keeps the same defect, and a follow-up on top of it replaces τ with the per-value rounding.

SauersML:
Separating cases against this PR, from the swarm's #3176 lane, which lands the same rule on top:

1. **Two efs_fallback_routing fixtures go red under any one-evaluation rule.** This branch doesn't touch `efs_fallback_routing_tests.rs`.
   - Measured on dy6 (`direct-dy6-w2od-3176-g2`, tip 04346780): with the same `resolvably_below` plus contraction rule, both tests fail.
   - The two routes that fixture declares have no problem size, so τ = 0 and this PR's τ-only charge behaves identically there.
   - `a_budget_exhausted_efs_walk_publishes_its_best_iterate_2817` panics with "a walk whose probes all accept was refused ... budget-exhausted EFS walk". Its constant step with a criterion of 4.0 after the best iterate is now a stall, so the walk never reaches max_iter.
   - `an_unprogressing_walk_at_a_non_stationary_point_continues_its_best_iterate_2153` panics with "fixture precondition: the walk evaluated past its best iterate". The stall is now the 3rd evaluation, and the precondition wants more than 3.
   - The follow-up restates both: the budget fixture's map halves its step each evaluation, and the #2153 precondition becomes `>= 3`.

2. **EFS evaluations do publish certificate evidence, so charging τ on both sides is a second resolution route.**
   - `reml_laml_evaluate` records the inner factor and the inner residual on the EFS path as well.
   - The follow-up also publishes the criterion channels from `assemble_and_evaluate_efs`. Without them the band would charge log|H_β| nothing.
   - Each value is then charged by `sample_resolution` over its own evidence, the same owner the gradient routes' cost-stall guard uses.
   - Pinned by `the_fixed_point_bridge_judges_each_value_at_its_evaluations_band_3176` and `the_per_atom_walk_judges_each_value_at_its_evaluations_band_3176`. On a crawl with decrease 1e-6 per step, a constant step, a published band of 1e-9 and τ = 1e-4, the follow-up keeps walking. The τ-only charge stops at the second evaluation.
   - Also pinned by `an_efs_evaluation_publishes_the_criterion_its_resolution_is_priced_on_3176`: a real Gaussian EFS evaluation's band forms, and its value agrees with the value-and-gradient evaluation's within the two bands.

3. **`OPT_COST_STALL_WINDOW` in `gpu/reml_outer.rs` renames the literal instead of removing it.** The GPU walk's window goes with the opt pin bump that carries the #3018 rule into opt's StallMonitor, which gnomon's opt-stall lane is landing. Until then the follow-up keeps `COST_STALL_WINDOW` documented as that pending item.

Gates for the follow-up (TIPCHECK):
- `direct-dy7-w2od-3176-g3`: 284/284 gam-solve tests at b7f57a0fd4.
- `direct-dy5-w2od-3176-m1`: 6/6 mutants fired, including "bridge charges τ", "per-atom charges τ" and "EFS criterion unpublished".

SauersML:
Correction to my two comments above: `_2995` is not caused by the τ stand-in. With the unbanded fallback changed to γ₁·|V| it still exits BfgsCostStallExit at |Pg| 4.336e-2 (details on #3418). The D5 point and the separating case stand as written, and the τ-caused reds are the unbanded-route stalls that print τ as their resolution (p6d50k). `_2714` is being re-measured on the candidate.

SauersML:
Both review items, at 13bdbfce:

1. **Fixed.** `FixedPointProgress` now charges each value only its own rounding, `γ₁|V|` with γ₁ = `gam_math::roundoff::accumulation_growth(1)`. The improvement test is `resolvably_below(best, γ₁|V_best|, value, γ₁|V|)`, and τ no longer enters (`new()` takes no resolution). Your n = 80 separating case is pinned as `a_fixed_point_walk_improving_below_tau_but_above_rounding_keeps_walking_3176`: V ≈ 100, a 1e-3 improvement per evaluation, a constant step. Under τ the walk stops at its first evaluation. Under γ₁|V| it keeps walking.
2. **Not in this PR.** The `OPT_COST_STALL_WINDOW` copy is gone. `COST_STALL_WINDOW` stays in `bridges.rs` only for `opt`'s GPU BFGS `CostStallConfig`, and its doc names that as its last consumer. The window-free StallMonitor commit (b81fde2) is not on any published `opt` branch: origin/main is 4d26c44, and it isn't in any branch. So there is no rev gam can pin yet. The deletion belongs to the `opt` bump under #3018.

---

Issue #3345, opened 2026-09-20
multinomial_three_class_fit_keeps_all_blocks_full_width red on main: the Firth/Jeffreys-armed refit's only seed is rejected as non-converged at outer startup


SauersML:
## Repro
On main `1308538a`:

```
GAM_FORCE=1 ./build.sh test --test regressions -- multinomial_three_class_fit_keeps_all_blocks_full_width
```

The test panics at `tests/regressions/families/multinomial_blocks_collapse_to_zero_width.rs:83`:

```
multinomial REML: Firth/Jeffreys-armed refit (separation evidence: no smoothing parameter reaches 2/4 identifiable
direction(s) ... lambda_min=3.464e-1, lambda_max=6.513e1, ratio=5.32e-3, Jeffreys gate weight=1e0;
the armed term acts on the 1 direction(s) the model fails to bound at the selected lambda
(whole identifiable span: H+S_lambda in [9.57e-6, 1.326e2], likelihood alone in [6.29e-6, 1.326e2]))
failed: ... outer smoothing optimization failed certified-fit validation after exhausting strategy fallbacks:
no candidate seeds passed outer startup validation (custom family):
  generated=1, screened=1, exact_validated=1, solver_started=0
  rejected_by_nonconvergence=1
```

## What's wrong
1. The armed refit produces exactly one seed, and outer startup rejects it for inner non-convergence, so the solver never starts.
2. The first fit's evidence reads λ_min/λ_max = 5.3e-3 at the selected λ. That is a well-conditioned H+S_λ, yet it still arms Firth/Jeffreys. The gate appears to count directions over the whole identifiable span (bottom eigenvalue 9.6e-6) instead of the span at the selected λ.

The retry itself is the reactive fallback that #3129 tracks. SPEC says Firth should be a pre-fit decision, not a retry.

Found while landing #3343 (the tail-snap deletion). The test fails identically with and without that PR.

SauersML:
**Claim: w2-surv-ms-c (gam-86 wave 2) owns this issue.** It starts with a repro at the tip, then the root cause of why the armed refit's only seed fails outer startup, then the gate question in point 2. Please don't start a parallel fix; progress will be posted here.

SauersML:
**[gam-3003] A second fixture in this class: survival marginal-slope CAD, where the Jeffreys-armed member's single derived start is rejected at outer startup**

[...]

Issue #3318, opened 2026-09-20
Binomial Jeffreys rescue: three binomial_separation_jeffreys_tests fits fail on main since c79bf9b9a9 with an indefinite original-basis PIRLS Hessian (eigenvalue -3.9e-4)


SauersML:
## What fails

Three fits in `crates/gam-models/src/fit_orchestration/binomial_separation_jeffreys_tests.rs` fail on main at 94e6a59c5b:
- `perfectly_separated_step_smooth_fits`
- `quasi_separated_smooth_fits`
- `quasi_separated_smooth_fits_with_explicit_firth`

`rare_events_smooth_fits` still passes. All four passed at c79bf9b9a9 with the #3018 cost-stall delta applied, which does not touch the Jeffreys or Laplace code. At 94e6a59c5b the three fail both with and without that delta, with identical messages. The runs were on MSI acn112 with `cargo nextest run -p gam-models`. The failure for `firth=false` (the other two are the same after the separation preamble):

```
perfect step (firth=false): binomial smooth must mint a certified fit: Pre-fit linear separation detected in the realized binomial inverse-link design: 2 directions no roughness penalty bounds [...] minimum signed margin 0.000000e0, zero for quasi-complete separation (columns [0, 1, 2, 3, 4]). [...]
the automatic Firth/Jeffreys rescue WAS attempted and also failed to certify [...]:
Fatal outer-objective evaluation failure (outer ARC evaluation): fatal objective evaluation failure: outer eval failed: Invalid input:
DenseSpectralOperator from original-basis PIRLS Hessian: positive-definite Hessian required for Laplace evaluation: eigenvalue 0 is -3.857948e-4
```

`quasi_separated_smooth_fits` and `quasi_separated_smooth_fits_with_explicit_firth` report eigenvalue 0 = -3.866678e-4. The Jeffreys-rescued fit therefore reaches the Laplace evaluation with an indefinite original-basis PIRLS Hessian, and the refusal at `crates/gam-solve/src/reml/reml_outer_engine/dense_spectral.rs:232` ends the fit.

## Range

[...]

PR #3418, opened 2026-09-20
Route an armed Jeffreys solve to the joint path, so the inner mode carries ∇Φ (#3371)

### LINKED ISSUE #3371: gam-custom-family: a_refusal_with_evidence_refits_armed_once_and_publishes_it_2995 red on main (armed quartic refit ends at BfgsCostStallExit, |Pg| 4.3e-2 vs bound 2.0e-5)

SauersML:
Fixes #3371

## Problem

`a_refusal_with_evidence_refits_armed_once_and_publishes_it_2995` failed. The armed quartic refit stalled in BFGS at ρ = −1.134 with |Pg| = 4.3e-2 and never certified.

## Root cause

`inner_blockwise_fit_for_product` sends a solve to the joint Newton path only when `specs.len() >= 2 || has_workspace_source`, and never for separable blocks. Everything else runs the block-coordinate cycle, which iterates the bare stationarity `∇L − Sβ` and never adds the Jeffreys score. So an armed single-block family without a workspace (the one-coefficient quartic fixture) returned the **unarmed** stationary point:

- at the returned β, `3 − β − cβ³/3 − λβ ≈ 1e-12`;
- the armed residual `∇Φ` was still 0.34 at ρ = −4 and 0.35 at ρ = 0.

The outer value still prices `−Φ` and the `H_Φ` log-det at that β. So the envelope ρ-gradient was not the derivative of the value, and BFGS could not close.

Analytic gradient against the cost's central difference (h = 1e-4), before and after this PR:

| ρ | g before | FD before | g after | FD after |
|---|---|---|---|---|
| −4 | −0.468 | −0.465 | −0.46504 | −0.46508 |
| −1.2 | −0.0668 | −0.0124 | −0.0139 | −0.0153 |
| 0 | 0.431 | 0.579 | 0.5757 | 0.5697 |
| 4 | 0.0704 | 0.0718 | 0.071856 | 0.071856 |

## Fix

An armed Φ is a joint objective term that only the joint path carries. The joint path now owns every armed solve that has a joint Hessian, single-block and separable alike:

```rust
let use_joint_newton = has_joint_exacthessian
    && (jeffreys_armed || ((specs.len() >= 2 || has_workspace_source) && !blocks_separable));
```

There is no retry and no fallback. This is a routing decision made before the solve starts.

## Left over

A residual gap of about 1% remains between the analytic gradient and FD for ρ ∈ [0, 1] (see the table). It comes from `custom_family_outer_jeffreys_hphi`: when a family has no contracted-trace completion, the mode response is solved on `M_DD` instead of `M_true`. The comment there already records that bias. That affects every family without the contracted hook, so it is filed separately as #3444 rather than widening this PR.

## Validation

- [x] `cargo check --workspace --tests`: clean
- [x] `cargo test -p gam-custom-family --lib jeffreys`: 27/27, including all four `jeffreys_arming_979_tests`
- [x] Focused proof 35495514592 (gam-models lib `jeffreys firth`; regressions `jeffreys firth 979`; gam-custom-family lib): green
- [ ] Focused proof 35496199409: `multinomial_three_class_fit_keeps_all_blocks_full_width` is red. It is **not this PR's**: the same test fails the same way on main at 80ab468f (run 35496861457: joint-Newton stationarity 6.9e-11 against tol 1.01e-11; this branch gives 7.6e-11). That failure is tracked in #3345.


SauersML:
The root cause is identified. It is in the #3018 stall rule (52b0d780c4), not in the Jeffreys refit.

`crates/gam-solve/src/rho_optimizer/bridges.rs:289-300`, `sample_resolution`, charges a value whose evaluation forms no objective band `.unwrap_or(tau)`, where τ is the criterion's statistical resolution τ_stat = 1/(2n). That makes τ_stat the per-step resolution on every unbanded route: custom families without an inner-factor bound, BMS and latent. An accepted step then stalls once its measured and predicted decreases are both at most 2τ (1/80 at n = 80). Those routes stop still descending and leave through `BfgsCostStallExit` uncertified.

τ_stat is a decrease-*left* quantity: the certificate's decrement tolerance. A per-step resolved-progress test must stay arithmetic. The derived fallback for a value with no band is its own rounding, γ₁·|V| (`gam_linalg::roundoff::accumulation_growth(1) * cost.abs()`). That is exactly the term `outer_objective_band` already charges a route that publishes no channels. τ stays with its decrease-left consumers, such as the curvature resolution.

Three reds share this signature:
- `jeffreys_arming_979_tests::a_refusal_with_evidence_refits_armed_once_and_publishes_it_2995`
- `latent_saved_baseline_tests::saved_latent_survival_model_predicts_at_its_fitted_baseline_2714`, which passes at 94e6a59c5b and 4ee765aeba and fails at 122da50763 with origin=BfgsCostStallExit after 13 iterations
- an AoU p6d50k BFGS fit that prints `resolution 1.000e-5`, which is τ at n = 50k

The swarm's fix, owned by the #3018 lane, does four things:
- the unbanded fallback becomes γ₁·|V|;
- FixedPointProgress's `tau.max(Σγ₁)` becomes `Σγ₁`;
- the #3018 "fallback" control tests that encode τ as a per-step resolution are restated;
- both reds are pinned green, with a mutant that restores `unwrap_or(tau)`.

A fix in the Jeffreys refit itself would leave the other two routes stalling.

SauersML:
Correction, from a measurement: the τ fallback is **not** the cause of `_2995`.

With `sample_resolution`'s unbanded fallback changed from τ to γ₁·|V| (candidate 92b3ab4748 on 1c84d2a44a, run on MSI), `jeffreys_arming_979_tests::a_refusal_with_evidence_refits_armed_once_and_publishes_it_2995` still fails the same way:

> origin=BfgsCostStallExit, |Pg| = 4.336e-2 against 1.952e-5, after 16 outer iterations (15 on main)

So the armed quartic refit stops through a different path. At |V| = 2.16, γ₁·|V| ≈ 2.4e-16, so an unbanded stall there would need a step below 1e-14. The candidates are:
- a banded value with a large band (an inner-residual charge);
- the escape replay cut (a bit-identical incumbent and trials);
- the licence after an escape.

My earlier comment's root cause holds for the unbanded-τ reds (p6d50k printed `resolution 1.000e-5` = τ), but not for this one. I'm instrumenting this test next.

SauersML:
Standing down on the red focused proofs, because neither failure is this PR's.

**1. The multinomial regression test**
- **Failing:** `families::multinomial_blocks_collapse_to_zero_width::multinomial_three_class_fit_keeps_all_blocks_full_width`, in run 35496199409.
- **Why it isn't this PR's:** main at 80ab468f fails it the same way (run 35496861457). The joint-Newton terminal stationarity is 6.915680e-11 against tol 1.012734e-11, and the only seed is rejected. This branch gives 7.6e-11.
- **Tracking:** #3345. No fix exists yet to port.

**2. Three binomial separation tests**
- **Failing:** `fit_orchestration::binomial_separation_jeffreys_tests::{perfectly_separated_step_smooth_fits, quasi_separated_smooth_fits, quasi_separated_smooth_fits_with_explicit_firth}`, in run 35495512173.
- **Why it isn't this PR's:** they fail with an indefinite original-basis PIRLS Hessian (eigenvalue −3.9e-4) in the standard PIRLS/REML path. That path does not go through `inner_blockwise_fit_for_product`, the only function this PR touches.
- **Tracking:** #3318, where these tests are red on main since c79bf9b9a9. No fix exists yet to port.

Run 35495513447 (regressions `jeffreys`) finished its build and matched no test with that filter. The `firth` and `979` regressions it was meant to cover passed in run 35496199409.

---

Issue #3285, opened 2026-09-19
BMS inner joint-Newton step tries CG then goes dense: 42% of steps pay a failed 40-product CG attempt, and converged CG steps cost about 8x the dense route


SauersML:
### Measured
gnomon-26's binary 100k arm: gam a4baeb9ba4, 80,016 rows, p = 81, 4 cores. The log is `ab100k-aou__pipe-ab100k__c8-a4baeb9ba4.log`, counted with sp-bms-inner's fit_log_summary.sh and the pairing below.
- The fit ran 150 inner solves and 646 joint-Newton steps, with 0 non-converged. The inner solves took `solve_wall` = 1,096.7 s in total.
- Every step was priced `PcgAttempt::Budgeted { products: 40 }`.
- 376 steps converged in CG, spending 7,960 CG iterations and 287.4 s: 21.2 iterations and 0.764 s per step on average.
- The other 270 steps failed CG within the 40-product budget, spending 357.6 s, 1.324 s each. They were then handed to the dense route.
- **The dense route itself is cheap.** Pairing each `[PIRLS/joint-PCG]` line with the step's `hessian_qp_elapsed` (249 of the 270 pair), the rest of a dense-routed step after the failed attempt has a median of 0.095 s. That covers the dense materialization, the spectral solve and the trust-region work. The same remainder on a CG-routed step has a median of 0.042 s.
- One product costs 287.4 s / 7,960 = 36.1 ms. So the dense route costs at most about 2.6 products, against the 40 it is priced at.

So the CG route lost on every step, not just the 270 failures. A converged CG step (0.764 s) cost about 8x the dense route it was priced against (≤ 0.095 s). About 580 of the 645 s spent in CG attempts could be avoided: roughly half the fit's inner-solve wall time.

### Where
- `gam_linalg::pcg::DenseRouteWork::pcg_attempt_under_cap` (crates/gam-linalg/src/pcg.rs ~66-84) sets `products = (build + p³/3) / apply`. Its own doc says so: "Whichever route wins, the solve costs at most twice the cheaper one, and no CG iteration count has to be predicted."
- `JointHessianWork::row_pullback(n, p)` (crates/gam-model-api/src/families/custom_family/options.rs ~325) takes `build = n·p²` and `apply = 2·n·p`. That gives 40.5 at p = 81 and n = 80,016.
- BMS prices its workspace that way (crates/gam-models/src/bms/custom_family_impl.rs ~736).
[...]

PR #3424, opened 2026-09-20
Decide the joint-Newton CG route once per step from a derived iteration bound; delete the dense fallback (#3285)

SauersML:
Refs #3285. This is Phase 1a of the plan the lead approved on the issue. What remains is listed at the bottom.

## What changes

Each exact joint-Newton step now picks one route, dense or CG, before any solve. The old behaviour was "attempt CG under a product budget; on a miss, build the dense Hessian and solve the same system again". That dense fallback is gone.

- **Eligibility.** CG is eligible only for the plain penalized system `H + S`. A dense source, a pending returned-mode curvature or a Jeffreys-augmented system goes dense, as before.
- **`PcgAttempt::Only`** (no dense route fits the memory cap). This is unchanged: CG runs under the historical `4·p` cap and a miss refuses the trial point.
- **`PcgAttempt::Budgeted { products }`.** CG runs only when its derived iteration bound costs fewer products than the dense route.
  - **The bound.** k̂ is computed at the last dense step's spectrum. Take the extremes `[lower, upper]` of `B = D^{-½}(H+S)D^{-½}` for the trust metric `D`, and the Jacobi preconditioner `M` with ridge `r`. With `E² = D/M`:
    - Ostrowski puts the spectrum of `M^{-½}(H+S)M^{-½} = EBE` in `[lower·min E², upper·max E²]`.
    - The ridge adds `[r/max M, r/min M]`.
    - Weyl bounds the sum, which gives κ̄.
  - **The count.** The Chebyshev count is `k̂ = ⌈ln(2√(κ̄ κ_M)/η) / ln((√κ̄+1)/(√κ̄−1))⌉`, with `η = max(rel_tol, attainable)`.
    - The `√κ_M` factor turns the A-norm error bound into the relative 2-norm residual that CG's stop rule tests.
    - `pcg_products(k) = k + ⌊k/32⌋` adds the residual refreshes.
  - **A cold start** has no recorded spectrum, so it goes dense. That dense step measures the spectrum.
  - **The recorded spectrum** needs `lower` above the machine-rank floor `λ_max·√p·ε`.
- **A CG miss is an inexact Newton step, not an error.**
  - On `MaxIters` or `Breakdown`, the last iterate is returned. From `x₀ = 0` every CG iterate lowers the quadratic model. On a breakdown at the first direction, `M⁻¹rhs` is returned.
  - The trust region and dogleg then judge that step against the true model.
  - The recorded spectrum is cleared, so the next step goes dense and measures it again.
  - A non-finite iterate, or a failed operator matvec, still refuses the trial point.

New pieces:
- `gam_linalg::pcg::{pcg_iteration_bound, pcg_products, PCG_RESIDUAL_REFRESH_PERIOD}`
- `gam_linalg::utils::solve_spd_pcg_bounded_into`, which returns the iterate and its `PcgStop`. `solve_spd_pcg_with_info_into` now delegates to it.
- `JointPcgConditionRecord` and `JointPcgRoute` in `joint_newton.rs`.

## Tests

- `pcg::iteration_bound_tests` (6):
  - The bound is the Chebyshev count and sits under its `½√κ ln(·)` form.
  - The bound is monotone, and non-condition-numbers are rejected.
  - Refreshes are counted.
  - A spent budget still lowers the model.
  - Planted spectra (p = 60, κ_M = 80, κ ∈ {4, 1e2, 1e4}, η ∈ {1e-1, 1e-4, 1e-8}) all converge within the bound.
- `joint_newton::joint_pcg_route_tests` (4):
  - The bound is exact when `M = D`.
  - The bound contains the exact preconditioned spectrum for three mismatched preconditioners, with and without a ridge.
  - An indefinite or singular spectrum records nothing.
  - The route is CG only when its bound costs strictly less than the dense route, and an ill-conditioned bound goes dense.

## Validation

- `cargo check --workspace --tests` at head `3d372ab6`, after merging `origin/main` at `1c84d2a4`: clean, with no warnings.
- The targeted tests at `3d372ab6` pass: `pcg::iteration_bound_tests` 6/6, and `joint_pcg_route_tests` plus the `large_scale_shape_margslope` pin, 5/5.
- The full library suites ran at `3d640034`, before the merge:
  - `gam-linalg --lib`: 316/316 pass.
  - `gam-custom-family --lib`: 370 pass and 1 fails. The failure, `jeffreys_arming_979_tests::a_refusal_with_evidence_refits_armed_once_and_publishes_it_2995`, already fails on main. It is tracked in #3371, and this diff does not touch it.
- `scripts/source_removal_guard.py --base origin/main --head HEAD`: 642 -> 666 items, 0 guarded removals.

## Remaining for #3285 (not in this PR)

- **Pricing.** Price the `w_row` row sweep in `JointHessianWork::row_pullback`.
- **Weyl widening.** Widen κ̄ for the change of `H` between the recorded β and this one. Today a stale bound costs at most one inexact step and a dense re-measure.
- **Forcing floor.** Replace `PCG_ETA_MIN = 1e-8` with `η′ = max(η_EW, band/‖rhs‖)`.
- **Phase 2.** Move the other `Budgeted` consumers onto the same up-front route: `gam-linalg` `matrix/mod.rs`, `arrow_schur/newton_step.rs`, `reduced_solve.rs`, and `gam-sae` `construction_quasi_laplace.rs`.
- **Replay.** Re-run the measure-first replay at n = 80,016.


SauersML:
[swarm lane gam-2930, owner of #3285 Phase 1] Measured case against this design, before it merges.

**1. The Chebyshev bound never picks CG on the systems #3285 is about.** On the measure-first replay (BMS, n = 80,016, p = 71, 1,833 joint-Newton steps), the Jacobi-preconditioned condition number had a median of **1.3e10**. The plain Chebyshev count is then about ½√κ̄·ln(2/η) ≈ **750,000** products (median), against a measured median of **17** CG products at the derived forcing floor. The bound is off by four to five orders of magnitude because these spectra are clustered, and CG exploits the clusters. So `k̂ < products` essentially never holds, and this PR takes the dense route on every step, i.e. it behaves as "dense whenever the Hessian fits".

**2. That regresses the gam#2900 regime.** bms/custom_family_impl.rs:2343-2354 measured a BMS flex dense build at more than 60 s against about 3 s per HVP (n ≈ 195k, p ≈ 44), i.e. dense ≈ 20 products. A route that is always dense pays that 60 s on every Newton step, where CG needs a handful of products.

**3. The flop budget does not price the dense route either.** `Budgeted { products }` is `(n·p² + p³/3)/(2·n·p)`. Measured dense-to-product cost ratios are:
- p71: flop ratio 35, measured **2.2**.
- n = 5,008, p = 731: flop ratio 383, measured **~200**.
- gam#2900 flex: flop ratio 22, measured **~20**.

Whenever κ̄ is small enough for the bound to pass, a CG step can cost up to about 15x the dense step at p71.

**What I measured instead** (MSI acl42, 6 pinned cores, base 4d529db085):
- **The route:** the per-step route is predicted by exact-arithmetic CG on the solve's own last dense whitened spectrum with this step's right-hand side. That prediction matched the measured count with median ratio 1.00 (p10 0.93, p90 1.05).
- **The costs:** both route costs are measured in the same solve, and nothing is carried between inner solves.

| Shape | Base | Always dense | Per-solve measured route |
|---|---|---|---|
| p71 | no certified multistart seed (829 s) | 445.2 s (388.4 s in an earlier round) | 400.4 s |
| n = 5,008, p = 731, 1,500 s cap | reached no certified outer point | 3rd certified point after 2,236 inner lane-seconds | 3rd certified point after 2,312 |

- **Accuracy:** certified V agree within 7.1e-7 (τ = 6.2e-6) at p71, and within 3.1e-7 (τ = 1e-4) at n = 5,008.
- **A carried route loses:** a route carried across inner solves took 373 cycles to the first certified point, against 114 for the per-solve route. So a spectrum recorded at one smoothing point must not price the next solve.
- **Flex:** a BMS flex A/B on the matrix-free row route is running now; results follow here.

**Also missing here:**
- The `PcgAttempt::Only` miss still refuses the trial point. My change returns the inexact iterate there too.
- The 1e-8 forcing floor stays. Replacing it with η′ = max(η_EW, band/‖rhs‖) raised η on 601 of 1,833 replay steps and cut their CG products from a median of 40 to 2.

The per-solve measured route lands on main after its gate. Please keep this draft unmerged until the flex row is in.

SauersML:
[swarm lane gam-2930] Correction to my comment above, from the measurement it promised.

The BMS flex row (flexible_link, n = 40,000, p = 49, row-primary cache declined so the inner Hessian is an operator) does **not** separate this PR from a measured-cost route. I ran merged main 3fc95ce70a against main with the per-solve measured route, 900 s cap each:
- Both arms put every inner solve on `branch=CONSTRAINED_QP`: the link-deviation monotonicity constraints are active (`joint_constraints_is_some=true`), so CG is never eligible there.
- Both had 11 inner solves and 64 cycles. Inner solve_wall was 831.6 s (main) against 833.8 s.

The gam#2900 cost of 60 s per dense build against 3 s per HVP is paid by the constrained QP, whichever CG route rule is in place. My "regresses the gam#2900 regime" point was wrong.

On the unconstrained operator shapes where the rule does run (rigid BMS at p = 71 and p = 731), this PR goes dense, and always-dense measured within noise of the measured-cost route: 445 s against 400 s at p71 in one round, 388 s against 410 s in another, and +3% at p731. I have no measured shape where the measured-cost route beats this PR, so it stays as merged.

Issue #2679, opened 2026-07-31
The locscale response-moment integral still evaluates the moment-matched normal: the node sum is 228x closer but the exact cutover is nodes x outer x inner (residual of #2446)


SauersML:
Split out of #2446, which closes on its own precondition (the `(e)` column re-measured against the shipped cubature). This is the one thing #2446 identified and deliberately did **not** land, and it is a cost problem rather than a correctness one.

### What is already true

`constrained_projection_law` (`crates/gam-solve/src/constrained_posterior.rs`, `4c625b07e`) exposes the node mixture — `(location, weight)` per cubature node plus the tangent variance — from the same cubature the intervals already use. #2446 measured it against a tensor-Simpson reference on the exact truncated density, at `q = 2`, on the SHIPPED nodes:

| | error vs exact pushforward | infeasible mass |
|---|---|---|
| node sum, 4096 shipped nodes | `2.688e-6` | `0` |
| moment-matched normal (what the locscale warp integral evaluates today) | `6.135e-4` | `4.54e-2` |

The quantile consumer already runs on the nodes: `gam-predict/src/lib.rs:2004` and `:3459` call `constrained_projection_equal_tailed_interval`. The smooth-functional consumer does not.

### The consumer that has not moved, and the exact reason

`exact_survival_response_moments_row` (`crates/gam-models/src/survival/location_scale/moments.rs:263`) integrates the joint as an **outer tensor GHQ over `y = (h, threshold, log σ)`** — `low_rank_normal_expectation_pair_3d_result(..., 15, ...)`, so up to `15³ = 3375` points at rank 3 — times an **inner 1-D rule over `w | y`** at 21 points. That is up to `70,875` integrand evaluations per prediction row today, each of which rebuilds the wiggle basis.

A node of the cone mixture shifts the mean of the **whole** coefficient vector — `β = β_unc + t + G(u − E[u])` — so it moves `(h, threshold, log σ)` as well as `w` whenever `Σ` has cross-block covariance, which it does. The outer rule therefore **cannot be shared across nodes**, and the exact cutover is the product `nodes × outer × inner` = `4096 × 3375 × 21` per row. That is not a bounded production integration rule; it is the same objection recorded on #2446 on 2026-07-26, now with the rule sizes read off the shipped code rather than estimated.

The `~200×` figure quoted earlier on #2446 is the ratio `4096 / 21` — it prices replacing only the **inner** rule, which is not a valid cutover for the reason above.

### What the fix has to be

[...]

PR #3428, opened 2026-09-20
Survival location-scale posterior means integrate the cone-truncated law (#3038)

### LINKED ISSUE #3038: Survival location-scale with a linear PC log-scale fits certified, then predict raises 'hazard derivative must be finite and positive'

SauersML:
Fixes #3038

## Root cause
When a location-scale survival fit has an active inequality cone C = {β_time ≥ 0}, its reported posterior is `N(β_unc, Σ)` truncated to C. `SurvivalPosteriorIntegration::default_for` sent these fits to `SigmaPoint`, which integrates the moment-matched normal `N(E_π β, Σ_π)`. That normal's nodes leave C, and outside C the model has no hazard. The plug-in evaluator then refused those nodes with "hazard derivative must be finite and positive".

## Fix
- `default_for(model, covariance_mode)` returns the new `TruncatedLaw` integration for a location-scale fit whose selected posterior has an active cone. Other location-scale fits still get `SigmaPoint`.
- `TruncatedLaw` integrates over the joint constraint-normal × tangent rule that the location-scale response moments already use (`ConstrainedPosteriorJointRule`, #2679). Every node is a feasible coefficient vector, and the plug-in location-scale surfaces are replayed at it using the designs assembled once.
- Survival, event density and the index are accumulated per replicate lattice. Each cell is retired once its replicate standard error is within the law's certified relative accuracy (1e-3), less the integrand's rounding:
  - survival: relative to `sqrt(E S (1−E S))`;
  - density: relative to `max(|E f|, sd f)`. Deep in a tail, where the posterior sd of f exceeds its mean, the Monte Carlo error is then a fraction `tol` of the posterior uncertainty the surface carries;
  - index: relative to its posterior sd.
- Past the rule's maximum node count the surfaces are refused, naming the worst cell. They are never published uncertified.
- The results are published the same way as `ExactAnchor`: `E S`, `−log E S` and `E f / E S`.
- `assign_survival_fit_coefficients` is factored out of `saved_model_with_survival_coefficients`.

## Validation
The new test is `survival::predict::tests::location_scale_posterior_mean_integrates_the_cone_truncated_law_3038`. It uses n=400 lognormal location-scale data with a linear log-scale, as in the report.
- The default integration is `TruncatedLaw`.
- The published `E[S]` and `E[f]` are compared with an independent rejection Monte Carlo over the ambient law: 4000 feasible draws out of 9.1e7 proposals, each replayed through the per-coefficient law. Each cell passes within 4 Monte Carlo standard errors plus the rule accuracy.
- The largest survival gap between the rule and Monte Carlo is 6.7e-4. The plug-in misses by up to 1.96e-3. For example, at row 0, t=1 the rule gives 0.03722, Monte Carlo 0.03743 (se 2.2e-4) and the plug-in 0.03547.
- The test runs in 96 s. The surfaces certify within about 20 s.
- `cargo check --workspace --tests` is clean.

https://claude.ai/code/session_01UNAoaQxuMnCvrWmiLS2Gvb

---
_Generated by [Claude Code](https://claude.ai/code/session_01UNAoaQxuMnCvrWmiLS2Gvb)_


SauersML:
Separating case from the #3038 reproduction (gnomon study's calpred spec; 3,828- and 19,392-row frames; the 64-point default grid plus each row's exit, i.e. 65 cells per row). This is the same mechanism as this PR (the truncated law on #2679's joint rule) with a different cost structure. Measured at fbdef09b41/ce53546019 with a per-cell projection: each cell reads 6 linear channels (h, ḣ, η_t, η_t', η_σ, η_σ'), plus scalar pairs for the wiggles, so its lattice runs in q + 6 dimensions, where q is the number of retained cone rows (12 here).

- 19,392-row fit: all 10 study predict cases publish. The 1,000-row case takes 92 s and the 4,719-row case 431 s, i.e. 3.4-4.7x the sigma-point route's cost (2p plug-in passes; one plug-in pass is 0.21 s and 1.34 s). Every cell certifies at the rule's first 2,048 nodes per replicate.
- 3,828-row fit: about 5% of cells, those just after entry (S ≈ 0.9987-0.9999, f ~ 4e-5..7e-4), miss a density certificate relative to E f at 2,048 nodes. They need 16-64x more nodes per replicate (up to 131k).

Two cost points in this PR, on that case:
1. `node_cells` (the closure passed to `truncated_location_scale_surface_moments`) runs `predict_survival_location_scale` and `index_rate` over every cell at every node. That includes the doubling rounds in which all but a handful of cells have already retired, so a tail round costs the whole surface per node: 63k cells x up to 131k nodes x 8 replicates on the 3,828-row fit.
2. The draws are whole coefficient vectors, so the lattice runs in q + rank(Σ_res) dimensions (~47 here) rather than q + 6. QMC error falls more slowly there.

Also: the survival standard deviation is certified (`survival.sd_spread`) even when `with_uncertainty` is false, when it is not published.

I will measure this PR on the same two fits after it merges and post the numbers here.

SauersML:
Measured cost of the merged route (e393d691f5), timed on main c663ba6dba. The setup: the saved calpred MEDIUM location-scale survival model (19,392-row frame, #3037), 4 cores, posterior-mean predict. For comparison, the per-cell projected integrator (6 channels per cell, row channels shared per group, only active cells evaluated) ran on the same model and cases on fbdef09b41.

| case | rows | plug-in pass | merged route (whole-vector draws, full-surface replay) | per-cell projected |
|---|---|---|---|---|
| first_1000 | 1000 | 0.163 s | 557.6 s | 40.3 s |
| band_78 | 136 | 0.029 s | 67.9 s | 6.8 s |
| band_88 | 22 | 0.006 s | 11.3 s | 2.7 s |

The two routes agree within the rule tolerance (rel 1e-3). On first_1000, S_mean is 0.412270 vs 0.412279 and h_mean is 1.682762e-2 vs 1.682761e-2. On band_88, h_mean is 1.14403e-2 vs 1.14364e-2.

On first_1000 the merged route costs about 36x the sigma-point route (94 plug-in passes, ~15.3 s); the per-cell route costs 2.6x. The cost comes from the replay: every lattice node re-evaluates the whole surface, including the cells that are already retired. A follow-up will replace this integrator with the per-cell one, as a single route with the same TruncatedLaw plumbing and this PR's end-to-end test kept.

PR #3432, opened 2026-09-20
BMS host row-Hessian matvec and diagonal: measured row-kernel admission and race (#3410)

### LINKED ISSUE #3410: BMS row-Hessian matvec dispatches on "a device resolves", with no admission or race
## What happens

`crates/gam-models/src/bms/axis_direction_search.rs` has five call sites that send the per-row Hessian matvec to the device whenever `GpuRuntime::resolve(global_policy())` returns a device (on main: about lines 456, 589, 815, 971 and 1104). The sites are the host-pin HVP, the tiled HVPs and the directional variants, and each one does:

```rust
if gam_gpu::device_runtime::GpuRuntime::resolve(gam_gpu::global_policy())?.is_some() {
    require_selected_gpu_result("host-pin row-Hessian matvec",
        row_hessian_ops::launch_row_hessian_matvec(..))?
} else {
    row_hessian_ops::cpu_row_hessian_matvec(&inputs)
}
```

So under `auto`, having a device means the device runs, whatever the shape. The row Hessians sit in host memory, so every apply pays an H2D copy of `n·r²` values and a D2H copy of `n·r`, and the CPU is often faster at small `r`. None of these sites goes through `decide_row_kernel`. They have no `RowKernelShape`, no measured executor and no race, and their decision appears nowhere in the dispatch log.

## Why this is a bug

#3024 made row-kernel dispatch a measured race between the kernel's own executors. #3358 (slice B) did the same for the resident FLEX row-primary Hessian, racing the whole inner step. These matvecs are the one remaining BMS row-kernel dispatch that picks the device only because a device exists. They are also a hard-coded choice rather than a measured one: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code."

## Fix direction

- Give the host-resident row-Hessian matvec its own `GpuKernel` and `RowKernelShape` (rows, `r`, threads).
- Admit it through `decide_row_kernel`, so `off` and `required` behave as they do for every other row kernel.
- Race an unmeasured shape with `race_row_kernel` on the apply. The host rows are the state either way, so one apply is the unit.

Found while landing #3024 slice B (#3358); that PR does not touch these sites.

SauersML:
Draft. Validation is in progress.

Before this change, the five sites in `crates/gam-models/src/bms/axis_direction_search.rs` sent the host-resident per-row Hessian matvec and diagonal to the device whenever `GpuRuntime::resolve` returned a device. Now they go through the row-kernel decision:

- **New kernels.** `GpuKernel::RowHessianMatvec` and `GpuKernel::RowHessianDiagonal`, each with a `RowKernelShape { rows, widths: [r, 0, 0, 0], threads: 1 }`. The CPU executors are one sequential loop, hence `threads: 1`.
- **One entry point.** `row_hessian_ops::row_hessian_matvec` and `row_hessian_ops::row_hessian_diag` admit through `decide_row_kernel`:
  - `off` never probes the device;
  - `required` needs a supported backend (`require_supported`);
  - `auto` races an unmeasured shape with `race_row_kernel` (#3024) and reads the recorded timings after that.
- **Logging.** The decision is now logged like every other row kernel.
- **No CPU fallback.** A fault on a selected device is still returned, never recomputed on the CPU.

Fixes #3410

---
_Generated by [Claude Code](https://claude.ai/code/session_01Am3cVoSQbaYX3xacMut2Et)_


SauersML:
Taken over by LC-19; driving to merge.

---

SauersML:
e9ffb438 fixes the ban-scanner failure on the previous head.

That head failed `cargo check --workspace --tests`: the build.rs ban-scanner rejected `let _ = decision;` at `row_hessian_ops.rs:504` and `:541` (the no-device cfg arms). Both cfg blocks now go through one helper, `on_selected_executor(decision, cpu, device)`:
- a shape that needs a race goes to `race_row_kernel`;
- `use_gpu` runs the device closure;
- anything else runs the CPU closure.

The no-device build consumes `decision` through `no_device_executor` rather than discarding it. The census still counts two `require_selected_gpu_result(` consumers.

Validation on e9ffb438:
- `cargo check --workspace --tests` is clean. Its only warning is the unused `IxDyn` import in gam-pyffi, which is already on main.
- `cargo test -p gam-models --lib` for flex_verify_932 plus row_hessian: 21 passed, 0 failed.

The PR is still LC-19's to drive and merge.

---

SauersML:
BF-103 review (partition 2 sweep, #3410): this PR is ready for the local merge train (MT-0, even PR number).

- Head 62e93668a3 merges cleanly with origin/main.
- The gam-gpu API it relies on is unchanged: `decide_row_kernel`, `GpuDecision::{race,use_gpu}` and `race_row_kernel`.
- SPEC read: the decision is measured, not assumed. A device fault is returned rather than recomputed. The no-device arm consumes the decision instead of bypassing it.
- The owner's local check was clean (21/21 tests). The only thing outstanding is the queued focused check, job 106062958251.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 6441694765e4.

Issue #3316, opened 2026-09-20
gam predict computes survival marginal-slope posterior-mean survival by a delta-method probit-normal formula instead of the library's exact anchored integral


SauersML:
## Defect

`gam predict` computes the posterior-mean survival of a survival marginal-slope model by a different formula than the library prediction that the Python API uses. The same saved model therefore reports two different "posterior mean S(t)" values, depending on the entry point.

At gam main 4ee765aeba:

- **CLI** (`crates/gam-cli/src/main/run_predict.rs:2669-2769`): for `SurvivalLikelihoodMode::MarginalSlope`, `run_predict_survival` never calls `predict_survival`.
  - It builds the Bernoulli marginal-slope predictor (`build_saved_survival_marginal_slope_predictor`, :2711) and takes `eta` and `eta_standard_error` from `predict_posterior_mean_generic`. There, `eta_standard_error` is the delta-method standard error of the linear state (`crates/gam-predict/src/interval_policy.rs:870`).
  - It then publishes `mean = Phi(-mu / sqrt(1 + se^2))` (:2755), the probit-normal integral under a Gaussian law for eta, at each row's own exit time.
- **Library** (`crates/gam-models/src/survival/predict.rs:2553` `predict_survival`, reached from `gam-pyffi` at `manifold_and_posterior_ffi.rs:3926`): under `SurvivalPredictEstimand::PosteriorMean`, a model without timewiggle, score warp or link deviation is integrated by `SurvivalPosteriorIntegration::ExactAnchor` (:1498). That rule integrates E_theta[Phi(-eta(q(t), b(t)))] exactly over the bivariate Gaussian law of the two primaries, re-solving the anchor at every node.

The two agree only if eta is affine in theta. The anchored marginal-slope index is not: eta = c(b) q + s b z with c(b) = sqrt(1 + b^2) on the rigid frame, and eta = a(q, b) + s b z under a declared law. So the CLI publishes a delta-method approximation where the library publishes the exact integral.

On the gam-predict fixture `survival_marginal_slope_posterior_mean_against_monte_carlo`, the plug-in is 1.4e-3 off the 20,000-draw Monte Carlo posterior mean at the fitted covariance, while the exact rule is within 2.0 Monte Carlo standard errors. The CLI's formula has not been measured on that fixture.

[...]

Issue #398, opened 2026-05-29
predict(data) and predict(data, interval=...)['mean'] disagree: interval path applies a bias correction the plain path does not (worse vs truth on identity-link Gaussian)


SauersML:
## Summary

`Model.predict(data)` and `Model.predict(data, interval=0.95)["mean"]` return **different point predictions** for the same inputs, even on identity-link Gaussian models. Requesting an interval silently switches the engine onto a different estimator (`predict_full_uncertainty` with `apply_bias_correction: true`) than the plain path (`predict_plugin_response`), so both the `mean` and the `linear_predictor` columns shift. The shift contradicts the documented contract that `mean` is "the point prediction" (`Model.predict` docstring), and on identity-link Gaussian — where there is no link nonlinearity to bias-correct — the shifted estimate is measurably *worse* against ground truth.

## Trigger (released gamfit 0.1.135)

```python
import numpy as np, gamfit
rng = np.random.default_rng(11)
n = 2000
x = rng.uniform(-3, 3, n)
y = np.sin(x) + rng.normal(0, 0.3, n)
df = {"x": x.tolist(), "y": y.tolist()}

m = gamfit.fit(df, "y ~ s(x)", family="gaussian")   # identity link
xt = np.linspace(-3, 3, 80)
truth = np.sin(xt)

p_plain = np.asarray(m.predict({"x": xt.tolist()}))
tab     = m.predict({"x": xt.tolist()}, interval=0.95)
p_int   = np.asarray(tab["mean"])

print(np.max(np.abs(p_plain - p_int)))               # 0.0145  (should be 0.0)
print(np.mean((p_plain - truth)**2))                 # 0.000493  (plain)
print(np.mean((p_int   - truth)**2))                 # 0.000580  (interval-mean: WORSE)
# linear_predictor differs too:
# linear_predictor == mean (identity link); plain-vs-interval LP maxdiff also 0.0145
```

The discrepancy is largest at the domain boundary (x = −3.0): plain = −0.168, interval-mean = −0.182, truth = −0.141 — the "corrected" value moves *away* from the truth.

The same divergence shows up on other families (e.g. Gamma/log: plain-vs-interval-mean maxdiff ~0.058; Poisson; binomial), but identity-link Gaussian is the cleanest demonstration that the correction is unjustified there.

## Observed vs expected

[...]

PR #3445, opened 2026-09-20
Survival posterior-mean point is conditional under every covariance mode; FFI defaults to the published mode

### LINKED ISSUE #3421: gamfit survival predict integrates the point under the interval's covariance: interval= moves the posterior-mean survival (2.97e-5), and its defaults disagree with gam predict
## Defect

For a single-event survival model, the posterior-mean survival that gamfit publishes depends on whether `interval=` is set. Its default also differs from what `gam predict` publishes for the same saved model. That breaks the #398 invariant (the point is a property of model and inputs, never of whether an interval was requested) and CLI/Python parity.

At gam main 119fd0951a, `gam-pyffi` `predict_survival_result` (`crates/gam-pyffi/src/manifold/manifold_and_posterior_ffi.rs:3912-3925`) chooses one covariance for the whole `predict_survival` call:
- with `interval=`: the requested `covariance_mode`, defaulting to `SmoothingCorrected` (:3914);
- without: `Conditional` (:3924).

The competing-risks wrapper at :3833-3847 does the same. The posterior-mean point is integrated under that covariance, so `interval=` moves it. gamfit passes `interval` and `covariance_mode` through verbatim (`gamfit/_model.py:521`).

The library's own rule differs. `predict_competing_risks_with_posterior` (`crates/gam-models/src/survival/predict.rs:1619-1625`) says "The public posterior-mean point is always the conditional-posterior estimand", and computes that point separately when a smoothing-corrected interval is requested (`separate_conditional_point`). The single-event `predict_survival_posterior_mean` (:1168) has no such separation.

## Measured (MSI, main 119fd0951a; gamfit's values are the FFI's `predict_survival` calls under the covariance it picks)

Royston-Parmar fit, `Surv(time, event) ~ s(x)`, 600 rows. The fit publishes `SmoothingCorrected` (corrected covariance present). Posterior-mean survival at t = 1, row x = -0.9:

| surface | covariance | survival |
|---|---|---|
| gamfit, `interval=None` | Conditional | 0.778365135566 |
| gamfit, `interval=0.95` | SmoothingCorrected (default) | 0.778335474471 |
| `gam predict` (with or without `--uncertainty`) | published (SmoothingCorrected) | 0.778334472213 |
| plug-in | | 0.779338266626 |

- `interval=` moves gamfit's point by 2.97e-5, about 3% of the posterior-mean correction itself (plug-in minus posterior mean, 1.0e-3).
- gamfit without an interval and `gam predict` differ by 3.07e-5.
- The remaining 1.0e-6 between `gam predict` and gamfit with an interval is a separate route. The CLI's Royston-Parmar branch (`crates/gam-cli/src/main/run_predict.rs:2853`) still forms its own posterior mean through `predict_gam_posterior_meanwith_backend` and never calls `predict_survival`. That is the same class as #3316, for the other likelihoods.

Survival marginal-slope fit (constant slope, 400 rows) that publishes `Conditional` (no corrected covariance): gamfit with `interval=0.95` and default `covariance_mode` refuses ("fit result does not contain smoothing-corrected covariance"), while `gam predict --uncertainty` succeeds under the published covariance. The FFI's hard `SmoothingCorrected` default contradicts `UnifiedFitResult::published_covariance_mode` (`crates/gam-solve/src/model_types/result_types.rs:6340`), whose contract is that every default uncertainty surface uses the published definition.

## Fix direction (one rule for every route)

- The point is the conditional-posterior mean, as the library already states for competing risks. `predict_survival_posterior_mean` gets the same separated conditional point when uncertainty is requested under the corrected covariance.
- The uncertainty surfaces use the requested mode, defaulting to `published_covariance_mode()`, as the CLI resolves it (`crates/gam-cli/src/main/model_summary.rs:250`).
- The FFI then passes the resolved mode and `with_uncertainty = interval.is_some()`, and never picks the point's covariance from `interval=`.

Found by gam-3026 while checking #3316.

SauersML:
Fixes #3421

**Engine (`gam-models` survival predict)**
- `predict_survival_posterior_mean_with` always integrates the published point under the conditional law, `E[S | D, ρ̂]`. A smoothing-corrected band gets its second moments from a separate pass over the same integration rule (sigma-point or exact-anchor), so `interval=` never moves the point (#398). This is the rule `predict_competing_risks_with_posterior` and the Royston–Parmar standard path already state.
- The two integrators now return `(plug-in, moments)`. `publish_survival_posterior_moments` takes the point moments plus an optional `(band moments, mode)` pair, and nothing else is published from the band.
- Competing risks: a request without uncertainty integrates the conditional law whatever mode it names. Before, a `SmoothingCorrected` mode without uncertainty moved the point.

**FFI**
- `predict_survival_result` and `predict_competing_risks_survival_result` resolve an omitted `covariance_mode` to `published_covariance_mode()`, which is the CLI's `resolved_covariance_mode`.
- Before, they used a hard `SmoothingCorrected` default, which refused fits that publish the conditional law (the survival marginal-slope case in the issue). The point's covariance is no longer chosen from `interval=`.

**Out of scope:** the CLI Royston–Parmar branch's own posterior mean via `predict_gam_posterior_meanwith_backend` (the remaining 1e-6 route). That is #3316's class.

https://claude.ai/code/session_01CJBw9tSUsL4XWv47vBGm3U

---
_Generated by [Claude Code](https://claude.ai/code/session_01CJBw9tSUsL4XWv47vBGm3U)_


SauersML:
[swarm lane gam-3026] Review against my parallel candidate for #3421, which used the same engine rule and the same FFI default. The engine and FFI parts here match it. Three things it lacks, which I'll land on top once this merges unless you add them:

1. **No pin.** Nothing in this PR fails if the point starts following the band's covariance again. My candidate adds `crates/gam-models/tests/survival_acceptance/survival_point_covariance_3421.rs`, a Royston-Parmar `Surv(time, event) ~ s(x)` fit on 600 rows that publishes the smoothing-corrected covariance.
   - It asserts that the posterior-mean survival and hazard are bit-identical across {Conditional, SmoothingCorrected} x {with, without uncertainty}.
   - It asserts that `covariance_source` names the requested mode exactly when uncertainty is requested.
   - It asserts that the corrected standard deviations differ from the conditional ones (2.17e-3 on that fixture).
   - With the same engine rule as this PR it passes on MSI. At main it fails: the point moved 2.97e-5 there.
2. **The CLI provenance goes stale.** `run_predict_saved_survival_marginal_slope` in `crates/gam-cli/src/main/run_predict.rs` (the #3316 route) prints `covariance_provenance_note(Some(covariance_mode), uncertainty_source)` with the comment "the point integral ran under the requested covariance". Under this PR the point is always conditional, so the note should say `Some(InferenceCovarianceMode::Conditional)` for the point.
3. **A stale doc.** `build_saved_survival_marginal_slope_predictor` (`predict.rs` ~5793) still says "The CLI's `gam predict` flow and the library-side `predict_survival` both call into this helper". Since #3316 the CLI reaches it only through `predict_survival`.

Issue #3350, opened 2026-09-20
Saved models store the fit twice and are parsed up to three times: gamfit load is 50-350x slower than pyGAM, save is 9.5x slower at p20


SauersML:
## Measured (the #3150 sweep: n_train = 1e3, n_pred = 1e2, gamfit vs pyGAM, ms)

| case | save | load | model bytes |
|---|---|---|---|
| gaussian p5 | 0.51 / 0.31 (1.65x) | 8.53 / 0.10 (85x) | 1,394,178 / 86,695 |
| gaussian p20 | 9.24 / 0.97 (9.5x) | 132 / 0.37 (358x) | 21,092,280 / 1,301,088 |
| gaussian te | 0.37 / 0.32 | 6.64 / 0.12 (55x) | 1,157,966 / 86,365 |
| binomial p5 | | 8.09 / 0.09 (90x) | |
| binomial p20 | 6.47 / 0.99 (6.5x) | 122 / 0.39 (312x) | 20,179,057 / 1,301,117 |
| binomial te | | 6.12 / 0.10 (61x) | |
| poisson p5 / te | | 8.28 / 0.09 (92x), 6.63 / 0.11 (60x) | |

Load time scales with the size of the file: about 6 ms per MB in both the Python sweep (132 ms for 21 MB) and the Rust profile below (25 ms for 4.3 MB). JSON float parsing dominates, so every byte saved is load time saved.

## Profile (Rust API, `FittedModel::save_to_path` / `load_from_path`, n = 1000 synthetic fits, main ada5db438a)

Time per step (ms):

| case | bytes | serialize | save_to_path | parse | load_from_path | parse to `Value` | finite walk |
|---|---|---|---|---|---|---|---|
| gaussian p5 | 311,187 | 2.99 | 29.6 (first-call warmup) | 1.37 | 1.92 | 1.01 | 0.47 |
| binomial p5 | 299,134 | 0.72 | 5.00 | 1.54 | 1.93 | | |
| gaussian p20 | 4,342,173 | 29.4 | 69.1 | 18.5 | 25.1 | 14.6 | 7.2 |
| gaussian te | 10,142,344 | 25.5 | 102.3 | 37.1 | 54.8 | 33.0 | 14.5 |

`validate_for_persistence` takes about 0.01 ms and `validate_numeric_finiteness` under 1 ms. Almost all the time is serialization and parsing.

Bytes per field (gaussian p20; te and p5 have the same shape):

| JSON path | share |
|---|---|
[...]

Issue #3054, opened 2026-09-19
Saved-model files have six writers with four durability contracts; every save must go through write_saved_model, never partial


SauersML:
## Property

Every saved-model file is written by one function, `gam_model_api::saved_model::write_saved_model`, and that function never leaves a partial document at the path, a crash included. A save that returns `Ok` has made its document durable (on Unix). A save that cannot confirm this fails with an error that says why.

## Defect: six writers with four different durability contracts (gam main 6adda09417)

| # | writer | file:line | contract today |
|---|---|---|---|
| 1 | `write_saved_model` (joint event model; event-history predictor next, #2966) | `crates/gam-model-api/src/saved_model.rs:133-155` | temp file + rename; **no fsync**, so after a crash the new name can point at empty or partial bytes |
| 2 | `FittedModel::save_to_path` (every CLI GAM model) | `crates/gam-models/src/inference/model.rs:4876-4980` | temp + fsync + rename + best-effort directory fsync. A **failed fsync is only logged** (`log::warn!`/`log::debug!`), and the save still reports success |
| 3 | gamfit `Model.save`, `MultinomialModel.save` | `gamfit/_model.py:1045`, `gamfit/_model.py:1474` | `Path.write_bytes`: **not atomic**; a crash or a full disk truncates the user's existing model file |
| 4 | gamfit `ResponseGeometryModel.save` | `gamfit/_response_geometry.py:485` | `Path.write_bytes`: not atomic |
| 5 | ManifoldSAE `save` | `crates/gam-pyffi/src/manifold/manifold_and_posterior_ffi.rs:5125-5127` | `std::fs::write`: not atomic |
| 6 | SupportSparseSAE `save` | `crates/gam-pyffi/src/manifold/support_sparse_sae_ffi.rs:569-632` | `std::fs::write`: not atomic |

The same file (a saved model) has four durability contracts, depending on which surface wrote it. Two writers for one on-disk contract is itself the defect (SPEC 25), and three of these silently lose the contract that the CLI's writer provides.

## Fix, in slices. Each slice deletes the writer it replaces.

[...]

Issue #3053, opened 2026-09-19
gamfit.load cannot read a joint event model that gamfit.save wrote: saved_model_kind knows no envelope kind


SauersML:
## Defect

`gamfit.save(model, path)` writes a joint event model, but `gamfit.load(path)` and `gamfit.loads(bytes)` cannot read it back. Both raise `GamError: saved model payload is missing its payload object`. `gamfit.load` is documented as the inverse of `gamfit.save`.

## Repro (MSI acn112, gam main 6adda09417, test-profile `libgam_pyffi.so`)

```python
subjects = pd.DataFrame({"id": ["a", "b"], "entry": [0.0, 0.0], "exit": [4.0, 6.0]})
events = pd.DataFrame({"id": ["a", "b", "b", "b"], "time": [4.0, 0.0, 1.0, 5.0],
                       "mark": ["cvd_death", "diagnosis", "visit", "visit"]})
model = gamfit.fit_joint_event_model(subjects, events, marks={"diagnosis": "once", "cvd_death": "terminal",
                                                              "other_death": "terminal", "visit": "recurrent"})
gamfit.save(model, path)          # writes {"kind": "joint", "version": 3, "model": ...}, 1990 bytes
gamfit.load(path)                 # GamError: saved model payload is missing its payload object
```

Output:
```
SAVED 1990 bytes; header {   "kind": "joint",   "version": 3,   "model": {     "schem
KIND scalar
LOAD_FAILED GamError saved model payload is missing its payload object
LOADS_FAILED GamError saved model payload is missing its payload object
```

## Root cause

- `gamfit.loads` (`gamfit/_api.py:1380-1430`) chooses a loader by the kind that the Rust `saved_model_kind` returns (`crates/gam-pyffi/src/model/model_ffi.rs:1537-1575`).
- `saved_model_kind` recognises three headers: the manifold SAE `schema`, the response-geometry `schema`, and the multinomial `model_class`. It returns `"scalar"` for everything else.
- It has no case for the shared saved-model envelope `{kind, version, model}` (`crates/gam-model-api/src/saved_model.rs`), which joint models have been saved in since 3a519357d9 (#2961, #2966). So a joint document goes to the scalar `Model` archive loader, which fails because the document has no `payload` object.
- The same gap affects every future envelope kind. The event-history predictor (#2966) is the next.

## Fix (gam-2966, landing now as its own commit)

[...]

PR #3454, opened 2026-09-20
Write every saved model durably through the one Rust writer

SauersML:
Part of #3054.

**Slice (a): durable, atomic `write_saved_model`** (`crates/gam-model-api/src/saved_model.rs`)
- It writes to a sibling temp file, then calls `write_all` and `sync_all`, then renames over the target. On Unix it also fsyncs the parent directory. If any step fails, the temp file is removed, so the target path is either the whole new document or untouched.
- `SavedModelError::Io { path, source: std::io::Error }` keeps the OS error kind instead of flattening it to a string.
- If the directory sync fails after the rename, it returns an error saying so: the model is in place, but its durability is unknown.
- The joint event-history model's save goes through this writer.

**Slice (c): the gamfit saves go through the same writer**
- The new `write_saved_model_file(path, model_bytes)` pyfunction backs `Model.save`, `MultinomialModel.save` and the response-geometry save. Python no longer opens and writes the file itself ("Python should be a thin wrapper over Rust").
- `saved_document_error_to_pyerr` maps `Io` to the matching Python `OSError` subclass, for example `FileNotFoundError`, and names the path. Any other saved-document error maps to `DataError`.

**Left out of this PR**
- Slice (b), `FittedModel::save_to_path`, is covered by #3350.
- Slice (d), the SAE saves, waits on the gam-sae → gam-interp rename.

**Tests**
- Rust: `a_save_replaces_the_document_whole_or_leaves_the_path_untouched_3054`, plus tests that NotFound and InvalidInput come through as those kinds.
- Python: `tests/saved_model_writer_is_atomic_3054_test.py`.

https://claude.ai/code/session_01LjoXsgfjuoEh2yjtoyn9GY

---
_Generated by [Claude Code](https://claude.ai/code/session_01LjoXsgfjuoEh2yjtoyn9GY)_


SauersML:
Separating case for this PR (from the #3053/#3054 lane, whose L1-L3 cover the same slices (a) and (c)): how a refused saved document is raised.

`saved_document_error_to_pyerr` sends every refusal except `Io` to `DataError(refused.to_string())`. That drops two things main already fixed for load refusals in 1308538a8f (#3008):

1. **The schema-mismatch class.** `SavedModelError::Kind` and `SavedModelError::Version` are a document of another kind or version. The class gamfit raises for that is `SchemaMismatchError`, which is a `DataError`. `gamfit.loads` already raises `SchemaMismatchError` for an envelope kind it does not read. With this PR, the same condition reached through `load_joint_event_model` raises a plain `DataError`, so one refusal gets two classes.
2. **The variant and category.** #3008 made "the class of its category, with the `variant:` / `category:` lines and attributes" the contract for every load refusal. `saved_model_error_to_pyerr` does this for `FittedModelError`, and a SchemaMismatch keeps its own variant there. The envelope refusals here carry neither.

The case: save a joint event model, set the document's `"version"` to 0, and load it.
- With this PR: `DataError("the saved joint model has version Some(0); ...")`. `isinstance(err, SchemaMismatchError)` is False, and `err.variant` raises AttributeError.
- With L3: `SchemaMismatchError`, with `variant == "SavedModelError::Version"` and `category == "data"`, in the message and as attributes.
- `Malformed` and `Inconsistent` are `DataError` under both, but only L3 names the variant. `Io` is the same under both: the `OSError` subclass of its kind.

L3 does this with one exhaustive match (no wildcard arm). It sets the attributes through a helper it shares with `saved_model_error_to_pyerr`, so #3008's behaviour is unchanged. A Rust test covers every variant: a real reader refusal for kind, version and malformed input, an owner's `Inconsistent`, and an absent file giving `FileNotFoundError`, which is not a `GamfitError`.

Where this PR is better than L3: it releases the GIL for the write (`py.detach`) and takes the document as bytes. I will keep both.

Plan once this PR merges, rather than a second writer: L2 and L3's writer halves are dropped as duplicates. On top of this PR go #3053's fix (gamfit.load of every envelope kind, including the joint document reader), then the contract above applied to `saved_document_error_to_pyerr`, with its tests.

PR #3460, opened 2026-09-20
bms: name every selected-GPU consumer in the #932 fail-closed census

### LINKED ISSUE #3457: main red: bms::flex_verify_932_tests::selected_gpu_consumers_cannot_retry_on_cpu_932 counts 10 require_selected_gpu_result calls, expects 9 (since cf9fe687)
## Failure

On main 1c84d2a4, `cargo test -p gam-models --lib bms::flex_verify_932_tests::selected_gpu_consumers_cannot_retry_on_cpu_932` fails:

```text
panicked at crates/gam-models/src/bms/flex_verify_932_tests.rs:1800:5:
assertion `left == right` failed: seven HVP/diagonal dispatches plus the joint-gradient and dense cache-boundary adapters must share the fail-closed contract
  left: 10
 right: 9
```

## Cause

cf9fe687 ("gpu: BMS FLEX row-primary Hessian races the whole inner step", #3024 slice B) added one more `crate::bms::gpu::flex::require_selected_gpu_result(` call site in `crates/gam-models/src/bms/axis_direction_search.rs`. It did not update the census in `flex_verify_932_tests.rs:1800` or that census's message.

## Fix

Decide whether the new call site is a legitimate fail-closed consumer:
- **If it is:** raise the census to 10 and name the new consumer in the message.
- **If it is not:** fold it into an existing adapter.

The test's intent is that every selected-GPU consumer is enumerated. Only raising the number would weaken that, so do not do that alone.

Found while validating #3447, whose diff does not touch these files.

SauersML:
Fixes #3457

This fixes the census that went red after cf9fe687 (#3024 slice B) and makes it name each consumer.

- **The new call site is legitimate.** cf9fe687 added a tenth `require_selected_gpu_result` call in `axis_direction_search.rs`, the raced joint-Hessian HVP. `ReusedStateRace::apply` propagates the device executor's error through `?` and marks the ledger faulted. It does not retry on CPU, so the site is a genuine fail-closed consumer.
- **The census now names consumers.** It used to count call sites (`== 9`), so it failed without saying which consumer was new. It now collects the operation argument of every call, sorts them, and compares against an enumerated list:
  - eight labelled HVP/diagonal dispatches;
  - two `operation` entries, one each for the joint-gradient and dense cache-boundary adapters.

  An added or removed consumer now fails by name.

https://claude.ai/code/session_01CJBw9tSUsL4XWv47vBGm3U

---
_Generated by [Claude Code](https://claude.ai/code/session_01CJBw9tSUsL4XWv47vBGm3U)_


SauersML:
Review from local agent AP-5. I have not pushed to this branch.

The diff is sound. On origin/main c663ba6dba, `crates/gam-models/src/bms/axis_direction_search.rs` has exactly ten `require_selected_gpu_result(` call sites, at lines 21, 49, 309, 343, 460, 595, 821, 942, 975 and 1110.
- Their first arguments, sorted by byte order, are exactly the enumerated list: eight quoted labels, then two `operation`.
- The site at :343 (`"raced joint-Hessian HVP"`) returns its device error through `racing.race.apply(...)` and never recomputes on the CPU, so it is a real fail-closed consumer.

`Vec<&str> == [&str; 10]` compiles through `PartialEq<[U; N]> for Vec<T>`.

I dispatched focused-rust-proof run 35498629130 (gam-models lib, filter `selected_gpu_consumers_cannot_retry_on_cpu_932`) on head 0603e621. That dispatch displaced run 35498600561, which had identical inputs. I will merge on green if nobody has pushed.

---

Issue #2965, opened 2026-09-17
Event history: computed_joint replays the whole cohort path once per coefficient pair, p(p+1)/2 path evaluations per Hessian, over a G^K latent grid


SauersML:
## Defect

Reference-centred latent models, and models with a held zero-rate (static) atom, assemble the coefficient Hessian by replaying the complete cohort path once per coefficient pair. The latent state is a product grid whose size is exponential in the number of factors. Neither can be the principal fitting route for many diseases and a biobank-scale cohort.

## Source (read at origin/main 5b2af0f350)

- `family.rs:498-499`: `if self.atoms > 0 && (self.reference.is_some() || self.held_rates.contains(&Some(0.0))) { return self.computed_joint(...) }`.
- `objective.rs:108-134` `computed_joint`: `for i in 0..total { for j in 0..=i { … self.path_value(states, &seeded)? … } }`.
- Each iteration seeds `Mixed` jets on `(i, j)` and calls `path_value`. That call evaluates the reference law (`reference_values`) and every subject's marginal (`par_iter`) in full.
- Count from the loops (not a timing): `p (p+1) / 2` full path evaluations per Hessian, so **500,500 at p = 1000**.
- Product grid: `G^K` points. At G = 9 that is 59,049 at K = 5; 43,046,721 at K = 8 (344 MB for one f64 vector); 3,486,784,401 at K = 10 (27.9 GB for one vector).

## Required correction

- The principal route uses matrix-free derivative products that exploit temporal and state structure:
  - a gradient from one jet sweep or from analytic scores;
  - Hessian-vector products or blockwise curvature for the Newton/trust-region step;
  - no dense all-pairs replay.
- Where a dense Hessian is still needed at small p, one sweep per column (p evaluations) is the ceiling. Pair replay is not.
- The exact product-grid engine stays as the small-state numerical reference for the SAME law. The scalable route is the structured state-space posterior in the joint model issue.
- Numerical infeasibility stays distinct from evidence for a low rank. A memory guard may refuse; it must not select a structure.

## Measurement plan

Report the wall time and peak RSS of one Hessian at p ∈ {20, 80, 320}, before and after, on the same fixture and sha. Confirm agreement with the pair-replay Hessian at small p.

Related: #2961 (acceptance A10); #2901 (directional-Hessian duals gated against the hand cumulant route).

SauersML:
evh-hessian: before-arm measurement, at main `645a27b743`

[...]

PR #3462, opened 2026-09-20
Event history: place each node's smoothed marginal on its own grid (#3013 slice 1)

SauersML:
Refs #3013 (slice 1: the re-placed smoother).

## Root cause
`backward_smoother` re-used each node's filtered grid for the smoothed marginal. On a slow atom, an early node's filtered density is as wide as the prior. The smoothed marginal has seen the whole history: it sits about 2σ off the filtered grid's centre and is about 5× narrower. A degree-(G−1) representation on the filtered grid resolves it only at high order.

## Change
- **Grid placement.** Each node n < N−1 gets its own smoothed grid, placed one-shot at the Rauch-Tung-Striebel moments. These combine node n's filtered moments with node n+1's smoothed moments through the gap's φ, q:
  - v̂ = φ²v_f + q, J = φv_f/v̂;
  - m_s = m_f + J(m′ − φm_f), v_s = v_f q/v̂ + J²v′.
- **Density on that grid.** p̂_n · lik_n · β_n is evaluated there:
  - p̂_n by the forward operator from filtered grid n−1, or the prior at n = 0;
  - lik_n exactly;
  - β_n by the inner Gauss-Hermite rule, with ln β_{n+1} interpolated on grid n+1's smoothed grid.
  - The last node, and every node of a static chain, keep the filtered grid.
- **Louis sweep.** Node m's expectations and gap innovation moments are taken on the smoothed grid. The carried vector stays on the filtered grids for propagation. The same forward operators also move it onto the next smoothed grid for the contractions, so it is never interpolated without its envelope.
- **`latent_state_moments`** reads each node's smoothed grid.

## Tests
New test: `slow_atom_first_smoothed_marginal_is_resolved_at_the_filter_accuracy_3013`.
- **Fixture.** The issue's chain: 41 recurrent events on [0, 6], η = 0.53, a = 0.845, ν = 2.6e-3.
- **Reference.** A dense forward-backward pass. Its own error, measured by doubling dz, is 6.9e-15.
- **Bar.** The forward filter's own error at the same order, read at the last node, where smoothed and filtered coincide.

| G | node-0 mean err | node-0 sd err | bar | mutant (filtered grid) |
|---|---|---|---|---|
| 9 | +1.2e-6 | +3.8e-6 | 4.3e-4 | −3.6e-1, fails |
| 17 | +7.6e-7 | +1.6e-7 | 1.4e-4 | |
| 33 | +3.0e-7 | +2.9e-8 | 7.2e-6 | |

Other checks:
- `cargo test -p gam-event-history --lib`: 134 passed, 0 failed (1959 s).
- `cargo check --workspace --tests`: clean, no warnings.

Still open on #3013: slice 2 (the Louis carried vector on smoothed grids), slice 3 (a stable spectral ln β representation) and S0.


SauersML:
Cross-check from #2965 step B (reverse sweeps, gating now on 8256e9b70f, landing ~09Z UTC). Read at head 792915265e; not run yet.

**No collision with step B.**
- This PR changes only `marginal.rs`, in `filter_start`'s body, `subject_marginal`, `filter_nodes`, `backward_smoother` and `latent_state_moments`.
- Step B changes `adjoint.rs` (new), `family.rs`, `objective.rs`, `preserve.rs` and `scalar.rs`. Its reverse sweep re-runs `filter_start`, `filter_step`, `condition`, `node_likelihood` and `transitions_across` on a tape.
- None of those signatures change here. The hoisted `prior_density` does the same multiplications in the same order (constant 1, then one `normal_density` factor per atom), so the taped value stays bit for bit the forward filter's.

**On #3013's S0 cost gap.**
- The parked S0 sends every dynamic atom to the computed path. It is measured on #3013:
  - `a_multi_mark_rank_two_cohort_does_not_run_away` takes 238.6 s on the Louis sweep, and runs past a 900 s cap on the computed path;
  - the target is a reverse tape at a cost ratio of about 3.
- This PR does not touch the computed path, so it does not close that gap. What it can do is remove the reason for S0: S0 exists because the Louis Hessian is wrong for dynamic (slow) atoms.
- Whether it does is measurable once step B is on main. Two measurements decide it:
  1. Louis H against the computed path's H (reverse sweep, cross-checked by the forward pair replay) on a slow dynamic atom, main against this head.
  2. The wall time of `a_multi_mark_rank_two_cohort_does_not_run_away`, since each node now places its own smoothed grid and runs extra forward operators.
- I'll run both on my cores after step B lands and post the numbers on #3013.

Issue #3339, opened 2026-09-20
An interior box on an unpenalised coefficient refuses: P-IRLS gradient_natural_scale is the score norm, which cancels at the optimum (constrain(x, min=-1, max=1, double_penalty=false))


SauersML:
## What fails

A box that does not bind, on an unpenalised coefficient, refuses to fit on main (366df1d1f8):

```
y ~ constrain(x, min=-1, max=1, double_penalty=false)
Parameter constraint violation: KKT residuals exceed tolerance: primal=0.000e0, dual=0.000e0, comp=0.000e0,
stat=3.664e-15 (stat_rel=9.789e-1 vs tol=5.000e-6; gradient_scale=3.743e-15); active=0/2; unreachable=3.664e-15 blocked=0.000e0
```

The data is the #3268 fixture (`tests/regressions/misc/bounded_default_prior_shrinks_a_null_coefficient.rs`): n = 200, `x` on an even grid over [-1, 1], `y = 1 + noise` with the noise's least-squares slope pinned at 0.0305. On the same data:
- `y ~ linear(x, double_penalty=false)` fits: slope 0.030465909502;
- `y ~ constrain(x, min=1, max=3, double_penalty=false)`, where the box binds, fits: 1.00817, the truncated mean (#2705).

So the constrained route refuses exactly where its answer should equal the unconstrained one. Lane probe: dy7 `w2saeb-p3289c`, `/scratch.global/sauer354/sw-status/direct-dy7-w2saeb-p3289c.log.test`, lines tagged `[P3289]`.

## Root cause

- d19e847aac (via #3232) made the constraint-KKT gate purely relative: `residual <= tol * gradient_scale`.
  - Its premise: the scale "stays large at an optimum where g cancels: ‖Hβ‖ + ‖rhs‖ in the metric projection, ‖score‖ + ‖S_λβ‖ (gradient_natural_scale) for P-IRLS".
  - The metric projection's half is true, since its operands are `Hβ` and `rhs`. The P-IRLS half is not. `score_norm` is `‖Xᵀ W(η − z)‖` (`pirls/gam_working_model.rs:1501`, `pirls/loop_driver.rs:1231`), which is the cancelled score itself, not its operands.
- At an exact interior optimum of an unpenalised coefficient, `S_λβ = 0` and the score is rounding-level. So `gradient_scale` = 3.7e-15 and the stationarity ratio reads about 1 on a fully converged fit.
[...]

PR #3464, opened 2026-09-20
Delete the never-realized coefficient prior mean, so the score test's Hβ̂ = XᵀWz identity holds for every fit (#3443)

### LINKED ISSUE #3443: Smooth score test reads H·β̂ as X'Wz under a prior-mean-centred penalty (DenseWithMean)
Found while fixing #3346 (PR #3389, merged as ed78ff56).

**What the code does.** `ScoreTestFit::of` (`crates/gam-solve/src/estimate/smooth_term_summary.rs`) reads the smooth score from `b = H·β̂` through the stationarity identity `H·β̂ = XᵀWz`. That identity holds only for a penalty centred at the origin.

**What #3389 covered.** It fixes the gauge's own affine shift: under `β = Tθ + a` the score is read from `H·(β̂ − a)` via `UnifiedFitResult::beta_from_gauge_shift`.

**What is still open.** A penalty with a non-zero prior mean `μ` gives the same kind of stationarity error, and nothing corrects for it. The source is `PenaltySpec::DenseWithMean`, which coefficient-group priors emit (`crates/gam-terms/src/smooth/penalty_priors.rs`, `CoefficientPriorMean::constant`).

Stationarity for that penalty is `XᵀW(z − Xβ̂) = Σ_k λ_k S_k (β̂ − μ_k)`, so `H·β̂ = XᵀWz + Σ_k λ_k S_k μ_k`. Two consequences:
- **Wrong score:** read from `H·β̂`, it carries the extra term `Σ_k λ_k S_k μ_k`.
- **Wrong null:** the `τ = 0` null pins the tested term at `μ`, not at 0.

**Not yet verified:** whether a term carrying a `DenseWithMean` penalty reaches `smooth_term_summary_rows` today. If it does, its p-value is miscalibrated.

**Proposed fix.** Carry the realised prior mean in the fit and measure the score from it. That is, test with `β̂ − μ`, the same way `beta_from_gauge_shift` handles `a`. A fit that does not carry `μ` should be refused with `FitCurvatureUnavailable` instead.

SauersML:
Closes #3443.

## Finding
The issue asks whether a term with a `DenseWithMean` penalty ever reaches `smooth_term_summary_rows`. No production route builds a non-zero coefficient prior mean:
- The only emitter of `PenaltySpec::DenseWithMean` is `realize_coefficient_groups`, which read `CoefficientGroupSpec::prior_mean`.
- That field is `#[serde(skip, default)]`, and no Rust code builds a `CoefficientGroupSpec`, so the mean was always zero.
- Every `BlockwisePenalty` constructor sets `CoefficientPriorMean::Zero`.

Several consumers already drop the mean silently: `BlockwisePenalty::to_penalty_matrix`, the GAMLSS/survival/parameter-block `Dense | DenseWithMean { matrix, .. }` arms, `joint_hyper`, and `gam-inference` sampling. If a non-zero mean were ever wired in, the smooth score test would be miscalibrated and those routes would fit a different objective than the PIRLS path does.

## Fix

This PR deletes the option end to end:
- `CoefficientPriorMean` and `PenaltySpec::DenseWithMean`.
- The `prior_mean` fields on `BlockwisePenalty`, `CanonicalPenalty` and `CoefficientGroupSpec`.
- The centred `PenaltyCoordinate` variants.
- The PIRLS linear/constant penalty shift, along with the matching GPU dispatch arguments and CUDA kernel parameters.

Afterwards every penalty is centred at the origin, and `Hβ̂ = XᵀWz` holds exactly. The one remaining shift is the gauge's affine `a`, which #3389 already handles.

I rejected the issue's proposed "test with β̂ − μ". It would move the null to β_j = μ_j, whereas the test's null is f_j = 0. Subtracting Σλ_k S_k μ_k from Hβ̂ fixes the score but not that null.

**Tests.** Deleted the tests that only exercised the removed centring:
- `penalty_prior_mean_hard.rs`
- `coefficient_group_constant_prior_mean_shrinks_toward_mean`
- `coefficient_group_zero_prior_mean_matches_default_bits`
- The nonzero-mean case in `coefficient_groups_hierarchical_hard.rs`

The other tests just drop the field.

**Merge with main.** Merging main brought in #3339's `penalized_gradient_natural_scale`. At every PIRLS/GPU site that took it, `s_beta` is now plain `S·β`.

## Checklist
- [x] Group penalties emit `PenaltySpec::Dense`
- [x] Delete `CoefficientPriorMean` and `DenseWithMean` through gam-problem / gam-terms / gam-solve / gam-models / gam-inference / GPU
- [x] Score-test docs: the identity is exact for every fit
- [x] `cargo check --workspace --tests` is clean (the one warning, an unused `IxDyn` in gam-pyffi, is already on main)
- [x] `cargo test -p gam-solve --lib -- pirls::`: 168 passed
- [ ] Focused proofs pending: inference `coefficient_group`, basis_smooth `curvature_certificate_penalty_invariance`, regressions `bug_hunt_2_3`


SauersML:
BF-55 (continuing this PR after the #3443 takeover): I merged origin/main f10039cc74 in 903d37f53b. Two things needed resolving:
- `standard_conformal_penalty` in `model_payload_builders.rs` is new on main and read `penalty.prior_mean`. Its `shifted_prior` gate is removed, because every penalty is now origin-centred. The inequality-constraint exclusion stays.
- `reml/objective.rs` had a conflict in a test module. Main had already deleted `adaptive_ift_max_drho` together with its test, so I took main's side.

I grepped the merged tree: no `CoefficientPriorMean`, `DenseWithMean`, `*RootCentered` or penalty `prior_mean` remains.

On the previous head, the inference `coefficient_group` proof (run 35499935074) went 12 of 13. The one failure is `nested_groups_with_gamma_priors_apply_per_level_shrinkage`: an outer non-stationary refusal with the parent ρ railed at the lower end, `|Pg| = 2.47e-3` against a bound of `1.55e-3`. That fixture's groups all had `prior_mean: Zero`, so the fitted objective is unchanged by this PR. I've queued an attribution run of the same test on main f10039cc74 next to the branch proofs:
- `BF-55-3443-1`: inference `coefficient_group`
- `BF-55-3443-main`: same test on main
- `BF-55-3443-2`: basis_smooth `curvature_certificate_penalty_invariance`
- `BF-55-3443-3`: gam-solve lib `pirls::` / `penalty_invariance`

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, dcde73b2038f.

PR #3478, opened 2026-09-20
Bounded Gaussian profiled dispersion: certified bracketing root solve on ln φ (fixes #3450)

### LINKED ISSUE #3450: bounded_default_prior_shrinks_a_null_coefficient_to_zero red on main: bounded Gaussian dispersion profile does not reach its fixed point in 50 refits (last phi = 0.2215)
## Symptom

On main (reproduced at `ecb7adcc`, and again on a branch merged with `53dfd647`):

```
cargo test --test regressions bounded_default_prior_shrinks_a_null_coefficient_to_zero
thread 'misc::bounded_default_prior_shrinks_a_null_coefficient::bounded_default_prior_shrinks_a_null_coefficient_to_zero' panicked at tests/regressions/misc/bounded_default_prior_shrinks_a_null_coefficient.rs:103:23:
`y ~ bounded(x, min=-1, max=1)` must fit: Invalid input: bounded Gaussian dispersion did not reach its profiled fixed point in 50 refits (last phi = 0.22149869057589253)
```

The other seven `bounded_` regressions pass on the same build.

## Where

`crates/gam-models/src/fit_orchestration/drivers/design_construction.rs`, at lines 3185-3240. The loop is `for _ in 0..BOUNDED_GAUSSIAN_SCALE_MAX_ITER`, and `BOUNDED_GAUSSIAN_SCALE_MAX_ITER = 50` is set at line 2675.

On each pass the loop:
1. profiles `phi_next = rss / (n - edf)`;
2. refits at `FixedDispersion { phi: phi_next }`;
3. stops when either `|ln phi_next − ln phi| ≤ 1e-8` or beta stops moving by `1e-8·(1+|b|)`.

In this test neither criterion is met within 50 refits.

## Why this is a defect, not only a budget

SPEC.md: "A fit object must only ever come from a converged optimization." and "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code."

The fixed-point map `phi -> rss(beta(phi)) / (n - edf(phi))` is iterated as plain successive substitution. When the map's slope at the fixed point is close to 1 it converges linearly and slowly. It can also oscillate, if the inner fit's lambda rescaling makes it non-monotone. In either case a 50-count cap is arbitrary, and the tolerance check can then mistake a stall for convergence or report failure on a sequence that is actually converging.

The principled fix is to stop treating the dispersion as an outer fixed point. It should be profiled out of the objective, i.e. the scale-profiled Gaussian REML/LAML criterion that the standard Gaussian path already uses, so that one converged outer optimization produces `phi`. Failing that, the loop should at least solve the scalar fixed point `phi = g(phi)` with a certified 1-D root solve on `ln phi` (for example a safeguarded secant on `ln g(phi) − ln phi`) instead of a counted substitution.

Found while validating #3339 (PR #3420). That PR does not touch this path, and the failure is identical with and without it.

SauersML:
Fixes #3450

## Diagnosis
The bounded fit profiled its Gaussian dispersion by iterating φ ← RSS/(n − edf), with a budget of 50 refits. It stopped when ln φ moved by ≤ 1e-8 or when β stopped moving by 1e-8(1+|β|).

Each g(φ) is read off a refit whose λ is resolved only to the outer tolerance. On the null-coefficient fixture the REML criterion is flat in ρ, so edf wanders by about 5e-4 between refits at essentially the same φ (1.00411–1.00499), and h = ln g − ln φ carries about 1e-6 of refit noise.

Instrumented trajectory (u = ln φ):

| u | h | edf |
|---|---|---|
| 0 | −1.507 | 1.00465 |
| −1.507331003 | −3.9e-7 | 1.00447 |
| −1.507331395 | +1.15e-6 | 1.00499 |
| −1.507331060 | +3.8e-7 | 1.00479 |
| −1.507331014 | −7.5e-8 | 1.00461 |

The first substitution already lands within 4e-7 of the root. After that, neither stopping rule can fire, and the fit was refused ("last phi = 0.2215" = e^{−1.5073}).

## Fix
- The dispersion is the root of h(u) = ln g(e^u) − u. Along the profile, dV/du = ((n−edf)/2)(1 − e^h), so this root is a stationary point of the profiled criterion.
- g is bounded, so h → −∞ as u → +∞. Unless the fit is exact, h → +∞ as u → −∞. A sign change therefore exists on the side h points to.
- Bracket: the first step is the substitution; later steps take the secant if it points forward and reaches farther, and otherwise double the last step.
- Narrow: Illinois regula falsi until the bracket is ≤ 1e-8 wide on ln φ. The probe with the smaller |h| is reported.
- The exact-fit exit (g = 0) is kept. The 50-refit budget and the β-motion stop are deleted.

On the fixture, the bracket closes in 8 refits.

## Validation
`cargo test --test regressions bounded_`: all 7 pass, including `bounded_default_prior_shrinks_a_null_coefficient_to_zero`.

https://claude.ai/code/session_01W5v9mtzjtTtPqec9DNtojF

---
_Generated by [Claude Code](https://claude.ai/code/session_01W5v9mtzjtTtPqec9DNtojF)_


SauersML:
Review note (cloud-merge review, w2-mpd, gam-86 wave 2): this PR reuses an undeclared literal as its bracket width.

- The new Illinois loop narrows until `(far.u - near.u).abs() <= BOUNDED_GAUSSIAN_SCALE_TOL`, and `settled` accepts `|h| <= BOUNDED_GAUSSIAN_SCALE_TOL`.
- That constant is `1e-8` at `crates/gam-models/src/fit_orchestration/drivers/design_construction.rs:2677`. It predates this PR, where it was the beta-motion stop's tolerance. Its doc comment states what it is used for, not where it comes from.
- The PR's own trajectory table shows `h = ln g - ln phi` carrying about `1e-6` of refit noise: edf wanders by about 5e-4 between refits at the outer tolerance. So the bracket is narrowed to a width about 100x below the noise in the function it brackets. Once the bracket is inside that noise, the sign of `h` at each probe is set by the noise, not by the root.
- This does not break termination: every Illinois probe shrinks the bracket, so the loop ends. But the stop is a magic literal. The derived width is the resolution of `h` itself, which is what the refits' outer tolerance leaves in `ln g`. Refinement below that spends refits without moving the answer.

Nothing needs to change for correctness today. This is logged so the width gets derived from the refit's own resolution.

Issue #944, opened 2026-06-09
RESEARCH: Curvature as an estimand — a continuous constant-curvature manifold family with exact-REML κ̂ and confidence intervals


SauersML:
## The open problem

The machine-learning literature *chooses* latent geometry: hyperbolic embeddings for hierarchies, spherical latents for directional data, Euclidean by default — picked by architecture fiat or held-out loss, with no inferential statement attached. "What curvature does this data actually support, and with what uncertainty?" has no practical answer anywhere: existing curvature-estimation work gives point estimates in toy settings with no likelihood, no confidence intervals, no model-based coverage. Geometry is the last major modeling choice still made without statistics.

## Why gam is uniquely positioned

The geometry layer already implements the three constant-curvature spaces *separately* — src/geometry/sphere.rs (κ>0), euclidean.rs (κ=0), poincare.rs (κ<0) — with exact exp/log/distance, pinned by analytic ground-truth tests (recent commits pin the Poincaré radial isometry and Grassmann geodesics). The engine already moves **design-moving hyperparameters through exact LAML derivatives**: the ψ-derivative machinery in penalty_logdet.rs handles parameters that move the penalty matrices themselves, and the Matérn-κ work proved the full pattern (basis κ-derivatives verified correct). And the topology-selection layer (#768 stacking) already adjudicates discrete geometric candidates. All the parts exist; what's missing is the *continuous family* between them.

## The method

1. **One chart, one parameter**: implement `ConstantCurvature { kappa }` in src/geometry/ using the gnomonic/stereographic unified model, where exp, log, distance, and the volume element are analytic in κ and smoothly interpolate S^d(1/√κ) → ℝ^d → H^d(1/√−κ) through κ=0 (the κ→0 limits are removable — Taylor-stable expressions near zero, the same discipline as the C∞ sphere jets already in the tree).
2. **κ as a ψ-parameter**: basis functions, penalties, and (for latent fits) the seed geometry all become κ-differentiable; κ joins the outer optimization with *exact* LAML derivatives via the established ψ-channel. No new outer machinery — κ is one more design-moving coordinate, with FD verification per the standing recipe.
3. **Inference**: profile-likelihood CI for κ from the exact outer curvature; the test of κ=0 is the first likelihood-based "is my latent space flat?" test with controlled size. Sign changes of the CI endpoints give "hyperbolic vs spherical" as a hypothesis test rather than an architecture decision.
[...]

PR #3485, opened 2026-09-20
Curvature inference consumes the outer κ solve's stationarity certificate (#3453)

### LINKED ISSUE #3245: √ε and relative cost floors still set outer tolerances in response_geometry, spatial curvature inference and spatial_optimization
SPEC.md bans magic constants and loosened tolerances: a tolerance must come from a derived resolution, not from a machine constant or a relative cost floor. I removed the same floor from the constant-curvature κ route. There the canonical `tol = 1e-10` was replaced by `options.tol.max(√ε) = 1.49e-8`, which loosened that one route's band about 150x. These sites still build a tolerance the same way. They are outside my area (outer-cert-a), so I'm recording them here.

| file:line | construction | problem |
|---|---|---|
| `crates/gam-models/src/response_geometry.rs:473` | `with_required_projected_gradient_norm(Some(√ε · n_outputs))` | Absolute gradient band = machine constant × output count. It grows with `n_outputs` and has no units of the criterion. The derived band is the τ_stat / Higham running-error band that the outer certificate already builds. |
| `crates/gam-models/src/response_geometry.rs:1408` | `√ε · metric.nrows() · scale` | The same construction, used as an acceptance tolerance. |
| `crates/gam-models/src/fit_orchestration/drivers/spatial_curvature_inference.rs:176` | `relative_tolerance.max(√ε)`, then `score_tolerance = rel·(1 + \|V̂\|)` | A floor on a relative tolerance, then an `rel·(1+\|V\|)` criterion resolution. That is the form C3 deletes elsewhere. For inference endpoints the resolution is τ_stat = 1/(2·n_eff) plus the computed evaluation band. |
| `crates/gam-models/src/fit_orchestration/drivers/spatial_optimization.rs:257` | `options.tol.max(1e-8 · \|baseline\|).max(1e-12)` | A relative cost floor plus two magic constants in the joint-versus-baseline agreement gate. The line-320 diagnostic reports `\|baseline\|·√ε` beside it. |
| `crates/gam-models/src/multinomial_posterior.rs:87`, `:493` | `√ε`, `√ε · max(scale, 1)` | Symmetry and acceptance tolerances. They need either an error-analysis bound or to be dropped (a symmetric construction needs no check). |

**Repro.**
- Every site is a literal expression, so grep finds all of them: `grep -rn "EPSILON.sqrt()\|1e-8 \* baseline" crates/gam-models/src`.
- For response_geometry:473, fit any response-geometry model with `k` outputs. The required projected-gradient norm is `1.49e-8·k` whatever the criterion's scale, so a criterion rescaled by `c` certifies at a different stationarity.

**Rule:** SPEC.md, "no magic constants" and "never loosen tolerances". Replace each floor with the derived resolution: τ_stat from `OuterProblemSize`, plus the certificate's evaluation band.

### LINKED ISSUE #3453: curvature inference re-tests κ̂ stationarity against its own absolute bar, which conflicts with the outer solver's decrement certificate; blocks moving κ to ARC (#3201)
## Summary
`curvature_profile_ci_from_analytic_score` (`crates/gam-models/src/fit_orchestration/drivers/spatial_curvature_inference.rs:175-214`) re-checks the κ̂ that the certified outer κ solve returns. The check is `|V_p′(κ̂)| ≤ max(rt, √ε)·(1+|V_p|)`. The outer solve certifies stationarity differently: a curvature-denominated decrement against the criterion resolution (#2458, `decrement_bands`). So there are two certificates for one point, and they can disagree.

On main the κ search runs gradient-only BFGS, which happens to push |g| far enough to pass both. When κ runs ARC on its exact `V_p″` (#3201, draft PR #3406), ARC stops as soon as the canonical certificate holds, and the inference layer then rejects the point.

## Reproducer
On `claude/fix-3201-kappa-profile-exact-hessian` (b7d23c58; this removes `.with_prefer_gradient_only(true)` from `constant_curvature_kappa_problem`):

```
cargo test -p gam --test identifiability -- hyperbolic_truth_recovers_negative_kappa_and_rejects_flat profile_ci_covers_planted_hyperbolic_curvature_across_replicates
```

```
curvature inference rejected a non-stationary point estimate: kappa_hat=-1.175899226900997, score=-4.037109e-3, stationarity_bound=8.744898e-5; box=[-2.7734, 2.7734] ... classified=interior
curvature inference rejected a non-stationary point estimate: kappa_hat=-0.7573528899815112, score=-2.475995e-3, stationarity_bound=1.439961e-3; box=[-2.1607, 2.1607] ... classified=interior
```

Both tests pass on main (e8568910).

At the first point, `V_p″ ≈ 59` (the κ profile curvature seen in the same fixture family). The remaining Newton step is `g/H ≈ 7e-5` in κ, and the remaining decrease is `g²/(2H) ≈ 1.4e-7` in V_p. That is well inside the outer solver's criterion resolution, so the outer certificate is right to stop. The inference bar instead scales with `|V_p|`. It is not derived from the statistical resolution of the criterion, and it has a `√ε` floor, a magic constant.

## What a fix needs
The inference layer should consume the outer solve's stationarity certificate, or re-test with the same decrement-based criterion, rather than a second independent bar. The rail logic (`at_lower` / `at_upper`, where stationarity means an outward score) must be kept. It should be restated through the same projected, curvature-denominated certificate. Do not loosen the bar ad hoc.

## Context: pre-existing failures on main
These fail on both main and the branch, so they are not caused by this conflict:
- `flat_truth_does_not_reject_flatness`
- `kappa_zero_fit_recovers_planted_flat_signal`
- `flatness_test_holds_size_across_flat_replicates`
- `profile_ci_covers_planted_curvature_across_replicates`
- `production_like_binary_outcome_shared_matern_learned_kappa_starts_outer_solver`

Their messages are inner range-solve failures: ρ̂ railed at its bound, or η̂ left uncertified at ln ℓ ≈ 8.6–14.5 with `V_ηη < 0`. They are related to #3426.

SauersML:
Fixes #3453

## Problem
The #944 curvature-inference layer tested κ̂ a second time, against its own bar `|V_p'| ≤ max(rt, √ε)·(1+|V_p|)`. That bar has no derivation. It is in raw gradient units and scales with the additive level of the criterion. As a result it refused κ̂ values that the outer κ solve had already certified. The #3453 reproducer shows this: κ̂ = −1.1759, score −4.04e-3 against a bar of 8.74e-5, while the Newton decrement λ² = g²/H ≈ 2.8e-7 is far below τ_stat = 1/(2n).

## Fix
Every check in curvature inference now uses the outer solve's own resolution τ_stat = 1/(2n), for the same n that `constant_curvature_kappa_problem` declares through `with_problem_size`. τ_stat has one public definition: `gam_solve::rho_optimizer::outer_statistical_resolution`.

- **Stationarity.** The profile closure returns `(V_p, V_p', V_p'')`. κ̂ is refused only when the projected Newton decrement g²/H exceeds τ_stat. A concave point (H ≤ 0) with g ≠ 0 is never certified.
- **Rails.** κ̂ counts as on a rail when the second-order model cannot tell it from the rail at the statistical resolution: `|g|·gap + ½|H|·gap² ≤ τ_stat`. The score is projected onto the feasible side.
- **Endpoints.** Four checks are in the same units:
  - the direction-change refusal, which applies when the outward score is negative and the descent is resolvable;
  - the "bound undercuts V̂" refusal, which applies when `V(bound) < V̂ − τ_stat`;
  - Wilks-endpoint termination, which stops when the residual is ≤ τ_stat;
  - one final Newton refine inside the bracket.
- **Removed parameter.** `options.tol`, `x_tolerance` and the √ε floor are gone. The `options` parameter of `curvature_inference_forspec` is removed because nothing reads it. This is a breaking API change; callers in gam-pyffi and the tests are updated.
- **Also fixes #3245.** This resolves the `spatial_curvature_inference.rs` row of #3245 (`relative_tolerance.max(√ε)` / `rel·(1+|V̂|)`).

## Tests
- A new #3453 unit test covers three cases:
  - The certified point from the issue (g = −4.04e-3, H = 59, κ̂ = −1.1759 in [−3, 3]) is accepted at V levels 0, 5.8e2 and 1e7, so the check no longer depends on the additive level.
  - A point with g = 1.1·√(τH) is refused as non-stationary.
  - A concave point is refused.
- The quadratic-crossings test now asserts a Wilks residual ≤ τ_stat. It asserts endpoint error ≤ τ_stat/(H·half_width), the bound derived from convexity plus the tangent slope z√H at the crossing.
- Focused runs on the `identifiability` binary (`hyperbolic_truth_recovers_negative_kappa_and_rejects_flat`, `profile_ci_covers_planted_hyperbolic_curvature_across_replicates`, `constant_curvature`) and on the `gam-models` lib (`curvature_profile_score_tests`, `constant_curvature`).


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, ca361971d81f.

Issue #2904, opened 2026-09-12
survival location-scale inverse-link shape is chosen by a nested BFGS whose gradient is not its cost's derivative; select θ_link on family hyper axes of the joint LAML


SauersML:
Split out of #2695's status of record, where it was listed as an open mechanism owned by surv2695. None of #2695's named acceptance tests take this path.

## Defect

SAS, BetaLogistic and Mixture survival location-scale fits pick their inverse-link shape θ_link with a nested BFGS, `optimize_link_parameters` (`crates/gam-models/src/fit_orchestration/fit.rs`, reached when `run_survival.rs` sets `optimize_inverse_link`).

- **Cost:** `survival_inverse_link_profile_objective` (`fit.rs:197`) returns `−ℓ + ½·stable_penalty_term` at a complete fit where ρ is re-selected by REML at every θ probe.
- **Gradient:** `link_param_data_fit_gradient` (`survival/location_scale/row_kernel.rs:4103`) returns `∂(−ℓ)/∂θ` at the converged β̂. Its doc justifies this by the envelope theorem with the penalty independent of θ.

The envelope argument holds only at fixed λ and with no Jeffreys term in the inner mode. With ρ re-selected per probe, `dF/dθ` also carries `½·β̂ᵀ(∂S_λ/∂ρ)β̂·dρ̂/dθ`, and a Jeffreys-armed mode adds `∂Φ/∂β·dβ̂/dθ`. So the gradient handed to BFGS is not the derivative of the cost it minimizes. The cost itself is also a profile likelihood, not the LAML criterion that selects every other hyperparameter.

Also in the gradient kernel: `S = (1 − μ).clamp(f64::MIN_POSITIVE, 1.0)`, a production floor.

`1b97afdfb` removed this search's private `init ± 6` box, following the baseline-θ search in `survival/construction.rs`. The gradient mismatch above is unchanged.

## Plan

Select θ_link jointly with ρ by the LAML, as family-owned hyper axes:

1. Register θ_link as `family_axes` in the survival location-scale `CustomFamilyHyperLayout` (2 axes for SAS and BetaLogistic, `rho.len()` for Mixture). The precedent is BMS's learnable frailty σ axis (`bms/block_specs.rs:2782`, hooks in `bms/custom_family_impl.rs`).
[...]

PR #3487, opened 2026-09-20
Fit the survival LS baseline θ as family-owned outer hyper axes (#3413)

### LINKED ISSUE #3413: Survival location-scale baseline-θ search: the envelope gradient is not the gradient of its cost, because ρ is re-selected at every θ probe
**Where:** `crates/gam-models/src/fit_orchestration/materialize/survival.rs`, around lines 1190–1290 (the `optimize_survival_baseline_config_with_gradient_only` call for `SurvivalLikelihoodMode::LocationScale` with a non-Linear baseline target). Re #1561.

**Rules involved:**
- SPEC.md says REML/LAML only, and a fit comes only from a converged optimization. The objective the outer engine certifies must be the one whose derivative it is given.
- #1561 (convergence theory) requires that every certified stop be a stationary point of the criterion actually evaluated.

**The mismatch.**
- Each θ probe calls `fit_survival_location_scale_model`, which runs a full REML outer search over the threshold and log-σ smoothing parameters ρ. That search is warm-started from the previous probe's ρ̂ (`location_scale_smoothing_warm_start`, line 874).
- The cost returned to the θ search is the profile penalized NLL at the probe's own ρ̂(θ): `V(θ) = −ℓ(β̂(θ, ρ̂(θ)); θ) + ½ β̂ᵀ S_{ρ̂(θ)} β̂`.
- The gradient returned is the envelope contraction `Σ_i r_c,i ∂o_c,i/∂θ` (`baseline_chain_rule_gradient`, or `marginal_slope_baseline_chain_rule_gradient` for probit channels). That contraction is `∂V/∂θ` with ρ held fixed.
- The total derivative is `dV/dθ = ∂V/∂θ + Σ_k (∂V/∂ρ_k) · dρ̂_k/dθ`. At the REML optimum, `∂V/∂ρ_k = ½ λ_k β̂ᵀ S_k β̂` is not zero: REML stationarity sets it equal to `½ λ_k [tr(S_λ⁺ S_k) − tr(H⁻¹ S_k)]`, which vanishes only when the log-determinant terms cancel. `ρ̂` also moves with θ, because θ enters H through the offsets.
- So whenever the fit has at least one smoothing parameter, the gradient handed to BFGS is not the gradient of the cost it line-searches. The Wolfe conditions and the stationarity certificate then refer to two different functions.
- The warm start makes the cost path-dependent as well: when the REML surface in ρ has more than one local optimum, V(θ) depends on the order in which θ was probed.

**The criterion is also not the fit's criterion.** θ is chosen by penalized NLL, and ρ is then re-chosen by REML at that θ. Neither is LAML over (ρ, θ) jointly. For comparison, the transformation path in `fit_orchestration/fit.rs` (~3608) also selects θ by penalized NLL, but at fixed λ, so its gradient is consistent. It now carries the exact profile θ-Hessian (`transformation_baseline_profile_theta_hessian`, PR #3130).

**Repro sketch.** Any LS fit with `baseline_target=gompertz` and a penalized threshold smooth. At an interior θ probe, compare the analytic gradient against a central difference of the returned cost, where each ±δ probe runs a full `fit_survival_location_scale_model`. The difference is `Σ_k ½λ_k β̂ᵀS_kβ̂ · dρ̂_k/dθ`.

**Fix direction.** Put θ into the location-scale LAML outer vector alongside ρ, with its exact first and second θ-derivatives. The offsets enter the LS predictor additively, so `∂H/∂θ` is the same third-derivative contraction the ρ-path already uses, applied to the offset tangents. Then delete this separate θ search and its warm-start RefCell. This is not a retry, cap or fallback. It replaces two stages with one certified criterion.

SauersML:
Fixes #3413.

## Problem
The survival location-scale fit picked a nonlinear baseline's shape θ (Weibull, Gompertz, Gompertz–Makeham) with a nested BFGS in `materialize/survival.rs`. The cost it minimized was the profiled NLL. The gradient it used was an envelope formula built from offset-channel residuals and curvatures. Those are two different functions, so the nested search's value and gradient disagreed. Each probe of that search also re-ran a full LS fit.

## Fix
θ is now an ordinary family-owned coordinate of the outer criterion: θ_outer = [ρ | log κ | inverse-link shape | baseline θ].

- **Frozen offset chart** (`SurvivalLocationScaleBaselineChart`, construction.rs). The time stack is prepared once, at the seed baseline. The prepared offsets are split into the seed baseline b(seed) and a frozen residual: the derivative guard plus any user offset. Each outer point realizes `residual + b(θ)` on the entry, exit and exit-derivative channels, together with the per-row tangents ∂b/∂θ. The tangent is zero for rows entering at the origin. The probit-survival baseline channel is selected by the inverse link, as before. The θ domain comes from the existing `survival_baseline_theta_domain`; there is no hand-supplied box.
- **Exact ψ-terms through a pseudo column** (`location_scale/baseline_theta.rs`). An offset shift along tangent t is the same as a new time-block column t whose coefficient is held at 0. Prepending t as time coefficient 0 and reading the family's own exact row program at β₀ = 0 gives:
  - `objective_psi = −∂ℓ/∂β₀`
  - `score_psi = H[1..,0]`
  - `hessian_psi = D_βH[e₀][1..,1..]`
  - the Jeffreys drift `D²H[e₀,(0,u)][1..,1..]`

  This holds with and without the link and time wiggles. No second model of the likelihood is kept.
- **Outer setup** (`location_scale/fit.rs`). The baseline axes follow the link-shape axes in `CustomFamilyHyperLayout`. As with the link shape (#2904), the family serves first-order terms only, so a fit that carries them runs a gradient-only outer.
- **Materialize.** The nested optimizer is deleted. Collapsing the time warp while a nonlinear chart is in use is an error in both prepare and materialize; the collapse is not silently skipped.
- **Removed as dead code:**
  - `offset_channel_geometry`
  - `marginal_slope_baseline_chain_rule_gradient`
  - `location_scale_uses_probit_survival_baseline` from the fit_orchestration re-exports
  - the converged-geometry plumbing
  - the ρ warm-start seeds that only the nested search fed

  The transformation-normal path still uses its own `baseline_chain_rule_gradient` and is untouched.
- **Payload bug fixed along the way.** The saved LS payload persisted the seed baseline config instead of the fitted one. It now persists `fit.baseline_config`.

## Validation
- `location_scale_baseline_chart_tangents_match_finite_difference_3413`: chart tangents against central differences of the realized offsets on all three channels.
- `baseline_theta_joint_psi_terms_match_offset_finite_difference_3413`: `objective_psi`, `score_psi` and `hessian_psi` from the pseudo column against central differences of the family NLL, gradient and Hessian, with the offsets shifted along the tangent.
- The existing link-shape tests (#2904) and the LS tests are also run, through `focused-rust-proof.yml` on this branch.


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 4578ac3e6567.

Issue #2590, opened 2026-07-28
Every inner-solver Err(String) is laundered into InvalidInput, so rho-local refusals kill the fit instead of retreating


SauersML:
## Summary

`gam-custom-family`'s inner solver reports its refusals as `Err(String)`. A
blanket `From<String>` turns **every one of them** into
`CustomFamilyError::InvalidInput`, which is precisely the variant the outer
loop classifies as *not* a trial-point infeasibility. So "there is no Laplace
mode at this rho" — a statement about one rho — is fatal to the entire fit,
when the outer optimizer is explicitly built to step away from such a point.

## The chain, confirmed end to end

`crates/gam-problem/src/custom_family_error.rs`:

```rust
impl From<String> for CustomFamilyError {
    fn from(value: String) -> Self {
        Self::InvalidInput {
            context: "custom-family string boundary",
            reason: value,
        }
    }
}
```

`is_trial_point_infeasible()`, same file, with its own stated rule:

```rust
// The inner solve missed its KKT condition at THIS theta.
// The outer search can step away; the problem is fine.
Self::InnerSolveNotConverged { .. } => true,
// Everything else is a property of the configuration, the data, or the
// numerics, and does not become true or false by moving theta.
Self::InvalidInput { .. } => false,
```

`rho_optimizer/objective.rs`:

```rust
let kind = if err.is_trial_point_infeasible() {
    ObjectiveEvalKind::Recoverable
} else {
    ObjectiveEvalKind::Fatal
};
```

`rho_optimizer/run_plan.rs`, the seed loop — the branch that should have taken
this case:

```rust
Err(err) if err.is_recoverable() => { /* reject seed, continue 'seed_attempts */ }
Err(err) => {
    return Err(EstimationError::fatal_outer_evaluation("outer BFGS seed evaluation", ...));
}
```

## The observed failure names every link

```
outer smoothing optimization failed certified-fit validation after exhausting
strategy fallbacks: Fatal outer-objective evaluation failure (outer BFGS seed
evaluation): ... custom-family invalid input in custom-family string boundary:
joint Newton tentative convergence rejected by fresh exact returned-mode
curvature: lambda_min=-3.831474e-1 < -floor=7.711644e-15; an indefinite
coefficient point cannot define a Laplace mode (after 2 negative-curvature
escapes); last_evaluated_rho=[0.0, 0.0, 0.0, 0.0, 2.9828384627775835]
```

`custom-family invalid input in custom-family string boundary:` is the literal
[...]

Issue #2553, opened 2026-07-26
Outer-objective errors are classified fatal-vs-recoverable by MESSAGE SUBSTRING, so one variant gets both verdicts (a non-finite cost aborts a clean 120x8 circle fit in 90s)


SauersML:
## Summary

`ObjectiveEvalError` recoverability is decided **by substring-matching the rendered error message**,
so the same `EstimationError` variant is classified RECOVERABLE at one call site and FATAL at
another. A non-finite outer cost that one layer deliberately treats as an infeasible probe is
re-wrapped and killed as a fatal whole-fit failure by the layer above.

Measured: a clean 120x8 synthetic circle aborts the entire fit in 90 seconds.

## Repro (90 s, no GPU, gamfit 0.1.261)

```python
import numpy as np, gamfit
rng = np.random.RandomState(0)
n, p = 120, 8
th = rng.uniform(0, 2*np.pi, n)
X = np.zeros((n, p)); X[:,0] = np.cos(th); X[:,1] = np.sin(th)
X += 0.05*rng.randn(n, p)
gamfit.sae_manifold_fit(np.ascontiguousarray(X), K=1, d_atom=1,
    atom_topology="circle", assignment="threshold_gate",
    fisher_factors=None, n_iter=30, random_state=0)
```

```
RemlConvergenceError: Fatal outer-objective evaluation failure (outer fixed-point evaluation):
  REML smoothing optimization failed to converge:
  outer EFS eval failed: objective returned a non-finite cost
```

(The same script with `assignment="softmax"` or `"ordered_beta_bernoulli"` fails differently and
much later — 765 s and 647 s — so this is not the generic SAE wall; it is a distinct, fast, fatal
path. Details on #2228.)

## The two classifications

**Site A — `bridges.rs:3096-3100`, RECOVERABLE, deliberately:**

```rust
Err(err @ EstimationError::RemlOptimizationFailed(_)) => {
    return Err(ObjectiveEvalError::recoverable(format!("outer EFS eval failed: {err}")));
}
```

with a comment stating exactly why: *"Blanket-classifying it fatal turned one invalid startup seed
into a whole-fit `Fatal outer-objective evaluation failure`"* (#2367). The non-finite cost itself is
handled the same way twenty lines later at `bridges.rs:3105-3107` —
`ObjectiveEvalError::recoverable("outer EFS eval failed: objective returned a non-finite cost")`.

**Site B — `into_objective_error` in `objective.rs:1228-1253`, FATAL:**

[...]

PR #3488, opened 2026-09-20
Survival marginal-slope startup: keep typed ray evidence from seed validation (#3467)

### LINKED ISSUE #3467: Survival marginal-slope exact-joint startup refuses a seed whose only defect is a ρ-closable descending ray, and drops its typed evidence
## Defect

On the exact-joint route, the survival marginal-slope fit refuses its only derived seed at outer startup validation when the seed's one defect is a **ρ-closable** descending ray. The refusal names its own remedy, and the fit still refuses. The ray's typed evidence is also dropped before the #979 arming lifecycle can read it.

This is the class of #3345 (multinomial: only seed refused at outer startup), in the survival marginal-slope family.

## Repro

Both runs at gam main 89bea03263 plus a candidate that does not touch this route (sms-curv, #2945). n = 400, all events, `Surv(time, event) ~ x`, `slope_formula = "1"`, `baseline_target = "weibull"`, `time_num_internal_knots = 3`. Event times sit on the Weibull chart's own crossing, `T = λ·(−log Φ(z))^{1/k}` with log λ = 1.202 and log k = 0.516, so `Φ(−(q(t)·√(1+b²) + b·z))` fits the rows perfectly as b → ∞. Two runs:
- frozen standard-normal score;
- `latent_measure = "global-empirical"`, `frozen_score = false`.

Both fail in 1-3 s:

```
REML smoothing optimization failed to converge: no candidate seeds passed outer startup validation (n-block exact-joint spatial):
  generated=1, screened=1, exact_validated=1, solver_started=0
  rejection breakdown: ... rejected_other=1 (total=1)
  seed 0 (validation): ... custom-family inner solve did not converge after 44 cycle(s) [joint-Newton terminal cycle 43:
  stationarity_residual=2.667373e3 (tol=3.412203e-3) ... termination=residual 2.735813e3 stalled or grew against 3.375449e-3
  over 44 cycles while the accepted steps kept descending a direction with no finite minimizer in reach, so the seed is
  under-penalized rather than failed; block 0 is under-penalized along the accepted step (likelihood slope -3.643e-2,
  penalty slope 4.677e-4, block step 1.305e-4): the ray closes at 77.8782x its penalty strength, i.e. rho[0..1] += 4.3551]
  (this solver path emits no typed projected-KKT diagnostic ...)
```

`artifacts.jeffreys_arming_evidence` is never set, because the fit returns Err.

## Two parts

1. **Startup should move, not refuse.** The refusal proves that a finite minimizer exists at larger ρ (block 0, the time block: ρ[0..1] += 4.3551). So nothing here proves separation, and #979 ruling (b) does not call for arming. The seed stage should take the ρ step the diagnosis names instead of refusing the only seed.
2. **Evidence plumbing.** `fit_entry.rs` reads arming evidence from `refusal.jeffreys_arming_evidence()` or from `outer_refusal_evidence`, and that is captured only at outer-eval sites (~2189 and ~2211 on main). A refusal at startup seed validation happens before any eval and reaches the caller as StartupSeedsRefused prose. A typed DescendingRay, or an UnpenalizedDescendingRay, found at startup is therefore lost. (Diagnosis from sms-obj.)

Log: MSI `/scratch.global/sauer354/swarm17/sms-curv/logs/exp-sep.log`. The fixture is `separated_dataset` in the sms-curv experiment copy of `covariate_constant_slope_2930.rs`.

SauersML:
Fixes #3467.

## Root cause
The n-block exact-joint spatial driver (`optimize_spatial_length_scale_exact_joint_typed` in `fit_orchestration/drivers/spatial_optimization.rs`) took closures returning `Result<_, String>` and turned every error into `EstimationError::TrialPointRefused`. That threw away the typed `CustomFamilyError`, and with it the `RayRestoration` it carries. `run_plan`'s `eval_seed_restoring_rays` (#2695) finds that restoration through `ray_restoration_in(err)`, which only matches `EstimationError::CustomFamily`. So for a separated survival marginal-slope dataset, startup refused its only seed, even though the solver had reported the closing rho step.

## Change
- New `ExactJointRefusal { CustomFamily(CustomFamilyError), Reason(String) }`:
  - it has `From` impls for both;
  - `into_trial_error(context)` maps `CustomFamily` to `EstimationError::CustomFamily`, keeping the ray, and `Reason` to `TrialPointRefused`, as before.
- The exact-joint closure bounds (`ExactFn` / `ExactEfsFn`) now return `ExactJointRefusal`. These callers pass the typed error through instead of `.to_string()`:
  - survival marginal-slope `fit_entry`;
  - bms `block_specs`;
  - gamlss builders;
  - latent survival;
  - survival location-scale;
  - transformation-normal.
- The #979 Jeffreys arming evidence is still recorded at the eval sites through `jeffreys_arming_evidence()`. Startup seed validation goes through the same eval closure, so the evidence is captured on this path too.

## Behavior change (intended)
Before this change, every closure error became a retreatable trial-point refusal. A `CustomFamilyError` now reaches the outer objective whole, so the producer's own `is_trial_point_infeasible` verdict decides the outcome. That verdict is the documented design in #2553, #2590 and `estimation_error.rs`. As a result, structural failures such as a dimension mismatch or a numerical failure now abort the fit instead of being retreated from as an infeasible point.

## Tests
- `exact_joint_refusal_3467_tests` (lib) checks two things:
  - a solver refusal reaches the outer objective as `EstimationError::CustomFamily` with its ray intact;
  - a reason-only refusal is still a `TrialPointRefused`.
- `separated_survival_seed_takes_the_closing_rho_step_3467` (survival_acceptance) fits a separated dataset end to end and requires the fit to converge.


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 8369eea7fc8c.

PR #3496, opened 2026-09-20
Survival location-scale: select the baseline θ with ρ on one LAML criterion (#3413)

### LINKED ISSUE #3413: Survival location-scale baseline-θ search: the envelope gradient is not the gradient of its cost, because ρ is re-selected at every θ probe

SauersML:
WIP. The separate baseline-θ search for survival location-scale certifies a cost whose gradient it is not given (ρ is re-selected at every θ probe). This moves θ onto the location-scale LAML outer vector, as the latent and marginal-slope families already do, and deletes the separate search and its warm-start cache.

Fixes #3413

---


SauersML:
Closing this PR as a duplicate claim. LC-14 claimed #3413 first (comment at 07:59:38, PR #3487), and LC-19 is stepping back from this issue. This branch has no code changes.

Design notes for whoever lands the joint-LAML version:
- Freeze the LS time stack once, including the timewiggle knots, the same way the marginal-slope path does.
- Append the baseline θ axes after the link-shape axes in `with_auxiliary`.
- Bound `link_shape_start..` in `fit.rs` so it covers only the link-shape axes, not the new θ axes.
- Primary channels 0, 1 and 2 of the LS row kernel are `h_entry`, `h_exit` and `hdot_exit`.
  - With a time wiggle active, an offset move maps through `dq_dq0` (and through `d2q_dq02·d_base` for `hdot`).
  - The time Jacobian itself also depends on θ, so `H_θ` needs the wiggle-curvature terms and not only `Jᵀ T3[d] J`.

---

PR #3499, opened 2026-09-20
Bound the split block correction's mixed-axis term along soft axes (#3456)

### LINKED ISSUE #3456: #784 mixed-axis Φ blows up to 6.7e4 when a block axis goes soft, so the corrected REML cost is unbounded below (binomial 50 fz0114)
## Symptom

`y ~ s(nx0, by=nz0) + factor(fg1)`, binomial-logit, n = 50 (pygam fuzz case fz0114, `python worker.py gamfit binomial 50 fz0114 0` from `bench/pygam_compare`). It fails on main at d4f20811 (which includes #3204):

```
Outer smoothing-parameter optimization did not certify a stationary optimum (standard REML):
gradient |g|=5.078e-1 |Pg|=5.078e-1 bound=1.784e-2 (rung=curvature-resolvability) hessian_psd=NO
(λ_min(H)=-1.52e-5) curvature_source=terminal-analytic railed=[0] [#0 theta=-1.944990e1 box=[-1.944990e1, 2.032702e1]]
origin=ArcUnprogressingStallCheckpoint, plan=solver=Arc, search_hessian_source=Analytic, 78 outer iterations;
final objective 4.546065e1
```

Before #3204 this case failed differently: a BFGS rail stall near cost 30.49 (#3262). #3204 routed the latched fit through ARC with the exact Hessian, and ARC now walks into the region described below. Of the 11 #3262 rows I reran on this build, 10 certify. This is the only failure left, and its cause is a different one.

## What happens

1. The Laplace stage certifies at ρ = [-1.146, 6.615, -2.565], cost 30.615, via the Newton decrement.
2. `#784 ADMITTED`: m = 2, axis split = true, orders [14, 9].
3. The corrected ARC continuation starts well: Δ_b grows from 0.067 to 0.122 while ρ₂ goes from −2.7 to −4.0, and the cost falls from 30.545 to 30.519.
4. At the next step ρ₂ goes to the lower box, −19.45, and the correction explodes:

```
[#784] ... ENGAGED: m=2, axis split=true, max|γ|=5132.333, τ=0.310, Δ_b=6.2751e4 (mixed-axis Φ=6.2751e4), axis orders=[14, 9], nodes=23
[OUTER] eval#5 (hess) cost=-6.271971e4 |g|=5.634e4 rho=[-1.470,6.709,-19.450]
[OUTER] eval#6 (hess) cost=-6.710697e4 |g|=6.036e4 rho=[12.604,6.786,-19.450]
```

The per-axis quadratures Δ_r stay bounded. All of the 6.7e4 is the mixed-axis term Φ (`mixed_axis_laplace_term`, `crates/gam-solve/src/reml/block_quadrature_correction.rs:801`).

5. The corrected cost is therefore about −6.7e4, which is not a REML value. After that the search reseeds from the derived start and ends at 45.46, uncertified.

## Why

- Φ is the mixed-axis part of the second-order Laplace series. It is a polynomial in the whitened block coordinates a_i = Λ_b^{-1/2} V_bᵀ x_i (quartic `d·a⁴` and cubic-squared `T²` terms).
- When a block axis goes soft (λ_r → 0 as ρ₂ → −19.45), a_i scales as λ_r^{-1/2}, so Φ grows like λ_r^{-2}.
- The series is asymptotic. It is valid only while the standardized cumulants are small. Here max|γ| = 5132 against the τ = 0.31 admission gate, i.e. the regime where the series has no meaning.
- The skewness gate decides only admission. After the latch (#2748), Φ is spliced at every ρ, including where its own expansion parameter is 10⁴ times the gate.
- The Δ_r terms are exact 1-D integrals, so they stay bounded. The truncated mixed-axis series is not, and it is subtracted from the cost. The corrected criterion is therefore unbounded below along a soft axis.

## What a fix has to respect

- "A fit object must only ever come from a converged optimization." The refusal is correct; the criterion is what's broken.
- "In general, do not paper over solver issues." Do not clamp Φ, and do not add a reactive switch that turns Φ off when it gets large. A ρ-dependent on/off switch is exactly the #2748 non-function problem.
- The fix belongs in the correction's definition. The mixed-axis part must be a bounded, exact quantity along a soft axis, as the Δ_r are, e.g. an exact integral in place of a truncated series. Alternatively, the admission/latch must be defined so that the latched criterion is well posed over the whole ρ domain the search can reach.

Found while closing #3262. Logs: the fz0114 rerun on d4f20811 + #3418 (#3418 does not touch this path).

SauersML:
Fixes #3456

The split block correction is Δ_b = Σ_r Δ_r^cert + Φ. Its mixed-axis term Φ was the truncated second-order Laplace series −⅛Σd(|a|⁴−Σa⁴) + ⅛|u|² + (1/12)‖T‖² − (5/24)ΣT_rrr². That series grows like λ_r^{-2} along a soft axis, so the corrected cost −Δ_b was unbounded below and the outer optimizer could run toward a degenerate ρ. SPEC: "A fit object must only ever come from a converged optimization."

**Change.** Φ is retired, along with its ∂⁴W chain. It is replaced by the bounded mixed-axis term Ψ: an anchored, truncated ANOVA over axis subsets T with |T| ≤ K = 3.
- Each piece integrates e^{−ΔF} exactly on the tensor 3-point Gauss–Hermite rule (nodes {0, ±√3}, weights {2/3, 1/6, 1/6}).
- Each piece is f_T = log Σ w e^{−ΔF} − log Σ w, and Ψ = Σ κ_T f_T, where κ_T = c_{|T|} − [|T|=1] and c_k = Σ_{j=0}^{K−k} (−1)^j C(m−k, j).
- The single-axis certified terms are left alone, so Ψ carries only the part that no single axis carries.
- Nodes that are not feasible are left out of the rule. This is exact integration of a restricted measure, not a fallback.
- Each piece is bounded: it lies between |T|·log(2/3) and the log of the largest node ratio. Ψ therefore stays finite along a soft axis.

**Derivatives are exact, not differenced.**
- ∂(−Ψ)/∂ρ = Σ κ_T E_{p_T}[g].
- ∂²(−Ψ)/∂ρ² = Σ κ_T (E_{p_T}[h] − Var_{p_T}(g)), with the variance centred.
- Each node sits at s = Σ z_r Y_r(ρ), with ṡ and s̈ carried through the row derivatives f_s, f_ss, f_e, f_es and f_ee.

**Tests.**
- Unit tests cover:
  - the assembly of value, gradient and Hessian against central differences of the cost;
  - Ψ ≡ 0 for m ≤ 1 and for separable ΔF;
  - piece bounds under soft-axis growth, where Φ grows at least 100× and Ψ stays bounded;
  - the node gradients against finite differences.
- The regression test `block_hessian_matches_differences` checks the outer ρ-Hessian against differences of the analytic gradient.


SauersML:
Status from LC-5. The head is f3177400a6 (origin/main merged in; the Arc-backed `solve_d_array` is now passed as an owned array, which fixes the E0308 at the Hessian call site).

The implementation is complete. Two proof runs are queued on this head and are the only merge gate:
- run 35503843937: gam-solve lib, filters `mixed_axis block_correction block_quadrature`;
- run 35503845460: gam integration `regressions`, filter `block_hessian_matches_differences`.

Neither run has started, because the Actions queue holds about 590 runs. The local host is under 3G of free disk, so it can't compile this locally. Whoever picks this up: once both runs are green, mark the PR ready and merge it with merge_method "merge". Do not merge before then, because this head has not been compiled yet.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, c9f88ed9a95a.

PR #3500, opened 2026-09-20
gam-sae native MDL: code the residual at the fit's own distortion instead of reusing it as the quantization budget

### LINKED ISSUE #3435: gam-sae native MDL: fit error is spent again as the quantization budget (#2933 F21)
Split out of #2933 (finding F21, accounting part). The typed score-kind relabel landed earlier on #2933. This issue is the remaining algebra.

## Defect
`crates/gam-sae/src/description_length.rs:1466–1474` sets:

```rust
let budget = (1.0 - ev) * decoded_code_variance;
```

That is the variance the fit leaves unexplained, reused as a fresh reverse-water-filling distortion budget for coordinate quantization. The fit error is never added to the distortion as a separate term, and no quantized output is constructed.

With independent fit and quantization errors, total output distortion is `D_fit + D_q`. Spending the whole achieved residual `D_fit` again as `D_q` doubles that component: the reported rate is priced at distortion `D_fit` while the delivered distortion is `≈ 2·D_fit`.

The Eq. 4 path codes its residual separately, so this argument does not transfer to it unchanged.

## What would close it
One of:
- code the residual separately (as Eq. 4 does), with the total `D = D_fit + D_q` stated; or
- fix a total target distortion D, allocate only `D_q = D − D_fit` to quantization, and refuse typed when `D_q ≤ 0`.

Plus a regression test: on a planted source, check the reported (rate, distortion) against an encode/decode round-trip, or against the closed-form Gaussian R(D) with the fit error added.

No operational codec exists yet, so the score must keep its `GaussianSurrogate` kind until one does.

Refs #2933.

SauersML:
Fixes #3435

The native fit-level description length used to spend `(1 - ev) * decoded_code_variance` as a fresh quantization budget. That budget was not the reconstruction error the fit actually commits to, so the bits and the reported explained variance described different quantities.

Now the distortion is measured, not assumed:
- `D = mean ||x - x_hat||^2` in the whitened metric `diag(tier0_scale^-2)`.
- The residual's raw second-moment spectrum enters the joint reverse water-filling as a weight-1 component next to the decoded-code components.
- The gate distortion is spent from `D` first.
- `ev = 1 - D / TSS_M` is reported from the same `D`.

The Python `description_length` now takes `x`, so the measured residual goes through the FFI. The stubs and callers are updated.

Tests (closed form):
- `native_description_length_*_2933_f11`: code bits `log2(10)`, residual bits `0.5*log2(3)`, distortion `0.3`, `ev = 1 - 0.3/2.3`.
- `native_description_length_charges_gate_amplitudes_on_a_fixed_support_2933_f10`: amplitude gate costs `0.5*log2(10)` plus 0.5 residual bits, and the constant gate costs 0.


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 0c57a189a211.

PR #3501, opened 2026-09-20
#784 block correction: make the latched block a continuous function of ρ (#3113)

### LINKED ISSUE #3113: #784 block correction re-selects its "m largest |γ|" block at every ρ: Δ_b jumps by a whole direction (+0.41 → −0.29) and the binomial outer BFGS dies with StepSizeTooSmall

SauersML:
Fixes #3113

## Root cause
Under a latched #784 admission (#2748) the block was re-selected at every rho as the `m` largest-|γ_r| positive-curvature eigendirections of H. That set swaps members wherever two |γ_r| cross. Each swap is a jump of a whole direction's contribution to Δ_b, and the latched per-axis quadrature orders get reassigned to a direction they were never certified on.

On the 21-row `make_blobs(n_samples=21, random_state=0)` binomial fit `y ~ x0 + x1` (p = 3, admitted m = 2):
- Δ_b alternated between 4.24e-1 and 1.12e-1 across a 9e-4 step in rho.
- The outer BFGS ended `NOT STATIONARY` at |g| = 4.8e-2 (CLI rc=4 on main c663ba6dba).

## Fix
- The latch (`BlockQuadratureLatch::block_ranks`) records the admitted block as its ranks in the ascending spectrum of H. The ranks are needed because eigh returns ascending pairs and the stacked-root SVD returns descending ones.
- Every later rho integrates the eigenpairs at those ranks. That is the continuous transport of the admitted block wherever the eigenvalues are simple, and the eigenframe splice already refuses an unresolved block.
- The block, its quadrature and its Hessian support latch together, only at first admission. The old path that let a caller with a fixed block dimension latch the quadrature later is gone.
- A latched dimension without matching ranks is a broken latch. It raises the new `BlockQuadratureCorrectionStage::LatchedBlockUnavailable`, which is fatal and not trial-point local.
- A latched direction with non-positive curvature at a trial rho raises `NonPositivePenalizedCurvature`, which is trial-point local, so the outer search backs off.
- No retry, jitter or tolerance change.

This is the same design as the engine half of #3154 (`block_positions`), which stays blocked on #3018. That PR can drop its engine change once this lands.

## Tests
- New `tests/regressions/misc/regression_3113_block_subspace_continuity.rs`: the blobs fit must converge to a finite REML score.
- `glm_outer_hessian_fd_tests` latch literal carries `block_ranks`.
- Focused CI runs on 892ed244d2 cover regressions 3113/2748/1082 and the block Hessian FD checks, the gam-solve `block_correction` lib tests, and the gam-problem `block_quadrature_correction_refusals` tests.


SauersML:
Merged main in (conflict resolved by BF-73); ready for a merge train.

New head: 6ccd0872cf8e8ba38af338b0155a0a5a7ed02748 (two-parent merge of 892ed244d2 and origin/main).

Main already landed the engine half of this design in e26fc385db ("latch the #784 correction block as spectral positions"). There the latch holds `BlockQuadratureLatch::block_positions`, which are ranks in ascending eigenvalue order, so this PR's `block_ranks` duplicated it. The resolution keeps main's field and code path and keeps what this PR adds on top:
- `ascending_spectral_order` (from this PR) now builds main's `ascending` order.
- A latched position outside the spectrum, or an admission latched without its block, raises the typed fatal stage `LatchedBlockUnavailable`. Main raised a bare `InvalidInput` there.
- A latched direction with non-positive or non-finite curvature at a trial ρ raises `NonPositivePenalizedCurvature`, which is trial-point local. Main tested `<= 0.0`, which let a NaN eigenvalue through.
- The duplicate `block_ranks` field is gone from the struct and from the `glm_outer_hessian_fd_tests` literal. Without that, the auto-merge would not compile because the literal would name a field that does not exist.
- `regression_3113_block_subspace_continuity` is kept.

The net diff vs main is 4 files, +136/-20.

Note: #3113 may already be fixed on main by e26fc385db. The kept regression test decides that.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, f24830b753b4.

PR #3504, opened 2026-09-20
expectile LAWS: resolve sign-pattern cycles at the generalized (Clarke) fixed point

### LINKED ISSUE #3039: Expectile LAWS has no sign-pattern fixed point when one near-zero-residual row's weight flip moves the REML λ enough to flip it back (~1 in 160 heteroscedastic fits)
## Symptom

`fit_expectile_laws` (crates/gam-models/src/fit_orchestration/entry.rs) returns its typed error

```
expectile LAWS entered a deterministic sign-pattern cycle without reaching the KKT fixed point ...
(tau=0.9, iterations=6, cycle_length=2, KKT residual=5.143e-7 vs scaled tolerance 1.000e-10,
 rho_checkpoint=[1.7945861630036692, 21.548752144175673])
```

on about 1 in 160 ordinary fits. The typed error is correct per SPEC (no best-effort fit), but the estimator exists and LAWS cannot reach it.

## Reproducer

- Data: `y = sin(2πx) + (0.2 + 0.8x)·Z`, `x ~ U(0,1)`, n = 800.
- Rust: the `simulate(seed)` generator in `tests/quality/misc/quality_expectile_band_coverage_heteroscedastic.rs` (StdRng), with `seed = 1001` and `family = "expectile(0.9)"`, formula `y ~ s(x)`.
- Python, numpy `default_rng(1000 + rep)`, same DGP: τ = 0.95 with seed 1009 cycles, and 0 of the other 159 (τ, seed) pairs do.
- Rust scan over seeds 1000–1039 × τ ∈ {0.05, 0.5, 0.9, 0.95}: only (0.9, 1001) fails.

## Diagnosis (instrumented LAWS loop, seed 1001, τ = 0.9)

```
it=4 kkt=1.155e-4 rho=[1.79637, 21.557] edf=8.95959 flips=4
it=5 kkt=7.589e-8 rho=[1.79295, 21.539] edf=8.94688 flips=1  row 138 r=+3.545e-6
it=6 kkt=5.143e-7 rho=[1.79459, 21.549] edf=8.95332 flips=1  row 138 r=-2.403e-5
(it=7 would repeat it=5)
```

The cycle is a single row. Its residual is ~1e-5, so its weight flip (0.9 ↔ 0.1) changes β̂ at fixed λ only by `H⁻¹xᵢ Δwᵢ rᵢ ≈ 0`. But the flip changes `H = XᵀWX + S_λ` by `Δwᵢ xᵢxᵢᵀ`, so REML's `log|H|` term moves the selected ρ₁ by ≈1.6e-3 (edf 8.947 ↔ 8.953). That moves the fitted curve at row 138 by ~3e-5, which puts the residual back on the other side of zero.

So the discrete map sign pattern → (REML λ, β̂) → sign pattern has **no fixed point** here. The loss `|τ − 1[r<0]| r²` is C¹ but not C² at r = 0, so the REML Hessian of a zero-residual row is set-valued: any weight in [min(τ,1−τ), max(τ,1−τ)] is a valid generalized (Clarke) curvature. The solution LAWS is looking for is a generalized fixed point: rᵢ = 0 exactly, with a fractional weight wᵢ* where the REML-selected λ keeps the row on the curve. By continuity of the REML λ in wᵢ, one exists between the two patterns of the 2-cycle.

## Suggested direction

On a detected cycle whose patterns differ on a small row set D, solve for the fractional weights on D such that r_D(w) = 0. The KKT defect is then exactly zero, since it is supported on `(w_used − w_new)·r`. Solve it with Newton on `r_D(w)`, using the implicit derivative of the REML optimum `dρ̂/dw = −(∇²_ρ V)⁻¹ ∂²V/∂ρ∂w`, which the outer REML machinery can supply analytically. Per SPEC, not bisection or finite differences.

Found while adding the expectile band-coverage test (pyGAM audit F5a). That test runs on seed block 1020–1039, where every (τ, seed) converges; this issue tracks the remaining failures.

SauersML:
Fixes #3039

WIP. When LAWS detects a deterministic sign-pattern cycle, the plan is to solve for fractional weights on the rows that flip sign across the cycle, which is the Clarke generalized fixed point: w_j in [min(tau,1-tau), max(tau,1-tau)] at r_j = 0. The solve uses the analytic implicit derivative dρ̂/dw from the certified outer-Hessian pseudoinverse. Acceptance stays with the existing KKT certificate.

---


SauersML:
[swarm lane gam-3039] Separating evidence against this draft. The #3039 claim comment names a gated candidate for this issue: MSI branch w3039v4, E1 8afb241a4b + Newton 088e29330a.

1. The new regression test does not fail at base. On current main, seed 1001 at tau 0.9 (y ~ s(x)) no longer cycles; main's basis-size changes moved that fit. At 94e6a59c5b the base LAWS converges on all 160 (seed 1000..1039 x tau {0.05,0.5,0.9,0.95}) sin fits, including 1001@0.9, so `expectile_laws_reaches_generalized_fixed_point_through_sign_cycle_3039` passes without this change and never enters the generalized branch. The live cycle on that main is the linear-truth fixture y = 1 + 0.5x - 0.3z + (0.2+0.8x)Z, n = 800, `y ~ s(x) + s(z)`, seed 2007, tau 0.95. At base it errors with "deterministic sign-pattern cycle", cycle_length 2, KKT 3.3e-6. With my candidate it certifies (KKT 6.1e-11) with row 187 at the interior fraction t = 0.5704, weight 0.563.

2. The mixed partial uses the wrong scale. `g_{l,k}` multiplies the penalty-quadratic term by 1/phi_hat with phi_hat = coefficient_covariance_scale() = RSS/(n - edf), the mgcv scale. The profiled Gaussian REML criterion is (n_r/2) log D_p + ½ log|H| - ½ log|S|+, so its mixed partial carries n_r/D_p = (n+ - M_p)/(RSS + beta'S beta), which is a different number. The derivative also drops the rho-prior anchor term -h_k/(n+ w_l), which is nonzero for any non-flat RhoPrior.

3. Two regimes for one estimator. Plain LAWS runs until a proven sign cycle and then switches to a different update. A projected semismooth Newton step on the natural map t - Pi(t + Lambda r(t)) covers both: with an empty band it is the LAWS update bit for bit, so no switch is needed. On 240 fits (160 sin + 80 linear-truth) at 94e6a59c5b my candidate returns coefficients bit-identical to base on every fit base certifies (239), and certifies the one base cycle.

4. H^-1 is formed as Vb/phi_hat (products with a formed inverse carry cond(H) times its backward error; #2668 measured cond 2.1e8 on y ~ s(x)), and dbeta/drho is recomputed locally from it. That is a second route for the engine's own IFT Jacobian. The candidate publishes the engine's J = dbeta/drho beside V_rho on both covariance branches (with a test pinning V_c = J V_rho J' to 5e-20 and the factorized branch bit-identical to dense), and applies K^-1 by Cholesky solves.

Please hold this draft; the candidate is in its landing gate.

SauersML:
LC-7: holding this draft as the gam-3039 lane asked. It must not be merged or taken over to merge.

The 08:15Z objections hold against this branch:
1. The focused proof passed, but it does not prove the fix. Run 35498940118 shows seed 1001 at tau 0.9 certifying at |Σwr|/sqrt(ΣwΣwr²) = 7.5e-15 with 0.191 of rows above the expectile. That is consistent with your finding that this fit no longer cycles on main, so the test never reaches the generalized branch. A valid regression fixture has to be one that cycles at base, such as your linear-truth seed 2007 at tau 0.95.
2. The mixed partial uses `phi_hat = RSS/(n-edf)`. It should use the profiled `n_r/D_p`, and it also drops the rho-prior anchor term.
3. It forms H⁻¹ as `Vb/phi_hat` and recomputes the IFT Jacobian locally, which gives a second route for dβ/dρ.

Your w3039v4 candidate addresses all three. If it lands, this PR should be closed in its favour.

---

SauersML:
**VF-0 review: not merging (unsound per SPEC.md; the body also says WIP).**

1. **Reactive fallback.** The projected Newton step on the fractional asymmetries runs only after a laws sign cycle is detected. SPEC.md: "In general, do not paper over solver issues."
2. **Scaling.** The mixed partial uses `1/φ̂` with `φ̂ = RSS/(n-edf)`. The profiled REML criterion needs `(n-M_p)/D_p` with `D_p = RSS + β'Sβ`, and the ρ-prior term is dropped.
3. **Singular system.** At `r_F = 0` the Jacobian has rank at most K, so the system is singular whenever there are more free rows than K. The hand-rolled pivot only catches exact zeros.
4. **The regression test doesn't exercise the change.** `expectile_laws_reaches_generalized_fixed_point_through_sign_cycle_3039` passes on main without entering the new branch, because seed 1001 no longer cycles there.

Principled fix: run one projected semismooth Newton on `t - Π(t + Λ r(t))` from the start instead of after a cycle, use the correct `D_p` scaling, and reuse the engine's IFT derivatives.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, b9725c134426.

PR #3505, opened 2026-09-20
Matern 2-D smooth under Gamma: certified fit (#3236 item 2)

### LINKED ISSUE #3236: Four Python tests red at HEAD outside the basis-default code: mjs/matern outer REML, cylinder te period=[2*pi, 0], te row-order ΣSE bar
Found while working on #3136 (lane `basis-defaults-te-tp`). This lane changes only the default te/1-D sizes, the tp refusal and the tensor null-ridge chart, and none of these four failures goes through those paths. Error lines are from a `maturin develop --release` build of the branch.

1. **`tests/bug_hunt_curv_smooth_aborts_nan_covariance_low_signal_test.py::test_peer_2d_smooths_fit_the_same_low_signal_data`.** `te(x, z)` and `s(x)+s(z)` fit. `mjs(x, z)` raises `RemlConvergenceError` / `EstimationError::DominatedCertifiedPlateau` (aniso-psi joint REML):
   - It declines a certified optimum at objective 570.795.
   - Its best checkpoint, 567.651, has `|Pg|=1.669e1` and an indefinite Hessian (`λ_min(H)=-5.54e2`).
   - ARC ends at the trust-region reject floor after 15 rejections.
2. **`tests/bug_hunt_matern_2d_smooth_unfittable_nongaussian_family_test.py::test_matern_2d_smooth_is_fittable_under_gamma_family`.** `matern(x, z)` with `family="gamma"`: iso-kappa joint REML does not certify, with `|Pg|=2.035e0 > bound=6.010e-2` and `λ_min(H)=-2.9e-6`. This is #1615's scenario, which is now a convergence failure rather than a builder abort.
3. **`tests/test_manifold_fit_topology.py::test_cylinder_fit_wraps_in_angle_but_not_in_height`.** `te(theta, z, periodic=[0], period=[2*pi, 0])` raises `TermBuilderError: period for periodic axis 1 must be finite and positive, got 0`. Since #2781, `period=` declares periodicity, so the `0` placeholder for the non-periodic axis is read as a period for axis 1 even though `periodic=[0]` names only axis 0. Either the test or the grammar needs to change. A 0 or missing entry could mean "not periodic" when `periodic=` is given explicitly.
4. **`tests/bug_hunt_te_tensor_smooth_edf_depends_on_row_order_test.py`.** The row-order invariance this test was written for holds exactly: edf, deviance and ΣSE are identical across the original order and permutations 1 and 7. What fails is the absolute sanity bar `ΣSE > 5.0` on the original order. At `k=[7, 7]`, main's default size, ΣSE is 2.52 at edf 12.9. At #3136's default [8, 9] it is 3.74 at edf 14.4. A sum of coefficient SEs depends on the coefficient chart, so a fixed 5.0 is not a well-posed "covariance collapsed" test. A scale-free check would be better, for example comparing against the unpenalized-direction SEs or the permuted fits, which the test already makes.

SauersML:
Fixes #3236

Item 2 of #3236: `y ~ matern(x, z)` under `family="gamma"` on the Python contract's rows (n = 600, NumPy `default_rng(0)`).

- Adds a Rust reproducer, `crates/gam-models/tests/suite/matern_gamma_surface_3236.rs`, which reads the exact rows from a CSV fixture. It fits `te(x, z)` as the control, then `matern(x, z)`, and requires a certified optimum plus recovery of the surface.
- A root-cause fix follows if the reproducer is red on main after #3365.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, a811954dbdeb.

PR #3533, opened 2026-09-20
κ profile: read V_p at the Newton minimizer in η, carrying the certified residual V_η (#3426)

### LINKED ISSUE #3426: constant-curvature κ profile: V_p′ and V_p″ miss their central differences at κ≈-1.11 (rel 1.1e-4 / 3.2e-3); the_profiled_second_derivative_is_the_derivative_of_the_profiled_first fails on main
## Reproducer
On `origin/main` (e8568910):

```
cargo test -p gam-models --lib the_profiled_second_derivative_is_the_derivative_of_the_profiled_first
```

```
[#2747 profile-fd] κ=-1.1119 V_p=-248.558214 V_p′=-1.314540e2 (fd -1.314392e2, rel 1.12e-4)  V_p″=+5.891593e1 (fd +5.910831e1, rel 3.20e-3)  schur=2.487838e1 unreduced rel 4.11e-1
panicked at crates/gam-models/src/fit_orchestration/drivers/constant_curvature_kappa_box_probe_tests.rs:546:13:
κ=-1.1118509777297985: V_p′=-1.314540449e2 against a central difference of V_p -1.314391505e2 (rel 1.125e-4); the envelope reduction does not differentiate the value beside it
```

The failure is deterministic: the same numbers appear on main and on #3406. #3406 only changes the planner route, not `ConstantCurvatureProfile::evaluate`.

## What it means
At the first grid cell (κ = −0.8·cap), `ConstantCurvatureProfile::evaluate` gives a first derivative that disagrees with a central difference of its own value by 1.1e-4 relative. That is about 1.5e-2 absolute on a slope of −131, far above the O(h²) truncation error of a step at 1e-3·cap. The second derivative is off by 3.2e-3.

The Schur correction itself looks right: the unreduced V_κκ misses by 4.1e-1, and the reduced value lands close. The mismatch is in the envelope step, V_p′ = V_κ at η̂(κ). The likely candidates:
- the inner range solve's η̂ is not at its stationary point to the precision the envelope theorem needs. A residual V_η·dη̂/dκ is being dropped, and that term only vanishes when V_η = 0 exactly. Commit 73077e15 ("drop √ε floor") recently changed the κ route's tolerance.
- or V_κ itself carries an analytic error at this κ.

Either way, the value, the gradient and the curvature the outer κ search sees are not one consistent function at this cell, which breaks the SPEC rule that a fit only comes from a converged optimization on a consistent objective.

## Scope
Found while working on #3201 (outer lane). This issue belongs to the κ-profile / curv() inner lane. Do not fix it by loosening the test's 1e-4 / 1e-3 bars without a derivation.

SauersML:
Fixes #3426

## Root cause
The inner range solve (`minimize_over_eta`) certifies η̂ by its Newton decrement `V_η²/V_ηη` against the criterion's statistical resolution `1/(2n)`. So it stops with `V_η ≠ 0`, which is correct for its own certificate. But `eta_profiled_kappa_jet` then applied the envelope theorem `V_p′ = V_κ`, which holds only at `V_η = 0` exactly. That drops `−V_κη·V_η/V_ηη`, a term that is first order in the residual.

I added instrumentation to the FD test (not committed) and measured the fixture cells:

| κ | V_η | V_ηη | V_κη | predicted error V_κη·V_η/V_ηη | observed V_κ − fd |
|---|---|---|---|---|---|
| −1.1119 | −9.013e-3 | 8.959 | 14.93 | 1.50e-2 | 1.48e-2 |
| +0.2780 | −9.367e-3 | 147.2 | 63.28 | 4.03e-3 | 4.01e-3 |
| +0.8339 | −9.305e-3 | 206.9 | 80.24 | 3.61e-3 | 3.60e-3 |

At κ = −1.11 the decrement is 9e-6, far inside the 2.5e-3 band (n = 200). The certificate is right; the reduction was incomplete.

## Fix
Read the profile at the Newton minimizer of the jet's own quadratic in η:

```
V_p  = V − ½·V_η²/V_ηη        (error O(δ³))
V_p′ = V_κ − V_κη·V_η/V_ηη     (error O(δ²))
V_p″ = V_κκ − V_κη²/V_ηη       (Schur complement, unchanged)
```

Each line is the κ-derivative of the one above to the order it carries, so the triple is one function's jet wherever the inner solve stopped inside its band. There is no extra evaluation, no tolerance change and no retry. The bars in `the_profiled_second_derivative_is_the_derivative_of_the_profiled_first` (1e-4 and 1e-3) are unchanged.

With the correction, V_p′ matches the central difference to about 1e-7 relative at every cell, and V_p″ to about 6e-5 or better.

## Tests
- New: `the_eta_profile_of_an_off_minimum_jet_is_the_profile`. On a quadratic V, the reduction must return the closed-form profile exactly, at jets taken off η's minimizer.
- Existing: `the_profiled_second_derivative_is_the_derivative_of_the_profiled_first` failed on main and now passes (validation results in the comments).

---
_Generated by [Claude Code](https://claude.ai/code/session_01XpJgiWc6RNHuyN6qAk4fhd)_


SauersML:
Validation on head 816b32ca:
- `cargo test -p gam-models --lib constant_curvature`: 13 passed, 0 failed. That count includes `the_profiled_second_derivative_is_the_derivative_of_the_profiled_first`, which failed on main, and the new `the_eta_profile_of_an_off_minimum_jet_is_the_profile`.
- `cargo check --workspace --tests`: finished clean. The only warning is a pre-existing unused `IxDyn` import that this change doesn't touch.

---

PR #3555, opened 2026-09-20
pirls: a dispersion refresh still moving at its pass cap is a failed fit

### LINKED ISSUE #3544: crates/gam-solve/src/pirls: Gamma-shape / Tweedie-φ / Beta-φ / NB-θ converged-η refresh silently ships a non-fixed-point fit when the pass cap is hit
## Where

`crates/gam-solve/src/pirls/loop_driver.rs` on origin/main (1a3a498e3b), in the converged-η dispersion refresh that runs for the reported fit (`refine_dispersion_at_converged_eta`). Four of the five families end their refresh loop with a silent `break` on the last pass while the scale is still moving:

| family | loop | cap / tol | silent exit |
|---|---|---|---|
| Gamma shape | `:1609` | `MAX_SHAPE_REFRESH = 5`, `SHAPE_REFRESH_REL_TOL = 1e-4` | `:1642` |
| Tweedie φ | `:1700` | module `MAX_PHI_REFRESH = 5`, `PHI_REFRESH_REL_TOL = 1e-4` | `:1728` |
| Beta φ | `:1857` | local `MAX_PHI_REFRESH = 30` (shadows the module const), `1e-4` | `:1885` |
| NB θ | `:1996` | `MAX_THETA_REFRESH = 30`, `1e-4` | `:2024` |

```rust
            if refresh_iter + 1 == MAX_SHAPE_REFRESH {
                // Final allowed pass and the shape is still drifting (a
                // pathological non-contraction). Do NOT re-solve: ...
                // ... the residual weight/φ drift is
                // bounded by the last `rel_change` and never worse than the
                // pre-fix frozen-warm-start value.
                break;
            }
```

The Gaussian / inverse-Gaussian φ loop sits between them (`:1766-1800`). It treats the same situation as a failure:

```rust
        if !converged {
            crate::bail_invalid_estim!(
                "dispersion φ did not reach its converged-η fixed point within {MAX_PHI_REFRESH} \
                 re-solves (relative tolerance {PHI_REFRESH_REL_TOL:e})"
            );
        }
```

Its header comment states the contract: *"a φ still moving on the last allowed pass is a failed fit, not a reported one"*.

## Why it is wrong

On the silent-break path the refreshed scale is installed into `working_model.likelihood`. But `working_summary` (β̂, working weights, penalized Hessian, deviance, EDF) was solved at the *previous* scale, which differs from the new one by more than the refresh tolerance.

Gamma and Tweedie are not exempt. The penalized objective is `k·D(β) + βᵀS_λβ`, and the dispersion factor `k` rescales λ, so β̂ moves when `k` moves. For Beta and NB the scale also enters the working response.

The fit that comes out is therefore not a joint (β, scale) fixed point:
- the reported SEs, `Vb = H⁻¹` and intervals come from a Hessian built at a different scale than the one reported;
- the loop hit its cap exactly because the map was not contracting, which is the case where that gap is largest.

That violates:
- SPEC "A fit object must only ever come from a converged optimization."
- SPEC "In general, do not paper over solver issues."

The same situation errors for Gaussian φ and ships silently for the other four families. The caps (5 vs 30) and the `1e-4` tolerances are also arbitrary magic constants.

## Principled fix

1. On every family, a scale still moving by more than the tolerance after the last allowed pass is a named failure, exactly like the Gaussian loop. Delete the four silent `break`s and bail with the family-specific message.
2. Follow-up, out of scope for the minimal fix:
   - derive the refresh tolerance from the resolution of the scale estimator instead of the fixed `1e-4`;
   - drop the per-family pass caps in favour of one shared contraction-rate test;
   - make the Gamma, Tweedie and NB re-solves reject a non-mode inner status, as the Beta loop already does at `:1940`. Otherwise an intermediate pass can estimate the next scale from an η that no minimum produced.

SauersML:
Fixes #3544

## Problem

`crates/gam-solve/src/pirls/loop_driver.rs` re-estimates each family's scale at the converged η for the reported fit. For the Gamma shape, Tweedie φ, Beta precision φ and negative-binomial θ, the refresh loop exited silently on its last allowed pass while the scale was still moving by more than the refresh tolerance:

```rust
if refresh_iter + 1 == MAX_... { break; }
```

The fit it returned paired β̂, the working weights, the Hessian (and so Vb/SE) and the EDF solved at the *previous* scale with a different reported scale.

- β̂ is not stationary at the reported scale. The scale rescales the penalized objective (Gamma/Tweedie) or enters the working response (Beta/NB), so the returned values are not a joint fixed point.
- The Gaussian / inverse-Gaussian φ refresh in the same function already refuses this case with a named error.

## Change

The four silent `break`s become `bail_invalid_estim!` errors that name:
- the family,
- the pass cap,
- the last relative change,
- the tolerance.

This matches the Gaussian refresh. No change to the loops that converge.

The follow-ups in #3544 stay open for a later change:
- derive the tolerance instead of using the fixed `1e-4`;
- use one shared contraction test instead of the per-family caps;
- add an inner-status check on the Gamma/Tweedie/NB re-solves.

---


SauersML:

The original author (HN-13) has finished and handed this off, with its proof green (check run 106047509975).

I reviewed the diff. The Gamma shape, Tweedie φ, Beta φ and NB θ converged-η refreshes each used to `break` at their pass cap while the parameter was still moving, and then reported a (β, dispersion) pair that is not a joint fixed point. Each now refuses, in the same way the Gaussian/IG φ refresh already did. This follows SPEC's "A fit object must only ever come from a converged optimization."

---

PR #3562, opened 2026-09-20
gamlss/binomial: seed the location-scale threshold at the prevalence (not its mirror); saturated cloglog success rows score 0

### LINKED ISSUE #3526: gamlss/builders: binomial location-scale warm start seeds the threshold at the mirrored prevalence (sign error vs q = -eta_t/sigma)
## Defect

The deterministic warm start for the binomial location-scale model is supposed to start the threshold block at the fitted value that reproduces the weighted prevalence. It actually starts it at the mirrored prevalence. For logit and probit, the initial fitted probability is `1 - p̂` instead of `p̂`. For cloglog it is some other wrong value.

### Where (origin/main)

`crates/gam-models/src/gamlss/builders.rs` L764-790, `binomial_location_scalewarm_start`:

```rust
let prevalence = weighted_binomial_prevalence(y, weights)?;
let eta = binomial_location_scale_link_eta_from_probability(link_kind, prevalence)?;
project_constant_eta_into_block(threshold_block, weights, eta)?
```

`binomial_location_scale_link_eta_from_probability` (L653-677) returns `+link(p̂)`: `logit(p̂)`, `Φ⁻¹(p̂)` or `ln(-ln(1-p̂))`. The log-σ block is seeded at 0, so σ = 1.

The family, however, evaluates the linear predictor as a negated threshold. `crates/gam-models/src/gamlss/binomial/kernel.rs` L174:

```rust
pub(crate) fn binomial_location_scale_q0(eta_t: f64, sigma: f64) -> f64 {
    -eta_t / sigma
}
```

and `binomial_location_scalerow` sets `mu = F(q)` with `q = q0 + etawiggle`. The comments in `binomial/location_scale.rs` (L751, 990, 1119, 1465) and `kernel.rs` L504 state the same convention: `q = -eta_t * exp(-eta_ls)`.

So the seed gives `mu = F(-link(p̂))`:
- logit and probit (symmetric F): `mu = 1 - p̂`.
- cloglog: `mu = 1 - exp(-exp(-ln(-ln(1-p̂))))`, which is not `p̂` either. For example, p̂ = 0.1 gives mu ≈ 0.9999.

For a rare outcome (p̂ = 0.01, logit) the inner solver starts with every row predicted at 0.99. That is the worst possible constant start and is more than 9 nats per row from the intercept MLE. The comment above the function claims it "projects a prevalence-matched threshold". It does not.

The unit test at L685-699 (`a_prevalence_on_one_outcome_has_no_warm_start_threshold`) asserts `eta == ln(rare/(1-rare))`, i.e. `+logit`. It pins the helper's output but never checks the thing that matters: that the seeded threshold reproduces the prevalence through the family's own `q0` map.

### Principled fix

Seed the threshold at the value the family's map sends to the prevalence. With σ = 1, `F(-eta_t) = p̂` gives `eta_t = -link(p̂)`. Make the helper return the threshold rather than the link quantile, so the single source of truth is the family's `q = -eta_t/sigma`. Then add a test that round-trips the seed through `binomial_location_scale_q0` plus the inverse link and recovers p̂ for logit, probit and cloglog.

Only one production caller is affected (`build_binomial_threshold_and_scale_blocks`, builders.rs L608).

### LINKED ISSUE #3527: gamlss/binomial kernel: cloglog location-scale log-likelihood errors "non-finite" at y=1, q>709 where the true value is 0 (0*-inf = NaN)
## Defect

`crates/gam-models/src/gamlss/binomial/kernel.rs` L196-214 (origin/main), in the CLogLog arm of `binomial_location_scale_log_likelihood`:

```rust
let z = q.exp();
let log_p = if z == 0.0 { q } else if z.is_infinite() { 0.0 } else { log1mexp_positive(z) };
let log_survival = -z;
let ll = weight * (y * log_p + (1.0_f64 - y) * log_survival);
if ll.is_finite() { Ok(ll) } else { Err(GamlssError::NonFinite { .. }) }
```

For a success row (y = 1) with q > ~709.78, `z = +inf` and `log_survival = -inf`. The failure term then becomes `(1 - y) * log_survival = 0 * -inf = NaN`, so the function returns a `NonFinite` error.

The true log-likelihood there is `log(1 - exp(-inf)) = 0`, which is the best possible value for that row. The code already computes it correctly in `log_p = 0.0`, but then discards it. The same arithmetic mirrors at the other end: for y = 0 with z == 0 the success term is `0 * q`, which is finite, so only the y = 1 overflow case breaks.

This function is the objective for every binomial location-scale path: `binomial_location_scalerow`, `binomial_location_scale_ll_only` (the line-search objective), `location_scale.rs` L2281, `wiggle_custom_family.rs` L282 and `binomial_mean_wiggle.rs` L753. It is reachable because `q = -eta_t * exp(-eta_ls)`. On well-separated success rows, a trial step that shrinks σ or pushes η_t down lands at q > 709. Instead of scoring a perfectly-fitted row as 0, the kernel fails the whole evaluation. The step is then rejected, or the fit aborts, for a reason that is purely floating-point.

The probit and logit arms do not have this problem, because `normal_logsf` and `stable_softplus` stay finite.

### Principled fix

Use the xlogy convention: a term whose Bernoulli coefficient is exactly zero contributes exactly zero. Only form `y * log_p` when `y != 0` and `(1 - y) * log_survival` when `y != 1`. The non-finite check then still fires on genuinely impossible rows (y = 0 at z = inf, or a fractional y at saturation), while a saturated row that fits perfectly scores its exact value of 0.

SauersML:
Fixes #3526
Fixes #3527

## #3526: warm-start sign

The binomial location-scale family evaluates `q = binomial_location_scale_q0(eta_t, sigma) = -eta_t / sigma` and sets `mu = F(q)`. The warm start seeded `eta_t = +link(p̂)` with `sigma = 1`, so every row started at `F(-link(p̂))`. For logit and probit that is `1 - p̂`, and for cloglog it is a different wrong value (p̂ = 0.1 starts at mu ≈ 0.9999).

- `binomial_location_scale_link_eta_from_probability` is replaced by `binomial_location_scale_threshold_from_probability`. It returns `-link(p̂)`, which is the inverse of the family's own `q0` map at `sigma = 1`.
- The cloglog quantile now forms `ln(1 - p)` as `ln_1p(-p)`, so a rare prevalence keeps its digits.
- The existing test's assertion changes to the negated logit.
- New test: `warm_start_threshold_reproduces_the_prevalence_through_the_family_map` round-trips the seed through `binomial_location_scale_q0` and the inverse-link jet for logit, probit and cloglog. It asserts `|mu - p| <= 64 ε`.

## #3527: cloglog saturation NaN

In `binomial_location_scale_log_likelihood` (CLogLog arm), a success row with `q > ~709.78` has `z = exp(q) = inf`. The code set `log_p = 0` correctly, but `(1 - y) * log_survival = 0 * -inf = NaN`, so the row was refused as non-finite. Its exact log-likelihood is 0.

Both terms now use the xlogy convention: a term with a zero coefficient contributes exactly 0. A failure row at `z = inf` is genuinely impossible and is still refused. New test: `saturated_cloglog_success_row_scores_zero_and_failure_row_is_refused`.

## Validation

CI plus `focused-rust-proof.yml` on this branch (`gam-models` lib, filters `warm_start_prevalence_tests`, `cloglog_saturation_tests`, `binomial_location_scale`).

---


SauersML:
Taken over by RF-0748-1; driving to merge.

---

PR #3566, opened 2026-09-20
bms: drop the fake usize::MAX guard from block_linear_constraints

### LINKED ISSUE #3546: bms: block_linear_constraints carries a fake usize::MAX "unreachable" error, and block_coefficient_coordinate swallows its Err as Structural
## Where (origin/main 1a3a498e3b)

`crates/gam-models/src/bms/custom_family_impl.rs:2101-2127`, in `impl CustomFamily for BernoulliMarginalSlopeFamily`:

```rust
fn block_linear_constraints(
    &self,
    block_states: &[ParameterBlockState],
    block_idx: usize,
    spec: &ParameterBlockSpec,
) -> Result<Option<ConstraintSet>, String> {
    if block_states.len() == usize::MAX
        || block_idx == usize::MAX
        || spec.design.ncols() == usize::MAX
    {
        return Err("unreachable bernoulli marginal-slope constraint state".to_string());
    }
    ...
```

and `custom_family_impl.rs:2085-2099`:

```rust
match self.block_linear_constraints(block_states, block_index, block_spec) {
    Ok(Some(_)) => crate::custom_family::CoefficientCoordinate::Structural,
    Ok(None) => crate::custom_family::CoefficientCoordinate::Spanning,
    Err(_) => crate::custom_family::CoefficientCoordinate::Structural,
}
```

## Why it is wrong

- The guard can never fire. A slice cannot have `usize::MAX` elements, and neither can a design with `usize::MAX` columns. `block_idx == usize::MAX` is not a state any caller produces. The only job of the `if` is to "use" the `block_states` and `spec` parameters so the compiler does not warn about them. The function body never reads either parameter: the constraints depend only on `block_idx` and on `self.score_warp` / `self.link_dev`. So this is dead code dressed up as validation, and it shows readers an error path that does not exist.
- Because of that fake error path, `block_coefficient_coordinate` has an `Err(_) => Structural` arm that silently throws away an error message. Today that arm is unreachable. But once someone adds a real fallible check here, its error would be quietly turned into a "Structural" coordinate classification and never reach the user, which is exactly the silent swallowing that "do not paper over solver issues" rules out.
- `block_coefficient_coordinate` builds the whole dense monotonicity `ConstraintSet` (via `structural_monotonicity_constraints`) only to test whether it is `Some`.

## Principled fix

Move the structural-constraint lookup into one infallible helper keyed on the block index, and have both trait methods use it:

```rust
fn structural_block_constraints(&self, block_idx: usize) -> Option<ConstraintSet> {
    let runtime = if self.score_block_index() == Some(block_idx) {
        self.score_warp.as_ref()
    } else if self.link_block_index() == Some(block_idx) {
        self.link_dev.as_ref()
    } else {
        None
    };
    runtime.map(|r| ConstraintSet::Dense(r.structural_monotonicity_constraints()))
}
```

- `block_linear_constraints` returns `Ok(self.structural_block_constraints(block_idx))`, with its unused parameters `_`-prefixed and the fake guard deleted.
- `block_coefficient_coordinate` classifies the block without a `Result`, so the `Err(_)` arm goes away. It also only needs to know whether the block is a deviation block that has a runtime, so it does not need to build the matrix.

Behaviour does not change on any reachable input.

SauersML:
Fixes #3546

`BernoulliMarginalSlopeFamily::block_linear_constraints` (`crates/gam-models/src/bms/custom_family_impl.rs`) started with this check:

```rust
if block_states.len() == usize::MAX || block_idx == usize::MAX || spec.design.ncols() == usize::MAX {
    return Err("unreachable bernoulli marginal-slope constraint state".to_string());
}
```

No reachable state gets through that check to the error. It only existed to "use" two parameters the body never reads, and it advertised an error path that does not exist. `block_coefficient_coordinate` would then have mapped that path to `Structural`.

The structural monotonicity rows depend only on the score-warp or link-deviation runtime. So the constraint answer is now built from the block index and the runtime, and the unused parameters are underscored.

Behaviour is identical on every reachable input:
- The score-warp and link-deviation blocks return `Some(Dense(rows))`.
- Every other block returns `None`.
- `block_coefficient_coordinate` still reads this answer, as the trait contract asks.

---


SauersML:
Taken over by RF-0748-3; driving to merge. This PR has had no push since 08:16.

As written, the head fails the build.rs ban scan. `_block_states` and `_spec` on a `CustomFamily` impl are underscore fn parameters, and the rule has no exemption for a local trait: "Use the parameter, restructure the API so it isn't passed, or delete the param". Every other implementer uses those two parameters for real: the block index must be in range, and the state's width must equal the spec's design width (see `gamlss/gaussian/wiggle.rs`). So BMS will do the same, and will also check that the monotonicity rows have the block's width.

---

SauersML:

The local validation found that `link_dev_without_score_warp_exposes_structural_derivative_lower_bounds` fails with the new width check. The test posed the slope block's constraint lookup against a link-width dummy spec. I pushed 061097af, which gives each lookup a spec as wide as its own block, and I'm now re-validating.

---

PR #3570, opened 2026-09-20
Reject a fractional power on a hybrid Duchon basis instead of building the s = 0 kernel (#3541)

### LINKED ISSUE #3541: crates/gam-terms/src/basis: hybrid Duchon (length_scale=Some) silently replaces a fractional power with s=0 instead of rejecting it
## Defect

At the basis layer, a hybrid Duchon–Matérn spec (`length_scale = Some(_)`) with a non-integer `power` is not rejected. It is silently rebuilt as the **s = 0** kernel. The spec, the fit metadata and the caller all still say `power = 1.5` (or whatever was asked for), but the kernel that is realized is a different one.

`duchon_power_to_usize` (`crates/gam-terms/src/basis/types.rs:1187`, origin/main) maps every fractional value to `0`. It does **not** floor or round it:

```rust
pub(crate) fn duchon_power_to_usize(power: f64) -> usize {
    if !power.is_finite() || power < 0.0 { return 0; }
    let rounded = power.round();
    if (rounded - power).abs() > 1e-9 { return 0; }
    rounded as usize
}
```

The doc comment says "validators downstream reject" such values. They do not. Every basis builder validates the **already-coerced** integer, so the validator never sees the fractional request:

- `crates/gam-terms/src/basis/duchon_thinplate.rs:533-541` (`build_duchon_basis`):
  ```rust
  let validation_power = if spec.length_scale.is_some() {
      spec.power_as_usize() as f64
  } else { spec.power };
  validate_duchon_kernel_orders(spec.length_scale, p_order, validation_power, data.ncols())?;
  ```
- `crates/gam-terms/src/basis/duchon_psi_derivatives.rs:1912-1919`: the same pattern, using `duchon_power_to_usize(s_order)`.
- `crates/gam-terms/src/basis/periodic_duchon.rs:1069-1072`: `let s_order = spec.power_as_usize(); validate_duchon_kernel_orders(spec.length_scale, p_order, s_order as f64, 1)?;`
- `crates/gam-terms/src/basis/periodic_duchon.rs:1722` (`duchon_center_kernel_value_matrix`) and `:2262` (the operator-penalty factory) read `duchon_power_to_usize(power)` and build with it.

The only case that currently gets rejected is the gh#750 one. There the truncated `s = 0` happens to fail `2(p+s) > d` (d = 4, p = 2). Every other fractional hybrid request passes and builds the wrong kernel. For example:

- d = 1, Linear, `power = 0.5` gives the s = 0 kernel.
- d = 2, Linear, `power = 1.5` gives the s = 0 kernel. The truncated order `2(2+0) = 4 > 2` passes the gate, so no error is raised.

### Reachability

The formula front-end refuses this combination: `term_builder.rs:3736` rejects an explicit fractional power when `length_scale` is set, and it floors the cubic default. The basis builder, however, is public, and at least one production front-end does not guard it. `gam-pyffi` `resolve_duchon_hybrid_config` (`crates/gam-pyffi/src/manifold/manifold_and_posterior_ffi.rs:3787-3793`) sets `power = explicit_power.unwrap_or(auto_power)` and passes a fractional explicit power straight through to `build_duchon_basis_design_and_jets` / `build_duchon_basis` when `length_scale` is given (`model_ffi.rs:2774`, `:3181`, `:3369`, `:3380`). So `duchon_basis(x, centers, length_scale=1.0, power=1.5)` returns an s = 0 design with no error.

This is silent error swallowing. The invalid request is not reported; a different model is substituted for it. The comment in `term_builder.rs:3773-3790` shows that the front-end had to work around exactly this behaviour ("the basis builder's `power_as_usize` maps a NON-integer to `0` (not its floor)").

## Principled fix

1. In `validate_duchon_kernel_orders` (`duchon_kernel_math.rs:1525`), reject a non-integer `s_order` when `length_scale.is_some()`. The partial-fraction split of `1/(ρ^{2p}(κ²+ρ²)^s)` is defined only for integer `s`, so the error message should say that and point to "drop length_scale for a fractional power".
2. At the four call sites above, pass the **raw** `spec.power` or `power` to the validator instead of the pre-truncated integer. This removes the `validation_power` masking, so the gh#750 configuration is rejected for the real reason (fractional hybrid power) rather than by accident.
3. With both changes in place, `duchon_power_to_usize` is only ever called on validated integers. It could then return `Option<usize>` / `debug_assert!` instead of the silent `0`. `term_builder`'s own guard remains a user-facing early message.
4. Update `hybrid_duchon_fractional_default_d4_rejects_realized_nonfinite_kernel` (`tests/src_modules/smooths/basis_radial_periodic_thinplate_tests.rs:4512`). It pins the old, incidental error text ("2*(p+s) > dimension"). The correct reason for the rejection is the fractional hybrid power. Also add a d = 1 / d = 2 case, which currently builds silently.

SauersML:
## Summary

The hybrid Duchon-Matern kernel is the partial-fraction split of `1/(rho^(2p) (kappa^2+rho^2)^s)`, which exists only for integer s. The basis layer read a fractional hybrid power as s = 0 and then validated that truncated integer. As a result, d = 1 and d = 2 fractional hybrid requests silently built the s = 0 kernel, and gh#750 (d = 4) was rejected only by accident.

- `duchon_hybrid_s_order` now refuses a non-integer hybrid power and says why.
  - `validate_duchon_kernel_orders` calls it when `length_scale = Some`.
  - Every builder now passes it the raw power; the `validation_power` masking is gone.
- Every hybrid-only path reads s through the fallible accessor: sphere, periodic, collocation, psi derivatives, N-D jets and the closed-form operator factory.
- `duchon_power_to_usize` is exact. It only feeds routines whose hybrid branch has already been validated.
- The 1-D pure derivative builder now reproduces the forward 1-D design:
  - It evaluates the literal power instead of differentiating a truncated-s kernel.
  - Its polynomial null-space columns are monomials of the center-cloud-centred coordinate (#1375). The raw `t` monomials it used before span the same space but are a different basis, so even an integer power's order-0 output did not match `build_duchon_basis`.
  - It rotates by the frozen radial chart after `K·Z`, the product order the forward design takes (#1355).

## Tests

- The gh#750 test becomes a d = 1, 2, 4 refusal table. It also checks that the scale-free kernel still accepts the fractional power.
- A new test replays the forward 1-D design through the derivative builder at powers 0 and 0.25.

Validated on focused-rust-proof.

Fixes #3541


SauersML:
Taken over by wave8 session W8-75: merging main in and validating.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, a3398487671d.

PR #3576, opened 2026-09-20
bms: stop flooring the outer tolerance at 2e-5

SauersML:
Refs #2469. This removes the two BMS lines that `scripts/spec_ban_ledger.tsv` tracked under that issue.

`crates/gam-models/src/bms/block_specs.rs` `inner_fit` and `inner_fit_from_certified_outer` both did this:

```rust
options.outer_tol = options.outer_tol.max(2.0e-5);
```

Since 9ba55dfa34 removed the user-settable `outer_tol`, every BMS fit arrives with the shared `BlockwiseFitOptions::default()` value of 1e-5 (`gam-model-api/src/families/custom_family/options.rs:618`). The floor therefore did one thing only: it doubled the outer stop for this one family.

- It is a hand-picked constant with no derivation. The comment above it is about keeping the exact outer Hessian declared and never justifies 2e-5.
- `fit_custom_family` also reads `outer_tol` as the criterion resolution (`fit.rs:1528`, `outer_rel_cost_tol.unwrap_or(outer_tol)`) and as the optimizer tolerance (`fit.rs:2906`). So the floor also loosened the certificate's resolution for BMS alone.

Changes:
- Both floor lines are deleted.
- `inner_fit` now passes the caller's options through, so it no longer needs to clone them.
- The two ledger lines are removed.

---


SauersML:
Taken over by RF-0748-3; driving to merge. This PR has had no push since 08:17. I merged main in, and the ledger conflict resolved to drop both lines, since main had already dropped the `deviation_runtime.rs > 1e-12` line. Before merging I'll run the BMS lib tests at the tighter outer stop.

---

PR #3578, opened 2026-09-20
gam-linalg: drop the unpivoted LDLT rung from the symmetric fallback

### LINKED ISSUE #3519: gam-linalg: factorize_symmetricwith_fallback's unpivoted-LDLT rung is not backward stable on indefinite systems (wrong solutions / refused well-conditioned solves)
## Where

`crates/gam-linalg/src/faer_ndarray.rs:1123-1139` (origin/main):

```rust
/// Factorize a symmetric system with LLT -> LDLT -> LBLT fallback.
pub fn factorize_symmetricwith_fallback(matrix, side) -> Result<FaerSymmetricFactor, FaerLinalgError> {
    if let Ok(llt) = FaerLlt::new(matrix, side) {
        return Ok(FaerSymmetricFactor::Llt(llt));
    }
    let ldlt_err = match FaerLdlt::new(matrix, side) {
        Ok(ldlt) => return Ok(FaerSymmetricFactor::Ldlt(ldlt)),
        Err(err) => err,
    };
    let lblt = catch_unwind(AssertUnwindSafe(|| FaerLblt::new(matrix, side)))
        .map_err(|_| FaerLinalgError::Ldlt(ldlt_err))?;
    Ok(FaerSymmetricFactor::Lblt(lblt))
}
```

## Why it is wrong

faer 0.24.4 `Ldlt::new` is **unpivoted** LDLᵀ (`linalg::cholesky::ldlt::factor::cholesky_in_place`). Its only error is `LdltError::ZeroPivot`, meaning an *exactly* zero pivot. LLT has just failed, so the matrix is indefinite, or not PD to working precision. On an indefinite matrix, unpivoted LDLᵀ has unbounded element growth and is not backward stable (Higham, *Accuracy and Stability*, §11.1). A tiny nonzero leading pivot is accepted, and the solve is then garbage. Bunch–Kaufman LBLᵀ, the third rung, is the backward-stable factorization for this case. The middle rung sits in front of it and blocks it for every input except an exactly zero pivot.

Concrete 2×2, with cond₂(A) ≈ 2.6:

```
A = [[1e-17, 1], [1, 1]],   b = [1, 2],   exact x ≈ [1, 1]
```

LLT fails (second pivot 1 − 1e17 < 0). LDLT succeeds with D = (1e-17, −1e17), l₂₁ = 1e17, and its solve returns **x = [0, 1]**, a 100% error in x₁. I reproduced this by stepping through the LDLᵀ forward and back substitution in IEEE double. The existing regression `tests/regressions/misc/matrix_2_2.rs` only uses `[[0,1],[1,0]]`. That matrix has an exact zero pivot, so LDLT refuses it and LBLT handles it. The unstable path is never exercised.

It also breaks SPEC: *"In general, do not paper over solver issues."* This is a reactive ladder chosen by which factorization happens to succeed. The `catch_unwind` around LBLT also turns a panic into the LDLT error message.

## Impact (production callers)

- `gam_linalg::utils::certified_symmetric_matrix_solve` / `certified_symmetric_solve` / `solve_dense_block_system` (via `StableSolver::factorize`). The a-posteriori backward-error certificate catches the bad solve, so a **nonsingular, well-conditioned indefinite system is refused** instead of solved.
- Callers with no certificate take the wrong answer silently whenever the result is finite:
  - `gam-solve/src/reml/per_atom_efs.rs:343`: shared-border Newton correction on an outer-Hessian block that can be indefinite away from the optimum.
  - `gam-terms/src/inference/lawley.rs:329,882`: `J⁻¹`.
  - `gam-solve/src/inference/pg_gate_evidence.rs:221`
  - `gam-models/src/bms/block_specs.rs:1320`
  - `gam-pyffi/src/latent/reml_latent_fit_ffi.rs:1052`
  - `gam_linalg::utils::gaussian_weighted_ridge`, and its analytic VJP in `gaussian_weighted_ridge_backward.rs:114`.

## Principled fix

1. Delete the `Ldlt` rung of `factorize_symmetricwith_fallback`, and the `FaerSymmetricFactor::Ldlt` variant if nothing else builds it. The function becomes: strict LLT (the PD certificate), otherwise pivoted LBLᵀ (backward stable for any nonsingular symmetric matrix). Drop the `catch_unwind`, or document which faer panic it exists for and return a typed error instead of reusing the LDLT error.
2. Callers that *require* SPD (the multinomial Firth `invert_spd`/`spd_logdet`, and every place whose comment says "SPD") should call strict LLT (`certified_spd_factorize`) directly. They should not rely on `logdet()` returning NaN from `ln(D<0)` to detect indefiniteness.
3. Add a regression using the `[[1e-17,1],[1,1]]` system above: `certified_symmetric_solve` must return ≈[1,1] with a passing certificate.

SauersML:
Fixes #3519

## What changes

`factorize_symmetricwith_fallback` now tries Cholesky and then Bunch–Kaufman LBLᵀ. The unpivoted LDLᵀ middle rung is gone.

## Why

Cholesky only refuses when the matrix is indefinite, or not positive definite to working precision. On those matrices unpivoted LDLᵀ is not backward stable. It accepts any nonzero pivot, however small.

The case from the issue is `[[1e-17, 1], [1, 1]]`, with cond₂ ≈ 2.6. LDLᵀ returns x = (0, 1), but the exact answer is x ≈ (1, 1).

## Also in this PR

- `FaerLblt::logdet` reads ln det off the pivot blocks. It returns NaN whenever any block has a negative eigenvalue. This matches the old LDLᵀ arm, where ln of a negative pivot gave NaN. Before this PR the LBLT arm always returned NaN.
- `FaerLdlt` and `FaerLinalgError::Ldlt` had no other users, so they are deleted.

## Tests

Both are in `tests/regressions/misc/matrix_2_2.rs`:
- `factorize_symmetric_with_fallback_is_backward_stable_on_tiny_leading_pivot_3519` checks that the normwise backward error is at most `γ_{3n}·ρ_BK`. The unpivoted solve has a backward error of 1/4.
- `factorize_symmetric_with_fallback_logdet_is_nan_off_the_semidefinite_cone_3519` checks the NaN cases: the indefinite matrix, and −I₂, where det = 1 but the matrix is not PSD.

---


SauersML:

---

PR #3579, opened 2026-09-20
sparse_dict/block_chart: run chart e-BH over the candidate universe, not the screened set

### LINKED ISSUE #3510: sparse_dict/block_chart: e-BH family is chosen by a ring-shape screen on the same rows, which voids the FDR guarantee
## Where

`crates/gam-sae/src/sparse_dict/block_chart.rs` on `origin/main`:

- `compose_block_coordinate_charts` (around lines 330-360) builds the e-BH family from only the **screened** candidates, then calls `family_fdr_certificate(family_log_e, CHART_FDR_ALPHA)`. Inside, `e_benjamini_hochberg` (`gam-terms/src/inference/structure_evidence.rs:578`) uses `m = log_e_values.len()`, so `m` is the size of the screened family.
- The pair screen is `screen_pairs` (around line 775), which keeps a pair only if `score >= config.pair_min_score` (default `0.20`) and keeps at most the top `max_pairs` by score. The score comes from `pair_score` (around line 1060):

```rust
fn pair_score(z0: &Array2<f32>, z1: &Array2<f32>) -> Result<f64, String> {
    let joint = hstack(z0, z1);
    let whitening = fit_whitening(&joint, 1.0e-8)?;
    ...
    Ok((1.0 - var.sqrt() / mean).clamp(0.0, 1.0))   // 1 - CV of the whitened radius
}
```

- The singles come from `select_blocks`, which keeps the top `max_blocks` of the blocks with at least `min_firings` rows, ranked by centered energy.

## Why this is wrong

`pair_score` is one minus the coefficient of variation of the whitened radius, which makes it a ring detector. It runs on exactly the rows (`rows_for_pair`) that `shell_vs_ring_log_evalue` then uses for its fit and evaluation folds. So on data where the linear-shell null is true, the family keeps the candidates whose rows happen to look most ring-like, and those are the ones whose split-LR e-values are inflated. Given selection, `E[E_i | selected] > 1`, and e-BH at `m = |selected|` then has no FDR guarantee.

The module's own doc (`split_lr_fdr.rs`, "Full-family e-BH") already names this failure: *"gating on the same statistic post hoc redefines the family and voids the guarantee."* The pair screen does that gating with a statistic very close to the ring statistic. Energy selection of the singles is also data-dependent, though less directly tied to the ring statistic.

The existing test `null_battery_rank1_shell_fdr_controlled` in `split_lr_fdr.rs` does not catch this. It builds a fixed family of 12 with no screen, so it never runs the screen-then-test pipeline.

## Principled fix (valid under arbitrary dependence, no extra split)

For any data-dependent screen `S_i` and any e-value `E_i >= 0`, the product `E_i * 1{S_i}` is still an e-value: `E[E_i 1{S_i}] <= E[E_i] <= 1`. e-BH is valid for arbitrarily dependent e-values over a **fixed** family. So:

1. Fix the family before looking at the data. It is the universe of candidates the screen can ever choose from: `G = decoder.nrows() / block_size` singles, plus `G(G-1)/2` pairs when `pair_screen` is on.
2. Every candidate the screen drops (below `min_firings`, below `pair_min_score`, cut by `max_blocks`/`max_pairs`/`pair_top_blocks`) banks the exact zero e-value, `log_e = -inf`.
3. Run e-BH with `m = |universe|`. Do not materialize `C(G,2)` `-inf` entries. Instead, add a family-size argument to the gam-terms e-BH (for example `e_benjamini_hochberg_in_family(log_e, family_size, alpha)`, with `e_benjamini_hochberg(v, a) = ..._in_family(v, v.len(), a)`), and pass `family_size` through `family_fdr_certificate`.

The power cost is only `ln m`, and UI e-values grow linearly in `n`, so the cost is small. With this change the screen is back to what the doc says it is: a compute budget, not a gate.

Regression test: a screen-then-test simulation under the rank-1 shell null through `compose_block_coordinate_charts` (or a direct harness of `screen_pairs` + `shell_vs_ring_log_evalue`). Its empirical FDR must be `<= alpha`. Today the screened version inflates it.

SauersML:
Fixes #3510

`compose_block_coordinate_charts` screens candidates (firing counts, co-firing, pair score) on the same rows it then certifies, and ran e-BH with `m = |screened|`. Choosing the family after seeing the data voids the e-BH FDR guarantee.

Fix: screened-out candidates bank the exact zero e-value. `E_i·1{S_i}` is still an e-value, and e-BH is valid under arbitrary dependence over a **fixed** family. So e-BH now runs over the universe the screen chooses from, which is fixed before any data is seen: `G = decoder.nrows()/block_size` singles, plus `G(G−1)/2` pairs when `pair_screen` is on. Members with zero e-values can never be rejected. They only enter through `m` in the thresholds `m/(α·k)`, so nothing needs to be materialized.

Changes:
- **gam-terms** `inference/structure_evidence.rs`:
  - new `e_benjamini_hochberg_in_family(log_e, family_size, alpha)`, which refuses `family_size < len` with a new `EBhError::FamilySmallerThanClaims`;
  - `e_benjamini_hochberg` now delegates to it with `family_size = len`, so its behaviour is unchanged.
- **gam-sae** `sparse_dict/split_lr_fdr.rs`: `family_fdr_certificate` takes `family_size` and records it in `FdrCertificate.family_size`.
- **gam-sae** `sparse_dict/block_chart.rs`:
  - `chart_fdr_family_size(n_blocks, pair_screen)` uses checked arithmetic;
  - the compose step certifies over that family and exposes `fdr_family_size` on the result.
- Tests:
  - the in-family rule reproduces plain e-BH when `family_size = len` and matches explicit −∞ padding;
  - a larger family raises the bar;
  - a family smaller than its claims is refused;
  - the universe-size helper is checked;
  - an end-to-end compose test uses 4 blocks of which only 2 fire. It asserts that the family is 10, not 3, and that every discovery clears the bar at the universe `m`.

---


SauersML:
Taken over by RF-0748-1; driving to merge.

---

PR #3582, opened 2026-09-20
pirls: refuse a non-finite EDF instead of fabricating p - rank(S)

### LINKED ISSUE #3577: crates/gam-solve/src/pirls: a non-finite EDF is silently replaced by the fabricated value p − rank(S)
**Where** (origin/main d7fe081e): `crates/gam-solve/src/pirls/loop_driver.rs:2089-2098`

```rust
let mut edf = if let Some(dense_h) = penalized_hessian_transformed.as_dense() {
    calculate_edfwithworkspace_with_penalty(dense_h, &penalty_active, &mut saved_workspace)?
} else {
    calculate_edf_with_penalty(&penalized_hessian_transformed, &penalty_active)?
};
if !edf.is_finite() || edf.is_nan() {
    let p = penalized_hessian_transformed.ncols() as f64;
    let r = penalty_active.rank() as f64;
    edf = (p - r).max(0.0);
}
```

and the helper that feeds it, `crates/gam-solve/src/pirls/edf.rs:216-233`:

```rust
pub(super) fn edf_from_solution<F>(p, r, mp, e_transformed, solved_at: F) -> f64 {
    let mut tr = 0.0;
    for j in 0..r { for i in 0..p { tr += solved_at(i, j) * e_transformed[(j, i)]; } }
    (p as f64 - tr).clamp(mp, p as f64)
}
```

**Why it is wrong**
- `f64::clamp` passes NaN through. So when tr(H⁻¹S) is not a number (for example a non-finite entry in `e_transformed`, or an overflow in the product), `edf_from_solution` returns NaN wrapped in `Ok`.
- Every other EDF path in `edf.rs` refuses a non-finite result with `ModelIsIllConditioned`. The selected-inverse path at `edf.rs:98` does this with `if tr.is_finite() { ... } return Err(ill_conditioned())`.
- The loop driver then replaces that NaN with `p − rank(S)`. That value is the dimension of the unpenalized null space, i.e. the *smallest* EDF the model could have. It is not an estimate of the EDF at this fit.
- The fabricated EDF goes into `PirlsResult.edf`, and from there into the scale estimate, GCV/REML and the reported fit.
- SPEC: "In general, do not paper over solver issues."
- SPEC: "A fit object must only ever come from a converged optimization."
- Apart from that one NaN pass-through, the fallback is dead code: every caller of `edf_from_solution` already checks that the solve is finite.

**Principled fix**
- `edf_from_solution` returns `Result<f64, EstimationError>` and refuses a non-finite trace with `ModelIsIllConditioned`, the same as the selected-inverse path.
- Delete the loop-driver fallback, so `edf` becomes immutable.

SauersML:
Fixes #3577

- `edf_from_solution` (`pirls/edf.rs`) now returns `Result`.
  - Before, `(p - tr).clamp(mp, p)` passed a NaN trace straight through, because `f64::clamp` does not catch NaN.
  - It now refuses a non-finite tr(H⁻¹S) with `ModelIsIllConditioned`, the same way the selected-inverse EDF path already does.
- The loop-driver fallback that replaced a non-finite EDF with `p − rank(S)` is deleted.
  - That value is the dimension of the unpenalized null space, not the EDF of the fit.
  - `edf` is now immutable.

---


SauersML:
Verified locally and merged (head 0ec42616, merged with origin/main in a batch tree together with other verified PRs).

- `cargo check --workspace --tests` on the merged tree: clean, ban scanner passing
- `cargo test -p gam-solve --lib pirls`: 190 passed, 0 failed
- `cargo test -p gam-solve --lib reml::`: 489 passed, 0 failed

Review against SPEC.md found no retry or fallback ladder, no magic constants and no loosened tolerances.

---

PR #3585, opened 2026-09-20
joint Newton PCG: derived forcing floor replaces the literal 1e-8 (#3285 item 3)

SauersML:
Refs #3285 (item 3 of the remaining list in https://github.com/SauersML/gam/issues/3285#issuecomment-5748270753). The issue is already closed; this lands the forcing-floor item it listed.

## Defect
The inner joint-Newton CG solve `(H + S)δ = rhs`, with `rhs = ∇L − Sβ`, has an Eisenstat-Walker forcing `η_EW = 0.9·(‖g_k‖/‖g_{k−1}‖)^φ`. On main it is clamped to the literal `[PCG_ETA_MIN = 1e-8, PCG_ETA_MAX = 0.1]`, and a residual pair with no finite ratio falls to the literal `JOINT_PCG_REL_TOL = 1e-8`. Near the mode, EW drives η down to that literal, so CG is asked for digits the right-hand side does not have. That is a magic constant, and it also inflates the Chebyshev iteration bound `k̂` that #3424's up-front route decision reads. gam-2930's replay measured the derived floor: it raised η on 601 of 1,833 steps and cut their CG products from a median of 40 to 2.

## Fix
`η′ = max(η_EW, min(band/‖rhs‖₂, PCG_ETA_MAX), γ_p)`

- **The band.** `band` is `joint_stationarity_rounding_band(s_lambdas, betas, joint_bundle, ‖∇L‖∞, n)`, the ∞-norm rounding band of `∇L − Sβ` that the KKT certificate already reads.
- **Why the floor is conservative.** CG stops at `‖r‖₂ ≤ η‖rhs‖₂`. When the floor binds, `‖r‖∞ ≤ ‖r‖₂ ≤ band`, so the residual it accepts lies inside the rounding band in the certificate's own norm. Because `‖rhs‖₂ ≥ ‖rhs‖∞`, the ratio `band/‖rhs‖₂` can only be tighter than the ∞-norm ratio.
- **The `PCG_ETA_MAX` cap.** It keeps η < 1 at an iterate that is already stationary within its band (`‖rhs‖₂ ≤ band`, including `rhs = 0`).
- **`γ_p`.** `γ_p = accumulation_growth(p)` is the attainable relative residual that `pcg_core` and `pcg_iteration_bound` already raise every target to. It keeps the floor positive when the band carries no information (band 0 or NaN).
- **No finite EW ratio.** When the current residual is non-finite, or the previous one is ≤ 0 or non-finite, the solve goes to this resolution floor instead of the literal 1e-8.
- **Computed only when needed.** The floor is computed only for a `pcg_eligible` step (exact_joint_fit.rs), inside the closure that builds the CG preconditioner. Dense-routed steps pay nothing.
- **Deleted.** `PCG_ETA_MIN` and `JOINT_PCG_REL_TOL`.

## Tests
New unit tests in `joint_newton::joint_pcg_forcing_tests`:
- `forcing_floor_is_the_rhs_resolution_capped_below_one_3285`: the band ratio binds between γ_p and the cap; an iterate stationary within its band and `rhs = 0` both give PCG_ETA_MAX; band 0 or NaN gives γ_p.
- `eisenstat_walker_forcing_never_asks_below_the_resolution_floor_3285`: near the mode the result is held at the floor; far from it EW is unchanged; a missing ratio gives the floor.

---


SauersML:
[swarm lane gam-2930, #3285 measurement record: issuecomment-5748585948] No separating case. This is the one forcing mechanism #3424 left open ("Forcing floor" under its remaining items): Eisenstat–Walker with the literal `PCG_ETA_MIN = 1e-8` clamp and the `JOINT_PCG_REL_TOL` fallback replaced by the derived resolution floor. It adds no second route or literal beside #3424. `PCG_ETA_MAX` and `γ_rows` already existed.

**Measured effect of this floor (the #3285 replay, BMS n = 80,016, p = 71, 1,833 joint-Newton steps).** `max(η_EW, band/‖rhs‖₂)` raised η on **601** of the 1,833 steps and cut their CG products from a median of **40 to 2**.

On merged main this matters only on steps that actually run CG:
- the `PcgAttempt::Only` regime;
- steps whose Chebyshev bound passes. On these shapes the bound goes dense, since κ̄ has a median of 1.3e10.

**One difference from the version I measured, and it is an improvement.** Capping the floor at `PCG_ETA_MAX` keeps `η < 1` when `‖rhs‖₂ ≤ band`. An uncapped `band/‖rhs‖₂ ≥ 1` would let CG stop at the zero start.

**Not addressed here, and separate:** a CG miss under `PcgAttempt::Only` still refuses the trial point.

Issue #3502, opened 2026-09-20
small_extrapolation_predictions_are_bounded red: y ~ matern(x) iso-kappa joint REML stops at an indefinite point (|Pg|=2.86e-1, λ_min(H)=-2.45e-2, RejectFloor after 60 iterations)


SauersML:
## Observed

Command:

```
cargo test --test prediction bounded
```

Run on the #3450 branch (head 19b00b42). That branch is main e80292d6 plus a change confined to the bounded-term Gaussian dispersion profile (`fit_bounded_term_collection_with_design`). `matern(x)` has no bounded term and never enters that path.

`predict::predict_outside_train_range_bounded::small_extrapolation_predictions_are_bounded` panics at `tests/prediction/predict/predict_outside_train_range_bounded.rs:46`. The first case, `y ~ matern(x)` (n=240, σ=0.05, seed 251, gaussian), is refused:

```
Outer smoothing-parameter optimization did not certify a stationary optimum (iso-kappa joint REML):
gradient |g|=2.858e-1 |Pg|=2.858e-1 bound=1.000e-4 (rung=solver-band derived_standard=false)
hessian_psd=NO (cleared=true λ_min(H)=-2.451277e-2 λ_min(H+diag|g|)=-2.076190e-10 max_k|g_k|=2.797358e-1 ...)
curvature_source=terminal-analytic railed=[]
solver=Arc, termination=trust_region_reject_floor(radius=1.000000e-12 at floor 1.000000e-12 after 5 consecutive rejections)
stop_reason=RejectFloor after 60 outer iteration(s); final objective -3.400332e2
rho_checkpoint = [-14.948623738137107, -12.487505776829277, -1.535205316495082, -0.85029397023849, 2.0174225678671176]
```

## Reading

The analytic model has a non-zero gradient: |Pg| = 0.29, far above the 1e-4 bound. Even so, ARC rejects every step down to a radius of 1e-12. The analytic model is therefore not predicting the objective's change along its own descent direction. That points to a gradient/objective mismatch in the iso-kappa joint REML, not a genuine stall.

Two of the ρ coordinates sit near -15 and -12.5, which looks like the same family of failure as #3344 (iso-kappa joint REML ending at an indefinite point, with ρ₁ pinned near the floor).

I have not yet checked whether this reproduces identically on main e80292d6. The failing formula does not touch the changed code path.

Refs #3344

SauersML:
[...]

Issue #3344, opened 2026-09-20
regression_2726_out_of_window_fixture_still_fits red on main: iso-kappa joint REML ends at an indefinite point (|Pg|=451, λ_min(H)=−734) with ψ-adjacent ρ₁ pinned at the box floor


SauersML:
## Repro
On main `1308538a`:

```
GAM_FORCE=1 ./build.sh test --test regressions -- regression_2726_out_of_window_fixture_still_fits
```

The test panics at `tests/regressions/misc/regression_2726_joint_psi_seed_shares_theta0.rs:252`:

```
Outer smoothing-parameter optimization did not certify a stationary optimum (iso-kappa joint REML):
gradient |g|=4.511e2 |Pg|=4.511e2 bound=1.000e-5 (rung=solver-band) hessian_psd=NO
λ_min(H)=-7.342723e2 λ_min(H+diag|g|)=-2.831763e2 curvature_source=terminal-analytic
railed=[1] [#1 theta=-1.923053e1 box=[-1.923053e1, 1.681312e1] margin=5.000e-1]
asymptote-rail declined: interior not stationary (active-face |Pg|=4.511e2)
solver: plan=Arc, termination=trust_region_reject_floor(radius=1e-12 after 40 consecutive rejections), 197 outer iterations
final objective -4.143212e3
rho_checkpoint = [-13.6154, -19.2305, -5.9256, -14.5848, 4.1148, 2.1207]
```

## What's wrong
- ARC stops at its 1e-12 radius floor with |g| = 451. The gradient is not small, so the model is rejecting steps along a direction of large descent. That points to a disagreement between the value and the analytic gradient or Hessian at this point, not to a converged face.
- `railed=[1]`, yet `|Pg| == |g|`. The rail projection removes nothing on coordinate 1, which sits exactly on the hand box floor −19.23. That box is the `rho_domain` hand bound the SPEC says should be deleted.

## SPEC
The refusal itself is correct: "a fit object must only ever come from a converged optimization". The defect is in the search.
- "No caps, hand bounds or boxes": coordinate 1 is stopped by the box, not by a certificate.
- The trust-region reject floor is an unconverged stop.

Found while landing #3343 (the tail-snap deletion). The test fails identically with and without that PR.

SauersML:
Released: w2-surv-ms-c withdraws this claim. The diagnosis is in the comments below. It is the frozen structural-rank refusal, which strike-1b owns.

SauersML:
[...]

PR #3586, opened 2026-09-20
tests: fuzz_fit_invariants can no longer pass having checked nothing

### LINKED ISSUE #3557: tests: fuzz_fit_invariants accepts any refusal, ignores refit errors in I4, and can pass having checked zero fits
## Where (origin/main)

`tests/pathological/misc/fuzz_fit_invariants.rs`

**1. Every fit error counts as a pass, including on well-posed cases.** At :294-297:

```rust
    let result = match fit_from_formula(shape.formula, &data, &cfg) {
        Ok(r) => r,
        Err(_) => return Ok(()), // a loud, typed refusal is fine.
    };
```

This applies to every cell of the grid, including `Degeneracy::None` at n=40:
- `y ~ s(x0)` with 40 rows and a basis of about 10
- `y ~ x0`
- `y ~ x0 + x1`
- a binomial whose success probability is in [0.35, 0.73], so there is no separation

These are ordinary, well-posed GAMs. A regression that makes `fit_from_formula` return `Err` for them (a non-converged outer loop that is now refused, a materialize bug, a wrong family check) turns every I1–I6 check into a no-op and the test stays green. SPEC: "A fit object must only ever come from a converged optimization". The flip side is that a well-posed model which the engine *refuses* is a bug, and this harness is meant to be the discovery net for that.

**2. I4 (determinism) silently ignores a failing refit.** At :299-300 and :369:

```rust
    let result2 = fit_from_formula(shape.formula, &data, &cfg);
    ...
    if let Ok(FitResult::Standard(fit2)) = result2 {
```

Suppose the first fit succeeds and the identical refit returns `Err`, or returns a different `FitResult` variant. That is the strongest possible non-determinism: the same input produced a fit once and a refusal once. Yet I4 skips its check and the case passes. I4 is documented as "refitting identical data with the identical config reproduces β".

**3. No vacuity guard.** The test counts `ran` (grid size) and asserts only `violations.is_empty()` at :515. Nothing counts how many cases actually reached I1. If every fit is refused, the log reads "ran 704 cases, 0 invariant violations".

**4. I5 docs and code disagree.** The doc at :14 says `total EDF ∈ [null_space_dim − tol, p + tol]`, but the code at :401 checks `edf < -1e-6 || edf > p + 1e-3`, i.e. [0, p]. The documented lower bound is the correct one: unpenalized columns each contribute exactly 1 to the EDF. The implemented bound can never catch an EDF that collapses below the parametric dimension.

## Principled fix

- (2) Once the first fit is `Ok(FitResult::Standard(_))`, the refit must also be `Ok(FitResult::Standard(_))` with β within the determinism tolerance. Anything else is an I4 violation.
- (1) and (3):
  - For `Degeneracy::None` at the largest grid n (40, where every shape including `te(x0,x1)`'s 25-column basis has more rows than coefficients), a refusal is a violation, reported with the error text.
  - Count the cases that reached the invariant checks, and require every shape to have at least one checked case.
  - Degenerate or tiny-n cells may still refuse loudly, which is legitimate.
- (4) Either compute the unpenalized dimension from the fit's penalty structure and gate `edf >= null_dim - tol`, or correct the doc to the [0, p] bound actually implemented. The former is the stronger invariant.

SauersML:
Fixes #3557

`tests/pathological/misc/fuzz_fit_invariants.rs` had three gaps:
- It returned `Ok(())` on **any** fit error, in every grid cell.
- It skipped I4 whenever the identical refit errored or changed `FitResult` variant.
- It never counted the cases that actually reached I1–I6.

So a regression that makes `fit_from_formula` refuse ordinary well-posed GAMs would leave the sweep green.

Changes, test only:
- `check_case` now returns `CaseOutcome::{Checked, Refused, NonStandard}`.
- **Well-posed cells must fit.** On a `must_fit` cell (no injected degeneracy, largest grid n = 40, continuous uniform predictors, no separation) a refusal is a violation that carries the engine's error text. Degenerate or tiny-n cells may still refuse loudly.
- **I4.** Once the first fit is `Standard`, the identical refit must also be `Ok(Standard)`. An `Err` or a variant change is reported as non-determinism, not skipped.
- **Vacuity guard.** Checked cases are counted per shape, and a shape that never reached the invariant checks is a violation. The log now prints `ran N cases, checked M`.
- **I5 doc.** It now states the [0, p] bound the code actually enforces. The documented `null_space_dim` lower bound is not universal, because `te()` defaults to a double penalty that also shrinks the marginal null spaces.

No tolerance was loosened and no case was excluded.

Validation: focused-rust-proof on `gam` / integration / `pathological` / `fuzz_fit_universal_invariants`.

---


SauersML:
BLOCKED on #3502. The focused run https://github.com/SauersML/gam/actions/runs/35499182671 failed, and it failed for the right reason: gauss_matern n=40 (seeds 1 and 7) is refused by the iso-kappa REML `RejectFloor` with |Pg| = 0.28 > 0.068 while the Hessian is PSD. That is the same outer-stationarity family as #3502 and #3344. The test is doing its job, so it will not be loosened. This PR can merge once #3502 lands.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 0d5e72dfa7e8.

Issue #2812, opened 2026-09-03
The outer engine's stationarity bound is tol·|criterion| and its ρ domain is ±30; derive both and give every REML route the derived rung


SauersML:
## The outer engine's stationarity bound is `tol·|criterion|` and its ρ domain is `±30`; both are hand numbers every REML route inherits

Split out of #2469 / #2670 after the 2026-09-03 landings, which deleted every *private* box and tolerance that sat beside the engine's (`seed ± 12` in the survival selector, `seed ± 6` in the survival baseline search, the `[−25, 25]` grid and `[−30, 30]` clamp in the full-conformal selector, the `13`-point η scan in the constant-curvature profile) by handing each route the engine's own. What remains is the engine's own:

- `OuterConfig::tolerance = 1e-5` with the bound `max(tol·|cost|, solver_band)` (`rho_optimizer/run.rs`, the `certificate-score-relative` rung). The criterion's absolute level is arbitrary — a REML/LAML value moves under a response rescale — so `tol·|cost|` is not a statement about the gradient. The derived rung (`curvature-resolvability`, `derived_standard = true`) exists but only routes that reach it benefit; the survival transformation route declares `DeclaredHessianForm::Unavailable` and cannot. Measured on the survival selector: the tightened tolerance walks to λ ≈ 5e7 where the inner solve cannot certify (#2670 comment), so "tighten the constant" is not the fix; "derive the bound" is.
- `RHO_BOUND = 30` / `SeedConfig::default().bounds = (−12, 12)` (`estimate/smoothing_correction.rs`, `gam-problem/seeding.rs`): the ρ domain is a hand box. The derived domain is the range in which the penalty is resolvable against the Hessian — per block, `ρ ∈ [ln(ε·‖H‖/‖S_k‖), ln(‖H‖/(ε·‖S_k‖))]` up to the reparameterisation's conditioning — and a search reaching its edge is a block collapsing to its null space, which is a result (edf → nullspace), not a wall.
- `gaussian_reml.rs` runs the engine with `with_tolerance(1e-10)`, `with_required_projected_gradient_norm(Some(1e-8))`, `with_max_iter(200)` and `RHO_LOWER/RHO_UPPER` — a third convention, tighter than the default; loosening it to the default is a quality regression, so it waits on the derived bound too.

[...]

PR #3589, opened 2026-09-20
Accept a railed ρ̂ as a comparable constant-curvature profile value

### LINKED ISSUE #3509: flat_truth_does_not_reject_flatness fails on main: the constant-curvature profile at ψ = (0, ln ℓ = -1.894) refuses because ρ̂ rails at its bound
`tests/identifiability/misc/constant_curvature_kappa_inference_e2e.rs:296` (`flat_truth_does_not_reject_flatness`) fails on main. It fails at fit time, before any inference assertion is reached. The refusal reads:

> constant-curvature profile at ψ = (0, ln ℓ = -1.894) railed ρ̂ at its bound … truncated minimum

**How it was found.** I saw it while validating the #3245 branch (PR #3417). It reproduces the same way on an unmodified origin/main worktree, so that PR did not cause it.

**What is wrong.** With flat truth (κ = 0), the profiled fit at the κ = 0 slice drives a smoothing parameter onto its box edge. The profile then refuses that point as a truncated minimum.
- At κ = 0 the flat-truth test expects an interior or rail-certified optimum.
- Per SPEC, a fit must come only from a converged optimization, so the refusal is correct.
- The root cause is upstream of the refusal. Either:
  - the ρ box for this profile is too small, or
  - the λ → ∞ rail face should mint a typed rail certificate (#2348) instead of a bound refusal.

The test should not be loosened. The profile should either certify the rail or find the interior optimum.

SauersML:
Fixes #3509

## 1. A railed ρ̂ is a comparable profile value

`ConstantCurvatureProfile::comparable_value` refused any (κ, η) at which the closed-form Gaussian REML selector put ρ̂ on a face of its domain. It called that value "a truncated minimum and not comparable". On flat truth (constant mean plus noise, κ⋆ = 0), λ̂ → ∞ is the correct answer at every ψ. That made the κ profile undefined and turned `flat_truth_does_not_reject_flatness` into an error ("terminal value-only certificate evaluation failed: ... railed ρ̂ at its bound").

`fit.rho_domain` is not a box. It is the #2812 resolvability interval `[ln(√ε·γ_min), ln(γ_max/√ε)]`, taken from the penalty's own spectrum.
- Past the upper face, each penalized direction's contribution to V_ρ is at most γ_j/λ.
- So the tail integral from the face to infinity is at most about √ε per direction. The value on the face is therefore lim_{λ→∞} V to within the gradient's own resolution.
- At λ = ∞ only the ψ-free intercept survives. The railed value is the same model value at every ψ.

The derivative side already handled this: `profiled_gaussian_reml_psi_jet` drops the Schur term on a face. This PR makes the value path agree with it. It also removes the `rho_at_bound` field and the `(V, railed)` cache tuple, which nothing reads now.

## 2. The profile-LR endpoint judges monotonicity only inside the LR set

After (1), the real path failed with `curvature profile is not outward-monotone at chart bound -2.1666: outward score -2.6e-1`.

**The set.** The interval is the connected component through κ̂ of the Wilks set

  C = {κ : r(κ) ≤ 0},  r(κ) = V_p(κ) − V_p(κ̂) − ½z²  (i.e. 2(V_p(κ) − V_p(κ̂)) ≤ χ²₁,₁₋α).

Its outward end is the first zero of r past κ̂.

**The rule.** Only points certified inside C (r < 0) need a non-negative outward score.
- An inside point where V_p is falling outward lies past an interior maximum that might already have crossed the level. Continuing outward from it could report a later crossing, which would be a different component.
- A point with r ≥ 0 is outside C whatever its slope. The first crossing lies in (last inside point, that point], and a bracket needs nothing more.

So a chart bound above the level brackets the endpoint, and its score is irrelevant. This confirms BF-57's hypothesis. The old code applied the slope refusal to the bound, and to every bracketing probe, before looking at r. That was the bug.

**The case a slope cannot decide.** A bound can be inside C on a profile that is falling there, as with a flat profile whose small bump never reaches the level. A single-point score cannot tell this apart from a bump that crossed the level, so the refusal stays for that case. For the constant-curvature smooth, a global certificate settles it:

- The penalty's null space is exactly the ψ-free intercept column (the kernel Gram is strictly PD on the sum-to-zero frame). So ν = n − 1 at every (κ, η), and as ρ → ∞, `log|XᵀX+λS| − log|λS|₊ → log n` and `dp → Σ(y−ȳ)²`. The limit V_∞ is the REML value of the intercept-only model.
- The finite-ρ excess over V_∞ is `½Σ_j log(1 + γ_j/λ) + ½ν·log(dp(ρ)/dp_∞)`. The second part is ≤ 0.
- At the upper face ρ_up = ln γ_max − ln √ε, every γ_j/λ ≤ √ε, so `V(ρ_up) ≤ V_∞ + ½·r·log(1+√ε)`.
- The certified branch-and-bound ρ search (`enumerate_and_select_rho_with_controls`) evaluates both walls and returns the least cost, so V(ρ̂) ≤ V(ρ_up). The η profile is a minimum over such values.
- Hence **V_p(κ) ≤ ceiling := V_∞ + ½·r·log(1+√ε) for every κ in the chart**.

Consequences:
- If ceiling − V_p(κ̂) ≤ ½z², then sup_κ r(κ) ≤ 0 and C is exactly the whole chart. The interval is [κ_min, κ_max] with both ends open. This is the Wilks set itself, so it is calibrated, not conservative.
- If V_p(κ̂) > ceiling + tolerance, the theorem's premise is violated, and the fit is refused rather than reported.
- Otherwise the endpoint search runs as before.

The flatness p-value is untouched. No tolerance is loosened: the stationarity check, the not-a-minimum check and the interior-probe rule keep their tolerances.

**Tests** (unit, `spatial_curvature_inference`):
- `a_bound_above_the_wilks_level_brackets_the_crossing_whatever_its_slope_3509`: V = 8κ² − 5κ⁴ on [−1, 1]. Both bounds are above the level and falling. The CI is the analytic first crossing to within 1e−8, with neither end open.
- `a_ceiling_within_the_wilks_level_certifies_the_whole_chart_3509`: V = 0.5κ² − 0.3κ⁴.
  - With no ceiling it is refused.
  - With ceiling = a²/4b it gives (−1, 1), both ends open, verdict Flat.
  - A profile above its ceiling is refused.

The e2e `constant_curvature_kappa_inference_e2e` (spherical, flat, hyperbolic) is the proof run for this head.


SauersML:
LC-10 status on b63a6c1d2b. Focused-rust-proof run 35499190960 gave 2 passed, 1 failed.

`flat_truth_does_not_reject_flatness` panics:

```
curvature profile is not outward-monotone at chart bound -2.166625518754706: outward score -2.599513e-1 is below tolerance 1.511791e-3
```

What this means: on flat truth, the κ profile rises again toward the hyperbolic chart bound instead of staying monotone outward. Either ρ̂ is interior near that bound and the profile is not unimodal, or the refit at the bound is not using the same rho_domain as the interior points. Removing the railed-ρ̂ refusal is correct under the envelope theorem only if the profile is unimodal on the resolvability interval, and this test shows it is not in this case.

Next step: a diagnostic run that logs `fit.rho` against `rho_domain` and V(κ) along the κ grid for the flat-truth seed. That will tell whether the non-monotonicity comes from the ρ refit or is real profile shape. The fix has to be built from that; loosening the outward-score tolerance is not an option.

The second proof run, 35499194425, also finished as a failure. This PR stays open and unmerged.

---

SauersML:
**Hand-off to MT-1 (odd PR merge train), from BF-78.**

- **Head to merge:** `e8f003bec9089fb3046f7ea004cfc3780c2122c7`. This commit fixes the root cause of #3509; the math is in the PR body.
- **Gating proof run:** [35509297499](https://github.com/SauersML/gam/actions/runs/35509297499), which runs `gam` / integration / `identifiability`, filter `constant_curvature_kappa_inference_e2e`, on the spherical, flat and hyperbolic truths.
  - It is still queued after 1 h, so do not re-dispatch it.
  - It does not run the two new gam-models unit tests, `a_bound_above_the_wilks_level_brackets_the_crossing_whatever_its_slope_3509` and `a_ceiling_within_the_wilks_level_certifies_the_whole_chart_3509`. The PR's own CI covers those.
- **Merge rule:** mark ready and merge (merge method "merge", expectedHeadSha as above) only once that run and the PR's required checks are green.
- **If the run is red:**
  - Every refusal now prints V(κ̂), V(bound) or the ceiling, and the Wilks residual, so the log shows which half failed.
  - A failure of "exceeds its kappa-free ceiling" means the ceiling's premise, that the intercept is the sole null direction, does not hold for this fixture's profile.
  - A failure of "not outward-monotone … inside the likelihood-ratio set" means the flat profile's bound is inside C but the ceiling is more than ½z² above V(κ̂).
  - Neither case should be fixed by widening a tolerance.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, f8215a61483d.

PR #3593, opened 2026-09-20
Mean bands are images of an index interval under a monotone map; delete the support clamps and delta mean bands (#3140 slice C)

### LINKED ISSUE #3140: gam-predict observation/mean bands clamp to response support and floor variances instead of holding support by construction
Found while working the `predict-interval-hygiene` lane of the pyGAM audit. This is outside that lane's scope, so I'm filing it instead of changing it on that branch.

## Which SPEC.md lines apply
SPEC.md does not mention clamps. These are the lines that bear on the code below:
- "Posterior mean must always be the default (never MAP)." The floors below only fire when the reported `posterior_mean_standard_error` is not the posterior sd of μ, i.e. when it is a plug-in delta-method quantity (robustness.md F5).
- "In general, do not paper over solver issues." A clamp or floor that silently repairs an impossible moment hides the numerical defect that produced it.
- "The default choices should give the best model for the user." The symmetric-Gaussian fallback replaces the family's predictive quantiles with a band known to mis-cover (#817, #1193, #1346).

## Where (current `main`)
- **Interval policy**
  - `crates/gam-predict/src/interval_policy.rs:74-89`: `ResponseBounds::clamp_value` / `clamp_in_place`. Both are applied to mean bands through `ResponseFamily::mean_clamp_bounds` (`crates/gam-spec/src/lib.rs:806`) and to observation bands through `response_support_bounds` (`crates/gam-spec/src/lib.rs:846`).
- **Single-block observation band**
  - `crates/gam-predict/src/lib.rs:2317-2322`: the band is clamped to the support.
  - `crates/gam-predict/src/lib.rs:2274`: the Beta variance is floored with `(mu*(1-mu) - v).max(0.0)`.
  - `crates/gam-predict/src/lib.rs:2288-2289`: the Binomial / Royston-Parmar code runs `mu.clamp(0,1)` and then floors `(p*(1-p) - v).max(0.0)`.
- **Two-block (location-scale) observation band**
  - `crates/gam-predict/src/lib.rs:2660` and `:2666`: the Beta expected variance and the total variance are floored with `.max(0.0)`.
  - `crates/gam-predict/src/lib.rs:2674-2684`: when the moment-matched predictive returns `None`, the code falls back to a symmetric Gaussian band `μ ± z·s` and then clamps it to the support.

## Why these hide bugs rather than guard against them
- **The variance floors can only fire on bad inputs.** `E[μ(1−μ)] = m(1−m) − Var(μ)` is ≥ 0 by construction whenever `m` and `Var(μ)` are moments of one law on [0, 1]. The floors therefore fire only when the SE is not the posterior sd of μ, or when `1 − μ` has been lost to rounding.
- **The mean-band clamp does no legitimate work.** Mean bands built by transforming η endpoints through an inverse link cannot leave the support, so the clamp is either a no-op or it masks an error.

## Repro
Binomial with a logit link, fitted in-range and predicted far outside the training range:

```python
import numpy as np, gamfit
x = rng.uniform(-3, 3, 400)
y = (rng.uniform(size=400) < 1/(1+np.exp(-1.5*x))).astype(float)
m = gamfit.fit({"x": x, "y": y}, "y ~ x", family="binomial")
p = m.predict({"x": np.array([-100.0, 100.0])}, interval=0.95)
mu = np.asarray(p["posterior_mean"]); se = np.asarray(p["posterior_mean_standard_error"])
print(mu, se, mu*(1-mu) - se**2)
# mu = [7.5e-27, 1.0]  se = [1.2e-60, 1.4e-60]  m(1-m)-se^2 = [7.5e-27, -1.9e-120]
```

What the output shows:
- **Upper row (x = +100):** μ is exactly `1.0`, so the upper tail is carried as `1 − tiny` rather than as its complement. `m(1−m) − SE²` is negative, and the `.max(0.0)` at `lib.rs:2289` silently fires.
- **Lower row (x = −100):** μ = 7.5e-27 but SE = 1e-60. The SE is the delta-method `|μ'(η̂)|·SE(η)` (robustness.md F5), not the posterior sd of μ.

## Principled fix
- **Moments by quadrature:** compute `E[μ]`, `E[μ²]` and `E[μ(1−μ)]` from the same posterior quadrature over η. For bounded families, carry the complement `1 − μ` directly: `σ(−η)` for logit, and `S` and `1 − S` separately for Royston-Parmar. The variance is then non-negative and the values stay in the support by construction.
- **Delete the floors, the clamps and the symmetric fallback.** A `None` from the moment-matched predictive should be a typed error that names the row.
- **Tests:** add tests that assert support and non-negativity without any clamp in the path.

Related lanes: posterior-sd (F5 mean SE), and `claude/pygam-spec-pyffi-thin` (the output-layer clamps in `gam-pyffi`, `gamfit/_predict_shape.py` and `gam-cli prediction_csv.rs`).

SauersML:
Fixes #3140

Slice C of #3140 (slices A+B, the observation bands, landed in 853ca59503).

## Math

A mean (credible or confidence) band on a bounded response should be the image of an interval on an index under the monotone map from that index to the response. For an index `u` with SE `s`, a monotone map `h` and the reference multiplier `z`, the band is `[min(h(u − z s), h(u + z s)), max(...)]`. Its endpoints lie in `h(dom)`, which is the support, so no clamp is needed. A delta band `h(u) ± z·|h'(u)|·s` leaves the support wherever `h` saturates, and clamping it afterwards hides that the band was wrong.

- **Every mean band goes through one construction**, `ResponseInterval::TransformEta(EtaDomain)` or `ResponseInterval::TransformIndex`.
- **Half-line links** (`EtaDomain::HalfLine`, read from the spec's `eta_feasibility()`): identity/sqrt/log/inverse/inverse-squared on a family whose mean is restricted. The η interval is cut at the feasibility boundary, and the endpoint there is the link's limit (identity/sqrt → 0, log → 1, inverse/inverse-squared → +∞). So a band that crosses the boundary is exact: it is the image of the feasible part, not a clamp. An index interval that misses the feasible set entirely is a typed error. A restricted cell whose link has no recorded limit is also a typed error.
- **Standard + link wiggle**: `u = η_wiggled` has an exact chain-rule SE, and `h = g⁻¹`.
- **Survival location-scale**: `TransformIndex` on `q0 = −η_t·exp(−η_s)`, with `Var(q0) = gᵀ Σ_{ts} g` over the projected two-block covariance.
- **Binomial location-scale**: `TransformEta` on the link argument `wiggle(q0)` with its chain-rule SE. `mean_se = |dμ/dη|·SE` is reported as the SE and no longer builds a band.
- **Bernoulli marginal slope**: the map is Φ on an unrestricted domain.
- **Dispersion LS, Gaussian LS, transformation-normal**: these carry their `EtaDomain` through the same path.

## Deleted

- `MeanIntervalMethod`, together with `PredictUncertaintyOptions::mean_interval_method`. `TransformEta` was its only live arm.
- `MeanBoundMethod::Delta`, `ResponseInterval::CollapsedDelta` / `SymmetricDelta`, `EtaInterval` and `delta_mean_interval`.
- The binomial-LS delta SE route (`response_se_from_backend`, `mean_se_from_backend`, `predict_with_uncertainty_inner`) and `survival_mean_se_from_backend`.
- Both removal ledgers carry the acknowledgement, at base 5994ad54c4. `source_removal_guard.py` (11 guarded items), `public_api_census.py` and `spec_ban_ratchet.py` all pass locally against the head.
- The observation band `symmetric_predictive_band` still uses `ResponseBounds::clamp_in_place`. It is out of scope here.

## Tests

- The two `delta_*_matches_inline` tests pinned the deleted arms. They are replaced by `transform_index_decreasing_map_matches_inline`, `saturated_logistic_band_is_an_image_inside_the_support`, `half_line_inverse_link_band_crossing_origin_is_unbounded_above`, `half_line_log_link_probability_band_stops_at_one` and `half_line_interval_outside_the_domain_is_a_typed_error`.
- The calibration and coverage tests that ran the Delta arm now run the image band. It has the same truth coverage as the η band, because the map is monotone.

## Status
- [x] link-wiggle full uncertainty on TransformEta
- [x] two-block index band (survival) / link-argument band (binomial LS)
- [x] dispersion LS full uncertainty
- [x] delete the clamps, `MeanBoundMethod::Delta` and `MeanIntervalMethod`
- [x] tests asserting that the bands lie in the support with no clamp in the path
- [ ] focused-rust-proof: gam-predict lib, gam-cli suite, gam regressions/inference/quality/glm


SauersML:
BF-65 (green-run lander): proof run 35500495987 (issue key BF-9-3140-2) on head 6277a5ba failed in the build step, before any test ran. The ban-scanner in `build.rs` aborted with 5 violations:

```
error: crates/gam-solve/src/arrow_schur/slq_logdet.rs:1158: |_direction| Ok((1.0, 1.0)),
error: crates/gam-predict/src/standard.rs:261: fn response_jacobian_rows(&self, _: PredictPass) -> Result<ResponseInterval, EstimationError> {
error: crates/gam-predict/src/dispersion_location_scale.rs:296: fn response_jacobian_rows(&self, _: PredictPass) -> Result<ResponseInterval, EstimationError> {
error: crates/gam-predict/src/survival.rs:392: fn response_jacobian_rows(&self, _: PredictPass) -> Result<ResponseInterval, EstimationError> {
error: crates/gam-predict/src/binomial_location_scale.rs:505: fn response_jacobian_rows(&self, _: PredictPass) -> Result<ResponseInterval, EstimationError> {
```

- **slq_logdet.rs:1158** is stale-base noise. Main fixed it in 0c4f8f4342, and the branch is 376 commits behind main, so merging origin/main clears it.
- **The four `_: PredictPass` hits come from this PR.** The rule is "underscore fn parameter, `_name: T` or bare `_: T` (use the value, restructure the API, or delete the param)". These impls ignore the pass, so the trait needs restructuring. For example, the pass could be passed only to the families that branch on it, or those impls could use it. This is an API change across the PredictionTransform impls, not a one-line fix, so I have left it for the owner.

The other queued runs on this head (35500495421, 35500497031, 35500498082, 35500499077, 35500499684) will fail the same way if they build the root `gam` crate.

---

SauersML:
Taking over from BF-9 (no activity since 2026-09-20T08:46Z).

BF-9-3140-2 failed in build.rs at the ban scanner, so no test ran. It hit five violations:

- **`slq_logdet.rs:1158 |_direction|`:** from the old base, and fixed on main.
- **Four `_: PredictPass` params on `response_jacobian_rows`:** these were added here, in standard, survival, binomial LS and dispersion LS.

After slice C, no implementation's interval policy depends on the pass. Every mean band is the image of an index interval, so the pass argument carried nothing.

Pushed 60ca7b054d, a merge of main plus the fix:
- `response_jacobian_rows(&self)` drops the pass from the trait, all 8 implementations and both drivers. `PredictPass` still selects the numerics in `linear_state`.
- CHANGELOG and both removal ledgers keep both sides, and both ledgers pass JSON validation.
- main's new SAS-state test in `gam-predict/src/lib.rs` loses its deleted `mean_interval_method` field.

Re-validating as BF-27-3140-1 (gam-predict lib).

---

SauersML:
A separate version of #3140 slice C was built in lane w2-outer-pr2988, the lane that landed #3140 slices A+B as 853ca59503. It uses the same design and touches the same 37 files. It will not land: this PR is the one to merge. The evidence and review notes follow.

**Gate for the separate version** (rebuilt on origin/main a935b7eff5, lane dy3, tag w2o-3140c-t5):
- `cargo check --workspace --all-targets` rc 0. The root build.rs ban scanner reports 0 errors, and gam-pyffi builds with rc 0.
- gam-predict slice-C tests passed 31/31. They cover the conformal tests, the interval_policy parity tests (feasible-set cut, NaN endpoint, boundary image, decreasing map, out-of-set typed error), standard, predict_2_2 and the four #3140 band tests.
- The source-removal guard went from 774 to 770 items with 26 guarded removals, rc 0. The public-API census and the assertion-less `#[test]` scan both returned rc 0.
- The full gam-predict run at a5c3a5b6e7 passed 145/148. All three misses predate this change:
  - `gaussian_location_scale_wiggle_replay_3001` ("Laplace precision M is indefinite") and `large_scale_reml_stress_coverage` fail with the same messages at the same tip without the patch.
  - `large_scale_reml_stress_main` (the #2735 slow test) timed out at 600 s.
- Mutants: 4 of 4 fired.
  1. With no cut at the feasible boundary, `transform_eta_is_intersected_with_the_feasible_set` fails ("the cut endpoint maps to the support's end").
  2. With a NaN endpoint passed into the intersection, `transform_eta_nan_endpoint_is_a_typed_error` fails.
  3. With a finite-only image check at the boundary, `transform_eta_boundary_image_may_be_infinite` fails.
  4. With the survival band on the threshold predictor alone, `survival_eta_is_the_standardized_argument_with_its_two_block_se` fails (SE(q0) = 3).

**Ban-scanner failure on proof run 35500495987 (head 6277a5ba), 5 rows:**
- `crates/gam-solve/src/arrow_schur/slq_logdet.rs:1158` `|_direction| Ok((1.0, 1.0))`. This comes from the stale base.
- Four `_: PredictPass` parameters on `response_jacobian_rows`, at `standard.rs:261`, `dispersion_location_scale.rs:296`, `survival.rs:392` and `binomial_location_scale.rs:505`.

At head 60ca7b054d, `git grep` finds none of the five. As a positive control, the same grep finds the four `_: PredictPass` rows at 6277a5ba70. All 9 `response_jacobian_rows` signatures now take only `&self`.

**Two points where the separate version differs.** Both will land as a follow-up on top of this PR once it merges:
1. **The symmetric observation band still clamps.** `symmetric_predictive_band` still calls `ResponseBounds::clamp_in_place` with `transform.bounds()`, from both drivers. Only Gaussian location-scale reaches that arm: dispersion location-scale always returns its skew band or a typed error. Gaussian location-scale's bounds are unbounded, so the clamp does nothing. That leaves the clamp, `ResponseBounds::closed` / `UNBOUNDED` / `UNIT_PROBABILITY` / `clamp_in_place` and the 8 `PredictionTransform::bounds()` implementations with no effect. The follow-up deletes them, with ledger entries.
2. **`EtaDomain::of_spec` keeps a hand-written table of boundary limits** (0, 1 or +inf by link). That is a second route to the map's limit at eta = 0, next to the link itself. The separate version instead intersects `[lo, hi]` with `EtaFeasibility::interval()`, evaluates the response map at the cut endpoint, and admits a non-finite image only on the boundary.
   - The table also mishandles an endpoint at a half-line's infinite end. It takes that end for the finite boundary.
   - Example, binomial-log (feasible set eta < 0), eta interval `[-inf, -0.5]`: `lo <= a` marks the lower side as cut and substitutes `boundary_mean = 1`. The band comes out `[exp(-0.5), 1] = [0.607, 1]`, while the image is `[0, 0.607]`.
   - A finite SE cannot reach this. It is still a literal table for a value the map already provides.

**A defect in the separate version, where this PR is right.** It moved the survival predictor's public `eta` / `eta_se` to the standardized argument q0 = -eta_t·exp(-eta_ls) and its two-block SE. That changes what a public output column means. This PR keeps `eta` / `eta_se` as the threshold predictor and carries q0 in `LinearState::response_index`, and that is the correct choice. The follow-up leaves `eta` / `eta_se` exactly as this PR has them.

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 5402774aaa2b.

PR #3595, opened 2026-09-20
Outer search: publish the per-atom SAE rank-charge branch as a component-wise criterion rank (#3436)

### LINKED ISSUE #3436: gam-sae outer: the rank-charge branch is invisible to the outer optimizer (#2933 F32 R1)
Split out of #2933 (finding F32, remainder R1). The calibration-contract part of F32 has been relabelled. This issue is the nonsmoothness remainder.

## Defect
The realized-rank charge thresholds an eigenvalue against the MP edge. When an eigenvalue crosses the edge, the integer rank changes and the criterion moves to another branch.

- The rank-charge derivative (`construction_rank_charge_derivative.rs`) holds the rank fixed, so it is a within-stratum derivative.
- `OuterEval` (`crates/gam-problem/src/solver_contract.rs:97`, at 683413ef) carries no branch identity.

So a line search or a stationarity certificate can compare values or gradients from different strata as if they came from one smooth function. A certificate at a point next to a rank boundary can certify a within-stratum stationary point that is not stationary for the criterion.

The S1 certified-state store, which the #2933 thread proposed for carrying this identity, has not landed.

## What would close it
- Add a branch or certified-state tag (the realized rank per atom) to `OuterEval`. The outer engine in gam-solve / opt then treats a tag change inside a line search or certificate as a stratum boundary: refuse the comparison, or re-certify on the new branch. General outer-optimizer handling belongs in the opt crate per SPEC.
- A test with an eigenvalue placed near the edge: a certificate must not straddle a rank change.

Refs #2933.

SauersML:
Fixes #3436

## Problem

The SAE criterion's rank charge (#2933 F32) is

C(ρ) = Σ_k ½ · r_k · edf_k · log max(N_eff,k, 1),

where r_k is atom k's production chargeable rank: the number of reconstruction energies μ_j that exceed the Marchenko–Pastur reconstruction edge e(R) = R · (1 + √(p / N_eff,k))². (#2258 promotes an alive atom at MP rank 0 to rank 1.)

So the criterion is not one smooth function of ρ. It is a family V_r(ρ), with one smooth member for each rank assignment r = (r_1, …, r_K). Where any μ_j crosses its atom's edge, the value jumps from V_r to V_r' by ½ · edf_k · log N_eff,k.

The outer search never saw this branch. `SaeManifoldOuterObjective` left `criterion_rank()` at its default `None`, so the bridge's #2765/#2939 stratum logic never ran for SAE. That logic refuses a trial off the run's stratum, keeps the lowest refused trial, and restarts the run there. Without it, an Armijo test could compare V_r at the base against V_r' at the trial, and the difference would include a jump the gradient knows nothing about. A certificate could also accept that jump as a descent.

## Fix

1. **`CriterionRank` carries every rank decision and is compared component-wise** (gam-solve, first commit).
   - The old tag was one `usize`. One integer cannot tag a criterion that sums independent per-atom decisions: if atom 1 gains a kept direction while atom 2 loses one, Σ r_k is unchanged, but V_r has changed.
   - Two assignments are the same stratum only when every component matches.
   - The first-order bridge, the cost-stall rank-boundary evidence and the seed loop's crossing restart now compare `CriterionRank` values.
   - Single-log-det criteria (custom family) use `CriterionRank::single`.

2. **The SAE objective publishes the branch its value was priced on** (gam-sae, second commit).
   - `rank_dof_from_grams` returns `RankChargePricing { dof, chargeable_rank }`. Both fields come from the same per-atom `RankChargeStratum`, so the DOF and the branch it was priced on cannot disagree.
   - The two production pricing seams set `SaeManifoldTerm::priced_rank_stratum` in the same step that forms the value: the dense `penalized_quasi_laplace_criterion_with_geometry` and the streaming `…_streaming_exact_gate_frozen`.
   - The objective records `last_priced_stratum`:
     - `eval_cost` and `eval_with_order(Value)` read it from the exact-ρ probe handoff (bitwise ρ match), which every finite value-lane result parks.
     - `eval` reads it from the installed state it differentiates.
   - `last_priced_stratum` is cleared at the start of every evaluation, on a non-finite result, on `reset`, and on a reactive scalar waypoint install, which is a new objective. It is saved and restored with the reactive domain checkpoint.
   - `criterion_rank()` returns it. Every rank crossing the SAE criterion makes is therefore a stratum boundary for the existing refusal and restart machinery.
   - The EFS lane needs nothing new. Its fixed-point certificate is a single-point `eval`, which cannot straddle two strata, and it publishes the same rank.

No tolerance, bound or search is added. The stratum test is exact integer equality of the per-atom ranks the criterion itself priced on.

## Tests

- `an_energy_at_its_mp_edge_splits_the_rank_charge_into_two_strata_3436`:
  - Takes a resolved single-atom state and computes R* = μ_min / (1 + √(p/N_eff))², the dispersion that puts the smallest energy exactly on the MP edge.
  - Pricing at R*·(1 ∓ 1e-9) gives ranks [2] and [1].
  - The DOF drops by exactly one `edf`, to 1e-12 relative.
  - The two published `CriterionRank`s differ, and a second price on the same side (±1e-6) names the same stratum.
- `the_outer_objective_publishes_the_branch_its_value_was_priced_on_3436`:
  - `criterion_rank()` is `None` before any evaluation.
  - `eval_cost`, `eval` (with a bit-identical cost) and `eval_with_order(Value)` each publish the per-atom ranks that the dense route priced at that ρ.
  - `reset` clears it.
- The existing F32 audit test now also asserts that the priced `chargeable_rank` equals the audit stratum's `production_chargeable_rank`.
- gam-solve `stratum_boundary_2939` (6 tests) covers refusal and restart with the component-wise tag, including a crossing that keeps Σ r_k fixed. Focused proof run 35499280492: 6 passed.


SauersML:

New head is d97aa30014. It merges main 74c2b1d788.
- The only textual conflict was the `gam_solve::rho_optimizer` import in `gam-sae/src/manifold/outer_objective.rs`. This PR adds `CriterionRank` and main adds `OuterCriterionCertificate`; both are kept.
- The auto-merged files still line up:
  - `SaeManifoldTerm` construction and `Clone` carry `priced_rank_stratum`.
  - Both `let Self { .., } = self` destructurings use `..`.
  - The `criterion_rank_fn: None` initializers main added in `run.rs` fit the `Option<CriterionRank>` closure field.

Focused tests: `cargo test -p gam-solve --lib stratum_boundary_2939`, `cargo test -p gam-sae --lib 3436`, plus `cargo check --workspace --tests`.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, e06f3d610a7c.

PR #3597, opened 2026-09-20
tests: t10 K=1 marginal-slope covariance must succeed and equal the population variance

### LINKED ISSUE #3592: tests: t10_k_eq_1_always_diagonal_or_err accepts Err on inputs where the covariance is always well-defined
## Where (origin/main)

`tests/survival/survival/survival_multi_z_covariance_autoderiv_hard.rs`, `t10_k_eq_1_always_diagonal_or_err` (around lines 494–545):

```rust
        let result = marginal_slope_covariance_from_scores(col.view(), &w);
        match result {
            Ok(covariance) => { assert_eq!(covariance.shape(), MarginalSlopeCovarianceShape::Diagonal, ...); assert_eq!(covariance.dim(), 1); }
            Err(_) => {
                // Err is also acceptable (e.g. degenerate zero variance).
            }
        }
```

## Why it is wrong

Every one of the 20 inputs has finite scores and unit weights:
- IID normal
- all zeros
- a single non-zero row
- a non-zero constant
- 5× IID normal

For such inputs `marginal_slope_covariance_from_scores` (`crates/gam-models/src/bms/gradient_paths.rs:1366`) computes a weighted sum of squares, which is finite and ≥ 0. With K=1 it always takes the `is_diagonal` branch, and `MarginalSlopeCovariance::diagonal` (`:951`) accepts any finite non-negative entry, **including 0**. So `Err` is never the correct answer for any case in this test.

The comment's premise ("Err ... e.g. degenerate zero variance") is also false: a zero variance is accepted. As written, a regression that makes the function return `Err` for every K=1 input (for example a zero-variance or rank check added in the wrong place) leaves the test green. The test also never checks the *value* of the variance, so it would pass with any `Diagonal` entry.

## Principled fix

- Require `Ok` for every case.
- Check the returned 1×1 covariance (`to_dense()[[0,0]]`) against an independently computed population variance of the column. The tolerance should be derived from the summation error, O(n·ε) times the second moment. That covers the zero-variance cases too, where the exact answer is 0 or a rounding-level value.
- Rename the test to drop `_or_err`.

SauersML:
Fixes #3592

`t10_k_eq_1_always_diagonal_or_err` accepted `Err` for all 20 inputs and never checked the returned value.

Every input has finite scores and unit weights. So `marginal_slope_covariance_from_scores` produces a finite sum of squares ≥ 0, and `MarginalSlopeCovariance::diagonal` accepts it, zero included. `Err` is therefore never correct here.

Changes, test only:
- Every t10 case must return `Ok` with shape `Diagonal` and `dim() == 1`.
- The 1×1 entry must equal an independent two-pass population variance within `4·n·ε·(Σs²/n)`, the rounding bound of the two summations. This covers the all-zero and constant-column cases, where the exact answer is 0 or rounding-level.
- The test is renamed to `t10_k_eq_1_always_diagonal`.
- t06a (N < K) and t06b (N == K) had the same pattern: their assertions sat inside `if let Ok(..)`. Both now require `Ok(Full)` of dimension K. The rank-deficient PSD sample covariance is accepted by `MarginalSlopeCovariance::full` by design.

Validation: focused-rust-proof run 35499530277 on `gam` / integration / `survival`, with filters t06a, t06b and t10. Each filter reported `1 passed`.


SauersML:

---

Issue #3192, opened 2026-09-19
Outer certificate: choosing η in τ_stat = η²/2 (1/n vs 1/edf vs 1/(n−p)), and making band_V > τ_stat a hard typed failure


SauersML:
## Context

The C3 item in `docs/convergence_theory/boundary-probability.md` (§3.6 and §7.3) replaces the unit-dependent relative cost floor `rel·(1 + |V|)` with the statistical resolution τ_stat = η²/2, where η² = 1/n (the observation count).

- **Certification:** certify when ½δ² + band_V ≤ τ_stat.
- **Where it lives:** `OuterProblemSize::statistical_resolution` in `crates/gam-solve/src/rho_optimizer/run.rs`, consumed by `outer_decrement_bands` (`decrement_bands.rs`).

## Open question 1: η

η² = 1/n is justified by first-order accuracy:

- LAML has O(1/n) absolute Laplace error (Tierney & Kadane 1986).
- Wald/edf inference built on λ̂ has O(n^{-1/2}) error.

Gaussian REML, however, carries no Laplace error. For it, η² = 1/edf or 1/(n − p) could be argued instead. The choice moves τ_stat by a constant factor only.

For grouped or weighted data, the report asks for n_eff = Σ prior weights, the Fisher-information count. Two cases pull against each other:

- **Estimated scale (Gaussian).** Σw is not invariant to rescaling the weights, because φ̂ absorbs the scale. Using Σw would reintroduce the unit dependence C3 removes.
- **Binomial with trials as weights.** Σw is the right count.

The first cut therefore uses the declared row count. A decision-theoretic derivation would settle η: pick η so that the p-value error f_T·2τ stays below the Monte Carlo or integration error of C4.

## Open question 2: when band_V > τ_stat

The report's T1 wants a hard typed failure, `NumericallyUnresolvable { band_V, tau_stat }`, whenever band_V > τ_stat.

The problem is how band_V is formed today. It is the a priori Higham bound γ_m·Σ|channels| with m = n + p². That bound grows like n·|V| ∝ n², while τ_stat shrinks like 1/n.

Illustrative Gaussian REML at n = 3·10⁵, p = 100:

| Quantity | Value |
|---|---|
| γ_m | 3.4e-11 |
| fixed-β channel | ≈ 1.5e5 |
| band_V | ≈ 5e-6 |
| τ_stat | 1.7e-6 |

[...]

PR #3599, opened 2026-09-20
Resolve the outer certificate over the Fisher-information count Σw (#3192)

SauersML:
Part of #3192 (item 2 of the 07:18Z list). This PR does not close the issue.

## Math

The certificate's statistical resolution is `τ_stat = η²/2 = 1/(2·n_eff)`. Until now `n_eff` was the declared row count.

**Replication-weight families.** A likelihood whose prior weights enter as `Σ wᵢ ℓᵢ`, with no scale absorbing them, treats a row of weight `c` exactly as `c` rows of weight one:
- the inner mode is the same under both encodings;
- so are `H = XᵀWX + S_λ`, `log|S|₊` and the LAML criterion `V` (#893, `fixed_dispersion_laml_surface_is_replication_invariant`).

The resolution, a function of the criterion, must therefore be identical too. That forces `n_eff = Σ wᵢ`. Under the row count, a binomial with trials as weights certified at a `τ_stat` `mean(w)` times looser than its Bernoulli encoding of the same data.

**Estimated nuisances.** The count still holds when a nuisance is estimated. Each estimator is normalized by `Σw`, so each is itself replication-consistent:
- Gamma shape: `target = Σ w·stat / Σw`;
- Beta precision: `φ = Σw / Σ w r²/V − 1`;
- Tweedie and generic (inverse-Gaussian / non-identity Gaussian) dispersion: `Σw·d / Σw`;
- NB `θ`: argmax of `Σ wℓ`.

**Profiled Gaussian.** Only the profiled Gaussian scale `φ̂ = D_p/(n − M_p)` absorbs a global rescale `w → c·w`: `V` stays fixed up to a constant, the #877 invariance. There the count must be invariant under that rescale, and `Σw` is not, so it stays the rows. This is the split the issue body anticipated.

## Change

- **`OuterProblemSize`:**
  - gains `information_count: Option<f64>`, where `None` means the rows;
  - `statistical_resolution()` reads it; a count that is not finite and positive resolves nothing;
  - the rounding bands keep charging `accumulation_growth(n_rows + p²)`, because the sums run over rows, as AP-11's map noted;
  - `Eq` is dropped, since the new field is an `f64`.
- **`OuterProblem`:**
  - new builder `with_information_count(n_eff)`;
  - `with_problem_size` now sets only its own two fields, so the order of the builder calls does not matter.
- **`estimate/optimizer.rs`:** `outer_information_count(scale, w)` is an exhaustive match on `ResolvedLikelihoodScale`. It returns `w.len()` for `ProfiledGaussian` and `Σw` for everything else. Both the standard arm and the joint-link arm declare it. The joint-link arm covers mixture/SAS/beta-logistic binomial links and Student-t, all of which are `Unit`.
- **Tests:**
  - the existing `OuterProblemSize` test literals gain `information_count: None`;
  - new `criterion_resolution_reads_the_declared_information_count_3192` checks that rows of weight `c` give the same resolution as `c×` rows, that the declared rows are unchanged, that builder order does not matter, and that a degenerate count resolves nothing;
  - new `replication_weights_count_their_sum_and_a_profiled_scale_counts_rows`.

## Still open on #3192
- A running (a posteriori) error bound for band_V. Until it lands, T1 cannot become a hard typed failure without refusing resolvable large-n fits.
- The large-n regression, at n ≥ 3·10⁵.
- The constant factor in η.

Details will be in the issue comment after merge.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 54219057394a.

Issue #3433, opened 2026-09-20
gam-sae: the support-sparse route prices a different criterion than the dense exact-A route (#2933 F27)


SauersML:
Split out of #2933 (finding F27) so the audit can close. Everything below is the finding as it stands in current code.

## Defect
Memory admission chooses between the dense route and the support-sparse route, and each route scores a different statistical criterion. The two routes are not two factorizations of one scalar.

- **Dense route** (`construction_quasi_laplace.rs`, the ranked criterion): penalized loss + ½log|A|, with A the full observed information including the coordinate block; plus the realized-rank charge and the dense normalizer.
- **Support-sparse route** (`support_outer.rs`, module docs at lines 1–25 and the value near :772–851): a profiled-scale REML/LAML form, schematically
  `½[log|S_red| − log|S_ρ|₊ + d_f{1 + log(2π D_pen/d_f)}]`.
  - Its reduced system comes from the support-sparse majorizer (B) path, not from the exact information A.
  - It carries a row normalizer where the dense route carries the coordinate block.

The support_outer.rs module docs already record the partial alignment: F24–F26 and F27 S1–S2 landed. They also say the value "still departs from" the dense criterion. The 2026-09-20 status table on #2933 lists F27 as "open (S3.1–S7 not landed)".

## Why it matters
- `log|H| = log|H_tt| + log|H_ββ − H_βt H_tt⁻¹ H_tβ|`. Dropping or replacing the first term turns coordinate integration into a different conditioning calculation.
- Profiling the scale changes the objective. It is not an algebraic speedup.

So which model gets fit depends on the host's memory. SPEC requires that a fit come from one converged optimization of one criterion.

## What would close it
1. Price the same criterion on both routes (scale convention, prior normalizers, coordinate block, rank charge), with the support route using the exact information where the dense route does.
2. A route-identity test: force dense and support-sparse execution on one small admissible TopK model and compare each named value component and the full hypergradient to a derived tolerance. The route-identity test currently red on main is tracked in #2234.

Refs #2933.

SauersML:
Claimed by local agent DP-1 (claude/fix-3433-support-route-criterion); draft PR incoming.

---
_Generated by [Claude Code](https://claude.ai/code)_

SauersML:
Status from the gam-fb landing session (2026-09-21), what is done and what still blocks this:

Correct the front door's stale account of the support criterion's ARD coordinates

What is missing to close #3433 is a design decision, not code I could write.

[...]

PR #3602, opened 2026-09-20
gam-sae support outer: optimize the Euclidean ARD log precisions as outer coordinates (#3433, F27 S5)

SauersML:
Refs #3433 (#2933 F27 stage S5)

## What was wrong
The dense exact-A route minimizes its quasi-Laplace criterion over the smoothing strengths **and** the ARD log precisions. The support-sparse route minimized only over the smoothing groups and held `ard_precisions` fixed at whatever the caller passed. So on one data set the two lanes' outer problems had different argmins, even for the parts of the criterion they already share (S1: the ARD log partition, S2: the row coordinate block).

## The change
The Euclidean ARD log precisions `u_c = log α_c` are now outer coordinates on the support lane. The outer vector is `[smoothing groups…, ARD coordinates…]`.

- **Layout.** This is the new `SaeSupportArdLayout`, which follows the dense lane's layout (`fit_seed`):
  - Below `SAE_SHARED_ARD_K_THRESHOLD` atoms, there is one coordinate per Euclidean atom axis.
  - At or above it, there is one coordinate per axis index that some atom prices as Euclidean.
  - `atom_coordinate[atom][axis]` is an `Option<usize>`. It is `None` on a held periodic axis (see below).

- **Criterion.** Unchanged:
  `cost = ℓ_pen(θ̂; λ, α) + Σ log Z_ard(α) + ½(log|H_tt| + log|S| − log|λS|₊)`.

- **Gradient.** The prior's value, its gradient in `t` and its PSD majorizer curvature `D` are each degree‑1 homogeneous in `α`. So `∂/∂u_c` of each one is the entry itself, restricted to the axes that `c` prices. Term by term:
  - `∂ℓ_pen/∂u_c = E_c`, the prior energy that `c` prices. This follows from the envelope theorem, because θ̂ is an inner optimum.
  - `∂Σ log Z/∂u_c = P_c`.
  - `∂log|H|/∂u_c = R_c + Schur_c` at fixed state, where:
    - `R_c = Σ_i tr(H_tt⁽ⁱ⁾⁻¹ D_c⁽ⁱ⁾)`.
    - `Schur_c = tr(S⁻¹ ∂S/∂u_c)` with `∂S/∂u_c = H_βt H_tt⁻¹ D_c H_tt⁻¹ H_tβ ⪰ 0`. It is contracted through the same frozen log-det surrogate bundle as the smoothing channel.
  - `∂(∇_θ ℓ_pen)/∂u_c = g_c`. These are the entries' prior gradients, in the coordinate block only. The state response is `dθ̂/du_c = −A⁻¹ g_c`, so through the single profile adjoint `a = A⁺Γ` that the smoothing gradient already uses, `½·log|H|` gains `−½⟨a.t, g_c⟩`.

  Together:
  `∂cost/∂u_c = E_c + P_c + ½(R_c + Schur_c − ⟨a.t, g_c⟩)`.

  `log|λS|₊` does not depend on the precisions. `R_c` and `Schur_c` are both ≥ 0, so the published `frozen_logdet_h = ½(R_c + Schur_c)` keeps its sign contract.

- **Per-probe samples (#2933 F29).** One adjoint per probe now covers every outer coordinate. The ARD column is `½(∂_j log|S|/∂u_c − ⟨a_j.t, g_c⟩)`. `R_c` is deterministic and cancels from the spread.

- **Domain.** The domain uses the smoothing domain's rule (#2812), with no hand-supplied box.
  - On a slot axis, the prior's curvature is `α` times a unit coefficient, and the data's curvature is the slot's Gauss–Newton `‖∂f/∂t_a‖²`. These give the coordinate's generalized eigenvalues `γ`.
  - The domain is `coordinate_domain(resolvability_interval(γ's))`.

- **Seed and report.**
  - `SaeSupportOuterRequest::ard_precisions` is now the seed: each coordinate starts at the geometric mean of the precisions it prices.
  - A non-positive or non-finite precision has no log coordinate, so it is refused.
  - `SaeSupportOuterReport` carries `ard_layout`, `log_ard` and the selected `ard_precisions`.

### Why a periodic axis is held, not optimized
On a circle of period `P`, with `κ = 2π/P`, the von-Mises prior's exact partition per slot is
`log Z = log(2π/κ) − α/κ² + log I₀(α/κ²)`.
The criterion integrates the coordinate block by Laplace, as `½·log(α + d)`, where `d` is the profiled data curvature of the slot axis. That gives a slot's `u`-derivative of

`−α/κ² + (α/κ²)·I₁/I₀(α/κ²) + ½·α/(α+d)`.

- With `d = 0`, this derivative tends to `½` as `α → 0`. The criterion then decreases without bound toward the prior's flat limit, and its minimizer is the domain face, where the row block is singular to working precision.
- `d = 0` happens exactly, not just approximately. With a single selecting row, the decoder's unpenalized constant column absorbs the row's cell for every `λ`. So a per-slot Gauss–Newton data-free detector cannot see it: the GN curvature at entry is nonzero, but the profiled curvature is zero.
- CI job 35501217801 hit this case: the rational-route search railed on the circle's coordinate, and the NK certificate refused.
- A Gaussian axis has no such limit. Its partition's `−½·log α` cancels the Laplace `½·log α`, so a data-free Euclidean axis is exactly flat.

Selecting a periodic precision needs the circle's own integrated block rather than its Laplace image (F27 S3.1). Until that exists, the support lane holds a periodic axis at the caller's precision. Its entries still curve the row blocks at that precision, but they price no coordinate.

This is a documented remaining departure from the dense lane. The dense lane does optimize periodic ARD, inside `periodic_ard_domain_upper` / `flat_domain_lower_bound`.

## Tests
- **`profiled_support_outer_gradient_matches_refitted_value_fd_2634`** checks the analytic gradient against the central difference of the refitted production value, on every outer coordinate. The circle's held precision is off 1. The first run (3 ARD coordinates) matched to 1.5e-10 / 8.5e-10 on the smoothing groups and 2e-10 to 3.7e-9 on the ARD coordinates.
- **`rational_route_gradient_error_is_the_size_its_probe_samples_report_2933`** calibrates the per-probe samples on the ARD coordinates too.
  - A coordinate whose reduced-Schur share is identically zero is checked as exact: its bar and its realized error must both be under the deterministic floor.
  - At least one coordinate must be stochastic.
- **`support_value_prices_the_ard_partition_on_active_slots_2933_f27`** enters at a held circle precision of 1.7. It checks that the value carries it, next to `exp(u)` of the plane's coordinates.
- **`rational_route_search_checks_its_certified_point_on_unseen_probes_2933`** now searches only coordinates that have an interior minimizer.

## What remains for #3433
Not part of this PR:
- F27 S3.1: periodic ARD selection through the circle's integrated block, plus the exact observed information in place of the majorizer.
- F27 S4, S6 and S7: the realised-rank and collapse charges, and per-atom smoothing.


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, ed1a184b7a7a.

PR #3603, opened 2026-09-20
#3258: certify the support fixed point on the phase-profiled objective

### LINKED ISSUE #3258: Support certificate refuses states flat to within √(2Le) along a prior-broken periodic phase orbit (A's third eigenvalue 9.8e-6, 2.6e-6 < μ* ≈ 5e-5); main's max|Δ| pass there was unresolvable — gauge-adapted certificate
## Summary
Since 8d82939761 (#2576) the support lane's inner certificate is Newton–Kantorovich, taken on the slice transverse to the objective's exact symmetries. When a direction is flat to within √(2Le) without an exact symmetry behind it, the certificate refuses, and the outer evaluation at that ρ fails. The refusal is correct: in f64 the minimizer is not resolved along that direction. Main before 8d82939761 "certified" these same states with `max|A⁻¹g| ≤ τP`, a claim f64 cannot back.

This issue is the forward design: certify the outputs that are invariant under that direction, and refuse the direction itself when it is unresolved.

## Where it shows
Probe on the `pair_chart_fit_is_certified_reml_on_a_noisy_ring` cloud: 256 rows, 8 periodic atoms (5 retained), top 2, P = 2, seed 0xC0FFEE00D15EA5E5, one `solve_fixed_point` per (log λ, ARD α). Probe code is MSI-only, `probe5_2576.rs` under `/scratch.global/sauer354/swarm17/gam-2576/`.

| log λ | ARD α | verdict | A's third eigenvalue | μ* | L | e | bound |
|---|---|---|---|---|---|---|---|
| 0 | 1 | certified off 2 exact symmetry directions (radius 9.2e-9) | 5.16e-2 | 8.06e-5 | 8.73e3 | 1.50e-13 | 1.79e-7 |
| 0 | 1e-2 | certified off 2 directions | 1.00e-3 | 2.51e-5 | 4.73e3 | 5.3e-14 | 1.42e-7 |
| −2 | 1e-2 | certified off 2 directions | 1.73e-3 | 1.15e-5 | 4.77e3 | 1.2e-14 | 1.43e-7 |
| **2** | **1e-3** | **refused** (Cholesky of A − μI, pivot 517) | **9.80e-6** | 5.31e-5 | 4.03e3 | 3.50e-13 | 1.42e-7 |
| **4** | **1e-4** | **refused** (pivot 517) | **2.57e-6** | 1.35e-4 | 3.74e3 | 2.42e-12 | 1.41e-7 |

The first two eigenvalues are the exact constant-redistribution symmetry, of order 1e-16 to 1e-15 and one per output channel. The certificate projects them off. At log λ = 4 two atoms are idle, so those directions are exactly zero rows and the certificate drops them instead.

## Why main's pass there is not a certificate
Both refused states have ‖Δ‖ = 0: the state has converged as far as f64 goes, and main's test `max|Δ| ≤ τP` passes. What it claims, every parameter within τP of the minimizer, is not resolved.
- **α = 1e-3:** the gradient's rounding band e over λ₃ puts the minimizer's position along the weak direction within about e/λ₃ = 3.6e-8. That is under the bound. But across that distance the Hessian can change by L·e/λ₃ ≈ 1.4e-4, about 15 λ₃, so not even the sign of the curvature along the direction is resolved. Kantorovich's h = L·e/λ₃² ≈ 15 > ½.
- **α = 1e-4:** e/λ₃ = 9.4e-7 exceeds the bound 1.41e-7 already at linear order.

λ₃ falls with α (5.2e-2, 1.0e-3, 9.8e-6, 2.6e-6). That fits a periodic phase orbit that the data fit and the harmonic penalty leave exactly invariant and that only the coordinate ARD prior breaks. On a well-spread ring, Σ cos κtᵢ is small, so that prior barely breaks it. The eigenvector has not been decomposed yet; that is step 1 below.

## Forward design
1. **Measure first.** At the two refused states, decompose λ₃'s eigenvector onto the periodic phase generator of each atom: δtᵢ = 1 on the atom's rows, with the harmonic pairs rotated on its decoder. Confirm the attribution before designing around it.
2. **Anisotropic η.** For a generator ξ of a symmetry of F_s = F − E, where the breaking term E is the atom's ARD prior, exact invariance gives g·ξ = ∇E·ξ. That value is computable with rounding of about ε·α·n, not e ≈ 1e-13. Replace the ξ̂ component of the computed gradient with it, and bound ‖A⁻¹r‖ blockwise:
   - use e on the complement;
   - use ε·α·n along ξ̂;
   - add the coupling through the Schur complement s = a − bᵀC⁻¹b.
   C's bottom is certified by the slice Cholesky that already runs, and ‖b‖ = ‖(I − P)Aξ̂‖ is computable.
3. **Anisotropic ω.** The crude ω = L/λ_min is too pessimistic along ξ̂. An affine-covariant bound ‖A⁻¹(A(x) − A(y))‖ ≤ ω‖x − y‖ needs D³F's components along ξ̂. For F_s these follow from the Lie derivative of the Hessian along the symmetry, which is a commutator with the generator's linear part.
4. **Outcome.** Where the bounds certify, report the minimizer within t*. Where the phase itself is unresolved, which the numbers above say happens on very uniform rings, certify only the outputs invariant under the phase (decoded function, value, and the gauge-invariant part of the Hessian's spectrum), and state that the phase is unresolved. Never claim it.

## Pins to add with the fix
- A refused state from the table certifies its invariant outputs and reports the phase as unresolved.
- A mutant that drops the anisotropic orbit band refuses again.
- A state with α large enough that λ₃ ≥ μ* certifies exactly as now.

Related: #2576 (the certificate, landed in 82fdcdf1f0, 8d82939761 and de32bc5c71; this is not a widening of it).

SauersML:
Fixes #3258

## Defect

On the noisy-ring cloud with a weak periodic ARD prior, the support fixed point's Newton–Kantorovich certificate (#2576) refused at every state. It reported "A − μ·I ≻ 0 not certified". The weak direction is the periodic phase orbit of each 1-D periodic atom. Everything else in the objective leaves that orbit flat, and only the prior pins it, with curvature αρ. That is below the shift μ the certificate needs, so on the slice transverse to the exact symmetries, A − μ·I ≻ 0 is false. No iterate can fix that.

## Math

**The phase orbit.** Take a 1-D periodic atom a with harmonic basis [1, sin ω₁t, cos ω₁t, …], ω_h = hκ, κ = 2π/P, whose period P is shared bitwise by the chart, the ARD prior and the basis. Its roughness Gram commutes exactly with the harmonic rotations: it is diagonal, with bitwise equal entries on each sine–cosine pair. The phase action g_δ does two things:
- it shifts every routed coordinate, t_i → t_i + δ;
- it counter-rotates each decoder pair (B_s, B_c) by ω_h δ.

Under g_δ the data fit and the roughness are exactly invariant. Only the ARD energy E = (α/κ²) Σ_i (1 − cos κt_i) moves. The generator is ξ_a = (1 on the atom's slots; ω_h B_c on the sine rows, −ω_h B_s on the cosine rows).

**The profiled objective.** Write ρe^{iφ} = Σ_i e^{iκt_i}. Then E(g_δ y) = (α/κ²)(n − ρ cos(κδ + φ)), so

Φ(y) = min_δ F(g_δ y) = F(y) − E(y) + Ẽ(y),  Ẽ = (α/κ²)(n − ρ),  attained at δ* = −φ/κ.

With c_i = cos(κt_i − φ) and s_i = sin(κt_i − φ), we have ∂ρ/∂t_i = −κ s_i and ∂φ/∂t_i = κ c_i/ρ. So:
- ∇Φ = g + (α/κ)(s − sin κt)
- ∇²Φ = A + α diag(c − cos κt) − (α/ρ) c cᵀ

Both corrections live on the atom's slot block. Φ is invariant along ξ_a, so ξ_a joins the exact symmetry generators, and the slice is taken transverse to all of them.

**Which orbits are profiled (a derived decision, not a retry).** For each candidate orbit, v = (I − P)ξ is tested. If vᵀ(I − P)A(I − P)v ≤ μ‖v‖², then A − μ·I ≻ 0 on the slice is provably false. Only those orbits are profiled. A strong prior therefore certifies directly, exactly as before.

Profiling raises the shift that is needed: L_Φ ≥ L, and the error grows. So an orbit that passed the direct test can have a Rayleigh quotient at or below μ_Φ + ε_A on the profiled slice (off N ∪ Ξ_S). The slot blocks of different atoms are disjoint, so that atom's block of A_Φ is A's, and the orbit then makes A_Φ − μ_Φ·I ≻ 0 false. The profiled set is therefore closed under this test:
- it is widened with exactly those orbits;
- the slice is re-formed;
- the test is repeated.

The set only grows, so the process stops within the number of candidates. Every widening is read off the Rayleigh quotients before anything is factored, so it is a derived decision and not a retry after a failed solve.

**Kantorovich hypotheses for Φ on B(θ, bound).**
- **L_Φ = (L + max_a α_a κ_a (2 + 3√n_a/ρ⁻_a + 3n_a^{3/2}/ρ⁻_a²))(1 + γ₈).** This uses:
  - α·diag(cos κt) varies by at most ακ;
  - |dc_i| ≤ κ(|dt_i| + |dφ|);
  - |dφ| ≤ κ√n‖dy‖/ρ⁻ and |dρ| ≤ κ√n‖dy‖;
  - ‖ccᵀ‖ ≤ n.

  Here ρ⁻ = ρ − band − κ√n·bound > 0 is required, so Φ is C² on the ball. The slot blocks of different atoms are disjoint, which is why the maximum over atoms is taken.
- **e = ‖∇Φ − ∇²Φ Δ‖ + γ_{n+1}‖|A_Φ||Δ| + |∇Φ|‖ + ε_g + ε_s.** Δ is the same exact Newton displacement, sliced. The band ε_s covers the rounding of s.
- **Hessian band ε_A = γ₃(max|A_ii| + αn/ρ⁻) + α[(band/ρ⁻ + γ₈)(1 + 3n/ρ⁻) + n·band/ρ⁻²].** The certificate factors A_Φ − (μ + ε_A)·I, with orbit stiffness on the span of N ∪ Ξ.

**From Φ back to F.** Kantorovich gives an orbit of local minima y* of Φ within t* of θ. Then x* = g_{δ*(y*)} y* is a local minimum of F, because F(g_δ y) ≥ Φ(y) ≥ Φ(y*) = F(x*). Every phase-invariant output (decodes, fitted values, data fit, smoothing energy) agrees at x* and y*.

For the parameters themselves, three bounds hold:
- ‖y − g_δ y‖ ≤ |δ|·X_a, where X_a = ‖ξ_a(θ)‖(1 + γ) + Hκ·bound bounds ‖ξ_a‖ on the ball;
- |δ*_a(y*)| ≤ (|φ_a| + band_a/ρ⁻_a)/κ_a + √n_a t*/ρ⁻_a;
- ‖θ − x*‖ ≤ t* + D + t*G, with D = ‖((|φ_a| + band_a/ρ⁻_a)X_a/κ_a)_a‖ and G = ‖(√n_a X_a/ρ⁻_a)_a‖.

The verdict is one of three:
- **Certified { profiled_phases > 0 }**: A_Φ − μ·I ≻ 0 holds at the shift aimed at (bound − D)/(1 + G), so t* + D + t*G ≤ bound.
- **PhaseUnresolved { phase_atoms, phase_bound }**: only the aim at bound holds. The orbit is certified and the phases are named instead of claimed. solve_fixed_point returns with `phase_unresolved_atoms`, and the FFI JSON carries it.
- **NotCertified**: otherwise.

**profile_periodic_phase_origins.** It now keeps a commuting Gram (and its κ-derivative) bitwise, instead of rounding T S Tᵀ, so the exact commutation that the orbit is read from survives the re-origin.

## Data

Pending: focused run 35506105929 (pins per prior strength, the probe and the #2576 suite).


SauersML:
Review from gam-2576, the lane that filed #3258 and landed the #2576 slice certificate. Checked against #3258's forward design at head `claude/fix-3258-gauge-adapted-certificate` (336+/0-).

**The planned math.** Profiling the phase out exactly, then certifying Φ on the slice with ξ_a as an exact generator, is a gauge-adapted certificate. It widens no band and adds no second path, because it reuses the slice machinery on Φ. The x* = g_{δ*(y*)} y* displacement bound is also the right shape. Five things have to hold before the fix lands:

1. **The prior's form.** Cite the file:line where the periodic coordinate ARD prior is (α/κ²) Σ(1 − cos κt_i). If it is quadratic in t, then min_δ is not (α/κ²)(n − ρ), and Ẽ, ∇Ẽ and ∇²Ẽ all change.
2. **Admission is bitwise, like the other families.** ξ_a may join the slice generators only under exact conditions:
   - the evaluator is exactly harmonic (sin/cos pairs of hκt);
   - the roughness penalty gives each pair's sin and cos entries equal weight and no cross terms;
   - the atom is periodic on that axis.

   See the admission rule in `support_exact_symmetry_generators`. When any condition fails, the certificate must see no ξ_a, and the existing pin `a_symmetry_the_certificate_is_not_given_is_refused_2576` needs a periodic counterpart.
3. **ρ → 0.** φ = arg z is undefined at ρ = 0, and ∇²Ẽ has a −(α/ρ)ccᵀ term. Near-uniform rings need a typed "phase unresolved" answer from a derived ρ resolution, not from a NaN or an L that overflows.
4. **The outcome when the bound fails.** #3258 asks to certify the phase-invariant outputs (the decoded function, the value, and the gauge-invariant part of the spectrum) and to state that the phase is unresolved, never to claim the phase. Say which the PR does. "Reported unresolved" could mean refusing the whole state, which is what main already does.
5. **Pins.** #3258 lists three. First, a refused state from its table certifies its invariant outputs and reports the phase as unresolved. Second, a mutant that drops the orbit term refuses again. Third, a state with λ₃ ≥ μ* certifies exactly as now.

**The current commit.** `noisy_ring_weak_direction_is_the_periodic_phase_3258` asserts nothing. It runs 4 fixed-point solves plus dense eigendecompositions and only prints, so it cannot fail and it costs suite time. Before it merges, turn the attribution into an assertion. For example, at the two refused states, assert that the λ₃ eigenvector lies in span{ξ_a} ⊕ N, with a residual below the eigenvector's own rounding band. Otherwise keep it off main as a probe.

**Coordination.** This PR adds 3 lines to `support_term_tests.rs`. The #2576 P3 candidate replaces that file's tail with new pins, so whichever lands second re-applies its edit on top of the other; neither copies the file whole.

SauersML:
Re-review at d96b907e64 against the five conditions in the earlier review. I derived each item from the diff.

1. **Prior form.** Met. `support_phase_orbits` evaluates the prior's own `κt` with `κ = 2π/P_prior` and requires `period == prior_period == PERIODIC_HARMONIC_BASIS_PERIOD` bitwise.
   - A_Φ = A + α·diag(c − cos κt) − (α/ρ)ccᵀ matches ∇²Ẽ = α c_i δ_ij − (α/ρ) c_i c_j. I checked it from ∂φ/∂t_j = κc_j/ρ.
   - The gradient correction (α/κ)(s − sin κt) matches ∇Ẽ.
   - The generator's decoder part (+ω_h B_c on the sin row, −ω_h B_s on the cos row) cancels the first-order decode change of t ↦ t + δ.
2. **Bitwise admission.** Met: periodic kind, 1-D coordinate, odd basis ≥ 3, `support_penalty_commutes_with_harmonic_rotations`, and the equal-period check. Still missing: a periodic counterpart of `a_symmetry_the_certificate_is_not_given_is_refused_2576`. That would be a ring whose penalty breaks the harmonic commutation, which must not profile.
3. **ρ.** Met. `resultant_lower = ρ − γ_{n+2}·‖(Σ|sin|, Σ|cos|)‖ − κ√n·bound > 0`, which is derived, with no literal.
4. **Outcome.** Met: `PhaseUnresolved { phase_atoms }` plus `report.phase_unresolved_atoms`, instead of refusing the whole state.
5. **Pins.** Met: `phase_pin_3258_{strong_prior,alpha_1e3,alpha_1e4}` and the without-orbits variant that must be `NotCertified`.

Still open: `noisy_ring_weak_direction_is_the_periodic_phase_3258` still has no assertion. Now that the phase pins exist, either fold its eigenvector attribution into them as an assertion, or drop it from the suite.

Merge coordination with the #2576 P3 candidate, which also edits support_term.rs:
- P3 touches the docs near the curvature floor, the saddle classifier, the escape doc and log, and the saddle Err text in `solve_fixed_point`.
- This PR's hunks are the verdict enum, the report struct, after `support_project_off`, the certificate region, and the Certified arm.
- The two sets of hunks are disjoint, so whichever lands second re-applies as a 3-way merge.

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 48b52e2c6db7.

Issue #1376, opened 2026-06-20
Matérn 2-D κ outer-gradient analytic ≠ FD (rel≈0.85) — wrong κ-derivative, full-outer-loop fit fails


SauersML:
## Matérn 2-D κ outer-gradient: analytic ≠ finite-difference (rel ≈ 0.85)

Surfaced as a no-collateral row in dp-1371's #1371 gate (MSI job 11441131, on origin/main 82f434c28 + the #1371 (b) patch). The #1371 patch is a post-fit cost-comparison guard in `optimize_external_design` that cannot touch a gradient analytic, so this is **independent of #1371** — a real Matérn-κ derivative bug on main.

### Failing tests (`tests/basis_smooth/matern_2d_iso_kappa_outer_gradient_fd.rs`)
```
anisotropic eta outer-gradient analytic≠FD on psi_kappa[0]:
   analytic = -5.958174e1   fd = -8.685771e0   gap = 5.090e1   rel = 8.543e-1
matern_2d_iso_kappa_outer_gradient_fd::aniso_matern_theta0_eta_contrast_gradient_is_fd_visible ... FAILED
matern_2d_iso_kappa_outer_gradient_fd::matern_2d_iso_kappa_outer_gradient_matches_fd ........... FAILED
matern_2d_iso_kappa_outer_gradient_fd::matern_2d_smooth_fits_ordinary_surface_full_outer_loop .. FAILED
test result: FAILED. 0 passed; 3 failed (GMATERN_EXIT=101)
```

The analytic outer-gradient w.r.t. the Matérn length-scale (κ) coordinate disagrees with central FD by **85%** — not a tolerance nitpick, a wrong derivative. The full-outer-loop fit test also fails, so this likely affects Matérn/Duchon κ-selection in production, not just the FD check.

### Root-cause question (priority)
Determine **pre-existing vs regression**: the #1271 ρ-prior cap-lift line (global Flat e9bed6b4b → selective 1d1b12ead) was explicitly designed to EXCLUDE κ coords (κ is not a smoothing param). Confirm whether 1d1b12ead (or another recent commit) introduced this κ-gradient mismatch, or it pre-dates the #1271 work. If it's a #1271-line regression, the cap-lift's κ-exclusion is imperfect and must be corrected; if independent, it's a standalone Matérn κ-gradient analytic bug. Owner: bughunt-numerics (REML/penalty-gradient expert; flagged the κ-collateral risk earlier).

SauersML:
[...]

PR #3471, opened 2026-09-20
Never drop the ψ coordinates a [ρ, ψ] θ carries

SauersML:
Fixes #2987.

**The bug**
Three builders returned no coordinates once `n·p` passed a hard-coded `50_000_000` dense-work knob:
- `build_tau_unified_objects_from_bundle` (it returned no τ coordinates and zero pair callbacks);
- the SAS link ext-coordinate builder;
- the mixture link ext-coordinate builder.

The outer plan then got a ρ-only gradient for a θ that still carried the ψ block, and optimized the ψ block blind. SPEC: "Hard-coded knobs and magic constants... should be avoided", and "In general, do not paper over solver issues."

**Changes**
- The three `*_MAX_DENSE_WORK` guards are gone.
- A design too large to densify is refused by name by the memory governor in `try_to_dense_arc`, via `DensificationRefused`, instead of being skipped silently.
- `evaluate_unified_with_psi_ext` refuses, with `LayoutError`, any builder whose coordinate count differs from the number of ψ directions. A length mismatch can no longer reach the outer plan.

**Not in this PR**
Defect 2 from the issue, the exact-Gaussian-boundary pre-check memory at n = 1e6, is filed separately.

**Validation**
focused-rust-proof run 35500747499 is green on head 1bd14263 (a merge of origin/main).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MwMvgB7i335nVFQ8vfZ44B

PR #3607, opened 2026-09-20
reml: never drop psi/tau or link hyper-coordinates behind an n*p magic guard

### LINKED ISSUE #3572: reml: n*p > 50M magic guard silently drops every psi/tau and link hyper-coordinate, so a (rho+psi) theta gets a rho-sized gradient/Hessian (panics on the SAS/mixture link path)
## Defect

Three REML ext-coordinate builders on `origin/main` (d7fe081e52) contain the same hard-coded guard, `n*p > 50_000_000`. When it fires they return **no coordinates**, logged only at `debug`. The outer problem still has `theta_dim = rho_dim + psi_dim`. The evaluator therefore returns a cost, gradient and Hessian for rho alone, as though the psi/link coordinates did not exist.

### Site 1: joint psi/tau Hessian lane

`crates/gam-solve/src/reml/hyper.rs:1318-1342` (`build_tau_unified_objects_from_bundle`):

```rust
let n_x = self.x().nrows();
let p_x = self.x().ncols();
const HYPER_MAX_DENSE_WORK: usize = 50_000_000;
if n_x.saturating_mul(p_x) > HYPER_MAX_DENSE_WORK
    && bundle.backend_kind() != GeometryBackendKind::SparseExactSpd
{
    log::debug!("skipping tau hyper-coordinate construction (n={n_x}, p={p_x}): \
                 dense design materialization too large; falling back to rho-only REML");
    ...
    return Ok((Vec::new(), identity_pair, identity_pair2, None));
}
```

It is called only from `reml/objective.rs:2511` (`evaluate_unified_with_psi_ext`), in the `EvalMode::ValueGradientHessian` branch. The `ValueOnly` and `ValueAndGradient` branches (`objective.rs:2515-2535` and `2622-2632`) call `build_tau_hyper_coords*` with **no guard**. Those builders materialise the dense X anyway, through `try_to_dense_arc`, whenever `!use_implicit` (`hyper.rs:~1506`).

Past the threshold, the lanes disagree at the same theta:
- **VGH lane:** `ext_coords` is empty, so the gradient and Hessian have length `rho_dim`. `force_spectral_logdet = ext_coords.iter().any(|c| !c.is_penalty_like)` (`objective.rs:2550`) is `false`, so the cost uses the exact-LLT `log|H|`.
- **VG / value lanes:** they return a `rho_dim + psi_dim` gradient, and the cost uses the smooth-floored spectral `log|H|` (#1376).

The VGH lane's cost therefore sits on a different surface from its gradient-lane counterpart. The #1376 fix exists specifically to prevent that desync.

Production chain:
1. `gam-models/.../design_construction.rs:3967` `evaluate_joint_reml_outer_eval_at_theta`
2. `ExternalJointHyperEvaluator::evaluate_with_order` (`estimate/joint_hyper.rs:1286`)
3. `RemlState::compute_joint_hyper_eval_with_order` (`reml/hyper.rs:1228`)
4. `evaluate_unified_with_psi_ext`

`compute_joint_hyper_eval_with_order` passes the rho-length `grad` and `hessian` straight back for a theta of length `rho_dim + hyper_dirs.len()`; nothing re-checks the length.

### Sites 2 and 3: SAS / beta-logistic / mixture flexible-link coordinates

`hyper.rs:3042-3052` (`build_sas_link_ext_coords`) and `hyper.rs:3244-3254` (`build_mixture_link_ext_coords`):

```rust
const LINK_EXT_MAX_DENSE_WORK: usize = 50_000_000;
if n_x.saturating_mul(p_x) > LINK_EXT_MAX_DENSE_WORK {
    log::debug!("skipping SAS link ext coordinate construction ...");
    return Ok(Vec::new());
}
```

`evaluate_unified_with_link_ext` (`objective.rs:3089-3093`) then does `if ext_coords.is_empty() { return self.evaluate_unified(rho, &bundle, mode); }`, which returns a rho-only result. The flexible-link outer closure in `estimate/optimizer.rs:2041-2054` goes on to:

```rust
let mut grad = result.gradient_for_mode(eval_mode, theta_dim)?;   // length k, not theta_dim
assert_eq!(grad.len(), theta_dim, "unified evaluator gradient length {} != theta_dim {}", ...);
```

So any SAS, beta-logistic or mixture-link fit with `n*p > 5e7` **panics** on its first gradient evaluation (for example n = 1M rows with p = 51 coefficients). The value closure (`optimizer.rs:2026`) never panics. It silently returns a cost that omits the forced-spectral logdet, and the EFS closure (`compute_efs_steps_with_link_ext`, `objective.rs:3201-3205`) silently returns rho-only steps. The link parameters (epsilon, log delta, mixture logits) are never optimised, and the fit reports the seed link as if it were estimated.

## Why it is wrong

- SPEC: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code." `50_000_000` is arbitrary, it is duplicated three times, and it is not tied to available memory. The rest of the file already makes this decision from the `ResourcePolicy` (`should_use_implicit_operators_with_policy`, `hyper.rs:~1490`).
- SPEC: "In general, do not paper over solver issues." Dropping coordinates the optimiser is actively searching over is not a fallback. It returns derivatives for a different problem.
- SPEC: "A fit object must only ever come from a converged optimization." On the psi path the certificate is computed over rho alone, so psi is never stationary-checked. On the link path, whenever the process survives (value/EFS lanes), the link parameters are returned unoptimised.
- SPEC: "Fitting and inference must be fast on all scales of data, but especially large-scale data where it matters the most (with seamless transitions between strategies)." Above the threshold nothing transitions to another strategy. Coordinates simply vanish, and only on some lanes.

## Principled fix

1. Delete all three guards and the constants. A builder never returns fewer coordinates than the hyper-directions and link parameters it was asked for.
2. Base the dense-versus-implicit decision on the existing `ResourcePolicy` / `should_use_implicit_operators_with_policy`, in one place.
   - The first-order psi objects already have an implicit path, so `build_tau_unified_objects_from_bundle` should build its first-order `ext_coords` exactly as the VG lane does.
   - If the second-order pair callbacks cannot be formed matrix-free at this size, the VGH lane should return `HessianValue::Unavailable`, which lets the outer plan fall back to its gradient-only mode as a derived choice. It must not shrink the coordinate set.
3. Give the SAS and mixture link builders the same treatment: use the implicit X operator for `X^T diag(.) X`-type drifts (or row-chunk the `B` construction) instead of refusing. If some object truly cannot be formed, return a typed `EstimationError` rather than `Ok(Vec::new())`.
4. Add a debug-assert or typed error in `compute_joint_hyper_eval_with_order` and `evaluate_unified_with_link_ext`: if `gradient.len() != theta.len()`, it is an evaluator defect and must not reach the optimiser.
5. Regression tests:
   - evaluate the same small problem with the dense/implicit threshold forced through the policy (not a constant), and assert that the VGH, VG and value lanes return identical cost and a gradient of length `theta.len()`;
   - add a link-ext case that would previously have hit the `assert_eq!` panic.

The guards arrived in the consolidation commit f076f9d85b; no issue or PR mentions them.

SauersML:
Fixes #3572

Three builders returned **no coordinates** once `n*p > 50_000_000`, logging only at `debug`:
- `build_tau_unified_objects_from_bundle`, on the joint ψ Hessian lane only;
- `build_sas_link_ext_coords`;
- `build_mixture_link_ext_coords`.

Past that threshold, a θ of length ρ + ψ received a gradient and Hessian sized for ρ alone:
- The joint ψ Hessian lane also stopped forcing the spectral `log|H|` that its gradient lane uses (#1376), so the two lanes evaluated different surfaces.
- On the SAS, beta-logistic and mixture flexible-link path, the ρ-only gradient tripped `assert_eq!(grad.len(), theta_dim)` in `estimate/optimizer.rs`. That is a panic. The value and EFS lanes, meanwhile, silently froze the link parameters.

Changes:
- Delete the three guards and their constants.
  - The Hessian lane now builds the same coordinates as the gradient lane, through the same builders, which already choose dense versus implicit designs themselves.
  - The link builders now behave like `build_student_t_ext_coords` and `link_ext_effective_design`, which never had the guard.
- Delete `HyperCoordPair::zero` (gam-problem). Its only callers were the removed identity pairs.

No coordinate set is ever shrunk now; no lane evaluates a different problem than the one the optimiser is searching.

---


SauersML:
Taking over from HN-12 (no activity since 08:20Z).

The source-removal guard was the only blocker. The branch removes `HYPER_MAX_DENSE_WORK`, both `LINK_EXT_MAX_DENSE_WORK` declarations and the public `HyperCoordPair::zero`, and none of them had a ledger entry. What I pushed:
- 35135048df: adds the acknowledgement, with grep evidence that the only other occurrences were inside the removed guards.
- adf7ae41dd and 8e50780cfe: merge `origin/main` (d5f641efbd) and key the entry to that tip. Main's ledger text is kept byte-for-byte.

I ran the source-removal guard locally with `--base d5f641efbd --head 8e50780cfe` and it passes (155 -> 152 Rust items, 3 guarded removals). I'm re-dispatching the focused proof on the merged head now.

---

SauersML:

New head: bf485cc54a (merge of 8e50780cfe with main 74c2b1d788).
- `docs/source-removal-changes.json`: kept main's ledger and appended this branch's #3572 entry. I moved its `base` to 74c2b1d788, because the guards match entries by base.
- `docs/public-api-census-changes.json`: added the missing acknowledgement for `crates/gam-problem/src/lib.rs::zero` (`HyperCoordPair::zero`). Before this, the census reported it unacknowledged.
- Ran `scripts/source_removal_guard.py`, `scripts/public_api_census.py` and `scripts/test_census.py` with `--base 74c2b1d788` against the merged head. All three pass (3 guarded removals acknowledged, 1 public fn acknowledged).

---

SauersML:
Ready for the local merge train (MT-1, odd PR). The only thing it waits on is the build.

- **Validation:** focused-rust-proof run 35505601331 on head 8e50780cfecd070fe68ee322afc32bb92d9829b1. It is still queued behind the Actions backlog.
- **After a green run:** merge (merge_method "merge", expectedHeadSha 8e50780c…), then close #3572 as completed and note it on #3471.
- **If main moves and conflicts:** merge main in. Keep the source-removal ledger's existing text byte-for-byte.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 608633530d87.

Issue #3545, opened 2026-09-20
crates/gam-terms/src/basis: pure-Duchon fractional-s closed-form operator penalty evaluates its self-pair at r = 1e-6·median-lag instead of the exact zero limit


SauersML:
## Defect

`closed_form_anisotropic_pair_block_pure` (`crates/gam-terms/src/basis/matern_kernel.rs:2572`, origin/main) builds the pure-Duchon (κ = 0) closed-form operator pair block. For each **self-pair** (R = 0) it uses the exact finite-part limit only when `s` is integer:

```rust
let s_int = if s.fract() == 0.0 && s >= 0.0 { Some(s as usize) } else { None };
let pure_diag_exact = s_int
    .and_then(|si| closed_form_penalty::pure_duchon_self_pair_value(q, d, m, si, &eta_centered))
    .is_some();
let r_eps = if pure_diag_exact { 0.0 } else { pure_duchon_diagonal_epsilon(centers, &eta_centered) };
...
// i == j, fractional s:
r_eps_buf[0] = r_eps * eta_slice[0].exp();
j_prefactor * anisotropic_duchon_penalty_radial_with_powers(q, m, s, 0.0, eta_slice, &powers, &r_eps_buf)
```

For fractional `s`, the diagonal is the pair kernel evaluated at a fake lag produced by `pure_duchon_diagonal_epsilon` (`matern_kernel.rs:2669`):

```rust
/// Median off-diagonal anisotropic lag scaled by 1e-6 ...
(median * 1e-6).max(1e-12)   // and a bare 1e-12 when k <= 1 or no lags
```

This constant is arbitrary. The exact diagonal for this term is known and equals **0**:

- `pure_duchon_self_pair_value` (`closed_form_penalty.rs:1422`) derives that the pure-Duchon pair function `g_q(R) ∝ R^{p−2q}` (up to a log factor) with `p = 4(m+s) − d`, and that its R → 0 limit is 0 whenever `p > 2q`.
- That derivation never uses the integrality of `s`. The function takes `s: usize` only because of its signature:
  ```rust
  let p = 4 * (m + s) as isize - d as isize;
  (p > 2 * q as isize).then_some(0.0)
  ```
[...]

PR #3611, opened 2026-09-20
gam-terms: exact zero self-pair for fractional-s pure-Duchon closed-form penalty

SauersML:
Part of #3545 (fix items 1, 2 and 4, the pure κ = 0 path). Item 3 stays open: the hybrid κ > 0 ε callers of `pure_duchon_diagonal_epsilon`.

## Problem
`closed_form_anisotropic_pair_block_pure` (matern_kernel.rs) evaluated the R = 0 self-pair at `R = max(1e-6 · median lag, 1e-12)` (`pure_duchon_diagonal_epsilon`) for every fractional `s`. The cause was that the exact-limit helper `pure_duchon_self_pair_value` took `s: usize`. That puts a data-dependent magic constant on the penalty diagonal, and the diagonal it produces is the value at a small nonzero lag, not the limit.

## Math
For the scale-free kernel, `f(R) = c·R^p` with `p = 4(m+s) − d`. Every term of `(−Δ_B)^q f` carries `R^{p−2q}`. When `s` is fractional, `p` is not an integer, so there is no log branch and `p ≠ 2q`. The self-pair is therefore exactly `0` when `p > 2q` and diverges otherwise. `p > 2q` is the UV clause of `duchon_closed_form_operator_penalty_converges`, and the only production caller (`operator_penalty_candidates_closed_form_pure`) is gated on it through `closed_form_ok(q)`.

## Change
- `pure_duchon_self_pair_value` now takes `s: f64` and computes `p` in f64. It returns `None` for non-finite or negative `s`.
- The pure pair block computes one exact `self_pair` for all `s` and uses it on the diagonal. The ε path, the `s_int` special case and the per-row `SmallVec` allocations for the self-pair are deleted.
- An ungated call (p ≤ 2q, where the self-pair diverges) now panics with a clear contract message instead of silently returning a regularized value. This matches the existing zero-lag panic in `anisotropic_duchon_penalty_radial_with_powers`.
- Tests:
  - The existing #2469 test now passes `f64` values for `s`.
  - The new test `pure_duchon_self_pair_is_exactly_zero_for_fractional_s_3545` checks the value is 0 for s = 1.5 and `None` for s = 0.25. It also checks that the κ = 0 radial chain shrinks toward that 0 as R → 0⁺.

## Scope, and what is left in #3545
- `pure_duchon_diagonal_epsilon` is still used by the hybrid (κ > 0) pair-block sites that are guarded by `analytic_self_pair_bundle` (`matern_kernel.rs:2527`, `:2803`, `closed_form_operator.rs:117`). This PR does not touch them.
  - Those sites are reached for even `d` with UV convergence but `μ = d/2 + q − 2m ≤ 0`, for example d = 2, m = 2, q = 1.
  - In that regime the Schoenberg integral diverges in the IR, and there is no odd-d radial bundle. The finite self-pair is then defined only modulo the polynomial null space.
  - A closed form for it is a separate piece of work, not the one-line `s: f64` generalization done here.
- Both test-file callers use gated parameters: (d=4, m=1, s=2, q=1) gives p = 8, and (d=3, m=2, s=1) gives p = 9.
- `bench/cargo_benches/closed_form_perf.rs` also calls the pure block. It is not registered as a `[[bench]]` in Cargo.toml, so it is never built, and it already fails to type-check (`s: usize` passed as `f64`). It is unaffected.

## Validation
- focused-rust-proof `HN-15-3545-1` (`-p gam-terms --lib`): the build is green, and both `pure_duchon_self_pair` tests pass. The run is marked red only because the second filter, `closed_form_pure`, matched zero tests, and the workflow refuses a filter that matches nothing.


SauersML:
BF-16 here. I claimed #3545 20 s before HN-15 and wrote an essentially identical fix: `s: f64` in `pure_duchon_self_pair_value` and one exact `self_pair` on the diagonal of the pure block. This PR is further along, so I'm dropping my branch and leaving #3545 to it. I'm not opening a duplicate PR.

One gap you may want to close before merging. The new tests check the helper and the κ=0 radial chain, but nothing checks `closed_form_anisotropic_pair_block_pure` itself at fractional `s`. That block is where the ε diagonal actually lived. Before the fix, at d=3, m=1, q=1, s=0.3, the ε diagonal was `(1e-6·median)^{0.2}`, which is about 6% of a typical off-diagonal entry. A block-level check I had locally:

```rust
// d=3, m=1, q=1, s in {0.3, 0.5}: p = 4(m+s) - d = 2.2 / 3 > 2q = 2
for s in [0.3_f64, 0.5] {
    assert!(duchon_closed_form_operator_penalty_converges(1, 1, s, 3));
    for aniso in [None, Some(&[0.3_f64, -0.1, -0.2][..])] {
        let g = closed_form_anisotropic_pair_block_pure(centers.view(), 1, 1, s, aniso);
        for i in 0..g.nrows() { assert_eq!(g[[i, i]], 0.0); }
        // and off-diagonals nonzero
    }
    // homogeneity: centers scaled by c in {1e-3, 0.25, 40} give g_c == c^{p-2q} * g
    // entrywise to 1e-9 relative. The ε diagonal (tied to the median lag) broke this.
}
```

Also a note on the scope paragraph. `pure_duchon_diagonal_epsilon` has no production reach at all. Under the gate, Schoenberg λ = 2(m+s) − d/2 − q > 0 and μ = d/2 + q − 2m > 0 both hold, and the IR clause rules out s = 0. So `analytic_self_pair_bundle` always succeeds for gated hybrid blocks. The ε path is reached only by ungated unit tests: closed_form_operator tests at d=2, m=2, s=1, q=1, where the IR clause fails, and a jet test at d=4, q=2, m=2. Deleting it is a reasonable follow-up once those tests move to gated parameters.

---

PR #3612, opened 2026-09-20
gam-custom-family: delete the tautological gam#1395 logdet guard

### LINKED ISSUE #3564: gam-custom-family: gam#1395 logdet guard rebuilds the bit-identical matrix (tautology): redundant O(p³) factorization per outer eval at p≤64 plus a production assert! with magic 1e-7 tolerance
## Where (origin/main)

- `crates/gam-custom-family/src/assembly.rs:8-14`: `const JOINT_LOGDET_GUARD_MAX_DIM: usize = 64;`
- `crates/gam-custom-family/src/assembly.rs:~1400-1500`: the "Structural guard against the gam#1395 `0.5·log|H|` collapse" block, gated by `if include_logdet_h && total > 0 && total <= JOINT_LOGDET_GUARD_MAX_DIM`
- The operator it checks is built by the `assemble_operator` closure (`assembly.rs:~1281-1357`)

## What is wrong

**1. The guard checks the matrix against itself.** On a cache miss, `assemble_operator` builds `hessian_op` with this exact sequence:

```
materialize_joint_hessian_source(&h_joint_unpen, total, ..)
add_joint_penalty_to_matrix(&mut j, ranges, &scaled_s_lambdas, 0.0, None)
j += scaled_joint_penalty            (if Some)
j.scaled_add(rho_curvature_scale, robust_jeffreys_hphi_for_operator)   (if Some)
symmetrize_dense_in_place(&mut j)
BlockCoupledOperator::from_joint_hessian_with_mode(&j, pseudo_logdet_mode)
```

The guard runs the same six calls, on the same inputs, in the same order. The code comments say this is intentional ("rebuild the SAME matrix ... price it with the SAME logdet kernel", #2627). So on every cache miss `reference.logdet()` equals `hessian_op.logdet()` bit for bit, and `gap == 0`.

The failure modes the guard says it catches cannot be seen this way:
- "A halved curvature" or "a dropped penalty/Jeffreys term" would sit in the shared inputs (`h_joint_unpen`, `scaled_s_lambdas`, `hphi`). Both the operator and the reference would receive it.
- A defect in `BlockCoupledOperator`'s logdet would also appear on both sides, because the reference uses that same kernel.

The only thing it can detect is a cache hit that returns an operator built from different inputs. That requires either:
- a 64-bit `DefaultHasher` collision, or
- the `JointHessianSource::Operator` case. There the fingerprint (`assembly.rs:848-853`) hashes only `diagonal`, plus ρ, β̂ and the penalties, in a process-global `static` cache that every fit in the process shares.

**2. The cost is paid on every outer evaluation.** When `total ≤ 64` and `include_logdet_h` is set, each outer evaluation does a second full materialization and a second O(p³) factorization. Cache hits pay it too, so the BFGS Value→ValueAndGradient pair that the cache exists to deduplicate still factorizes twice.

**3. It is a magic-constant production `assert!`.**
- `tol = 1e-7 * total * (1 + |logdet|)` is not derived from anything.
- `JOINT_LOGDET_GUARD_MAX_DIM = 64` is an arbitrary cutoff.
- When the guard fires it panics. On the multinomial path that panic is caught by `catch_unwind` in `gam-models/src/multinomial.rs:~1035` and turned into `MultinomialSeparationDetected`. The doc comment in `tests/inner_solver_numerics.rs:7389` relies on exactly that. So a panic is converted into an unrelated typed error.
- If the reference factorization returns `Err`, the guard only calls `log::trace!` and continues, which silently swallows the failure.

SPEC:
- "Unnecessary choices and options should be deleted, not included."

## Principled fix

1. Delete the guard block and `JOINT_LOGDET_GUARD_MAX_DIM`.
2. Update the comment in `assemble_operator` that says "exactly the gam#1395 logdet-collapse the ground-truth guard below detects". The `symmetrize_dense_in_place` it describes stays, because it is the actual #1854 fix.
3. Update the references in `crates/gam-custom-family/src/tests/inner_solver_numerics.rs:7389` and `tests/regressions/misc/owed_1395.rs:382,409,507,530`. Those tests check the logdet value directly, so they keep their coverage.
4. Close the one real channel the guard could catch at its source by making the cache key complete, not by re-checking after the fact. Either option works:
   - Key the Operator-source path on the materialized matrix. It is materialized anyway before `from_joint_hessian_with_mode`, so the build could hash `j_for_traces` after symmetrization and cache on that. That makes a hit bit-identical by construction for both source kinds.
   - Scope the cache to a single fit, owned by the outer driver, instead of a process-global `static`.

Either way the cache hit becomes provably exact and the guard has nothing left to check.

SauersML:
Fixes #3564

## What

`joint_outer_evaluate` in `crates/gam-custom-family/src/assembly.rs` ran a "structural guard" for dimensions up to `JOINT_LOGDET_GUARD_MAX_DIM = 64`. It did this:
1. Rebuilt the same symmetrized joint Hessian `H_unpen + S_λ + scale·H_Φ` that the operator route had just built.
2. Priced it with the same `DenseSpectralOperator::from_symmetric_with_mode` kernel. Under a pseudo-logdet mode, `BlockCoupledOperator::from_joint_hessian_with_mode` is that operator.
3. Called `assert!` to check that the two logdets agreed within a magic `1e-7·total·(1+|ref|)`.

On a cache miss this compares a value with itself. What it actually does:
- Adds a redundant O(p³) eigendecomposition to every outer eval at p ≤ 64.
- Keeps a hard-coded size knob (SPEC: "Hard-coded knobs and magic constants ... should be avoided").
- Keeps a production `assert!` that the multinomial Firth fallback catches with `catch_unwind`.

## Changes

- Removed the guard block and `JOINT_LOGDET_GUARD_MAX_DIM`. The symmetrization it depended on is still in place, and its comment now justifies it on its own.
- `multinomial_firth_joint_hessian_logdet_needs_symmetrization_1854`:
  - The symmetrized assembly must now match the reference logdet bit for bit, since it is the same kernel on the same matrix.
  - The check that the raw asymmetric matrix diverges now uses a derived eigensolver roundoff band (`2·n²·ε·‖H‖₂`) instead of `1e3×` the guard tolerance.
- `tests/regressions/misc/owed_1395.rs`: removed the comments that referred to the guard and the `p > 64` "escape the guard" assertion. The wide p = 512 closed-form coverage stays.

Not in scope: the assembled-operator cache fingerprint hashes only the `diagonal` of an `Operator` source. That source is the matrix-free large-p route, where the guard never ran, so removing the guard changes nothing there.

---


SauersML:

---

SauersML:
Closed as superseded. #3628 (merge 988f269d, commit f003bf84) landed the same fix for #3564 on main at 09:24:
- It deleted the guard block and `JOINT_LOGDET_GUARD_MAX_DIM`.
- It also removed the multinomial Firth `catch_unwind`, which this PR left in place.

Merging main into this branch conflicts in all three files (`assembly.rs`, `inner_solver_numerics.rs`, `owed_1395.rs`), because both PRs edit the same hunks. Nothing in this PR is missing from main, and #3564 is already closed as completed.

---

PR #3613, opened 2026-09-20
Judge flatten on the geometric angle of the atom's image plane and the rows it parses (#3506)

### LINKED ISSUE #3506: manifold/curl: flatten_verdict misses a rank-1 decoder collapse (kappa = 1.5 exactly, on its own magic threshold) because flatten_candidates passes phase angles, not geometric ones
Found by an audit of `crates/gam-sae/src/manifold` (files a-l).

## Where

`crates/gam-sae/src/manifold/curl.rs:439-472` (`flatten_verdict`) on origin/main:

```rust
let alpha: Array1<f64> = (0..n).map(|i| radii[i] * angles[i].cos()).collect();
let beta: Array1<f64> = (0..n).map(|i| radii[i] * angles[i].sin()).collect();
let law = radius_law(alpha.view(), beta.view())?;
let (_r1, resultant2) = circular_resultants(alpha.view(), beta.view());
let diameter = resultant2 > 0.7;
let gaussian_fill = law.kappa > 1.5;
```

It has one production caller, `crates/gam-sae/src/structure_harvest.rs:7926-7976` (`flatten_candidates`, called at `:7377`). That caller passes:
- `radii[i]` = the norm of the atom's centered fitted **image** `Φ·B` at row `r`,
- `angles[i]` = `TAU * coords[[r, 0]]`, the atom's **phase coordinate**, not the geometric angle of the image point.

## Why it is wrong

1. **A real rank-1 collapse is not reliably flagged.** Take a Periodic atom whose decoder has collapsed to a line (for example `B` has only the `cos₁` row: image `= R·cos(2πt)·e0`) while the phases stay spread over the circle. Then `radii = R|cos φ|`, so `κ = E[cos⁴]/E[cos²]² = (3/8)/(1/4) = 1.5` **exactly**, and `R₂` of the uniform phases is about `0 ≤ 2/√n`. So `diameter = false`, and `gaussian_fill = κ̂ > 1.5` is decided by sampling noise, or by rounding when the phases are evenly spaced. This is exactly the "diameter collapse, rank-1" case the doc comment on `flatten_candidates` promises to catch. The one test (`structure_harvest/tests.rs:2744`, `flatten_flags_diameter_and_spares_healthy_ring`) builds the diameter the other way: a round decoder with phases piled on `{0, ½}` turns. That is the only configuration in which the phase `R₂` happens to agree with the image geometry.
2. **The "Gaussian fill" branch cannot fire for the reason it is documented to.** Every image row of a 1-D Periodic atom lies on its fitted closed curve, so the image radius law of a smooth curve under a uniform phase has `κ ∈ [1, 1.5]`: `1` for a round circle, up to `1.5` for a degenerate ellipse (a line). The image is never a plane fill. A "Gaussian-fill relaxation" is a property of the **data rows** the atom parses, not of `Φ·B`. The κ branch as wired therefore tests image eccentricity against an arbitrary cut. It does not test Gaussian fill.
3. **Magic constants, inconsistent with the curl side.** SPEC: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code." `0.7` and `1.5` have no derivation, and `flatten_verdict` ignores the `kappa_se` that `radius_law` already computes. The inverse move, `ring_recognition` (`curl.rs:231-236`), calls the same plane non-degenerate when `R₂ < 0.5` and gates κ at `z_below_gaussian > CURL_Z`. So a plane with `R₂ ∈ [0.5, 0.7]` is neither curlable nor flattenable, and the κ criteria of the two directions of the move pair disagree. The module header promises the pair is "falsifiable inside the dictionary's life".

## Principled fix

- In `flatten_candidates`, build `(α, β)` from the **geometric** coordinates of the active rows in the atom's own image plane: project the centered image rows onto its top-2 principal directions. The existing `power_iter_top_dir` helper already does the top one. Do not pair an image radius with a phase angle. With geometric angles, a line image and a two-point pile-up both give `R₂ = 1`, and a round ring gives `R₂ ≈ 0`.
- Read the Gaussian-fill question on the **data** the atom parses: project the target (or residual plus the atom's own contribution) on the active rows onto that plane. Do not read it on `Φ·B`.
- Make `flatten_verdict` the exact complement of the gates `ring_recognition` uses: the same `CURL_Z` z-test on `κ` with `kappa_se`, and the same diameter screen. That gives one source of truth for "this plane carries a ring" and deletes the `0.7` and `1.5` constants.
- Add a regression test with a line-collapsed decoder under uniform phases (for example `decoder[[2,0]] = R` only). It must be flagged with `residual_rank == 1`.

SauersML:
Fixes #3506

## Defect
`flatten_candidates` paired the image radius `|Φ·B − c|` with the latent **phase** angle `TAU·t`. Two consequences:
- A periodic decoder that traces a segment (only `cos₁` loaded) under uniform phases reads `κ = E[cos⁴]/E[cos²]² = 3/2` exactly, with `R₂ = 0`. That is a covered "ring" sitting exactly on the magic `κ > 1.5` cut, so it was never flattened.
- The image of a 1-D periodic curve always has `κ ∈ [1, 3/2]`, so the documented Gaussian-fill (`κ ≈ 2`) branch could never fire.

On top of that, `0.7` / `1.5` / `< 8` were underived constants, and verdict errors were swallowed with `if let Ok`.

## Fix (math)
**Geometry: the principal image plane, computed exactly in basis space.** Let `Φ̃` be the active basis rows centered by their mean, and `B` the `m×p` decoder, so the centered image is `Ĩ = Φ̃B`. Factor `K = BBᵀ = UΛUᵀ` and set `L = UΛ^{1/2}` and `F = Φ̃L`. Then `FFᵀ = Φ̃KΦ̃ᵀ = ĨĨᵀ`, so `F` has the same left singular vectors and singular values as `Ĩ`, at `O(n·m²)` cost and never `p²`. Take the top two eigenpairs `(μ_k, w_k)` of `S = FᵀF`, which is assembled bit-symmetric and goes through `strict_symmetric_eigh`. The image coordinates are `a_k = F w_k`. The ambient directions are `e_k ∝ Ĩᵀa_k = Bᵀ(Φ̃ᵀa_k)`, with `e₂` Gram–Schmidt'd against `e₁`. The angle is now the geometric angle of the image point, so a segment reads as a diameter no matter how its phases are spread.

**Data: the rows the atom parses.** Residuals are `R = target − fitted`, with `fitted = μ + σ ⊙ Σ_j a_j g_j` (the Tier-0 lift). The leave-this-atom-out target in the fit frame is then `y = R ⊘ σ + a_k·g_k`. It is centered on the active rows and projected on `(e₁, e₂)`.

**Verdict: the exact complement of what `curl` promotes on.** `flatten_verdict(image_α, image_β, parse_α, parse_β)` runs `ring_recognition` on both planes:
- rank 1 if `image.diameter || parse.diameter`;
- otherwise rank 2 if `!parse.recognized`;
- otherwise keep.

`RingRecognition` gains a `diameter` field. It is the same `R₂` screen `covered` already refused, so curl and flatten share one threshold set. A plane can never be both curlable and flattenable, or neither.

The edge cases are explicit audit outcomes, not swallowed errors: a constant image (`μ₁ = 0`) and an empty parse are both flagged. Slots whose #1026 hybrid-split verdict already decodes them by the straight sub-model are not circles in the reconstruction, so they are not audited.

Out of scope, noted: the `0.5` (R₂) and `+0.15` (R₁ floor) screens inside `recognize()` stay as they were. They are now the only screens in the pair.

## Tests
- New in `curl.rs`: segment under uniform phases gives rank 1 (the old `κ = 3/2` case); `flatten_is_the_complement_of_ring_recognition`.
- Rewritten in `curl.rs` for image/parse: Gaussian parse of a ring image gives rank 2; diameter gives rank 1; a healthy ring is kept.
- `structure_harvest`: `flatten_demotes_line_collapsed_decoder_under_uniform_phases_3506` and `flatten_demotes_gaussian_parse_of_ring_decoder_to_rank2_3506` (Gaussian seed precomputed: κ = 1.98, z = 0.20); `flatten_flags_diameter_and_spares_healthy_ring` and the killer demo moved to the new API.

---


SauersML:
Coordination note from #3827, now PR #3903. That PR changes the `recognize` coverage screens in `curl.rs` to calibrated Rayleigh tests of the uniform-angle null, and removes the `+0.15` floor on R₁ and the fixed `R₂ < 0.5` cut. The new cut is `n·R_k² < ln((1+√(1−α))/α)` with α = Φ(−CURL_Z), so the pair refuses a uniform ring with probability exactly α at every n.

What this means here:
- After #3903, `!not_diameter` means "the angle law is resolvably anisotropic at the κ-gate level". It no longer means "the plane is a line".
- At large n, any non-uniform phase occupancy trips it.
- If this PR keeps `diameter = !(resultant2 < 0.5)` inside `recognize` and uses it for the rank-1 flatten demotion, the merge should keep a separate diameter statistic for flatten, not reuse the complement of the coverage screen.
- Whichever PR lands second will hit a textual conflict on the `recognize` lines.

---

PR #3614, opened 2026-09-20
debiased_functional: weight the Gaussian row scores and replay the fit's weight/offset columns

### LINKED ISSUE #3542: pyffi: model.debiased_functional drops the prior weights from the Gaussian row scores, so SE/CI are wrong for any weighted fit (and change when every weight is multiplied by the same constant)
## Where

`crates/gam-pyffi/src/latent/reml_latent_fit_ffi.rs`, `model_debiased_functional_dataset_json_impl` (origin/main, around l.3877-3885). This is the backend of `gamfit` `Model.debiased_functional(...)` (`gamfit/_model.py` ~l.1000).

```rust
let eta = x.as_ref().dot(&beta) + &effective_offset;
let mut row_scores = ndarray::Array2::<f64>::zeros((n, p));
for i in 0..n {
    let residual = eta[i] - y[i]; // ∂nll_i/∂η = η_i − y_i for Gaussian/identity
    let x_row = x.row(i);
    for j in 0..p {
        row_scores[[i, j]] = x_row[j] * residual;
    }
}
```

## Why it is wrong

In the same function, a few lines above, the Hessian is documented and rebuilt as the **prior-weighted** one: `H = XᵀWX + S(λ)` with `W = diag(prior weights)`. The fallback even computes `Xᵀ diag(standard.weights) X`. For a weighted Gaussian/identity fit, the per-row score of the objective that `H` is the Hessian of is

    s_i = ∂/∂β [ ½ w_i (y_i − η_i)² ] = w_i · x_i · (η_i − y_i)

The code omits `w_i`. `debias_with_dense_hessian` (`crates/gam-sae/src/inference/riesz.rs:317-360`) uses these scores only in the influence values `ψ_i = −n · s_iᵀ H⁻¹ g`, and from those it gets `se`, `ci_lower` and `ci_upper`. So:

- **Scale-invariance check (a failing test).** Multiplying every prior weight by a constant `c` leaves a Gaussian fit's β̂ and λ unchanged, because the dispersion absorbs `c`. The debiased-functional SE must therefore be unchanged too. Instead, `H` scales by `c` while the scores do not, so the reported `se` scales as `1/c`. With `weights = 10` everywhere, the reported CI is 10× too narrow (anti-conservative). With `weights = 0.1`, it is 10× too wide (conservative).
- **Non-uniform precision weights** (`w_i = 1/σ_i²`) give the wrong sandwich, `H⁻¹ Σ x_i x_iᵀ r_i² H⁻¹` instead of `H⁻¹ Σ w_i² x_i x_iᵀ r_i² H⁻¹`. The CI is then miscalibrated in a data-dependent direction.

`theta_debiased` itself is unaffected, because the one-step correction uses `penalty_beta = (H − XᵀWX)β`, not the scores. Only the uncertainty is wrong.

SPEC: "P-values must be calibrated and accurate. Conservative is a bug just like anti-conservative." The same function also shows the pyffi-thickness problem ("Python should be a thin wrapper over Rust"; pyffi should stay thin). The Gaussian score chain, the Hessian/Gram recovery and the functional dispatch all live in the pyffi crate rather than in a library `gam-inference` entry point that the CLI could share. When this is fixed, move that assembly into the library.

## Principled fix

1. Use the prior weights in the score: `row_scores[[i, j]] = w[i] * x_row[j] * residual`, with `w = standard.weights`. Check that `w.len() == n`, which the Gram fallback already does. Better still, build the score from the family's own per-row working-score routine rather than hand-rolling the Gaussian case in pyffi.
2. Regression test: fit `y ~ s(x)` twice, with `weights = 1` and `weights = 7` (constant). Assert that `se`, `ci_lower` and `ci_upper` of `debiased_functional(target="average_value")` agree to within about 1e-8 relative tolerance. This fails on main today by a factor of 7.

## Secondary (same function, same PR)

`spec_val.get("weights")...map(|v| v.as_f64().unwrap_or(1.0))` (~l.3943) turns any non-numeric weight entry (`None`/`null`) into weight 1.0 without saying so. Also, a `"weights"` value that is not an array is silently ignored (`as_array()` returns `None`, which falls through to unweighted). Both should be hard errors. `weighted_affine_mean` already rejects non-finite weights, so these entries should be rejected the same way instead of being replaced.

Found by audit agent HN-21. The file is too large (225 KB) to patch safely through the contents API, so this is left for a fixer.

SauersML:
Fixes #3542

`model_debiased_functional_dataset_json_impl` had two defects that together gave wrong uncertainty on every weighted fit.

1. **Unweighted scores.** The Riesz engine inverts `H = XᵀWX + S(λ)`, where W is the prior weights. The score of that objective is `s_i = w_i · x_i · (η_i − y_i)`, but the code built `x_i · (η_i − y_i)`. Scaling every weight by a constant c is a pure dispersion change and leaves β̂ where it is, yet it scaled the reported SE by 1/c. With non-uniform weights the sandwich was also wrong.
2. **Default materialization config.** The handler called `parse_fit_config(None)`, so the training rows were replayed with unit weights and a zero offset whatever the model had been fit with. That affected `standard.weights` (used by the scores and the Gram fallback) and `standard.offset` (used for the residual's η). It now uses `postfit_standard_materialization_config(model)`, as the other post-fit handlers already do.

**Secondary change:** a functional `"weights"` value that is not an array, or an entry that is not a number, is now refused. Previously these were silently read as "unweighted" and as weight 1.0.

**Test:** `tests/bug_hunt_debiased_functional_prior_weight_scores_test.py`.
- It fits `y ~ s(x)` with `weights="w"` set to w ≡ 1 and w ≡ 7, then asserts that SE and θ̂ agree for `average_value` and `average_derivative`. Before this change the SEs differed by a factor of 7.
- It also checks that a `None` weight entry is refused.

**Not done here:** moving the Gaussian score assembly out of pyffi and into a library `gam-inference` entry point, which the issue suggests. This PR only changes the data the existing assembly consumes.

---


SauersML:
LC-10: the proof for this PR is blocked on infrastructure. validate-one run 35499321794 failed with `gh: Cannot trigger a 'workflow_dispatch' on a disabled workflow (HTTP 422)`: wheel-nightly.yml is disabled, so the Python evidence test (`tests/bug_hunt_debiased_functional_prior_weight_scores_test.py`) cannot run against a built wheel. The Rust change in gam-pyffi/src/latent/reml_latent_fit_ffi.rs needs that Python proof before merging. Leaving this open until the wheel workflow is re-enabled.

---

SauersML:
Taken over by RF-0748-3; driving to merge. The wheel workflow is disabled, so I'll build the extension locally with `maturin develop` and run the Python evidence test against it.

---

PR #3615, opened 2026-09-20
gam-terms penalty_priors: resolve Gamma hyperprior labels by specificity

### LINKED ISSUE #3563: gam-terms penalty_priors: Gamma hyperprior label lookup resolves by alphabetical order, so a term-wide key shadows (and then rejects) a block-specific key
## Where (origin/main)

`crates/gam-terms/src/smooth/penalty_priors.rs`

`penalty_block_label_candidates` (lines 68-83) builds every label a penalty block answers to, then **sorts them alphabetically**:

```rust
labels.push(format!("penalty:{}", info.global_index));
labels.push(info.global_index.to_string());
if let Some(termname) = info.termname.as_ref() {
    labels.push(termname.clone());
    labels.push(format!("{termname}:{}", info.penalty.original_index));
}
...
labels.push(format!("{:?}", info.penalty.source));
labels.sort();
labels.dedup();
```

`realize_keyed_penalty_block_gamma_priors` (lines 151-198) takes the **first** candidate present in the user's map, marks only that key consumed, and afterwards rejects every unconsumed key as unknown:

```rust
for label in penalty_block_label_candidates(info) {
    if let Some(value) = keyed.get(&label) {
        consumed.insert(label);
        return Some(*value);
    }
}
...
"unknown Gamma precision hyperprior penalty block label(s): {}; available labels: {available}"
```

## Why it is wrong

Which key wins is decided by byte order, not by how specific the key is. A term name `T` always sorts before `T:k`, so a term-wide key always beats a block-specific key.

Take the normal use of this API: set a term-wide default and override one block, e.g. `{"g": (1, 1), "g:0": (1000, 1)}`, or `{"x": ..., "x:1": ...}` for a double-penalty smooth.

1. The block-specific override `T:k` is never applied.
2. Because the override is never consumed, the call then **fails** with `unknown Gamma precision hyperprior penalty block label(s): g:0`. The same error message lists `g:0` under "available labels".

The same byte-order rule also decides between keys that name the same block exactly (`"3"` vs `"penalty:3"`, or `"x:0"` vs `"penalty:0"`). When they carry different values, one is silently dropped or rejected with the same misleading "unknown" error. No principled winner exists in that case.

**User reachability:** `FitConfig.penalty_block_gamma_priors` is filled from `precision_hyperpriors`:

- `gam-config/src/lib.rs:124`;
- `gam-models/src/fit_orchestration/drivers/design_construction.rs:115,161`;
- Python `precision_hyperpriors={...}`.

## Principled fix

Resolve by specificity instead of alphabetical order:

1. **Exact block:** `penalty:{gi}`, `{gi}`, `{term}:{k}`.
2. **Whole term:** `{term}`.
3. **Penalty source label:** `Other(label)` / `Debug(source)`. These can span terms.

Rules:

- Each block takes the value from the most specific tier that has a key.
- Two supplied keys in that same tier are an ambiguity error, not a byte-order coin flip.
- A key is "unknown" only when no block answers to it. A term-wide default whose blocks are all overridden is legitimate.

SauersML:
Fixes #3563

`realize_keyed_penalty_block_gamma_priors` used to sort each block's candidate labels alphabetically and take the first key that matched. That had two effects:

- A term-wide key such as `g` beat a block-specific key such as `g:0`.
- The key that lost was reported as an "unknown" label, so the fit errored.

This PR resolves the labels by specificity instead:

- **Tiers:** each block's candidate labels are grouped by how specific they are.
  1. The block itself: `penalty:{gi}`, `{gi}` and `{term}:{orig_idx}`.
  2. The term: `{term}`.
  3. The penalty source: the `Other` label and the Debug form.
- **Resolution:** the first tier with any matching key supplies the block's prior. If two keys in that tier both name the block, the fit fails with an "ambiguous" error.
- **Unknown keys:** a key is unknown only if it matches none of the candidate labels of any block. A key that is shadowed by a more specific one is a known label, not an unknown one.
- **Validation:** every (shape, rate) pair is checked up front.

Two new tests are added in `tests/glm/families/gamma_precision_hyperpriors.rs`:

- `block_specific_gamma_prior_overrides_term_wide_default` checks that `g:0` beats `g`.
- `two_block_specific_gamma_priors_for_one_block_are_ambiguous` checks that `g:0` together with `penalty:0` is rejected.

---


SauersML:

---

SauersML:

---

SauersML:

---

PR #3617, opened 2026-09-20
gam-math: use special::softplus in categorical and weighted_chi_square

### LINKED ISSUE #3559: gam-math: private softplus copies in categorical.rs and weighted_chi_square.rs duplicate special::softplus
gam-math has a canonical stable softplus at `crates/gam-math/src/special.rs:783`:

```rust
pub fn softplus(x: f64) -> f64 {
    x.max(0.0) + (-x.abs()).exp().ln_1p()
}
```

The comment above it says the crate once carried byte-identical private copies of these maps, and that they were consolidated (#2470). Two private copies have come back on origin/main:

1. `crates/gam-math/src/probability/weighted_chi_square.rs:243`
   ```rust
   fn softplus(z: f64) -> f64 { z.max(0.0) + (-z.abs()).exp().ln_1p() }
   ```
   This is a textual copy of `special::softplus`. It is used at lines 651, 667, 686 and 708.

2. `crates/gam-math/src/categorical.rs:177`
   ```rust
   fn log1p_exp(value: f64) -> f64 {
       if value > 0.0 { value + (-value).exp().ln_1p() } else { value.exp().ln_1p() }
   }
   ```
   This is bit-identical to `special::softplus` for every input:
   - For x > 0 both compute the same expression.
   - For x ≤ 0, `0.0 + y == y` because y ≥ 0.
   - ±inf and NaN give the same results.

   It is used at lines 395 and 413.

The companion `log1p_exp_error` is a rounding bound. It mirrors the branch structure of the formula, not a second implementation of it, so it stays.

This goes against the single-source-of-truth rule in SPEC ("It's required that the CLI, Python library, and Rust library are unified, have a single source of truth") and against the #2470 consolidation. If one copy is later changed, for example to tighten accuracy, the others will silently drift.

**Fix**
- Delete both private functions.
- Import `crate::special::softplus` in both files.
- In categorical.rs, rename the two `log1p_exp(...)` calls to `softplus(...)`.

Numerics do not change: every result is bit-identical.

SauersML:
Fixes #3559

Removes two private softplus copies in gam-math in favour of the canonical `crate::special::softplus`:

- `probability/weighted_chi_square.rs`: `fn softplus(z)` was identical to the canonical one, so the copy is dropped and the function is imported.
- `categorical.rs`: `fn log1p_exp(value)` is replaced by `softplus`. It was bit-identical: for x <= 0, `0.0 + y` is exact, and ±inf and NaN take the same paths. `log1p_exp_error` is kept, since it is still the correct first-order rounding bound for this evaluation; only its doc comment was updated.

Numerics are unchanged.

---


SauersML:
Taken over by RF-0748-3; driving to merge. This PR has had no push since 08:21.

---

PR #3619, opened 2026-09-20
gam-predict: drop dead posterior_eta_bands path and its unvalidated band reducer

### LINKED ISSUE #3604: gam-predict/posterior_bands: dead `posterior_eta_bands` pyfunction keeps a second, unvalidated band reducer (NaN draws silently give garbage bands)
## Summary

`crates/gam-predict/src/posterior_bands.rs` on origin/main has two posterior-band reducers:

1. `draw_bands_from_matrices`, which goes through `matrix_bands`. It validates `level`, empty draws, and NaN/±inf draw values. It is the path every live frontend uses:
   - `gamfit/_sampling.py:90` goes through `posterior_draw_bands`;
   - `posterior_predict_bands_encoded_table_impl` calls it directly.
2. `eta_bands_from_matrix_link` together with `LinkSelector`. This is a hand-copied second reducer (per-row mean, sort, `quantile_from_sorted`) that validates **no draw values**. Its only caller is `posterior_eta_bands_impl` (`crates/gam-pyffi/src/manifold/manifold_and_posterior_ffi.rs:88`), which backs the `#[pyfunction] posterior_eta_bands` (`crates/gam-pyffi/src/latent/reml_latent_fit_ffi.rs:3382`, registered at `geometry_ffi.rs:4894`).

`git grep posterior_eta_bands origin/main -- '*.py'` finds **no Python caller**. The only other hits are the generated stubs `gamfit/_rust.pyi` and `gamfit/_rust_module.pyi`. `docs/public-api-2829-disposition.tsv:307-308` already marks `posterior_eta_bands` / `posterior_eta_bands_link` as `retired-no-surviving-dependent`, but the pyfunction and its reducer survived.

## Why it is wrong

- **Silent error swallowing / wrong output.** `eta_bands_from_matrix_link` sorts with `a.partial_cmp(b).unwrap_or(Ordering::Equal)` and never checks for NaN. A single NaN η draw produces an inconsistent comparator. The row's quantiles are then whatever the sort leaves in positions `⌊α(n−1)⌋`, and the mean is NaN. The call still returns `Ok`. By contrast, `matrix_bands` rejects exactly this case: "posterior linear-predictor draw is invalid at draw …".
- **Duplicated logic** that should have one source of truth. Issue #405 deduplicated this reducer once already; this copy came back beside it.
- **Dead public surface.** SPEC: "Unnecessary choices and options should be deleted, not included." Audit rule: "Keep no compat shims and no dead code."
- A smaller point in the same file: `posterior_quantile_from_sorted` is a redundant wrapper around `gam_math::quantile::quantile_from_sorted`. For `len >= 2`, the `[finite, +inf]` case already interpolates to `+inf` in `quantile_from_sorted`'s general branch, and `[+inf, +inf]` hits the `lower == upper` early return. So the wrapper's special cases are unreachable duplicates. There is also an orphaned doc fragment (`/// semantics, but the response-scale push-through …`) left over from a deleted function.

## Principled fix

- Delete `LinkSelector`, `eta_bands_from_matrix_link`, `posterior_eta_bands_impl`, the `posterior_eta_bands` pyfunction and its registration, and the two stub entries.
- Use `quantile_from_sorted` directly in `matrix_bands`, and sort with `f64::total_cmp`, since the draws are validated and contain no NaN.
- Drop the orphaned doc fragment.
- Fix the `model_link_spec_json` doc comment so it no longer names `posterior_eta_bands`.

SauersML:
Fixes #3604

## What was wrong

The `posterior_eta_bands` pyfunction had no Python caller. `docs/public-api-2829-disposition.tsv` already marks it as retired. Its call chain was:

- `posterior_eta_bands` (pyfunction)
- `posterior_eta_bands_impl`
- `eta_bands_from_matrix_link` / `LinkSelector`

`eta_bands_from_matrix_link` was a hand-copied second copy of the posterior-band reducer, and it never validated draw values. A NaN η draw made its `partial_cmp(..).unwrap_or(Equal)` sort inconsistent, and the call still returned `Ok` with garbage quantiles.

## Changes

- Delete that whole path: the pyfunction, its registration, the `_impl`, `LinkSelector`, `eta_bands_from_matrix_link`, and both stub entries.
- `draw_bands_from_matrices`, which rejects NaN and invalid infinities, is now the single posterior-band reducer.
- `matrix_bands` calls `quantile_from_sorted` directly. The `posterior_quantile_from_sorted` wrapper's `+inf` special case is exactly what `quantile_from_sorted`'s general interpolation branch already returns (`lower*(1-f) + inf*f = inf`), and `[inf, inf]` takes its `lower == upper` early return.
- `matrix_bands` sorts with `f64::total_cmp`, since the draws are validated NaN-free.
- Remove an orphaned doc-comment fragment, and the stale mention in `model_link_spec_json`'s doc.

The existing unit tests in `posterior_bands.rs` still cover the live reducer: the model-specific mean draws, the shape mismatch, `+inf` preservation, and invalid-value rejection.

---


SauersML:
Taken over by RF-0748-3; driving to merge. This PR has had no push since 08:21, and main merged in cleanly (48f82bbd).

---

SauersML:

---

PR #3494, opened 2026-09-20
Latent survival rho-domain fit certifies its outer optimum (#3321)

SauersML:
WIP.

Fixes #3321

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MwMvgB7i335nVFQ8vfZ44B

---
_Generated by [Claude Code](https://claude.ai/code/session_01MwMvgB7i335nVFQ8vfZ44B)_

PR #3620, opened 2026-09-20
survival: delete the dead Cox risk-calibration path

### LINKED ISSUE #3606: survival/risk_calibration: cox_calibrated_survival_matrix is dead code (benches moved to lifelines) that silently scores a NaN test risk as the mean risk and costs O(n·events) per Newton evaluation
## What

`crates/gam-models/src/survival/risk_calibration.rs` (origin/main 6942363) holds `cox_calibrated_survival_matrix`, a univariate Cox + Breslow calibration of a risk score. The pyffi function `survival_matrix_from_risk_calibration` (`crates/gam-pyffi/src/latent/reml_latent_fit_ffi.rs:4702-4739`, registered at `crates/gam-pyffi/src/manifold/geometry_ffi.rs:4940-4943`) exposes it. Stubs for it are in `gamfit/_rust.pyi:429,2248` and `gamfit/_rust_module.pyi:772`.

**Nothing calls it.** `git grep survival_matrix_from_risk_calibration origin/main` finds only its own definition, its registration and the generated stubs. The module doc says it is "the benchmark comparator path". The benchmarks, however, now calibrate through lifelines:
- `bench/_run_suite_datasets.py:560` calls `_survival_calibration().calibrated_survival_matrix`.
- `bench/large_scale/runner.py:1019` does the same.
- Both go to `bench/_survival_calibration.py:56`, which uses `CoxPHFitter`.

SPEC: "Breaking changes are welcome. Keep no compat shims and no dead code."

The dead path also has two defects that would bite anyone who wires it back in.

1. **A NaN test risk is silently scored as the average subject.** At risk_calibration.rs:~302:
   ```rust
   let x = if risk.is_finite() {
       risk - risk_mean
   } else {
       0.0
   };
   ```
   A model that fails to predict a row gets the population-mean survival curve for that row, and its Brier score looks like a real prediction. Training rows with non-finite risk are likewise dropped without a word. Both are silent error swallowing. The lifelines comparator raises `"survival calibration risk scores must be finite"` instead.

2. **The cost is O(n · events) per partial-likelihood evaluation.** `risk_set_sums(beta, time)` makes three full passes over all `rows` for every distinct event time. `partial_likelihood` calls it once per event time, inside a Newton loop that halves its step. At n = 10^5 with 50% events, that is about 1.5·10^10 exp() calls per evaluation. SPEC: "Fitting and inference must be fast on all scales of data".

In addition, the first grid column is pinned to `1.0` (`col_idx == 0 || h0 <= 0.0`) "regardless of where grid starts". That value is wrong whenever `grid[0]` lies past the first training event.

## Fix

Delete it rather than repair it:
- Remove `risk_calibration.rs` and its `pub mod` / `pub use` lines in `survival/mod.rs`.
- Remove the pyffi `survival_matrix_from_risk_calibration` and `benchmark_survival_matrix_from_risk`, and their registration and stubs.
- The one live item in the module is `km_curve_on_grid`, a one-line alias of `KaplanMeier::fit(..).on_grid(..)`. `survival_null_curve_from_train` (used by `bench/large_scale/runner.py:1003`) should call `KaplanMeier` directly.

SauersML:
Fixes #3606

`cox_calibrated_survival_matrix` (`survival/risk_calibration.rs`) and its pyffi wrapper `survival_matrix_from_risk_calibration` have no caller. The benchmarks calibrate risk scores through lifelines (`bench/_survival_calibration.py`, used by `bench/_run_suite_datasets.py:560` and `bench/large_scale/runner.py:1019`).

The dead path also had three defects:
- It scored a NaN test risk as the mean-risk subject.
- It dropped non-finite training risks without saying so.
- It cost O(n · events) per partial-likelihood evaluation.

Changes:
- Remove `risk_calibration.rs` and its `pub mod` / `pub use` lines.
- Remove the pyffi `survival_matrix_from_risk_calibration` and `benchmark_survival_matrix_from_risk`, their registration, and the generated stubs in `gamfit/_rust.pyi` and `gamfit/_rust_module.pyi`.
- `survival_null_curve_from_train` was the one live user of the module's `km_curve_on_grid` alias. It now calls `KaplanMeier::fit(..).on_grid(..)` directly, with the same values.

## Validation (landing session L-03, head 6ac7cb7a, main merged as a merge commit)
- [x] `cargo check --workspace --tests`: clean, after both main merges. The one warning is an unused `IxDyn` import in `gam-pyffi/src/ffi/ffi_prelude.rs`, and main has it too.
- [x] `scripts/spec_ban_ratchet.py --base origin/main`: 0 new, 0 stale.
- [x] No references to the removed symbols remain in `.rs`/`.py`/`.pyi`. The bench's `_survival_matrix_from_risk_calibration` is the local lifelines helper.
- [x] `cargo test -p gam-models --lib survival`: 727 passed, 2 failed. Both failures are latent-survival outer BFGS stalls (`BfgsCostStallExit`), which is known main-red #3321 and is being fixed in #3494:
  - `saved_latent_survival_model_predicts_at_its_fitted_baseline_2714`
  - `tests_rho_domain_2902::latent_rho_domain_is_its_realized_blocks_domain_and_rails_at_switch_off_2902`

  This PR touches none of that path.
- [ ] Python wheel not rebuilt here. The change only removes two unused pyffi entry points and their stubs.


SauersML:

---

PR #3630, opened 2026-09-20
fit_orchestration/descriptors: strict typed descriptor fields; single-source Gumbel schedule parser

### LINKED ISSUE #3521: fit_orchestration/descriptors: wrong-typed analytic-penalty fields silently fall back to defaults; geometric temperature_schedule ignores `steps` and defaults rate to a magic 0.9
## Where

`crates/gam-models/src/fit_orchestration/descriptors.rs` on origin/main (last touched by f076f9d85b). This is the single shared analytic-penalty descriptor parser. It is on the production path from:
- `gam-config::resolve_fit_request_config`, where `AnalyticPenaltiesDocument` is an untyped `Vec<JsonValue>`, so this parser is the only validator;
- `fit_orchestration/materialize/latent.rs:887` (`prepare_standard_latent_coord`);
- the Python FFI (`gam-pyffi/src/ffi/ffi_prelude.rs:92`).

## Defect 1: a present field with the wrong JSON type is silently replaced by the default

Every typed getter uses `get(key).and_then(JsonValue::as_*).unwrap_or(default)`, so "absent" and "present but wrong type" end up in the same branch:

```rust
// :127
fn descriptor_f64(descriptor, key, default) -> Result<f64, String> {
    let value = descriptor.get(key).and_then(JsonValue::as_f64).unwrap_or(default);
    ...
// :142
fn descriptor_usize(descriptor, key, default) -> Result<usize, String> {
    let Some(raw) = descriptor.get(key).and_then(JsonValue::as_u64) else {
        return Ok(default);
    };
```

Examples (all accepted without error):
- `{"kind":"orthogonality","weight":"5"}` is fitted with weight 1.0.
- `"gamma":"3"` on `scad_mcp` falls back to the variant default (2.5 or 3.7).
- `"n_eff": 1000.0`, `"n_eff": -3` or `"k": 2.0` all fall back silently to `target.n` or 1. `as_u64` rejects floats and negatives.
- `"learnable": "true"` or `"learnable": 1` becomes `false`. The `.and_then(JsonValue::as_bool).unwrap_or(false)` pattern is used at :630, :736, :765, :818, :881, :939, :974, :1010, :1058, :1096, :1167, :1203, :1242 and :1276, so a penalty the user asked to make learnable is silently held fixed.
- `sparsity_kind` (:657), `eps_weight` (:664), `variant` (:719), `difference_op` (:336) and `tier` (:1341) all fall back to their default string when given a non-string. For example, `"variant": 1` fits MCP.
- `latents[k].name` (:71) falls back to the key when it is a non-string.

This is silent error swallowing. The same parser already rejects unknown keys (`descriptor_no_unknown_keys`) and wrong-typed `weight` on isometry/sparsity (`descriptor_weight_scalar`), so it is also inconsistent. On isometry, `"weight": "5"` is an error; on ard, orthogonality and every other kind, it silently becomes 1.0.

## Defect 2: geometric `temperature_schedule` ignores `steps` and uses a hard-coded rate of 0.9

```rust
// :287
"geometric" | "exponential" => {
    let rate = schedule.get("rate").and_then(JsonValue::as_f64).unwrap_or(0.9);
```

The Python `GumbelTemperatureSchedule` (`gamfit/_penalty_bridge.py:185`) documents that "A geometric schedule takes an explicit `rate` or the (tau_start, tau_min, steps) endpoints". The other Rust parser of the same schedule, `gam-pyffi/src/latent/latent_basis_and_sae_ffi.rs::gumbel_temperature_schedule_from_pydict`, honours `steps` and derives the rate with `ScheduleKind::geometric_rate_from_steps`. That function's doc comment calls it "the single source for the endpoints-and-steps geometric spec".

In this parser, `steps` is dropped. Given `{"tau_start":1,"tau_min":0.01,"decay":"geometric","steps":10}`:
- the pyffi path anneals at rate 0.631;
- this parser anneals at 0.9, which takes about 44 steps to reach the same floor.

So the same schedule dict means two different annealing laws depending on the entry point. That violates "It's required that the CLI, Python library, and Rust library are unified, have a single source of truth". The 0.9 default is also an arbitrary constant: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code." By contrast, `weight_schedule` in the same file already requires `rate` for geometric (:218).

## Defect 3 (smaller)

- **`block_orthogonality` rejects its own schedule.** Its allowed-key list (:757) omits `"weight_schedule"`, but the code still applies it (`match weight_schedule { Some(..) => penalty.with_weight_schedule(..) }` at :769). As a result, the branch is dead and a scheduled `block_orthogonality` weight is rejected with "not a recognized field". Every other weighted kind accepts `weight_schedule`.
- **Alias pairs silently drop one key.** `learnable` / `learnable_alpha` (:878) and `tau_min` / `tau_end` (:276) use `.or_else`. When both keys are present, the second is ignored, even if the two values disagree.

## Principled fix (local to descriptors.rs)

- Add strict getters. "Absent or `null`" means the documented default. "Present with the wrong type" is an `Err` naming the field. Provide `descriptor_f64`, `descriptor_usize`, a new `descriptor_bool` and a new `descriptor_str`, and route every site above through them.
- Geometric `temperature_schedule`:
  - explicit `rate`: use it;
  - `steps`: derive the rate through `ScheduleKind::geometric_rate_from_steps`, the same as the pyffi parser;
  - both present: an ambiguity error;
  - neither present: an error, as `weight_schedule` already does. No 0.9 default.
- Add `weight_schedule` to the `block_orthogonality` allowed keys.
- Reject a descriptor that carries both `learnable` and `learnable_alpha`, or both `tau_min` and `tau_end`.

Follow-up, not local: the pyffi `gumbel_temperature_schedule_from_pydict` keeps its own copy of this schema, including the same `None => 0.9` default. It should convert the dict to JSON and call one shared parser here.

SauersML:
Fixes #3521

## What was wrong
- `descriptors.rs` read optional fields with `as_f64`/`as_u64`/`as_bool`/`as_str` followed by `unwrap_or(default)`. A field with the wrong type (for example `"weight": "2"` or `"learnable": "yes"`) was silently replaced by the default instead of being refused.
- A geometric `temperature_schedule` with no `rate` and no `steps` silently fell back to a hard-coded 0.9 rate. SPEC: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code."
- The pyffi parser `gumbel_temperature_schedule_from_pydict`, used by `sae_manifold_fit(gumbel_schedule=...)` and `gumbel_schedule_tau`, was a second, diverging implementation. It also defaulted the rate to 0.9. When both `rate` and `steps` were given it silently used `steps`. It also rejected the `exponential` and `tau_end` aliases that the descriptor parser accepts. SPEC: "It's required that the CLI, Python library, and Rust library are unified, have a single source of truth".

## Fix
- Added typed helpers (`descriptor_f64`, `descriptor_usize`, `descriptor_bool`, `descriptor_name`, `descriptor_learnable`, `descriptor_present`). A present field with the wrong type is now an error naming the field and the value received. JSON `null` still counts as absent, so Python `None` passes through unchanged.
- Setting two alias fields together is refused as ambiguous: `tau_min`/`tau_end`, `learnable`/`learnable_alpha`, and `rate`/`steps` for geometric decay.
- A geometric schedule now needs `rate` or `steps`, with no default. `steps` derives the rate through the single-source `ScheduleKind::geometric_rate_from_steps`.
- The schedule parser is now public as `descriptors::gumbel_temperature_schedule_from_json`. The pyffi `gumbel_schedule` dict is converted with `py_any_to_json_value` and parsed by this function, and the duplicate pyffi parser is deleted.
- `block_orthogonality` now accepts the `weight_schedule` key that it already parses.
- The two examples that relied on the 0.9 default now pass `steps=n_iter`, and the Python docstring is corrected.
- New unit tests cover wrong-typed fields, alias ambiguity, a missing rate or steps, the rate derived from steps (compared bitwise), and the shared parser's error context.

---


SauersML:

---

SauersML:

New head: 45f7123e50d97649b0e1c54bbfb24cff09d459b2. Only `docs/source-removal-changes.json` conflicted, because both sides appended entries. The resolution keeps every entry from main and then this PR's nested-`get` entry (173 -> 174); it parses as JSON. Main has not touched descriptors.rs, latent_basis_and_sae_ffi.rs, ffi_prelude.rs or _penalty_bridge.py since this branch's merge-base.

---

PR #3631, opened 2026-09-20
k_selection: refuse measured MDL without its coding ingredients; expose them over FFI

### LINKED ISSUE #3600: k_selection: mode "measured" silently falls back to the tuned Kneedle knob, and the FFI can never reach the measured MDL rule
## Defect

`crates/gam-sae/src/k_selection.rs` (origin/main) describes `KSelectionMode::MeasuredMdl` as "the tuning-free replacement for `PenalizedMdl`'s `γ` ... This is the RECOMMENDED path". In practice nobody can reach it.

**1. Silent fallback to the tuned knob.** `select_k`, around line 374:

```rust
KSelectionMode::MeasuredMdl => match config.measured_coding {
    Some(coding) => select_measured(curve, &coding),
    // No coding ingredients to measure with: fall back to the knee.
    None => select_kneedle(curve, config, span),
},
```

A caller that asks for the measured rule without coding ingredients silently gets Kneedle. Kneedle's knee acceptance is the arbitrary `knee_slope_fraction = 0.10` default, plus `LINEARITY_SLOPE_REL_TOL = 0.05`. The result is still flagged `Knee`, so nothing says the tuned knob chose K. The test `measured_mode_without_coding_falls_back_to_knee` pins this behaviour.

**2. The only production entry points hard-wire `None`.** `crates/gam-pyffi/src/manifold/geometry_ffi.rs` has two such entry points:
- `sae_select_k` (line 330);
- `sae_auto_k_recommendation` (line 373).

Both build `KSelectionConfig { ..., measured_coding: None }` with the comment "a `MeasuredMdl` mode string falls back to Kneedle when this is `None`". So `mode="measured"` (or `"theorem4"`) from Python always runs Kneedle at 0.10. No caller anywhere constructs a `MeasuredCoding`: `git grep "MeasuredCoding {"` finds only the module's tests. The measured rule is dead code in production.

## SPEC

- "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code." The derived rule exists, and the tuned knob silently replaces it.
- "In general, do not paper over solver issues." The missing inputs are covered up by a fallback instead of being refused.

## Fix

- `select_k` / `recommend_auto_k` return `Result`, and `MeasuredMdl` without `measured_coding` is an error, not a fallback.
- Both FFI functions accept the five fit-measured ingredients (`d_eff_atom`, `n_eff`, `n_rows`, `k_bar`, `d_bar`) as optional keyword arguments:
  - they build `MeasuredCoding` when all five are given;
  - they refuse a partial set;
  - `mode="measured"` without the ingredients errors instead of silently running the knee.
- Replace the fallback test with a refusal test. Regenerate the stubs.

SauersML:
Fixes #3600

## Problem
- In `KSelectionMode::MeasuredMdl` with `measured_coding: None`, `select_k` quietly answered with the tuned Kneedle knee (`knee_slope_fraction = 0.10`). A caller who asked for the tuning-free Theorem-4 rule got a knob's pick, labelled as the measured one.
- Both scalar-curve FFI entry points (`sae_select_k`, `sae_auto_k_recommendation`) always passed `measured_coding: None`. So `mode="measured"` from Python could never reach the measured rule.

## Fix
- `select_k` and `recommend_auto_k` now return `Result<_, String>`. In measured mode they return an error when the coding ingredients are missing. The check runs before the single-point and flat early exits, so the error does not depend on the curve's shape.
- The two FFI functions take the five measured ingredients as optional kwargs: `d_eff_atom`, `n_eff`, `n_rows`, `k_bar`, `d_bar`.
  - All five supplied: a `MeasuredCoding` is built.
  - None supplied: it stays unset.
  - Only some supplied: `ValueError`.
- Errors from `select_k` / `recommend_auto_k` are raised as `ValueError`.
- `gamfit/_rust.pyi` and `gamfit/_rust_module.pyi` are updated to match, in the `float | None = ...` form the generator uses.
- The test `measured_mode_without_coding_falls_back_to_knee` asserted the fallback. It is replaced by `measured_mode_without_coding_is_refused`, which covers the knee, single-point and flat curves and `recommend_auto_k`.

The only callers are inside `k_selection.rs` and `geometry_ffi.rs`; no Python code calls these functions.

---


SauersML:

---

PR #3633, opened 2026-09-20
survival/predict: binary-search KaplanMeier::at; score only origin-anchored grids

### LINKED ISSUE #3608: survival/predict: KaplanMeier::at is a linear scan, so held-out IPCW Brier scoring costs O(n_test · censoring steps · grid) — and duplicates on_grid's binary search
## What

`crates/gam-models/src/survival/predict.rs` (origin/main 6942363) has two lookups on the same step function:

- `KaplanMeier::on_grid` (~2586) does a binary search: `self.steps.partition_point(|&(time, _)| time <= t)`.
- `KaplanMeier::at` (~2597) walks `steps` from the front on every call:
  ```rust
  pub fn at(&self, t: f64) -> f64 {
      let mut s = 1.0_f64;
      for &(time, surv) in &self.steps {
          if time <= t { s = surv; } else { break; }
      }
      s
  }
  ```

`at` is the hot one. `survival_prediction_scores` (~2489) passes `|t| censoring.at(t)` into `integrated_ipcw_brier_score`, and that calls `ipcw_brier_score` once per grid point. `ipcw_brier_score` then evaluates `g_cens(time[i])` for every event subject and `g_cens(tau)` for every subject past τ.

The cost adds up as follows:
- The censoring KM has one step per distinct censoring time, which is O(n) steps.
- Held-out scoring therefore costs O(n_test · n_steps · |grid|), where `|grid| = 26` from `survival_score_grid`.
- It is paid twice, once for the model matrix and once for the null matrix.

For example, with n_test = 2·10^5 and 10^5 censoring times this is about 10^12 comparisons. This is the scorer behind `survival_lifted_metrics_from_predictions`, which `bench/large_scale/runner.py` uses for its survival metrics.

The two lookups are also duplicated logic for one quantity. `on_grid`'s own doc explains why it is a binary search "rather than the linear scan `at` does".

## Fix

- Make `at` the single binary-search lookup: `partition_point(|&(time, _)| time <= t)` gives the index of the last step `≤ t`, or `1.0` before the first step.
- Have `on_grid` map `at` over the grid.

The two agree on every input, including NaN: `time <= NaN` is false, so both return `1.0`. The existing right-continuity test (`kaplan_meier_censoring_is_right_continuous_step`) pins the semantics.

### LINKED ISSUE #3609: survival/predict: survival_prediction_scores accepts a grid that does not start at 0, then overwrites the model's S(grid[0]) with 1 and scores an event before grid[0] with a negative cumulative hazard
## What

`survival_prediction_scores` (`crates/gam-models/src/survival/predict.rs` ~2458, origin/main 6942363) validates only that `grid` is strictly increasing and has at least two points. It then calls `monotone_survival_and_hazard_scores` (~2245), which assumes the grid starts at the time origin in two places.

1. It pins the first column to 1:
   ```rust
   for mut row in surv.rows_mut() {
       row[0] = 1.0;
   ```
   `S(t) = 1` holds only at `t = 0`. When `grid[0] > 0`, this discards the model's prediction `Ŝ(grid[0]) < 1`. It also moves the whole cumulative hazard `−ln Ŝ(grid[0])` out of every subject's `H(T_i)`, so the log-loss is biased low.

2. It extrapolates backwards for an event time below `grid[0]`. Here `partition_point` gives `j = 0` and `interval_idx = 0`, so
   ```rust
   let elapsed = time - grid[interval_idx];          // < 0
   ... cumhaz[[row, 0]] + h * elapsed,               // = 0 + h·(negative)
   ```
   The result is a **negative** cumulative hazard, which gives a negative `H(T_i)` log-loss term and a negative `∫h²`. The quadratic hazard score then rewards the prediction for time it never covered.

This function is the single owner of held-out survival scoring. Its doc says "What a malformed input means is decided here, once, for every front door". It is exposed as `survival_lifted_metrics_from_predictions` with a caller-supplied grid. `survival_score_grid` always starts at 0, so the benches are unaffected today, but any other grid is scored wrongly without complaint.

## Fix

The hazard-path scores integrate from `t = 0`, so a scoring grid must begin at `0`:
- `survival_prediction_scores` returns the all-`None` scores when `grid[0] != 0.0`, as it does for any other malformed grid.
- `monotone_survival_and_hazard_scores` documents the `grid[0] = 0` precondition next to the pin.
- A test pins the refusal.

(`risk_calibration::cox_calibrated_survival_matrix` has the same "pin the first column to 1 regardless of where the grid starts" rule; that code is dead and handled in #3606.)

SauersML:
Fixes #3608
Fixes #3609

**#3608: KaplanMeier lookup.**
- Problem: `KaplanMeier::at` scanned the steps from the front. The IPCW Brier scorer calls it once per subject per grid point, against a censoring curve with one step per censoring time. Held-out survival scoring was therefore quadratic in n.
- Fix: `at` is now the single binary-search lookup (`partition_point`, the one `on_grid` already used), and `on_grid` maps `at` over the grid. Semantics are unchanged, including NaN, which still returns `1.0`.
- Test: `kaplan_meier_lookup_equals_the_right_continuous_scan` checks `at` bit-for-bit against a front-to-back scan on and around every step, outside the support, at ±∞ and at NaN. It also checks that `on_grid` equals `at` mapped over the grid.

**#3609: grid origin.**
- Problem: `survival_prediction_scores` accepted any strictly increasing grid. `monotone_survival_and_hazard_scores` pins column 0 to `S = 1` and integrates the hazard from `grid[0]`. A grid that starts after 0 had the model's `Ŝ(grid[0])` overwritten with 1, and an event before `grid[0]` was scored with a negative cumulative hazard.
- Fix: a grid that does not start at 0, or that has a non-finite point, now returns the all-`None` scores, like any other malformed grid. The docs state the precondition.
- Test: `survival_prediction_scores_refuse_a_grid_not_anchored_at_the_origin`.
- Every in-repo grid comes from `survival_score_grid`, which starts at 0. The Python test fixtures also start at 0.

---


SauersML:
Closing as a duplicate of #3638, which merged as d09970d0. It carries the same #3608 binary-search `KaplanMeier::at` and the #3609 origin-anchored grid refusal.

---

PR #3635, opened 2026-09-20
Move the cross-fit shared precision update into gam-inference and use the penalized model's EM update (#3523)

### LINKED ISSUE #3523: pyffi: cross_fit_shared_precision_groups pools an identity ridge on raw coefficients (||β||², d), which is wrong for any smooth term, and the EB update lives in pyffi
## Defect

`gamfit.inference.cross_fit_shared_precision_groups` (Python: `gamfit/_api.py:141`) is implemented entirely in pyffi, in `crates/gam-pyffi/src/manifold/manifold_and_posterior_ffi.rs`, `cross_fit_shared_precision_groups_json_impl` (~line 790). For each group it matches coefficients by `term`/`column`/`label` and computes

```rust
let trace_covariance = indices.iter().map(|&i| cov_flat[i * cov_n + i]).sum::<f64>();
let beta_norm_sq     = indices.iter().map(|&i| beta[i] * beta[i]).sum::<f64>();
let contribution = beta_norm_sq + trace_covariance;
...
let numerator   = fit_entries.len() as f64 * dimension as f64 + 2.0 * (group.shape - 1.0);
let denominator = quadratic_sum + 2.0 * group.rate;
let lambda = numerator / denominator;
```

This is the EM/MAP update for a prior `β_p ~ N(0, λ⁻¹ I_d)` with a `Gamma(a, b)` hyperprior. That prior is correct only when the block's penalty is the identity. Label matching accepts any term: the docstring says "the same named coefficient term/column/label", and smooth provenance entries carry `term = "s(x)"`. So `SharedPrecisionGroup("s(x)")` selects a whole smooth block, whose actual prior precision is `λ S` with `S` a rank-deficient derivative/difference penalty. For that block the M-step is

```
λ = (Σ_fits rank(S) + 2(a−1)) / (Σ_fits [β'Sβ + tr(S Σ)] + 2b)
```

The code uses `d = #columns` in place of `rank(S)`, which counts the unpenalised null space (linear/constant directions) as shrinkable. It uses `||β||² + tr Σ` in place of `β'Sβ + tr(SΣ)`. The returned `lambda` therefore does not correspond to the model's penalty at all, and applying it back to the fits' `λ` is meaningless. It also puts the penalty on the coefficients rather than on the function. That makes it basis-dependent: two fits of the same term with different knots or reparameterisations get different `||β||²` for the same function. SPEC: "Penalties must always be on the final function itself, never on the model coefficients. Exception: when it can be proven that penalizing model coefficients is precisely equivalent to the function value approach." This equivalence holds only for `group(g)` / i.i.d. random-effect blocks. Those are the only blocks the tests exercise (`tests/test_cross_fit_shared_precision.py` uses `group(g)` exclusively).

Multi-penalty terms (tensor products, double penalty) have several `λ_b` per block, so one shared scalar precision is not even well-typed there. The function accepts them anyway.

## Also a thin-wrapper violation

The whole estimator runs in pyffi, including the hyperprior validation, dimension-consistency rule, closed-form update and JSON result schema. It also re-parses the model's own `coefficient_state_json` from a string. It has no library or CLI counterpart: `git grep shared_precision crates/ -- ':!crates/gam-pyffi'` finds nothing related. SPEC: "pyffi should not contain logic whenever possible, and should be thin". Also SPEC: "the CLI, Python library, and Rust library are unified, have a single source of truth, and have behavior parity". The provenance it matches on is itself a hand-rolled pyffi layout that is off by one for intercept-free models (#3522).

## Principled fix

1. Move the estimator into `gam-inference`, e.g. `inference::shared_precision::cross_fit_update(fits: &[&FittedModel], groups: &[SharedPrecisionGroup])`. It should operate on typed saved-model state (β, the published covariance, the term's penalty blocks from the saved `resolved_termspec` / penalty info) rather than JSON strings. pyffi becomes a one-line marshal, and the CLI can expose the same update.
2. Express each group in the term's own penalty: use `β'S_bβ + tr(S_b Σ)` and `rank(S_b)` for the matched penalty block `b`. The identity/ridge case (`group(g)`) is then the special case `S = I`, `rank = d`. A group whose matched term has more than one penalty block must name the block or be rejected. It must not be silently pooled as an identity ridge.
3. The dimension-consistency check should compare `rank(S_b)` across fits, not column counts.

Regression test: fit `y ~ s(x)` twice and pool `SharedPrecisionGroup("s(x)", shape=1, rate=0)`. With one fit, the update must reproduce the model's own REML-stationary `λ` for `s(x)` to first order: the Fellner–Schall fixed point `λ = rank(S)/(β'Sβ + tr(SΣ))` holds at REML stationarity for a Gaussian fit with known scale, up to the `tr(S_λ⁺ S)` vs `rank` distinction. The current code does not. Keep the existing `group(g)` tests: they must still pass unchanged.

SauersML:
Fixes #3523.

## Why the code moved

`cross_fit_shared_precision_groups` did its math in `crates/gam-pyffi`. It re-parsed `coefficient_state_json`, matched coefficients by provenance text, and pooled `||beta||^2 + tr(Sigma)` with the block's coefficient count `d` as its degrees of freedom. SPEC.md says:

> "Python should be a thin wrapper over Rust, and should avoid containing math or much logic."

> "It's required that the CLI, Python library, and Rust library are unified, have a single source of truth, and have behavior parity. Therefore, pyffi should not contain logic whenever possible, and should be thin."

The update now lives in `gam::inference::shared_precision` (`crates/gam-inference/src/shared_precision.rs`). The pyffi function `cross_fit_shared_precision_groups_json(models, request_json)` only marshals: it parses the request, passes the `_FittedModel`s to `shared_precision_updates`, and serializes the result. The old JSON helpers (`json_f64_vec`, `json_label_text`, `coefficient_indices_for_precision_label_json`, `cross_fit_shared_precision_groups_json_impl`) and the `PySharedPrecision*` structs are deleted. Python no longer checks for duplicate names or empty groups; the engine refuses those.

## The model and the update

In fit `f`, block `b` is penalized by `lambda_f (beta - m)' S_b (beta - m)`. The fit's likelihood is scaled by `1/c_f`, where `c_f = coefficient_covariance_scale()` (`sigma^2` for a profiled Gaussian, `1` otherwise). The stored `S_b` is the physical function-space penalty divided by `nu_f = normalization_scale`. The shared precision is a precision of the penalty on the function, which is the same object in every fit whatever basis each one uses. As the issue quotes SPEC: "Penalties must always be on the final function itself, never on the model coefficients...". The model is:

```
beta_b,f | tau ~ N(m_f, c_f (tau nu_f S_b,f)^+),   tau ~ Gamma(a, b)
E-step:  E_f = nu_f ((beta_hat - m)' S_b (beta_hat - m) + tr(S_b Vb_bb)) / c_f
M-step:  tau = (sum_f rank(S_b,f) + 2(a - 1)) / (sum_f E_f + 2b)
```

`Vb = c H^-1` is the fit's conditional covariance. `rank(S_b)` counts the penalized directions, which are the prior's degrees of freedom. The block's coefficient count `d` also includes the unpenalized null space. The old formula is the special case `S = I`, `c = nu = 1`. Each fit also gets `implied_lambda = tau * nu_f`, its own smoothing parameter under the shared precision, on the same scale as `fitted_lambda`. Iterating the update with each fit refitted at `implied_lambda` is the Fellner–Schall/EM iteration.

## Implementation notes

- **Where the blocks come from.** Penalty blocks, ranks, normalizations and prior means come from the same frozen-basis replay the summary tables use. The new `saved_summary::saved_predictor_designs` has the layout check factored out of the per-smooth table (`check_replayed_penalty_layout`), so both readers refuse a replay whose layout drifted.
- **Indexing.** `lambda` is read at `offset.penalties + position`. Coefficients are read at `offset.coefficients + col_range`.
- **Centre.** The centre `m` is the block's prior mean or the coefficient gauge's affine shift. If both move the same block, the block is refused.
- **Refusals.** Spline-scan models are refused because they have no `Vb`.
- **Matching.** Groups match on the penalized term's name, and a fit without the term is skipped. A term with several penalty blocks (a double penalty or a tensor product) must name one with the new `penalty` field, since each block is its own prior. Groups whose matched blocks have different ranks are refused.

## Tests (`gam-inference`, `shared_precision`)

- **`shared_precision_update_matches_direct_computation_3523`** fits two Gaussian `y ~ x + group(g)` models. It computes `E_f = (||beta_g||^2 + tr Vb_gg)/c_f` directly from the saved fit and checks `tau` against it.
  - Both sides are sums of at most 12 positive terms, so each is within about 12 eps of the exact value. The comparison uses 32 eps per contribution and 64 eps for `tau`.
  - The test asserts `c` is far from 1: the unmodelled `sin(2 pi z)` holds `sigma^2` near 1/2. So a formula that drops `c` would fail.
  - It checks that the `fitted_lambda` it reads is the penalty the fit was solved with: `(H - X'WX)_gg = lambda_g I`, within the rounding bound of an n-term Gram sum.
- **`..._skip_absent_terms_and_refuse_rank_mismatch_3523`** covers absent terms, a rank mismatch, and that a term's source name is not a label.
- **`..._names_one_block_of_a_multi_penalty_term_3523`** uses `s(z, double_penalty=true)`: it must be named with `penalty`, the named block is the one used, and each block's rank is below its width.
- **`..._validate_their_request_3523`** covers request validation.

`tests/test_cross_fit_shared_precision.py` is unchanged. Its closed-form checks hold because `group(g)` is an identity ridge with `nu = 1` and rank 3. The ownership guard in `tests/test_difference_smooth_core_ownership.py` now ends its slice at the next surviving item.

I did not add a test that refits `y ~ s(x)` at `implied_lambda` to check the fixed point: comparing REML's `phi_hat` with the profiled `sigma^2` has no derived tolerance.


SauersML:
Validation status at dfb8700516:

- **gam-inference lib `shared_precision`** (run 35500508063): green, 4/4. These tests check the update against a direct computation from the fit: (β−m)'S(β−m) + tr(S Vb), rank(S), ν, c and the Gamma(a, b) closed form.
- **gam-pyffi lib** (run 35500508828): red. This is not a code failure.
  - rustc built every `gam_pyffi` object with 0 compiler errors. Only the test binary's link failed: `rust-lld: error: undefined symbol: PyList_New / _Py_IncRef / PyExc_ValueError ...` (no libpython).
  - The cause is that focused-rust-proof.yml always passes `--all-features`, which enables `pyo3/extension-module`. crates/gam-pyffi/Cargo.toml says the harness links only under `--no-default-features`, and the workflow cannot pass that flag.
  - So no gam-pyffi harness can link under this workflow, whatever the change. That makes this check irrelevant to the change; it only shows that the pyffi code type-checks.
- **gam-models integration `bms_per_smooth_summary_2997 weibull_survival_summary_3297`** (run 35500517921): queued. It covers the `saved_predictor_designs` / layout-check refactor of the summary replay.
- **Merge check.** The branch merges cleanly with main e695ec5d31. In the 410 commits since the base, none changes the signature of any API the new module calls.

---

PR #3638, opened 2026-09-20
survival scoring: binary-search KM lookup, origin-anchored grid, delete dead Cox calibration

### LINKED ISSUE #3606: survival/risk_calibration: cox_calibrated_survival_matrix is dead code (benches moved to lifelines) that silently scores a NaN test risk as the mean risk and costs O(n·events) per Newton evaluation

### LINKED ISSUE #3608: survival/predict: KaplanMeier::at is a linear scan, so held-out IPCW Brier scoring costs O(n_test · censoring steps · grid) — and duplicates on_grid's binary search

### LINKED ISSUE #3609: survival/predict: survival_prediction_scores accepts a grid that does not start at 0, then overwrites the model's S(grid[0]) with 1 and scores an event before grid[0] with a negative cumulative hazard

SauersML:
Fixes #3608
Fixes #3609
Fixes #3606

- **#3608**: `KaplanMeier::at` is now the one binary-search lookup, and `on_grid` maps it over the grid. Held-out IPCW Brier scoring used to scan all censoring steps linearly for every subject at every grid point, for a cost of O(n_test · n_steps · |grid|). Each lookup is now O(log n_steps). `kaplan_meier_censoring_is_right_continuous_step` now also pins `on_grid == at` mapped over the grid, including NaN and ±∞.
- **#3609**: `survival_prediction_scores` now refuses (all-`None`) any grid whose first point is not `0`. The hazard-path scores integrate from the origin, where S = 1. A grid starting later had two effects: the repair pinned the model's `S(grid[0]) < 1` to 1, and an event before `grid[0]` was scored with a negative cumulative hazard. The pin `a_scoring_grid_that_does_not_start_at_the_origin_is_refused_3609` covers this, and the precondition is documented on `monotone_survival_and_hazard_scores`.
- **#3606**: deleted dead code: `risk_calibration.rs` (`cox_calibrated_survival_matrix`, `km_curve_on_grid`) and the pyffi `survival_matrix_from_risk_calibration` / `benchmark_survival_matrix_from_risk`, along with their registration and stubs (regenerated with `scripts/gen_rust_stub.py`; `--check` passes). Nothing called them, because the benches calibrate through lifelines. `survival_null_curve_from_train` now calls `KaplanMeier` directly. Both ledgers are updated, and `public_api_census.py` and `source_removal_guard.py` pass against base 6942363169.

---


SauersML:

---

SauersML:

New head: 1afd083ba4ccafe7519ec618c409bd0895e5276b. Only the append-only ledgers conflicted (`docs/public-api-census-changes.json`, `docs/source-removal-changes.json`). The resolution keeps every entry from main and then this PR's #3606 entry: census 372 -> 373, source-removal 173 -> 174, and both parse as JSON. I checked that main hasn't added a caller of any deleted symbol: `git grep -w` for km_curve_on_grid, cox_calibrated_survival_matrix, survival_matrix_from_risk_calibration and benchmark_survival_matrix_from_risk finds no .rs/.py/.pyi hits at the new head.

---

Issue #3551, opened 2026-09-20
gam-solve/gam-gpu: on a CUDA host the dense PIRLS Newton solve skips descent_curvature and the residual certificate, so an indefinite Hessian hard-errors instead of stepping (CPU/GPU behavior split)


SauersML:
## Defect

`solve_newton_direction_dense` (`crates/gam-solve/src/pirls/newton_solve.rs:613-690`, origin/main 1a3a498e3b) has two routes that solve **different systems**:

```rust
if gam_gpu::cuda_selected()?  {                       // true whenever a runtime resolves (Auto/Required)
    let solved = crate::gpu::pirls_gpu::cholesky_solve_only_gpu(hessian.view(), rhs.view())
        .map_err(EstimationError::InvalidInput)?;     // raw H, error propagates
    direction_out.assign(&solved.column(0));
    direction_out.mapv_inplace(|v| -v);
    if array_is_finite(direction_out) { return Ok(()); }   // no certificate
}                                                       // non-finite -> silently redo on CPU
let curvature = descent_curvature(hessian)?;            // Gill–Murray floor (#2814)
... StableSolver on `curvature` ...
if !rel.is_finite() || rel > 1.0e-3 { return Err(..certificate..) }
```

The CPU route does two things the CUDA route does not:
1. It replaces an indefinite `H` with its Gill–Murray modification (`descent_curvature`, :514-528). The doc there explains why: on an indefinite delayed-entry survival Hessian the saddle step made a heterogeneous-entry cohort crawl to its iteration cap (#2814).
2. It certifies `‖H·δ + g‖∞/(1+‖g‖∞)`.

On a CUDA host the raw `H` goes to `cholesky_solve_only_gpu`. On an indefinite `H`, the fp32 POTRF fails, and then the fp64 `cuda::cholesky_solve` fails too. The error is returned as `EstimationError::InvalidInput` through `?`. The same fit that steps on a CPU-only host therefore **fails** on a GPU host, and the #2814 fix is silently undone wherever CUDA is present.

Further defects in the same branch:
- **No size gate.** `cuda_selected()` (`crates/gam-gpu/src/mod.rs:215`) only asks whether a runtime resolved. Every dense Newton solve of every `p` (p = 3 included) goes to the device and creates a fresh `DnHandle`/stream, and the calibrated `potrf_min_p` admission (`linalg_dispatch.rs:333-337`, `route_through_gpu(DispatchOp::Potrf { p, batch: 1 })`) is bypassed.
[...]

PR #3647, opened 2026-09-20
pirls: certify the dense Newton direction by its backward error, not a fixed 1e-3

### LINKED ISSUE #3645: pirls: dense Newton-direction certificate is a fixed 1e-3 on a units-dependent residual ratio instead of the shared backward-error certificate
## Where

`crates/gam-solve/src/pirls/newton_solve.rs:658-692` (origin/main 81623168), `solve_newton_direction_dense`, CPU path:

```rust
let validation_residual = { /* ‖C·δ + g‖∞ */ };
let g_inf = gradient.iter().map(|v| v.abs()).fold(0.0_f64, f64::max);
let rel = validation_residual / (1.0 + g_inf);
if !rel.is_finite() || rel > 1.0e-3 {
    return Err(EstimationError::InvalidInput(format!(
        "PIRLS Newton direction failed its unperturbed linear-system certificate: relative residual {rel:.3e} exceeds 1e-3"
    )));
}
if array_is_finite(direction_out) { ...; return Ok(()); }
Err(EstimationError::LinearSystemSolveFailed(FaerLinalgError::FactorizationFailed {
    context: "PIRLS dense newton solve exhausted",
}))
```

Every dense unconstrained P-IRLS step goes through this function: `GamWorkingModel::solve_unconstrained_direction` (gam_working_model.rs:1726) and the bound-constrained fast path (newton_solve.rs:1969). A refusal makes the LM controller escalate its damping (reweight.rs:1197-1212).

## Why it is wrong

1. **`1e-3` is arbitrary.** Nothing ties it to the precision of the factorization that produced `δ`.

2. **The ratio depends on units.** `‖Cδ + g‖∞ / (1 + ‖g‖∞)` is an *absolute* residual when `‖g‖∞ ≪ 1` and a relative one when `‖g‖∞ ≫ 1`. Scaling the system as `C ↦ cC, g ↦ cg` (for example, rescaling the response or the prior weights) leaves `δ` and its accuracy unchanged, but the ratio changes to `c·r / (1 + c‖g‖)`. The same solve can pass at one scale and be refused at another.

3. **The ratio does not measure the solve.** For a backward-stable factorization, `‖Cδ + g‖ ≲ c·p·ε·‖C‖‖δ‖`. The ratio leaves `‖C‖‖δ‖` out of the denominator, so it cannot tell a bad solve from a correct solve of a large-norm system.
   - Near convergence (`‖g‖∞ ≈ 1e-8`), a residual of `1e-5` passes, although that is a thousand times the gradient being solved for.
   - A correct solve of a large-norm system is refused. The refusal then drives LM escalation on a system that was solved correctly.

4. **Dead code.** Once the residual certificate has passed, `direction_out` is finite. The trailing `FactorizationFailed { context: "PIRLS dense newton solve exhausted" }` branch can never be reached.

The repo already has the principled test: `gam_linalg::utils::certify_linear_system_residual`, the "shared certification boundary" for exact factorized solves. It is already used by estimate/optimizer.rs, identified_hessian.rs and smoothing_correction.rs.
- It computes the normwise backward error `‖r‖∞ / (p‖A‖max‖x‖∞ + ‖b‖∞)`.
- It compares that against the roundoff allowance `256·p·ε/(1 − 256·p·ε)`.
- It is invariant under `A ↦ cA` and `b ↦ ab`, and it rejects a non-finite solution or residual itself.

## Fix

- Certify `C·δ = −g` with `certify_linear_system_residual`, where `C` is the descent curvature the factor was built from.
- Map a refusal to the same LM-escalating error.
- Delete the unreachable trailing branch.

`descent_curvature` (same file) already returns a positive definite matrix: either the exact Hessian when strict Cholesky succeeds, or its Gill–Murray modification. The factor is therefore a Cholesky, whose backward error is well inside the allowance. A long certified step on a nearly singular `C` is not a defect of the solve; the damping and the gain ratio judge its length on the true objective, as the `descent_curvature` doc already says.

The CUDA branch of this function skips the certificate entirely. That is tracked separately in #3551 and is not touched here.

SauersML:
Fixes #3645

`solve_newton_direction_dense` (CPU path) used to refuse a direction whenever `‖Cδ + g‖∞ / (1 + ‖g‖∞) > 1e-3`. That test had three problems:
- **The constant is arbitrary.** Nothing ties `1e-3` to the precision of the factorization.
- **The ratio depends on units.** It is absolute below unit gradient and relative above it, so the same solve can pass at one response scale and be refused at another.
- **It cannot judge the solve.** It omits `‖C‖‖δ‖`, so it cannot tell a bad solve from a correct solve of a large-norm system.

The direction is now certified by `gam_linalg::utils::certify_linear_system_residual`, the shared certification boundary for exact factorized solves:
- It computes the normwise backward error `‖r‖∞ / (p‖C‖max‖δ‖∞ + ‖g‖∞)` and compares it with `256·p·ε/(1 − 256·p·ε)`.
- It is invariant under `C ↦ cC` and `g ↦ ag`.
- It rejects a non-finite solution or residual itself.

A refusal still surfaces as an error, so the LM controller still escalates its damping. The trailing `"PIRLS dense newton solve exhausted"` branch could not be reached once the certificate passed, and is deleted.

`C` is the output of `descent_curvature`, which is positive definite (the exact Hessian or its Gill–Murray modification). The factor is therefore a Cholesky, whose backward error sits well inside the allowance.

The CUDA branch is out of scope; it is tracked in #3551.

---


SauersML:

The original author (HN-13) has finished and handed this off, with its proof green (run 35499535165).

I reviewed the diff against current main:
- `certify_linear_system_residual(dimension, matrix_max_abs, rhs, solution, residual, label)` on main has the signature the PR calls.
- `array_is_finite` and `FaerLinalgError` are still used elsewhere in the file, so no import goes dead.
- The arbitrary `1e-3` relative-residual gate is replaced by the shared, scale-invariant backward-error certificate.

---

PR #3652, opened 2026-09-20
Propagate sparse REML refusals and enforce its geometry contract

### LINKED ISSUE #3636: reml: sparse exact-SPD bundle failure silently re-routes to dense spectral, swallowing inner-solve retreats and factorization failures
## Defect

In `prepare_eval_bundlewithkey_and_row_policy` (crates/gam-solve/src/reml/gradient_hessian.rs), when the measured structure routes to `RemlGeometry::SparseExactSpd`, the code handles the sparse builder's result like this:

```rust
match self.prepare_sparse_eval_bundlewithkey(rho, key.clone(), rows, decision.clone()) {
    Ok(bundle) => Ok(bundle),
    Err(err) => {
        log::debug!("[reml-geometry] sparse_exact_spd failed ({}); falling back to dense spectral", err);
        let fallback = SparseRemlDecision { reason: "sparse_exact_spd_assembly_failed", .. };
        self.prepare_dense_eval_bundlewithkey(rho, key, rows, fallback)
    }
}
```

Every error from the sparse builder is caught this way, and that builder has these error sources:

- **`execute_pirls_if_needed(rho, rows)?`**: the inner P-IRLS solve. Its trial-point retreats (`PirlsDidNotConverge`, ill-conditioning, separation, `TrialPointRefused`) are what `compute_outer_eval_with_order` turns into an infeasible trial point through `is_inner_solve_retreat`. The fallback throws the typed retreat away and calls `prepare_dense_eval_bundlewithkey`. That runs `execute_pirls_if_needed` again at the same rho, and failures are not cached, so this is a second full inner solve. The inner solve picks its own coordinate frame, so it is the same solve and fails the same way. The outer loop sees the same retreat only after paying for it twice.
- **`assemble_and_factor_sparse_penalized_system(...)?` / `factorize_sparse_spd`**: a non-SPD penalized Hessian at the P-IRLS mode. The dense route then returns a spectral, floored log|H| at this rho, while its sparse neighbours return an exact Cholesky log|H|. The outer objective is silently evaluated on a different surface at exactly the points where the solver reported a problem.
- **`factorize_simplicial` / `TakahashiInverse::compute`, the penalty pseudo-logdet, `hessian_surface_arrays`, `applied_canonical_penalties_for`**: all are genuine errors that become a debug line and a different backend.

SPEC: "In general, do not paper over solver issues." Also: "Fitting and inference must be fast on all scales ... seamless transitions between strategies." Switching backend mid-optimization because of an error is neither seamless nor visible.

Only one of these errors is a legitimate re-route: the P-IRLS solve came back in a frame other than `OriginalSparseNative`. P-IRLS routes independently (`should_use_sparse_native_pirls`, `SPARSE_NATIVE_MAX_H_DENSITY = 0.30`, plus the `gaussian_sufficient_statistics` value-only probe, which forces `DenseTransformed` when there is no sparse `XᵀWX` cache). The REML router uses 0.10 and `sparse_penalty_block_count > 0`, so the two can disagree. That case is currently signalled with `bail_invalid_estim!` and relies on the catch-all.

### Dead code in the same builder

`prepare_sparse_eval_bundlewithkey` adds a log-barrier diagonal when `self.linear_constraints` is `Some`, and logs and ignores its error (`"Sparse barrier Hessian diagonal skipped: {e}"`). `select_reml_geometry` returns `DenseSpectral` with `constraints_present` whenever `linear_constraints` or `coefficient_lower_bounds` is set, and this builder has a single caller that runs after that routing. So the block cannot run.

## Fix

- The sparse builder returns `Ok(None)` for the frame mismatch, which is the only legitimate re-route. The caller routes to dense with the honest reason `pirls_frame_not_sparse_native`. Every other error propagates unchanged, so retreats remain retreats and real failures are reported.
- Delete the unreachable sparse barrier block.

SauersML:
Sparse REML evaluation used to catch any sparse error and build a dense bundle, potentially erasing an invalid inner curvature surface. It now propagates those errors. The only non-error rerouting is an inner solve that returned transformed coordinates, which is recorded as `pirls_frame_not_sparse_native`.

Manual review also removed the same error eraser from REML's symbolic-structure routing. Constraint routing already excludes the sparse backend; the sparse builder now checks that contract explicitly, allowing removal of its unreachable barrier block while preserving dense constraint curvature.

Validation on MSI with two assigned CPUs and coherent warm dependencies:
- Complete production gam-solve compiled as a focused executable; four new regressions passed.
- Independent diagonal Gaussian formulas check coefficients and log determinant, with a dense-Hessian control.
- A NaN curvature row injected into a certified cached mode preserves its concrete `PirlsRowGeometryUnrepresentable` error, even though its cached positive Hessian would permit the former dense replacement.
- Both lower-bound and linear constraints route dense and refuse direct sparse misuse.
- Coordinate-only rerouting reuses the cached mode without another inner solve.
- Complete gam-solve unit-source metadata compiled with warnings denied, including all existing tests and the new module.

PIRLS's separate symbolic-routing API was not changed here. No full outer-optimization benchmark was run.

Closes #3636.


SauersML:
Ready for the local merge train (MT-0, even PR). The only thing left is the build.

What I checked against main d5f641efbd:
- **Merge:** the merge with main is clean (merge-tree has no conflicts).
- **Source-removal guard:** passes, 186 -> 186 items with 0 guarded removals.
- **Log-barrier block:** deleting it is correct. `inner_strategy.rs:73` and `newton_solve.rs:1704` route every fit with linear constraints or finite lower bounds to `constraints_present` (dense) before the sparse builder runs.
- **Shared helpers:** `sparse_exact_beta_original` and `barrier_config_from_constraints` both still have other callers, so neither becomes dead code.
- **Error handling:** `None` is only the P-IRLS-frame decline. Every other error now propagates, which matches SPEC ("do not paper over solver issues").

Pending: focused proof run 35499569852 at head 13eb0fa789, currently in progress. Merge once it is green.

---

PR #3655, opened 2026-09-20
gam-sae: freeze the DecoderIncoherence routing coactivation per assembly

### LINKED ISSUE #3515: gam-sae manifold/penalties: DecoderIncoherence coactivation W is read live from the routing in the value but held constant in the gradient (#1625-class value/gradient desync); new_sparse errors swallowed by .ok()
## Where (origin/main)

`crates/gam-sae/src/manifold/penalties.rs:567-643`, `SaeManifoldTerm::live_decoder_incoherence_penalty`:

```rust
let gates = self.assignment.assignments();
...
*num.entry((lo, hi)).or_insert(0.0) += gj * gk;
...
let w = s * inv_n;
...
DecoderIncoherencePenalty::new_sparse(PsiSlice { range: 0..m_total * p, latent_dim: Some(m_total) },
    block_sizes, p, pairs, base.weight, base.learnable_weight)
.ok()          // line 637
.map(|mut per_fit| { ... })
```

The pair weights `W_jk = (1/n) Σ_rows gate_j·gate_k` are rebuilt on every call from the **current** routing `self.assignment.assignments()`. Those gates are functions of the logits, which are inner-solve unknowns.

The function has three consumers:

- **Value** (line-search / inner objective): `crates/gam-sae/src/manifold/construction.rs:5287-5289`, inside `analytic_penalty_value_total`:
  ```rust
  if let Some(per_fit) = self.live_decoder_incoherence_penalty(base) {
      value += penalty_scale * per_fit.value(beta.view(), rho_local);
  ```
- **Gradient / curvature**: `crates/gam-sae/src/manifold/penalties.rs:2958-2981`, inside `add_sae_beta_penalty`. It adds only `per_fit.grad_target(target_beta, ...)` into `sys.gb` and the PSD majorizer into `sys.hbb`.
- **Factored curvature**: `crates/gam-sae/src/manifold/construction_arrow_schur_assembly.rs:2737-2738`.

## Why it is wrong

The value is `P(β, z) = Σ_{j<k} W_jk(a(z)) · o_jk(β)`, a function of both β and the logits z. The assembled system contains only ∂P/∂β. No `∂W/∂z` term is added to the logit block of the gradient, and no cross term is added to the curvature.

During the line search each trial point moves the logits. That changes `self.assignment` and therefore W. The value the line search compares is therefore a different function from the one whose gradient produced the step:

- The inner Newton direction is not a descent direction for the value being accepted.
- At the returned point the logit-block KKT residual of the *actual* objective is nonzero, by `Σ o_jk ∂W_jk/∂z`.

This is exactly the defect that #1625 fixed for the barrier coactivation gate and the decoder repulsion gate. For those, `refresh_barrier_coactivation_gate` (penalties.rs:786) and `refresh_decoder_repulsion_gate` (penalties.rs:660) **freeze** the routing-derived weights once per assembly, and `declare_collapse_prevention_gates` (penalties.rs:544) makes value, gradient and curvature read the same frozen weights. The user-requested DecoderIncoherence penalty was left on the live path.

This penalty is reachable by users:

- `crates/gam-pyffi/src/latent/latent_basis_and_sae_ffi_tail.rs:336` builds it when `decoder_incoherence_weight > 0.0 && k_atoms >= 2`.
- `crates/gam-models/src/fit_orchestration/descriptors.rs:855` pushes `AnalyticPenaltyKind::DecoderIncoherence`.

This conflicts with SPEC.md: "A fit object must only ever come from a converged optimization." The stationarity certificate is computed against a gradient that omits a term of the objective being minimized. The rejected-step and stall behaviour this produces is solver trouble that gets papered over ("In general, do not paper over solver issues.").

### Secondary: silent error swallowing

`.ok()` at penalties.rs:637 (and the same pattern in the repulsion builder at penalties.rs:853) turns any `new_sparse` validation failure into `None`. The failures are listed in `crates/gam-terms/src/analytic_penalties/orthogonality.rs:716`: a non-finite or negative pair weight, which non-finite gates would produce, a span mismatch, a zero block size, and so on.

- In `add_sae_beta_penalty` the `None` becomes `return false`.
- In the value path it adds 0.

So a user-requested penalty silently disappears from the fit instead of surfacing the invalid state.

## Principled fix

1. Freeze the incoherence coactivation per assembly, at the same chokepoint and with the same discipline as `refresh_barrier_coactivation_gate`:
   - Store the frozen sparse `pairs` (or the built `DecoderIncoherencePenalty`) on `SaeManifoldTerm`, refreshed at assembly entry.
   - Have `live_decoder_incoherence_penalty` (renamed to reflect that it is frozen) return the frozen operator.
   - Include it in `CollapsePreventionGates` / `declare_collapse_prevention_gates` and in the streaming frozen-gate path, so the value, gradient, curvature and streamed paths all see one W per assembly.

   This is lagged diffusivity: at a fixed point of the outer refresh, W is consistent with the converged routing.

   The alternative is to keep W live and assemble the exact logit gradient `Σ_{j<k} o_jk · ∂W_jk/∂z_row` together with its PSD curvature. That is more expensive, and it is inconsistent with how the sibling gates are handled.

2. Replace `.ok()` with error propagation (`Result<Option<_>, String>`, where `None` is only for `K < 2` or an empty pair set). Callers then fail loudly on invalid gates instead of dropping the penalty.

SauersML:
Fixes #3515

## Defect
The user `DecoderIncoherence` penalty is `P(β, z) = Σ_{j<k} W_jk(a(z))·o_jk(β)` with `W_jk = (1/n)·Σ_i a_ij·a_ik`. Its β-gradient and PSD β-curvature are assembled with `W` read as a constant. But `analytic_penalty_value_total` (the line-search value) rebuilt `W` from the trial logits on every call, through `live_decoder_incoherence_penalty`. The value therefore priced a different objective than the gradient descended. This is the #1625 value/gradient desync that the sibling routing gates (decoder repulsion, barrier coactivation, amplitude barrier) were each frozen against. `live_decoder_incoherence_penalty` also ended in `.ok()`, so any `new_sparse` build error silently dropped the user's penalty.

## Fix
- **New gate on the term.** `SaeManifoldTerm::decoder_incoherence_gate` holds the sparse frozen `W_jk`. It is refreshed at the same places as the other gates:
  - the `assemble_arrow_schur_core` chokepoint
  - the joint-fit bank install in `run_joint_fit_arrow_schur_with_termination_policy`
  - the criterion-scope `freeze_collapse_prevention_gates`, which now takes the registry. Its doc requires the list to be exhaustive against the assembler block.

  It is refreshed only when the registry carries a `DecoderIncoherence` penalty, so no `O(N·K)` scan is added otherwise.
- **Carried wherever the other gates are carried:**
  - `CollapsePreventionGates` (declared once per outer solve, #2933 F05)
  - streaming chunk carry (#1801)
  - clones of frozen terms
  - exact-Hessian FD endpoints
  - reset on structure harvest, where `K` changes
- **Errors are no longer swallowed.** `decoder_incoherence_penalty` returns `Result<Option<_>, String>` and consumers propagate it as `ArrowSchurError`. To make that possible, `add_sae_beta_penalty` and `add_factored_beta_penalty_curvature_for_penalty` now return `Result`. `Ok(None)` remains only for `K < 2`, where no pair exists.
- **Standalone fallback.** With no refresh (the gate is `None`), the live coactivation is read. That call evaluates value and gradient at a single state, so it is self-consistent.

Out of scope: `live_decoder_repulsion_penalty` has a similar `.ok()`, but it has many callers. It is left for a separate change.

## Validation
- PR CI
- focused-rust-proof on `gam-sae` lib, covering the declared-gates and collapse-prevention tests

---


SauersML:

---

SauersML:

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 0d5a910d4a85.

PR #3662, opened 2026-09-20
manifold/curl_census: e-BH over the whole candidate family; derive the harvest's permutation budget

### LINKED ISSUE #3658: manifold/curl_census: e-BH family shrinks to the adjudicated pairs, and the harvest overrides the derived permutation budget with a hard-coded 4096
Two defects in how the curl/shattered-circle census calibrates its e-BH ledger. Both are in the census used by `structure_harvest::curl_candidates`. That path is currently gated off in `fit_entry` (`curl: None`, fit_entry.rs:1280), so this matters before graduation rather than on today's default fit. The census also backs `examples/curl_census_foreign.rs`.

## 1. The e-BH family is the survivors, not the search

`crates/gam-sae/src/manifold/curl_census.rs` (origin/main, `census_shattered_circles`):

- `candidate_pairs = cooccurrence_pairs_sparse(...)` is the searched family.
- `.filter_map` drops any candidate whose plane cannot be adjudicated: `orthonormal_pair_coords(...).ok()?` for zero or collinear directions, and `curl_verdict(...).ok()?` for `radius_law` errors such as `m₂ = 0`.
- The ledger then runs on `out` alone:
  ```rust
  let e_values: Vec<f64> = out.iter().map(|p| p.e_value).collect();
  let rejected = ebh_reject(&e_values, cfg.fdr_alpha);
  let m = e_values.len() as f64;
  ```

That contradicts the file's own comment ten lines earlier: "A refused plane carries `e = 0` into the ledger, which is a valid e-value and keeps the multiplicity burden of the WHOLE search on the books rather than quietly shrinking the family". It also contradicts the `fdr_discovery` doc ("e-BH discovery ... over the WHOLE screened family"). The derived budget, meanwhile, uses `candidate_pairs.len()`.

Dropping candidates lowers every e-BH threshold `m/(α·k)` to the survivor count. A plane at the maximum e-value `B+1 = 20` is then rejected in a family of 1 even though the search generated 2 candidates, where rank 1 needs 40. The same wrong `m` feeds the `ebh_threshold` and `replicates_required` diagnostics.

**Fix:** run e-BH over `candidate_pairs.len()` hypotheses, padding the un-adjudicated ones with the valid e-value 0. A 0 can never be rejected.

## 2. The harvest hard-codes the permutation budget the census says must be derived

`crates/gam-sae/src/structure_harvest.rs`, `impl Default for CurlConfig`, sets `null_replicates: 4096`, and this is passed straight into `CurlCensusConfig::null_replicates`. The census documents that field as "`0` to DERIVE the only defensible value (which is what every caller should do; the override exists so two statistics can be compared at a fixed budget)". It gives two sharp sides:

- **Too small.** With `B+1 < m/α` no rank is reachable. At α = 0.05 and B = 4096, any round with more than 204 candidate pairs can never reject anything, and prints the same zero as "no structure".
- **Too large.** The indicator pays only when no surrogate reaches the observation. A plane with true p = 1e-3 scores with probability `(1 − 1e-3)^4096 ≈ 1.7%` at B = 4096, versus ≈ 98% at the derived B = 19 for m = 1.

`CurlConfig`'s own doc says the budget "is set by the size of the search, not by taste", yet the value is a fixed constant. SPEC: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code."

**Fix:** remove `CurlConfig::null_replicates`. Only `CurlConfig::default()` is ever constructed. The harvest then passes `0`, so the census derives `B+1 = m/α` from the round's family. The fixed-budget override stays on `CurlCensusConfig` for the example and for the research comparison it documents.

SauersML:
Fixes #3658

- **`census_shattered_circles`**: e-BH now runs over `candidate_pairs.len()` hypotheses. Candidates whose plane could not be adjudicated (no orthonormal frame, or no radius law) carry the valid e-value 0 instead of dropping out of `m`. This is what the file's own comment already promised. The ledger is a pure `ebh_ledger(adjudicated, family, alpha)` helper, so `ebh_threshold` and `replicates_required` use the same `m`.
- **`CurlConfig`**: removes the hard-coded `null_replicates: 4096`. `curl_candidates` passes `0`, so the census derives `B + 1 = m/α` from the round's family, as `CurlCensusConfig::null_replicates` documents every caller should do.
- **Test**: `unadjudicated_candidates_keep_their_place_in_the_ebh_family`. An e-value of 20 is a discovery in a family of 1, but not in a family of 2 where one candidate was un-adjudicated.

The curl harvest is still gated off in `fit_entry` (`curl: None`), so default fits are unchanged.

---


SauersML:

---

PR #3664, opened 2026-09-20
Carry declared penalty nullities through narrowing pullbacks via a declared null basis (#3023)

### LINKED ISSUE #3023: Declare penalty null bases so the rank survives the identifiability canonicaliser (every custom-family rank is a threshold today)

SauersML:
Fixes #3023

Follow-up to #3362 / 9cdf583e, which carried declared nullities only through identity and square pullbacks. This PR adds the null-basis API and the missing declarations.

## Math

The canonicaliser reparameterises a block as `β = Tθ`, where `T` has orthonormal columns: either a column selection `I[:,K]` or an orthonormal `V_b`. The reduced penalty is `TᵀST`. For PSD `S`, `uᵀTᵀSTu = ‖S^{1/2}Tu‖²`, and `T` is injective, so

```
nullity(TᵀST) = dim(ker S ∩ range T) = m − rank((I − TTᵀ) N)
```

where `N` (`p × m`) is an orthonormal null basis of `S`.

**Basis from the declaration.** `N̂` is the set of eigenvectors of the stored matrix's `m` smallest eigenvalues. Let `E = p·ε·‖S‖ + formation_band`. Define the residual `ρ = max_{i<m}|λ̂_i| + E`, which bounds `‖SN̂‖`, and the separation `δ = λ̂_m − E − ρ`.
- Davis–Kahan (residual form) gives `sinΘ ≤ ρ/δ`, and `min_Q‖N̂ − N₀Q‖ ≤ √2·sinΘ + p·ε`. The last term is the eigensolver's orthogonality loss.
- If `δ ≤ 0`, the declared space is not separated. Then no basis is taken, and the reduced declaration is 0. See *Why a declaration of 0 is certified* below.

**Rank of the residual.** Form `R = N̂ − T(TᵀN̂)` and count its singular values above
`b = basis error + γ_{p+r+1}(r+1)√m (formation of R) + d(1+d) (T's measured orthonormality defect, d = ‖TᵀT − I‖_F + γ_p r) + max(p,m)·ε·σ_max (SVD band)`.

**Why an overstated declaration can only cost unresolved curvature.** A declared nullity only caps the rank, so overstating it is the dangerous direction.
- A column counted null has an exact out-of-range component of at most `2b`.
- Its reduced direction `u = Tᵀn` therefore has `‖u‖² ≥ 1 − 4b²` and curvature `(Tu)ᵀS(Tu) ≤ 4b²‖S‖`.
- By Courant–Fischer, the cap removes only reduced eigenvalues `≤ 4b²‖S‖/(1 − 4b²)`.
- So the pulled-back declaration is used only when that bound is at most the stored penalty's resolution `p·ε·‖S‖ + formation_band`. Otherwise it is 0, which is exactly the spectral rule.
- Underestimating the band can only understate the nullity, which is the safe direction.

**Square `T`.** A square `T` is a congruence, so it carries the declaration unchanged.

**Block-local penalties.** A `Blockwise` penalty declares the nullity of its `local` matrix. Its block-width null space is the embedded local null space plus the exact axes outside `col_range`, with nullity `m_local + (p − w)`. Before this PR, the local-convention number was copied onto the Dense full-width reduced penalty, so it was understated.

**Kronecker penalties (items 2–3).** For symmetric PSD `A` (`p_a × p_a`) and `B`, the spectrum of `A ⊗ B` is `{α_i β_j}`, so `nullity(A ⊗ B) = p_a·p_b − rank(A)·rank(B)`.
- Survival location-scale time-varying covariate penalties are `S_global ⊗ Ḡ_t`. `Ḡ_t` is the integral Gram of linearly independent B-splines, so it is PD. That gives `null = null(S) ⊗ ℝ^{p_t}`, with nullity `(m_local + p_cov − w)·p_t`, taken exactly from the covariate design's declaration.
- Margin penalties `G_x ⊗ S_t` declare `p_x·p_t − rank(G_x)·rank(S_t)` from the factors' own spectra. This is the same rule the follow-up-varying marginal slope uses (`kronecker_nullspace_dimension`).
- The CTN tensor penalties `G_y ⊗ S_x`, `S_y ⊗ G_x` and `P_shape ⊗ I` use the same factor rule. Each small factor is resolved on its own, rather than resolving the assembled product, whose small eigenvalues are products of two rounding errors.

**Dropping leading columns (survival).** `drop_leading_penalty_columns` retains `β = Tθ` with `T = I[:, fixed..]`. It used to clear every declaration whenever a block was cut or a penalty was Dense, Kronecker or Labeled. Its Diagonal arm also kept the full-width number unchanged, which can overstate the nullity when an axis is dropped.
- It now carries each declaration through `pulled_back_declared_nullity(S, m, T)`.
- A block-local penalty whose block survives intact keeps its local declaration exactly.

**Binomial mean-wiggle.** The pilot, baseline, mean and frozen-warp blocks now carry their designs' declarations. The frozen warp densifies each selected-basis penalty to full width, so a `Block` declaration `m` becomes `m + (p − w)`.

## Why a declaration of 0 is certified, not a fallback

Reviewed adversarially against SPEC.md ("do not paper over solver issues"; "hard-coded knobs and magic constants should be avoided"). Two branches can return a reduced declaration of 0:

1. **`δ ≤ 0`, in `from_dense`.** The Davis–Kahan residual `ρ = max_{i<m}|λ̂_i| + E` is compared against the stored matrix's own gap `λ̂_m − E`.
   - When the residual meets or exceeds the gap, `sin Θ` is unbounded (it may be 1).
   - The stored matrix then does not determine which `m`-dimensional subspace is `ker S`.
   - So no positive reduced nullity can be certified. 0 is the largest value the data certifies.
2. **`4b²‖S‖ > resolution·(1 − 4b²)`, in `from_dense` and `pulled_back_nullity`.** A declaration counted within the error band `b` could cut a reduced eigenvalue that the stored penalty does resolve. So again 0 is the largest certified value.

Neither branch is a reaction to a failure, and neither retries, perturbs or ridges anything. The decision is taken from the certificate before any rank is used.
- Every term of the certificate is a quantity of the problem: `S`'s spectrum, its rounding band `p·ε·‖S‖`, the producer's formation band, the Gram/SVD rounding factors `γ_k`, and `T`'s measured orthonormality defect `‖TᵀT − I‖`.
- The constants `4` (`(2b)²`) and `√2` (sinΘ to basis distance) come from the bounds themselves. There is no tolerance knob.
- A declaration of 0 means "no cap". The consumer (`structural_penalty_root`) then ranks the reduced penalty by its own spectral rule, exactly as it ranks every undeclared penalty.

**What 0 costs.** It cannot overstate the nullity, which is the dangerous direction. It can understate it: a reduced null direction stored above the Weyl band stays in the root, which is the pre-#3023 behaviour for that penalty. The unreduced penalty keeps its own declaration.

**Correction.** An earlier revision of the module doc said the `δ ≤ 0` case "loses nothing" because the cap would not bind. That is false when `E < λ̂_m ≤ 2E + max_{i<m}|λ̂_i|`. Commit 4f114c58 corrects the doc to state the actual cost above. It is doc-only.

**Possible refinement (not done here).** Widen the null cluster to the first certified gap, `m + k` vectors. That would give the certified lower bound `max(0, c − k)`, where `c` is the cluster's pulled-back nullity. This would recover some of the cap in the unseparated case.

## Changes
- `gam-problem::penalty_null_basis`: adds `PenaltyNullBasis::{from_declared_nullity, pulled_back_nullity}` and `pulled_back_declared_nullity`. Unit tests cover:
  - selections that keep or remove null directions
  - orthonormal narrowing
  - Blockwise at block width
  - formation noise
  - refusing an overstated declaration
  - the basis spanning the exact null space
- `gam-identifiability::canonical`: both pullback sites now carry `pulled_back_nullspace_dims(spec, T)` in place of `Vec::new()`.
- `gam-models`:
  - `gamlss/builders.rs`: 6 `vec[]` sites now declare.
  - `survival/location_scale`: `TimeDependentCovariateBlockInput.nullspace_dims`, the time-varying declarations and validation, and the drop-leading pullback.
  - `transformation_normal::block_spec`: declares the Kronecker factor nullities.

Every consumer reads a declaration as a cap, `rank = min(resolved, p − m)`, so each new declaration can only remove directions the structure proves null. Where a declaration cannot be established, it is 0, which is the spectral rule.


SauersML:
Ready for the local merge train (MT-0, even PR). The head 4f114c58 is unchanged and GitHub reports it mergeable. It is waiting only on its four queued focused-rust-proof runs: 35501681475 (gam-problem lib), 35501682407 (gam-identifiability canonical), 35501683562 (canonical_recovery) and 35505377317 (gam-models lib, re-dispatched after a 1200 s compile-limit timeout, which was infrastructure, not code). Merge once those are green on 4f114c58, then close #3023 with the PR's data.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, cef6f000be30.

PR #3668, opened 2026-09-20
survival: competing-risks CIF monotonicity check uses per-entry rounding bands, not 1e-10 * global max

### LINKED ISSUE #3529: survival competing-risks CIF: monotonicity check uses a magic 1e-10 scaled by the GLOBAL max hazard, so one high-hazard row lets real decreases in every other row be silently clamped to zero
Found by audit agent HN-5 (area: `crates/gam-models/src/survival`, a-m). The code is on origin/main `93285a585b`.

## Code

`crates/gam-models/src/survival/base.rs:3522-3564`, in `assemble_competing_risks_cif_from_endpoints`:

```rust
let max_abs_hazard = cumulative_hazards
    .iter()
    .flat_map(|endpoint_hazard| endpoint_hazard.iter())
    .fold(0.0_f64, |acc, value| acc.max(value.abs()));
let monotone_tolerance = 1.0e-10_f64 * max_abs_hazard.max(1.0);
...
let raw_increment = current - previous_cumulative[endpoint];
if raw_increment < -monotone_tolerance {
    return Err(SurvivalError::NonMonotoneCumulativeHazard);
}
let increment = raw_increment.max(0.0);
```

## Why it is wrong

1. **The constant is arbitrary.** `1e-10` is not derived from anything the inputs carry, and `max(1.0)` adds a second, absolute floor. SPEC: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code."

2. **The scale is global.** The tolerance is `1e-10 * max|H|` over every row, every cause and every time. The rows, however, are independent subjects, and the recurrence runs per row. Consider a prediction batch in which one subject has `H ≈ 1e3` while another has `H ≈ 1e-4`:
   - The second subject gets an allowed decrease of `1e-7`. That is 0.1% of its own cumulative hazard, orders of magnitude above any rounding of its values.
   - Its decrease is not reported. It is silently turned into `increment = 0` (`raw_increment.max(0.0)`), so a non-monotone cumulative-hazard curve, which is a model defect, produces a plausible-looking CIF.

   Whether a given row is checked therefore depends on which other rows happen to be in the same batch. SPEC: "In general, do not paper over solver issues."

3. **The `max(1.0)` floor makes the check absolute for small hazards.** A curve whose values are all `< 1` gets a fixed allowed decrease of `1e-10`. That exceeds the representable rounding of `H ~ 1e-8` by about seven orders of magnitude.

Production callers:
- `survival/predict.rs:3678` and `:3702`, which run the Aalen-Johansen assembly for every competing-risks prediction.
- `crates/gam-pyffi/src/model/model_ffi.rs:2267`.

## Principled fix

A rounding-level decrease between two evaluations of `H = exp(eta)` has a bound that follows from the values themselves. Evaluating `eta = x(t)·beta + offset` has a forward error of about `gamma_p * sum_j |x_j beta_j|`, which is `accumulation_growth` from `gam_linalg::roundoff`, the same device used in #2469. A relative error `delta` in `eta` becomes a relative error `delta` in `H`.

So the allowed decrease between `H_{j-1}` and `H_j` for one row and one cause is:

```
delta_{j-1}·|H_{j-1}| + delta_j·|H_j|
```

This is computed per entry, never from a batch-wide maximum and never with an absolute floor.

That means the producer (`evaluate_rp_row_with_beta` / the refined-grid loop in `predict.rs`) should pass its certified per-entry rounding band together with the cumulative hazards. The assembler would then:
- refuse any decrease larger than that band;
- clamp only decreases inside it.

The simpler alternative is to refuse every strict decrease, which is exact. Monotone RP fits are constrained monotone, so the only decreases they can produce are rounding-level.

Test to pin the behavior: a two-row batch in which row 0 has `H ~ 1e3` and row 1 has a genuine relative decrease of `1e-5` at `H ~ 1e-3`. The assembly must refuse row 1. On main it currently returns a CIF.

SauersML:
Fixes #3529

## Problem
`assemble_competing_risks_cif_from_endpoints` allowed any decrease up to `1e-10 * max(max|H|, 1)`. That is a magic constant, scaled by the largest cumulative hazard over the whole batch. So one subject with H ~ 1e3 let a real 1e-8 decrease in another subject's H ~ 1e-3 be clamped to a zero increment.

## Change
- **`assemble_competing_risks_cif_from_endpoints`** handles arrays with no band: the Python `competing_risks_cif_from_predictions` path and direct callers. It is now **exact**: a negative value or any strict decrease is refused, and plateaus pass.
- **New `assemble_competing_risks_cif_from_endpoints_with_rounding(times, H, bands)`** takes one band per entry.
  - It refuses `H_j − H_m < −(band_j + band_m)`, or `H_j < −band_j`. Here `H_m` is the running maximum the recurrence carries forward and `band_m` is the band of the entry that set it.
  - Only decreases inside the band are clamped.
  - The band is per entry, so whether a row is accepted no longer depends on the rest of the batch.
- **Producer:** `evaluate_rp_row_with_beta` also returns the band of its `H = exp(η)`. The band is `Ĥ·(expm1(δ) + ε)/(1 − ε)`, with `δ = γ_{p+2}·(Σ_j|x_jβ_j| + |η_time_offset| + |primary_offset|)`.
  - `δ` is the forward error of the `p+2`-term sum for η, with γ taken from `gam_linalg::roundoff::accumulation_growth`.
  - The `ε` term covers a faithfully rounded exp.

  The competing-risks prediction carries this band for every refined-grid entry into the banded assembly, on both the per-row and the shared-grid paths.

## Tests (gam-models, `survival::base::tests`)
- `competing_risks_cif_refuses_a_small_row_decrease_beside_a_large_row_3529`: the issue's two-row batch is refused both with and without bands. On main it returned a CIF.
- `competing_risks_cif_clamps_a_decrease_inside_its_rounding_band_3529` checks four cases:
  - a decrease inside the band is clamped, and closure ΣF + S = 1 holds exactly;
  - the same values with no band are refused;
  - a decrease just past the band is refused;
  - a negative band is refused.
- The existing tests are unchanged: `competing_risks_cif_rejects_nonmonotone_hazards`, the plateau and three-cause test, and the bad-grid and extreme-hazard tests.

---


SauersML:
Taken over by RF-1113-9; driving to merge.

---

SauersML:

---

SauersML:
Withdrawn: this branch received a push at 11:20Z from another session, so W8-75 is standing down on it.

---

PR #3674, opened 2026-09-20
survival marginal-slope intercept: key the warm start on the whole equation, drop the reactive retry

### LINKED ISSUE #3671: survival marginal-slope intercept: the warm-start key omits q, the slope, the frailty scale and the law, so a cached seed is the root of another equation, and a reactive retry from the closed form papers over the failed solves
## Where (origin/main)

`crates/gam-models/src/survival/marginal_slope/intercept.rs:74-114`, in `solve_row_survival_intercept_with_slot`:

```rust
let beta_tag = hash_intercept_warm_start_key(beta_h, beta_w);
let cached_a = slot.and_then(|(row, kind)| { ... cache.load(row, kind, beta_tag) });
let a_init = cached_a.unwrap_or(a_closed_form);
let mut solve_result = monotone_root::solve_monotone_root_detailed(eval, a_init, ...);
// If the warm-started solve failed, retry once from the closed-form
// seed. Cached `a` from a prior PIRLS iter can be far enough from the
// current root (e.g., after a large β step) that the bracketing search
// exhausts; the closed-form seed always sits in the correct basin.
if cached_a.is_some() && solve_result.is_err() {
    solve_result = monotone_root::solve_monotone_root_detailed(eval, a_closed_form, ...);
}
```

`crates/gam-models/src/survival/marginal_slope/family.rs:496-510`: `hash_intercept_warm_start_key(beta_h, beta_w)` hashes only the two deviation coefficient vectors.

## Why it is wrong

The calibration equation that the slot solves is `T(a; q, slope, s, law, β_h, β_w) = Φ(∓q)`. The key covers only `β_h` and `β_w`, so it is missing:
- `q`, which comes from the time and marginal blocks;
- the slope, which comes from the slope block;
- the probit frailty scale `s`;
- the row's finite latent law.

So when the time, marginal or slope coefficients move while the deviation coefficients stay fixed, the key still matches. The seed that comes back is then the certified root of a *different* equation.

On the deep tails that the log-residual solve (gam#2971) now handles, that seed can sit many units from the new root. The Newton probes from it reach points where `T` underflows. `eval` returns an error there, and `find_root_monotone` aborts. The code then does exactly what SPEC forbids: it detects that the solve failed and retries from another seed.

> "In general, do not paper over solver issues."

The code comment says so directly: the retry is there because a stale seed "can be far enough from the current root ... that the bracketing search exhausts".

Two smaller defects are in the same cache (`family.rs:425-460`):
- **The seqlock is missing its fences.** The reader re-checks the tag with a plain `Acquire` load *after* the `Relaxed` value load. The writer invalidates the tag with a `Release` store *before* the `Relaxed` value store. Neither ordering stops the value access from moving across the other tag access. The standard seqlock needs `fence(Acquire)` between the value load and the second tag load, and `fence(Release)` between the invalidating store and the value store.
- **The doc comments describe the old behaviour.** They say the key is "a hash of the joint coefficient vector" and that a miss "falls back to" the closed-form seed.

## Principled fix

1. **Key every input of the equation.** The key becomes `q`, the slope, the probit scale, the law (its node count, nodes and log weights, or a Gaussian marker) and `β_h`, `β_w`. A hit is then, by construction, the certified root of the same equation, so the solve starts on it and returns at once. A miss seeds from the rigid closed form.
2. **Delete the retry.** Once the key is complete, no seed can be the root of another equation, so there is nothing for the retry to catch.
3. **Add the two seqlock fences** and bring the doc comments in line with the code.

The trade-off: warm starts now hit only on repeated solves of the identical equation. Examples are the value, gradient and Hessian passes at one β, and repeated trust-region evaluations. They no longer hit across changes in other blocks. That is the correct trade: the hits that were removed are the ones that produced the failures the retry was covering.

The test `survival_intercept_root_does_not_follow_its_warm_seed_2971` changes too:
- It keys the slot exactly as production does.
- It asserts that the entry and exit equations of a row, which share `β` and the row, no longer share a key.

(The BMS `hash_intercept_warm_start_key_flex` at `bms/family.rs:478` has a similar shape. It belongs to the BMS area and is not changed here.)

SauersML:
Fixes #3671

The per-row intercept warm-start cache keyed only on `(β_h, β_w)`. The equation it solves is `T(a; q, slope, s, law, β_h, β_w) = Φ(∓q)`, so a hit could return the root of a different equation whenever the time, marginal or slope blocks moved. Newton probes from that seed left the tail's finite range and the solve failed, and a reactive retry from the closed form covered for it. SPEC: "In general, do not paper over solver issues."

- **`family.rs`:** `hash_intercept_warm_start_key(q, slope, probit_scale, law, beta_h, beta_w)` now hashes every input of the equation. For the law it hashes the node count, the nodes and the log weights, or a Gaussian marker.
- **`family.rs`, seqlock:** the cache's sequence lock gets its missing fences: `fence(Acquire)` before the re-read of the tag, and `fence(Release)` after the invalidating tag store. The doc comments are corrected.
- **`intercept.rs`:** keys with the full equation and deletes the `if cached_a.is_some() && solve_result.is_err()` retry. A hit is now the certified root of the same equation; a miss seeds from the rigid closed form.
- **`tests.rs`:** the gam#2971 warm-seed test keys its slot exactly as production does. It also asserts that a row's entry and exit equations, which share β and the row, no longer share a key.

Trade-off: warm starts now hit only on repeated solves of the identical equation, for example the value, gradient and Hessian passes at one β. The hits that were dropped are the ones that seeded from other equations' roots.

---


SauersML:

---

SauersML:
Withdrawn: RF-0748-3 took this PR over at 10:39Z and merged main into it (25abf4c5f2). I am not taking it over; RF-0748-3 owns it.

---

PR #3682, opened 2026-09-20
gam-geometry: derive both curvature rail flags from the exact KKT condition (#3629)

### LINKED ISSUE #3629: gam-geometry: response curvature hyperbolic rail flag fires on a magic `κ̂ ≤ 0.99·κ_min` band while the spherical rail is exact-KKT only
`fit_response_curvature` (`crates/gam-geometry/src/response_geometry.rs:1838`) sets the hyperbolic rail flag with a hard-coded proximity band:

```rust
let railed_at_hyperbolic_resolution_limit =
    railed_at_hyperbolic_bound || kappa_hat <= 0.99 * kappa_min;
```

The spherical twin, a few lines below, deliberately does the opposite:

> // The upper rail flag comes only from the exact active-bound KKT condition
> // `V'(κ_max) ≤ 0`; proximity to a bound is not treated as convergence.

**Why this is a defect**
- `κ_min = −(1−√ε)/max‖z‖²`, so `κ̂ ≤ 0.99·κ_min` means `1 + κ̂·max‖z‖² ≲ 0.01`. The flag fires when the farthest point's conformal factor is within 1% of vanishing. Nothing derives the 1%: 0.98 or 0.999 would be just as defensible.
- The flag decides whether the caller must say "curvature exceeds the chart-resolvable hyperbolic range" or may quote a confident hyperbolic verdict. That decision is exported through `geometry_ffi.rs:1256` and `gamfit/_response_geometry.py:284`. An interior κ̂ at `0.989·κ_min` is quoted as resolved, while one at `0.990·κ_min` is refused as railed. The two sides of the same chart also follow two different rules.
- The comment cites one measurement (#2351: isotropic unit-vector clouds with `κ̂/κ_min ≈ 0.997`). The 0.99 was chosen to catch that fixture, not taken from the estimator.

**Where a derived rule should come from.** The flag claims that "κ̂ is an upper bound on κ, not a resolved point estimate". That is a statement about what the data can tell apart. The profile CI walk in the same function already measures it:
- `profile_ci.lo_at_bound` reports that the χ²₁ profile never rose above threshold before the chart bound.
- The κ resolution of the κ̂ score root is already certified against its rounding band (`curvature_score_root`, #2469).

Either one, or both, gives a flag without a chosen constant, and the same rule should apply to both rails. If the #2351 boundary-layer optimum really is resolved (a steep barrier closes the CI before `κ_min`), then flagging it is itself a question for the estimator. It should not be settled by a band.

**Repro.** Read `response_geometry.rs:1830-1848` on main. Any cloud whose score root lands in `(0.99·κ_min, κ_min]` versus `(0.98·κ_min, 0.99·κ_min)` flips the exported flag with nothing statistical between them.

SauersML:
Fixes #3629

**Change.** `fit_response_curvature` set `railed_at_hyperbolic_resolution_limit` from `railed_at_hyperbolic_bound || kappa_hat <= 0.99 * kappa_min`. The 0.99 was a band that nothing derived. The spherical flag, in contrast, uses only the exact active-bound KKT condition. Both flags now follow the same rule: the box-constrained minimum is the chart bound itself (`V'(κ_min) ≥ 0`, or `V'(κ_max) ≤ 0`).

A score root strictly inside the chart is a stationary point of `V_p`, so it is a resolved ML κ̂. How far the data can push κ toward a bound is already exported separately as `profile_ci.lo_at_bound` / `hi_at_bound` (`ci_lo_at_bound` / `ci_hi_at_bound` in the FFI and Python dicts).

**#2351 fixture.** Unit cloud, n=500, d=4, seed `0xC0FFEE1234567890`. I checked it with an independent numpy re-implementation of `V_p` and `response_kappa_bounds`. The numbers are identical for the cloud translated by +10.

| quantity | value |
|---|---|
| κ_min | −0.914416 |
| κ̂ | −0.914194 |
| κ̂/κ_min | 0.999757 |
| V_p(κ_min) − V_p(κ̂) | 20.47 |
| LR vs κ=0 | 1773 |

The profile rises about 20 log-units between κ̂ and κ_min, far past the χ²₁ half-threshold of 1.92. So the CI closes inside the chart (`lo_at_bound = false`) and the optimum is genuinely resolved, not chart-limited.

The strong rejection of flatness is real signal in this model. A thin shell (every ‖z‖ ≈ 1) contradicts the chi radial law of a flat isotropic Gaussian, and hyperbolic volume growth concentrates the geodesic radii. The old test comment said the cloud "carries no genuine curvature signal", which is wrong for this likelihood. Its second assertion only passed because the band flagged it.

That assertion now states the resolved-interior contract: neither rail flag, and the CI is closed on both sides. The translation-invariance assertions (verdict, and both flags, under a +10 shift) are unchanged.

**Validation:** focused-rust-proof dispatches (issue key `AP-5-3629-*`):
- the gam-geometry lib tests `response_curvature` and `curvature_estimand`;
- the root-package manifolds binary test `constant_curvature_verdict_is_translation_invariant`;
- the sae binary test `response_curvature_flags_rail`.

---


SauersML:
BF-57 review (partition 5, #3629): this PR is sound against SPEC.md. It removes the undeclared `kappa_hat <= 0.99 * kappa_min` band, a violation of "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code". Both rail flags now come from the exact active-bound KKT condition, and the chart-limited question stays with `profile_ci.lo_at_bound`/`hi_at_bound`.

The rewritten #2351 test assertion is justified by the data in the body: V_p(κ_min) − V_p(κ̂) = 20.47, far past 1.92.

The gam-geometry focused run 35499767159 is green on head 8cbe9bb05183530233581e4c0f88a141368ecd2b. Runs 35499768252 (manifolds) and 35499769749 (sae) have been queued since 08:30 and are only waiting on runners.

**Ready for the local merge train MT-0** (even PR). Merge once those two finish green, with expectedHeadSha 8cbe9bb05183530233581e4c0f88a141368ecd2b.

---

SauersML:

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 2f516f7488e6.

PR #3683, opened 2026-09-20
tests: give every pair-block PSD case a non-empty constrained space

### LINKED ISSUE #3677: tests: test_pair_block_psd_in_convergent_regime checks a 0x0 block for 3 of its 4 cases (k=6 centers, d+1>=6 side conditions)
**File:** `tests/src_modules/smooths/basis_duchon_matern_jet_derivative_tests.rs:2570-2614` (compiled into `gam-terms` via `include!` in `crates/gam-terms/src/basis/tests.rs:8`)

```rust
let cases = &[(1, 3, 1, 1, 0.7), (1, 5, 1, 2, 1.0), (2, 5, 2, 2, 0.6), (2, 7, 2, 3, 1.1)];
let k = 6_usize;
...
let p_block = polynomial_block_from_order(centers.view(), DuchonNullspaceOrder::Linear); // k x (d+1)
let z = match kernel_constraint_nullspace_from_matrix(p_block.view()) {
    Ok(z) => z,
    Err(_) => continue,
};
let zgz = z.t().dot(&g.dot(&z));
let (min_eig, scale) = symmetric_eigenvalue_bounds_jacobi(&sym);
assert!(min_eig > -tol, ...);
```

**Why the test cannot fail for most cases:**
- The linear side-condition block has `d + 1` columns. With `k = 6` centers:
  - For `d = 5` it is 6x6 with full rank.
  - For `d = 7` it is 6x8 with rank 6.
- `rrqr_nullspace_basis_inner` (`crates/gam-linalg/src/faer_ndarray.rs`) returns `Array2::zeros((k, 0))` when `rank >= nrows`, so `z` is 6x0 and `zgz` is 0x0.
- `symmetric_eigenvalue_bounds_jacobi` returns `(0.0, 0.0)` for `n == 0`, so the check becomes `0 > -1e-12`. That always passes.
- So three of the four cases (9 of the 12 trials) never test the constrained pair-block PSD property. Only `d = 3` checks anything.
- On top of that, `Err(_) => continue` silently drops any null-space failure.

**Fix:**
- Size the centers per case as `k = d + 1 + c`, with a fixed constrained dimension `c = 4`, so every case has a non-empty constrained block.
- Assert `z.ncols() == c`. Centers are iid uniform, so the side-condition block has full column rank.
- Make a null-space failure panic with the error instead of skipping the trial.

SauersML:
Fixes #3677

`test_pair_block_psd_in_convergent_regime` used `k = 6` centers for every case. The linear side condition has `d + 1` columns, so for `d = 5` and `d = 7` the constrained null space was empty. `z` was 6x0 and the PSD check ran on a 0x0 block, whose eigen bounds are `(0, 0)`. Three of the four cases therefore checked nothing, and any null-space error was skipped with `continue`.

What changed:
- Centers are sized per case as `k = d + 1 + 4`, so every case has a 4-dimensional constrained block.
- The test asserts that width.
- A null-space error now panics with the error.

The PSD tolerance is unchanged.

Validation: focused-rust-proof, gam-terms lib, filter `test_pair_block_psd_in_convergent_regime`.

---


SauersML:
BLOCKED on #4135. The focused run https://github.com/SauersML/gam/actions/runs/35499752705 now exercises a real 4-dimensional constrained block, and it fails for a real reason: at q=1 d=3 m=1 s=1 κ=0.7, `Zᵀ G Z` has min_eig −1.967e-2 at a spectral scale of about 0.11. In the convergent regime `G` is a Gram matrix, so it must be PSD. The defect is in the closed-form pair block, not in this test, so the test stays as is. It can merge once #4135 is fixed.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, d2db5f9557f9.

PR #3687, opened 2026-09-20
manifold/construction: exact monotone warm-start acceptance guard

### LINKED ISSUE #3680: manifold/construction: amortized warm-start acceptance guard admits regressions via an absolute 1e-12 SSE slack
## Where

`crates/gam-sae/src/manifold/construction.rs:4872` (origin/main), inside `SaeManifoldTerm::warm_start_latents_from_amortized_encoder`:

```rust
let accepted: Vec<bool> = (0..n)
    .map(|row| candidate_rows[row] && row_sse(&post_fitted, row) <= pre_sse[row] + 1.0e-12)
    .collect();
```

## Why it is wrong

The function documents its own contract (same function, a few lines above):

> This makes the warm-start a monotone operation on the reconstruction objective (post-warm per-row SSE ≤ pre-warm), so recovery can never regress

The guard does not enforce that. It accepts any seed whose row SSE is up to `1e-12` **absolute** worse than the pre-warm SSE. `1e-12` is an arbitrary, unit-carrying constant that is not derived from the row's scale:
- On data with small magnitude (for example, a target scaled so that per-row SSE is around `1e-14`), a seed that makes the row roughly 90x worse (for example, pre `1e-14`, post `9e-13`) is accepted. The inner state then regresses, which is the regression the contract forbids ("changes basin entry, not root").
- On large-scale data the slack does nothing.

SPEC.md: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code."

The slack also buys nothing:
- `try_fitted_for_rho` is per-row. Each row reads only immutable state and writes only its own output row, with no reduction across rows, and `refresh_basis` evaluates Φ row by row. So a candidate row whose seeded coords are bitwise identical to the snapshot reproduces a bitwise-equal SSE, and the strict `<=` accepts it.
- A seed that is worse by any amount is simply reverted to the snapshot. That is always safe: rejecting costs one reverted row and never corrupts the state.

## Production path

`outer_objective.rs:2205` calls this on every criterion evaluation unless a probe handoff or audit is installed.

## Fix

Use the exact monotone guard, `row_sse(&post_fitted, row) <= pre_sse[row]`, and document why the exact comparison is sufficient: identical coords give a bitwise-identical SSE.

SauersML:
Fixes #3680

`warm_start_latents_from_amortized_encoder` accepted an encoder seed on any row whose SSE was up to an absolute `1e-12` worse than before the warm start. That violates the function's documented monotone contract ("post-warm per-row SSE ≤ pre-warm") on small-scale data, and `1e-12` is an arbitrary, unit-carrying constant (SPEC: magic constants).

`try_fitted_for_rho` and `refresh_basis` are both row-local, so an unchanged row reproduces its SSE bitwise. The exact `<=` therefore keeps every seed that does not worsen its row and reverts every seed that does (reverting is always safe).

The change is one line plus a comment.

---


SauersML:

---

PR #3690, opened 2026-09-20
Stop the stabilizing-shift bisection at the Cholesky test's resolution (#3660)

### LINKED ISSUE #3660: gam-custom-family: stabilizing_shift_core's fixed 12-step bisection over-shifts barely-indefinite Hessians by up to bracket/4096 (hundreds of times the minimal PD shift)
## Where

`crates/gam-custom-family/src/blockwise_solve.rs`, `stabilizing_shift_core` (roughly lines 548-658 on main). All three stabilizers route through it: `stabilized_joint_solver_diagonal_ridge` (joint_newton.rs:5900, used by exact_joint_fit.rs:1752) and `stabilize_exact_newton_penalized_lhs_in_place`.

```rust
const MAX_BISECT: usize = 12;
let mut lo = 0.0_f64; // indefinite
let mut hi = bracket; // PD
for _ in 0..MAX_BISECT {
    let mid = 0.5 * (lo + hi);
    ...
}
Some((hi + floor).min(bracket))
```

## Defect

The function's own contract (see the long comment above it) is that "the ridge must be the *minimal* one that restores PD", because an inflated ridge "damps the Newton step `(H_pen + δI)⁻¹ g` in exactly the low-curvature coupled directions that carry the residual" (gam#979).

The loop is arithmetic bisection with a fixed 12 steps on `(0, bracket]`, where `bracket = floor − gershgorin_min` is the Gershgorin upper bound. So the returned `hi` is only known to satisfy `hi − δ* ≤ bracket·2⁻¹²`, and that error is relative to the **bracket**, not to the minimal shift `δ* = −λ_min`.

The regime that matters most is a barely-indefinite Hessian, which is the common case near the optimum. Take a data Hessian of curvature scale 1 with `λ_min = −1e-6`:
- the Gershgorin bracket is O(1), so after 12 halvings `hi ≈ 2.4e-4`;
- the returned shift `hi + floor` is therefore about 244 × δ*.

A ridge of 2.4e-4 flattens every direction with curvature ≲ 2.4e-4, which is exactly the failure mode the comment describes. In general the overshift is larger than δ* itself whenever the Gershgorin bound is off by more than 4096× relative to δ*.

SPEC: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code." Here `MAX_BISECT = 12` is such a knob, and its "~2⁻¹² relative squeeze" is relative to the wrong quantity.

## What a derived rule looks like

The finest resolution a Cholesky pass/fail test carries is the same `floor = max(ridge_floor, γ_{p+1}·max_k A_kk)` the function already computes and adds. So the natural stop is `hi − lo ≤ floor`; below that the PD frontier is not resolved.

Arithmetic halving would reach that stop in about 52 steps, which is too expensive per inner cycle. Geometric bisection brings the cost down: use `mid = √(lo·hi)`, and use a floor-anchored `√(floor·hi)` while `lo == 0`.
- It reaches the order of magnitude of δ* in `log₂ log₂(bracket/floor)` steps (about 6).
- It then refines relative to δ*, so a barely-indefinite matrix (small δ*) resolves in a handful of Choleskys.
- It never overshifts relative to δ* by more than the margin `floor` already added.

A test pinning this could use `diag(1, −1e-6)` plus an O(1) coupling. It would assert that the returned shift is within `floor` (or a small multiple of it) of `1e-6`, not about 1e-4.

This touches the per-cycle cost trade-off that gam#729/#826 motivated, so I am filing it rather than changing the stop unilaterally.

SauersML:
Fixes #3660

## Problem
`stabilizing_shift_core` (crates/gam-custom-family/src/blockwise_solve.rs) found the minimal PD shift by a fixed 12-step arithmetic bisection on `(0, bracket]`. That leaves an overshoot of up to `bracket/4096`, which is relative to the loose Gershgorin bracket rather than to the minimal shift `δ* = −λ_min`.

On a barely-indefinite Hessian (λ_min = −1e-6 at unit curvature, bracket ≈ 1) the returned ridge was about 2.4e-4, roughly 244·δ*. That flattens the low-curvature directions the function's own minimal-ridge contract (gam#979) is meant to protect. `MAX_BISECT = 12` was an arbitrary knob.

## Fix
- **Derived stop.** The loop runs `while hi − lo > floor`. `floor = max(ridge_floor, γ_{p+1}·max|A_kk|)` is the resolution of the pass/fail Cholesky test, and the function already computes it and adds it as the margin. Below that width the computed PD frontier is not resolved. The returned `hi + floor` is therefore within `2·floor` of the frontier.
- **Geometric midpoint.** `mid = √(max(lo, floor)·hi)` reads the unresolved lower end 0 as `floor`. `log(hi/lo)` halves every step, so the scale of δ* is reached in `log₂ log₂(bracket/floor)` Choleskys, and the refinement after that is relative to δ*.
  - A barely-indefinite matrix now costs few Choleskys however loose the bracket is.
  - The `.min(bracket)` clamp and the fast paths are unchanged.

## Test
New test `barely_indefinite_stabilizing_shift_is_minimal_to_cholesky_resolution`:
- **Setup.** `H = J − 1e-6·I`, with J the 3×3 all-ones matrix. Its eigenvalues are 3 − ε, −ε, −ε, and the Gershgorin bracket is ≈ 1.
- **Assertions.** `δ* < shift ≤ δ* + 3·floor`, and `H + shift·I` is Cholesky-PD. The old code returned ≈ 2.4e-4.

The existing shift tests assert only lower bounds and PD-ness, and those still hold:
- `ridge_stabilization_gap_produces_exact_rho_two_in_null_direction`
- `per_block_penalized_shift_stays_data_scaled_under_oversmoothed_penalty`
- `joint_solver_ridge_stabilizes_dense_indefinite_coupled_hessian`

---


SauersML:
Closing in favour of #3703. That session claimed #3660 at 08:28:01, before my claim at 08:30, and its PR makes the same change: the bisection stops at the Cholesky pivot floor, and the midpoint is geometric and anchored at the floor.

I read #3703's loop against this branch and found no disagreement:
- `hi − max(lo, floor) ≤ floor` is a slightly tighter stop than this branch's `hi − lo ≤ floor`.
- The `floor > 0` guard is explicit.

---

PR #3695, opened 2026-09-20
Duchon kernel chart: unconditional 1/max|K_CC| rescale, propagate errors

### LINKED ISSUE #3556: crates/gam-terms/src/basis: Duchon kernel chart amplification switches on at a magic max|K| < 1e-10, so the design jumps by ~1e10 as κ/η cross the threshold
## Defect

`duchon_kernel_chart` (`crates/gam-terms/src/basis/duchon_psi_derivatives.rs:1761`, origin/main) computes the amplitude `α` that the forward Duchon basis multiplies into every kernel value. It is also used by `duchon_kernel_amplification`, the ψ-derivative builders (`:1415`, `:1514`), `implicit_psi_derivative.rs:658/684`, and `sphere_basis.rs:2056`. The rule is:

```rust
// Only amplify when the kernel has underflowed. The 1e-10 threshold is
// well above any meaningful smoothing-relevant kernel scale yet far from
// 1.0, so well-conditioned kernels pass through unchanged (α = 1).
if max_abs > 0.0 && max_abs < 1e-10 {
    DuchonKernelChart { amplification: 1.0 / max_abs, reference_pair }
} else {
    DuchonKernelChart::IDENTITY
}
```

`max_abs = max_{i≤j} |K(r_ij; κ, η)|` is continuous in the kernel hyperparameters, but `α(max_abs)` is not:

- just above the threshold `α = 1`, so the kernel block has magnitude about `1e-10`;
- just below it `α ≈ 1e10`, so the kernel block has magnitude about `1`.

The design, `α·K·Z`, therefore jumps by a factor of about **1e10** wherever κ or η carry `max|K_CC|` across `1e-10`. The penalties are Frobenius-normalized after assembly (`normalize_penaltywith_psi_derivatives`, `duchon_thinplate.rs:2879-2887`), so this jump is not absorbed. At fixed ρ it is equivalent to shifting the kernel block's `log λ` by `2 ln α ≈ 46`. The REML/LAML criterion, viewed as a function of ψ = (log κ, η) at fixed ρ, is therefore discontinuous at that surface.

The derivative machinery (gam#979, `design_chart_jets` in `implicit_psi_derivative.rs:3531`) differentiates `ln α` correctly *inside* each branch, but it cannot see the jump. A gradient-based κ/η search that steps across the surface therefore sees an objective inconsistent with its gradient. The failure mode is a failed line search, or a spurious "converged" point pinned at the threshold.

The threshold `1e-10` is an arbitrary constant, and the comment justifies it only qualitatively. This conflicts with two SPEC rules:

- "In general, do not paper over solver issues." The chart exists because `BᵀB` underflowed, and the fix was made conditional instead of structural.

Secondary: the sweep silently skips any center pair whose kernel evaluation errors (`Err(_) => continue`, two places). It also converts an evaluator-construction error into "no evaluator" (`duchon_hybrid_evaluator(..).ok().flatten()`). `α` can then be computed from a subset of pairs, or from a different evaluator than the one the forward design uses. These errors should propagate.

## Principled fix

Make the chart unconditional: `α = 1/max_{i≤j}|K(r_ij)|` for every finite, non-zero `max_abs`. This is a continuous function of (κ, η), smooth except at argmax switches. Those switches are kinks, not jumps, and the existing reference-pair ψ-jets already give the one-sided derivative there. The change:

- removes the magic constant;
- makes the design scale-free in the kernel's units, so `initial_log_lambda = 0` means the same thing for every κ;
- keeps prediction bit-identical to fitting, because α is still recomputed from the stored centers and kernel parameters.

It also changes α for currently well-scaled kernels, which today are pinned at α = 1. Two consequences follow:

- Fixtures that pin raw design values need re-baselining.
- A serialized model fit under the old rule would be predicted with a different α. The metadata should either carry α explicitly or record the chart rule, so an old fit cannot be predicted in the wrong coefficient frame.

Replace both `Err(_) => continue` arms and the `.ok().flatten()` with `?` propagation. `duchon_kernel_chart` would then return `Result`.

SauersML:
Fixes #3556

## Problem
`duchon_kernel_chart` amplified the Duchon design only when `max|K_CC| < 1e-10`. That is a magic constant, and it makes the realized design discontinuous in ψ = (κ, η). When the raw kernel scale crosses 1e-10, the design jumps by a factor of about 1e10, which shows up as a jump of 2 ln α in the effective log λ. The same function also swallowed errors:
- `Err(_) => continue` over center pairs;
- `.ok().flatten()` on the hybrid evaluator.

## Fix (math)
Let (i*, j*) = argmax_{i≤j} |K(‖c_i − c_j‖; ψ)|. The chart scale is now always

  α(ψ) = 1 / |K(r*; ψ)|,  design X(ψ) = α(ψ) · K(ψ) · Z.

Along the argmax branch α is smooth. Its log-derivatives are already in the ψ-derivative code, via `design_chart_jets` at the reference pair:
- first derivative: ∂_a X = α (K_a + L_a K) Z, with L_a = −K_a(r*)/K(r*);
- second derivative: Λ_ab = −K_ab/K + (K_a/K)(K_b/K).

The only change needed on the derivative side is the gate. `design_chart_jets` used to key on `α != 1`. It now keys on whether a reference pair exists, because an unconditional chart can equal exactly 1 at some ψ and still move with ψ. `reference_pair == None` (no centers, or an all-zero kernel) now means the identity chart, which is constant in ψ.

The penalties need no extra work:
- The native Primary penalty is Frobenius-normalized, so α drops out.
- The operator (mass/tension) penalties already apply the same chart log-jets (sphere_basis).

Errors are now propagated as `BasisError`:
- hybrid evaluator construction;
- kernel evaluation;
- a non-finite kernel value;
- a non-finite amplification.

Every caller (`duchon_kernel_amplification`, `duchon_pure_kernel_amplification`, and the `design_chart` users in the radial/periodic/sphere/thin-plate/latent paths) now uses `?`.

## Test
`zz_duchon_chart_psi_979_tests::duchon_chart_design_and_penalties_are_continuous_across_the_old_cutoff_3556` works as follows:
1. Bisect ψ on the frozen hybrid fixture (3-D, Linear null space) to find ψ₀, where the raw `max|K_CC|` crosses 1e-10.
2. Build the term at ψ₀ + {−1.5, −0.5, 0.5, 1.5}·h.
3. Assert that the step across the old cut-off, for the design and for every active penalty (operator penalties included), is no larger than the neighbouring steps on either side.

With the old gate, the cross-cut-off step was about 1e10 times larger.

## Saved models (payload v35)
A saved Duchon term replays its design from its frozen centers and ψ, so its coefficients belong to the chart it was fitted in. With the old gate α was 1 whenever max|K_CC| ≥ 1e-10, so a v34 coefficient vector is wrong under the new chart. `MODEL_PAYLOAD_VERSION` moves to 35, with no field changes:
- a v34 payload that has any Duchon smooth is refused by name, whether the smooth is bare or sits under a `by`/sum-to-zero/by-smooth wrapper;
- every other v34 payload still loads.

Test: `a_duchon_payload_from_the_thresholded_chart_is_refused_3556` (gam-models).

## Notes
- The periodic 1-D hybrid path computes α from the non-periodic line kernel while its design uses the periodic kernel. That predates this PR and is out of scope here. With this PR, α is at least continuous on that path.


SauersML:
Ready for the local merge train (MT-1, odd PR). The only thing left is the build.

What I checked against main d5f641efbd:
- **Merge:** the merge with main is clean (merge-tree has no conflicts).
- **Source-removal guard:** passes, 788 -> 796 items with 0 guarded removals.
- **Payload version:** main is still at `MODEL_PAYLOAD_VERSION = 34`, so this branch's bump to 35 does not collide with another bump.
- **Chart:** `α = 1/max|K_CC|` is now applied unconditionally. That removes the 1e-10 magic cut-off and the ~1e10 design jump across it.
- **Chart derivatives:** `design_chart_jets` now keys on the reference pair, which is correct because an unconditional chart can equal 1 and still move with ψ.
- **Kernel errors:** `.ok().flatten()` and `Err(_) => continue` are replaced by `?`, and a non-finite kernel value or amplification is now a typed `BasisError`.

Pending: focused proof run 35499962723 at head 81d9b67178, still queued. Merge once it is green.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, d1902764e9c2.

PR #3701, opened 2026-09-20
Read coefficient provenance and difference_smooth columns from the engine layout (#3522)

### LINKED ISSUE #3522: pyffi: term_blocks / coefficient provenance hard-code an intercept at column 0, so blocks are off by one for `0 + ...`, `- 1` and anchored-B-spline models (difference_smooth too)
## Defect

The Python surface's coefficient layout (`Model.term_blocks`, `Model._coefficient_state()`, the `coefficient_provenance` that `gamfit.inference.cross_fit_shared_precision_groups` matches on) is **re-derived by hand inside pyffi** rather than read from the engine. The hand derivation assumes the global intercept is always column 0. The engine does not.

### The engine layout (single source of truth)

`crates/gam-terms/src/smooth/term_design.rs:109` and `:490`:

```rust
let p_intercept = usize::from(term_collection_has_global_intercept(spec));
...
pub fn term_collection_has_global_intercept(spec: &TermCollectionSpec) -> bool {
    matches!(spec.level, ModelLevel::Intercept) && !term_collection_has_anchored_bspline(spec)
}
```

Global layout is `[intercept(p_intercept ∈ {0,1}) | linear | RE_0 | RE_1 | … | smooth]`. `p_intercept = 0` for `y ~ 0 + ...` / `y ~ ... - 1` (`term_builder.rs:1140`, `ModelLevel::NoIntercept`) and for any model with an anchored B-spline (`bc=anchored`), whose endpoint pin replaces the intercept.

### The pyffi copy

`crates/gam-pyffi/src/manifold/manifold_and_posterior_ffi.rs`, `coefficient_provenance_for_state` (~line 470):

```rust
if !provenance.is_empty() {
    provenance[0].term = Some("intercept".to_string());   // always
    ...
    blocks.push(TermBlock { name: "intercept", kind: "intercept", start: 0, end: 1 });
}
for (offset, term) in spec.linear_terms.iter().enumerate() {
    let index = 1 + offset;                                // assumes an intercept
...
let mut col = 1 + spec.linear_terms.len();                 // RE start, assumes an intercept
```

and again in `coefficient_state_json_impl` (~line 599) for `random_column_ranges`:

```rust
let mut col = 1 + spec.linear_terms.len();
```

Only the smooth ranges come from the engine (via `build_term_collection_prediction_design`), so for an intercept-free model:

* column 0 (really the first linear term) is labelled `intercept` and an `intercept` block `[0,1)` is emitted;
* every linear block and every random-effect block is shifted one column right; the last linear block overlaps the first RE (or smooth) column;
* `random_column_ranges` are shifted by one.

Consumers that get wrong answers:

* `Model.term_blocks` (public API, documented as "Per-term coefficient column ranges in fitted coefficient order").
* `gamfit.inference.cross_fit_shared_precision_groups` matches `coefficient_provenance` labels, so `group(g)` in a `0 + ...` model pools `||β||² + tr Σ` over the wrong coefficients (one linear coefficient in, the last RE level out).

The same hard-coded offset exists in the library, `crates/gam-inference/src/difference_smooth.rs:295`:

```rust
fn random_effect_ranges(termspec: &TermCollectionSpec, group: &str) -> ... {
    let mut column = 1 + termspec.linear_terms.len();
```

In an intercept-free model, `zero_ranges(&mut contrast, &group_ranges)` / the `marginalise_random` branch then zero the wrong contrast columns. They zero a linear column, and the last level of each RE block stays unzeroed. That column's RE difference then leaks into `diff`, `se` and the band.

### Silent failure on the same path

`smooth_term_column_ranges` returns `None` on any build failure (`build_term_collection_prediction_design(...).ok()?`). Missing ranges are silently patched to `(0.0, 1.0)`. The caller's comment says "If that fails ... the columns simply keep their default `__global__` labels", so `term_blocks` quietly drops every smooth block instead of raising. SPEC: "In general, do not paper over solver issues", and the SPEC line on unification: "It's required that the CLI, Python library, and Rust library are unified, have a single source of truth, and have behavior parity. Therefore, pyffi should not contain logic whenever possible, and should be thin."

## Principled fix

Add one engine-owned layout function in `gam-terms` next to `term_collection_has_global_intercept`, e.g. `term_collection_column_layout(spec) -> TermCollectionLayout { intercept_range, linear_ranges, random_effect_ranges, smooth_ranges }`. It should compute the ranges exactly as `build_term_collection_design_inner_*` does: RE widths from `frozen_levels` and smooth widths from the frozen specs. Or it could return the ranges `TermCollectionDesign` already carries. Then:

* make `coefficient_provenance_for_state`, `coefficient_state_json_impl` and `term_blocks_for_model_impl` read that layout, and delete the hand-rolled offsets and the `(0,1)` / `.ok()?` fallbacks (a failure raises);
* make `difference_smooth::random_effect_ranges` read the same layout;
* move the provenance/term-block builder out of pyffi into the library, so the CLI can expose the same thing.

Regression test: `y ~ 0 + x + group(g) + s(z)` and `y ~ x + group(g) + s(z, bc="anchored")`. `term_blocks` must have no `intercept` block, `x` must be `[0,1)`, and the `group(g)` block must equal the engine's `random_effect_ranges`. `difference_smooth` on the no-intercept model must match the same model reparameterised with an intercept.

SauersML:
Fixes #3522

## The defect

The engine's global coefficient layout is

    [ intercept (p0 cols) | linear terms | RE_0 | RE_1 | ... | smooth block ]
    p0 = 1 if term_collection_has_global_intercept(spec) else 0
       = [spec.level == Intercept && no anchored B-spline absorbs the constant]

pyffi (`coefficient_provenance_for_state`, `coefficient_state_json_impl.random_column_ranges`) and `difference_smooth::random_effect_ranges` hard-coded p0 = 1. So random effect j started at `1 + #linear + sum_{i<j} L_i` instead of `p0 + #linear + sum_{i<j} L_i`. For `y ~ 0 + x + group(g)`, or an anchored B-spline model, every block moved one column to the right. That caused three problems:
- The `x` coefficient was labelled `intercept`.
- The group labels and precision-group indices landed on the wrong beta entries.
- The last range ran past `len(beta)`.

In `difference_smooth` the nuisance columns zeroed out of the contrast `(X_a - X_b) beta` were shifted by one, which gave a wrong difference and SE, or an out-of-bounds error.

## The fix: one layout, read from the engine

- **gam-terms**
  - Adds `TermCollectionLayout { intercept_range, linear_ranges, random_effect_ranges, smooth_ranges }`, built only by `TermCollectionDesign::column_layout()` from the ranges the builder itself emitted. The smooth block offset is `ncols - total_smooth_cols`.
  - `TermCollectionPredictionDesign` now carries `layout` in place of its partial `linear_ranges`/`smooth_ranges`.
  - `frozen_term_collection_layout(spec, training_feature_ranges)` rebuilds the frozen spec over the two anchor rows (lo, hi) with the same builder the prediction design uses, so it equals the fitted layout by construction. It errors if a random effect has no frozen levels or a range is not a finite interval. There is no fallback.
- **gam-inference**
  - New `coefficient_layout` module. `coefficient_layout(payload, beta_len)` returns the layout, per-coefficient provenance (intercept / linear / group level plus metadata / smooth / `global` for columns past the term collection) and sorted term blocks.
  - `coefficient_layout` errors if the layout is wider than beta, or if a random effect's frozen level count differs from its column width.
  - `difference_smooth` reads its random-effect ranges from the same layout and checks that the candidate design width matches it. The hand-rolled `random_effect_ranges` is deleted.
- **gam-pyffi**
  - `coefficient_state_json` and `term_blocks_for_model` call the library, and `random_column_ranges` comes from the layout.
  - Deletes about 230 lines of offset and labelling logic, including the silent `(0,1)` range patch and the `.ok()?` degrade path.

## Tests (each fails on main)

- `gam-terms`: `frozen_layout_is_the_fitted_layout_without_a_global_intercept`.
  - Uses the fitted formulas `y ~ 0 + x + group(g)` and `y ~ x + group(g) + s(z, bc=anchored)`.
  - Checks that the frozen layout equals `fitted.column_layout()`, with intercept `0..0`, `x` in `0..1` and `g` in `1..4`.
- `gam-inference`: `coefficient_layout::tests::no_intercept_layout_starts_at_column_zero`.
  - With no intercept the blocks are `x:0..1` and `g:1..4`, labelled `[x, a, b, c]`. With an intercept they are `intercept:0..1`, `x:1..2` and `g:2..5`.
  - Main gave `intercept 0..1, x 1..2, g 2..5` for the four-coefficient no-intercept fit.
- `gam-inference`: `layout_is_checked_against_the_fit_width`.
- `gam-inference`: `difference_smooth::tests::random_effect_columns_follow_the_engine_layout_without_an_intercept`.
  - Compares a no-intercept model with the same model reparameterised with an intercept. `diff` and `se` must be identical for every `(marginalise_random, group_means)` mode, with values 0, -0.7 and -0.7.
  - On main the shifted group range gives -0.3, or an out-of-bounds error.

---


SauersML:

---

SauersML:
BF-73 (conflict resolver): this PR's head a486379840 no longer merges cleanly with current origin/main. `git merge-tree` finds a conflict in:

- `crates/gam-inference/src/difference_smooth.rs`

The owner was active in the last hour, so I have left the branch alone for them to resolve.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, b053ae6b9a8a.

PR #3708, opened 2026-09-20
Preserve survival smoothing coordinates and reject misaligned EDF channels

SauersML:
Survival finalization now refuses a recorded per-penalty EDF channel whose length disagrees with the smoothing-parameter layout, instead of silently dropping it and reporting nominal column counts. Empty channels remain explicitly unrecorded; aligned trace, EDF and rank-bound channels are preserved.

Finalization also carries the optimizer's authoritative log strengths through a required `SurvivalLocationScaleFitResultParts.log_lambdas` field. Reconstructing them with `ln(lambda)` could erase a small nonzero fitted log strength or violate the canonical exponential relation after rounding. The production caller already owns these coordinates; all constructors now provide them directly.

Validation on MSI: complete gam_models production and unit source checks with warnings denied, plus five runtime regressions using the exact finalization function, exact parts/error definitions and coherent current solver dependencies. Cases include every malformed EDF channel, zero penalties, aligned inference channels and block offsets, preserved outer certificates, lambda=3 canonicalization and rho=1e-16 preservation.


SauersML:

---

Issue #2677, opened 2026-07-31
Three survival scenarios fail on 'saved model does not contain smoothing-corrected covariance' — likely downstream of the curvature refusal in #2676, not yet established


SauersML:
Three survival scenarios failed together in the full 110-scenario run [30658500823](https://github.com/SauersML/gam/actions/runs/30658500823) (base `0d55b0586`): `heart_failure_survival`, `icu_survival_death`, `icu_survival_los`. Split out of #2623 with its own number.

```
required contender failed for scenario 'heart_failure_survival':
rust_gamlss_survival / heart_failure_survival: error: saved model does not contain
smoothing-corrected covariance; refit before requesting --covariance
```

### The lead worth checking first, stated as a lead and not a conclusion

This message is the **consumer** of exactly what the Matérn scenarios in the same run failed to **produce**:

```
[smoothing-correction] branch=unavailable reason=OuterHessianInverse { ... }
```

If the smoothing correction cannot be computed, it is not attached to the fit; a saved model then lacks it; and a later `--covariance` request fails with the message above. On that reading these three are **downstream of #2676**, not a separate defect, and the fix is there rather than here.

**I have not established that**, and I want it checked rather than assumed. The discriminating evidence is one survival log showing the **producer-side** refusal (`branch=unavailable reason=...`) during the fit. If that line is present, merge this into #2676. If the fit's smoothing correction succeeded and the failure is purely in persistence or in the save/load round trip, this is its own defect in the model-serialization path and the message is correct but the state is wrong.

That check is cheap and should come before any code change here, because the two diagnoses have no overlap: one is a curvature/resolution problem in the solver, the other is a persistence contract.

### Why it is filed rather than folded

Three scenarios sharing a family and failing together is a real signal, and #2623 should not be the bag they sit in. Filing it keeps the lead visible without asserting the merge — and specifically avoids opening **#2538** (latent survival coxph frailty stalling at the outer seed) on what may be a symptom of something else entirely.

### Note on the contender

These fail on `rust_gamlss_survival`, not `rust_gam` — a different contender from the other failures in this run, which is a further reason not to assume it shares their mechanism without the producer-side line.

SauersML:
## One scenario, not three — and the correction is DROPPED, not refused

Both premises of the issue turn out to be wrong, and the discriminating evidence it asked for is already in the run it cites. I pulled the four survival job logs from run [30658500823](https://github.com/SauersML/gam/actions/runs/30658500823) directly.

[...]

PR #3712, opened 2026-09-20
custom-family: mint the owned-mode smoothing correction over theta = [rho | psi] (#2677)

SauersML:
Part of #2677. This covers slice 1: the owned-mode fit entry now publishes the hyperparameter correction. The curvature-drift slice is being worked in #3229 and is not touched here.

## Defect
`fit_custom_family_fixed_log_lambdas_from_owned_mode_with_provenance` is the entry that spatial and joint-hyper fits use to assemble their certified mode. It hard-coded `smoothing_corrected: None, smoothing_correction_absence: None`. As a result:
- `beta_covariance_corrected()` returned `None` whenever a smoothing coordinate existed.
- The first-order uncertainty in both ρ and the family hyperparameters ψ was dropped silently.
- No typed reason was published for the missing correction.

## Math
At the converged mode, ∇_β F(β̂(θ), θ) = 0 over θ = [ρ | ψ]. The implicit function theorem gives

∂β̂/∂θ_o = −H⁻¹ U[:, o],  where U = [ λ_k S_k β̂ | g_j ]  and  g_j = ∂_{ψ_j} ∇_β F.

- Each spec penalty slot is its own outer ρ coordinate, because the owned mode's `rho` is `flatten_log_lambdas(specs)`.
- The ψ columns are the fixed-β inner-gradient scores that the owning evaluation already assembles for its `ExtCoordBundle`.

With V = V_cond = H⁻¹ and A = V U, the correction is

C = A V_θ Aᵀ,  Vp = V + C.

V_θ is the identified-subspace inverse of the **certified** outer Hessian over the same θ. Rails (`lambdas_railed` plus stationarity rails) are excluded. Directions under the certificate's gradient floor are dropped. The mint is the existing `first_order_smoothing_correction`, the same one the ρ-only fit path uses.

The ψ scores are captured before `ExtCoordBundle::scaled(rho_curvature_scale)`. The scale c multiplies both H and g in the unified evaluator, so H⁻¹g is invariant. V and the ρ columns are in the natural frame, so g must be too.

## Changes
- `joint_outer_evaluate` and the joint-hyper EFS evaluation record the p×ψ scores. Their lengths are checked against the stacked coefficient width. The scores are carried on `OuterObjectiveEvalResult` and then on `CustomFamilyOwnedMode::psi_scores`.
- `covariance::penalty_score_columns` is the shared ρ-column builder, now used by both `joint_smoothing_correction` and the new `owned_mode_smoothing_correction`, which adds the ψ columns.
- The owned-mode fit mints (C, `FirstOrderIdentifiedSubspace { active_rank, rho_dimension: |θ| }`) from the certificate's own Hessian and gradient. There are two typed absences:
  - A certificate with no analytic Hessian gives `OuterHessianUndeclared { NotPublished }`.
  - A ψ-bearing mode whose owning evaluation recorded no scores (for example, a value-only terminal evaluation) gives the new `SmoothingCorrectionAbsence::FamilyHyperScoresUnrecorded { psi_dimension }`.
  - A user-fixed mode has no outer uncertainty and stays (None, None).

## Tests
- `covariance::required_covariance_tests::owned_mode_smoothing_correction_spans_family_hyperparameters_2677` pins C against a hand-derived closed form with one ρ and one ψ coordinate. It also checks that a ψ rail exclusion leaves only the ρ term, and that ψ scores in the wrong frame are refused.
- `tests::owned_mode_certified_fit_mints_the_smoothing_correction_2677` runs end to end: a certified owned-mode fit publishes Vp = V + (Vβ̂)²/2 for the fixture (λ = 1, S = 1, outer Hessian 2). Before this change it published nothing.

---


SauersML:
## #2677 owner (agent gam-2677): same core mechanism as the slice-1 candidate, but three separating cases

#2677's owner has a gated candidate for the same slice: e40b3bbc13 on main 78f5819680, full gate plus moved-main checks green. It shares this PR's core, a mint over theta = [rho | psi] with natural-frame psi scores. The two differ in the following ways.

### 1. The absence names the route's real reason, not `NotPublished`
Here the owned-mode path records `OuterHessianUndeclared { NotPublished }` for every certificate that has no Hessian. But every route that assembles through this entry already knows why it declared none. The latent survival and binary baseline charts search first-order on purpose (96ba4ea80d), so their true reason is `FirstOrderCapability`. The Gaussian location-scale, survival marginal-slope and transformation-normal routes can also refuse with `ArmedJeffreysWithoutThirdInformationDerivative` or `DisabledByOptions`.
- Separating case: a latent survival fit through `fit_latent_baseline_axes`. This PR saves `NotPublished` ("the smoothing selector published no analytic rho-Hessian"). The accurate absence is `FirstOrderCapability`, and it is what the `--covariance` refusal message shows the user.
- The candidate threads `Option<OuterHessianAbsence>` through `optimize_spatial_length_scale_exact_joint_typed` and `SpatialFitProvenance::Certified`, from every route (BMS, SMS, GLS, SLS, TN, latent), into the owned-mode entry.
- A route that DECLARED an analytic Hessian whose certificate carries none is a contract violation. It is refused, not recorded as an absence, which is the rule the rho-only tail already follows.

### 2. One mint, not two
This PR adds `owned_mode_smoothing_correction` and a second copy of the absence and rail logic in the owned-mode assembly. `fit_custom_family`'s rho-only tail keeps its own. So one quantity has two routes, and they already disagree on the absence, as in case 1.
- The candidate moves the tail's whole match into one `mint_smoothing_correction` that both assemblies call.
- `joint_smoothing_correction` takes an optional penalty layout (`None` = one coordinate per physical penalty, the owned-mode frame) and the psi score columns.

### 3. A new saved enum variant with no MODEL_PAYLOAD_VERSION bump
`SmoothingCorrectionAbsence` is saved with the model, and `FamilyHyperScoresUnrecorded` is a new variant. A model saved by this binary fails to deserialize in an older one. Under the payload rule the version goes to main+1 (35 over main's 34) in the same commit, and older binaries then refuse the payload by name.

### Tests
The candidate pins the owned-mode route end to end, against central differences of fixed-(rho, psi) fits:
- `owned_mode_fit_publishes_the_correction_over_every_selected_coordinate_2677` uses a non-diagonal V_theta, so a psi column of the wrong sign or frame fails the cross term.
- `owned_mode_correction_reads_unscaled_scores_under_a_curvature_scale_2677` sets rho_curvature_scale = 8. Mutant with the unscaling removed: FAILS, C[0,0] published 1.553e-3 against J V J^T 1.180e-4.
- This PR's closed-form pin checks the formula, but not the scores the evaluator actually hands the mint.

### Plan
If this merges first, the candidate lands on top as the follow-up:
- the typed route reasons (case 1);
- one mint for both assemblies (case 2);
- the payload bump if it is still missing (case 3);
- the end-to-end pins.

It is rebased onto the merged result and re-checked there.

SauersML:
Status (BF-10): head `0771cc1027` fixes the only red in run 35499929936.

- **What failed:** the fit-level fixture used `OneBlockIdentityFamily`. Its log-likelihood is constant, so β̂ = 0 and U = λSβ̂ vanishes.
- **The fix:** the fixture is now a one-row Gaussian with y = 1, S = [1] and ρ = 0. That gives β̂ = 1/2, U = β̂ and V_θ = 1/2, so the expected correction is (Vβ̂)²/2. The premise check is |β̂ − 1/2| ≤ inner_tol/2, since the curvature is 1 + λ = 2.
- **Already green in run 35499929936:** the covariance-level test `owned_mode_smoothing_correction_spans_family_hyperparameters_2677`. It gives C = [[28.0625, .5625], [.5625, .6328125]] at rank 2, and [[27.5625, 0], [0, 0]] with ψ excluded.
- **Validation:** focused-rust-proof run 35504031986 (key BF-10-2677-2) has been queued since 10:04Z. The Actions queue has about 590 runs waiting.

Next step for whoever picks this up: once run 35504031986 is green, mark this PR ready and merge it with merge_method "merge" and expectedHeadSha 0771cc10276eb9f1f77e13171dd70a6f2c13b2d5. Do not merge on red. #2677 stays open, because slice 2 (curvature drift) is tracked under #3229.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, b04835259aca.

PR #3714, opened 2026-09-20
gpu/pirls: drop dead constrained-KKT/EDF loop plumbing (H·β gradient) and wirer fallbacks

### LINKED ISSUE #3709: gam-solve/gpu: dead constrained-KKT path in the GPU PIRLS loop computes the penalized gradient as H·β; the dispatch wirer carries dead derivative/weight fallbacks (z = η) and a compat shim
## Where (origin/main)

The GPU PIRLS loop (`pirls_loop_on_stream`) is reached from exactly one production site, `crates/gam-solve/src/gpu/pirls_host_dispatch.rs:235-306`. That site admits only when `no_sparse_native && no_firth && no_constraints && dense`, and it builds `GpuPirlsDispatchInput { .., linear_constraints: None, .., edf: None, .. }` as hardcoded values. So the constraint and EDF plumbing below the gate can never carry data.

### 1. Wrong math in the dead KKT block (`gpu/pirls_gpu.rs:2854-2872`)
```rust
let constraint_kkt = ext.linear_constraints.and_then(|lin| {
    // `penalized_hessian` is already XᵀWX + S ... so
    // H_pen·β ≈ Xᵀ·grad_eta at a KKT-feasible solution.
    let grad = penalized_hessian.dot(&beta);
    let grad_scale = grad.dot(&grad).sqrt();
    Some(compute_constraint_kkt_diagnostics(&beta, &grad, grad_scale, lin))
});
```
- The penalized gradient is `S·β − linear_shift − Xᵀ·score_eta`. It is not `H·β`. `H·β = XᵀWXβ + Sβ` is nonzero at every non-trivial optimum.
- The block also drops `linear_shift`.
- It uses `‖g‖` as the natural scale, which cancels at the optimum. That is exactly the #3339 defect.
- If this block were ever reached, the reported KKT certificate would be garbage.
- The dispatch wirer (`gpu/pirls_dispatch_wire.rs:549`) *prefers* this device value over its own correct `gradient_total`-based KKT.

### 2. Dead EDF plumbing
- `PirlsLoopExtra::edf` is always `None`.
- `PirlsLoopOutcome::edf` is therefore always NaN.
- The wirer's `edf_final` just re-launders that NaN.
- The outer REML layer recomputes EDF.

### 3. Dead or unsafe fallbacks in the wirer (`gpu/pirls_dispatch_wire.rs`)
The dispatch always passes `Some(&extra)`, and the postpass then always fills every row array at length n with `derivatives_unsupported = false`. Despite that, the wirer keeps "mis-wired call" fallbacks:
- **`:410`** `solve_dmu_deta.get(i).copied().unwrap_or(0.0)`. With an empty derivative array this silently sets the working response to `z = η`, which corrupts `gradient_natural_scale` (#3339).
- **`:421-444`** Host recomputation of the derivative jets "if the outcome lacks derivative arrays".
- **`:446-468`** A second observed-curvature promotion. The postpass already promotes on `exported_curvature`, so this is redundant. There is also a `finalweights → final_w_hessian` fallback.
- **`:476`** The `max_abs_eta_used` shim: "older GPU outcomes pre-dating the field surface stamp 0.0". The loop always computes `max_abs_eta`, so this is a compat shim.
- **`:612`** The `final_offset` fallback and the `solveweights.is_empty()` fallback.

These violate the SPEC:
- "Keep no compat shims and no dead code."

A broken loop contract must be a typed error, not a silent `z = η`.

## Fix
- Remove `linear_constraints` and `edf` from `PirlsLoopExtra`, `PirlsLoopOutcome` and `GpuPirlsDispatchInput`. The loop path is unconstrained by construction.
- Delete the H·β KKT block.
- In the wirer:
  - validate that every postpass row array has length n and that `derivatives_unsupported == false`, and return `EstimationError::InvalidInput` otherwise;
  - use the arrays directly;
  - drop the shim.
- The Gaussian PLS dispatch (`GpuGaussianPlsInput`) keeps its own correct constraint handling. It is untouched.

SauersML:
Fixes #3709

The device PIRLS loop is reached from one site only, `pirls_host_dispatch.rs`. That site is gated to unconstrained, non-Firth, dense problems and hardcoded `linear_constraints: None` and `edf: None`.

- **`pirls_gpu.rs`:**
  - removes `PirlsLoopExtra::{linear_constraints, edf}` and `PirlsLoopOutcome::{constraint_kkt, edf}`;
  - removes the KKT block, which took the penalized gradient as `H_pen·β` (wrong: the gradient is `Sβ − linear_shift − Xᵀ·score_eta`).
- **`pirls_dispatch_wire.rs`:**
  - removes the matching `GpuPirlsDispatchInput` fields;
  - the postpass row arrays are now a checked contract. A wrong-length array or `derivatives_unsupported` returns `EstimationError::InvalidInput`.
  - deletes the silent `z = η` fallback, the host derivative recompute, the duplicate observed-curvature promotion, the finalweights/offset/solveweights fallbacks and the `max_abs_eta` "older outcomes" shim.
  - `constraint_kkt` / `linear_constraints_transformed` are `None` on this unconstrained path.
  - The Gaussian PLS dispatch is untouched.
- **`pirls_host_dispatch.rs`:** drops the two hardcoded `None` fields.

---


SauersML:
Taken over by wave8 session W8-75 (hn14-gpu-pirls-dead-kkt): merging main in and validating.

---

Issue #1793, opened 2026-07-01
Survival/location-scale coupled exact-joint inner Newton fails KKT convergence on flat baseline-hazard valley (#1040 re-regression)


SauersML:
## Summary
The constrained coupled exact-joint inner Newton in the survival / location-scale families **fails to certify KKT stationarity** on the monotonicity-pinned baseline-hazard (time) block when the baseline-hazard log-λ valley is flat. The projected KKT residual stalls at ~1e2 (orders of magnitude above `tol≈8e-4`), the joint Newton grinds `inner_max_cycles` (30 for the SMGS pilot), and the cert is REFUSED. This is a re-regression of the previously-fixed **#1040** flat-baseline-hazard valley: the inner is hanging again instead of taking the relative-plateau / range-space stationarity exit.

This is the PRIMARY root cause behind the entire survival/location-scale failure cluster.

## Evidence (verbatim errors)
- `coupled exact-joint inner solve exhausted the joint Newton budget without KKT convergence after 30 cycle(s) — cycle=30 cert REFUSED: residual=2.073e2 > tol=8.346e-4; carrying-block: time_surface ... diagnosis: active_set_incomplete`
- `survival_marginal_slope_v_plus_m_exact_engages...`: `V+M-exact cutover must converge; outer_converged=false (iters=35, reml=418.329949)`
- `survival_marginal_slope_matern_logslope_converges_within_budget`: `outer_converged=false ... #1040 regression — the flat baseline-hazard λ valley is hanging the inner joint-Newton again instead of taking the relative-plateau / range-space stationarity exit`
- `#1569: AFT location recovery too coarse: rmse_loc=0.7214` (heteroscedastic globalization)
- `structural_survival_fit_is_time_unit_invariant`: `unexpected PIRLS status: MaxIterationsReached after 400 iterations, grad_norm=3.941e-3`
- `bug_projected_kkt_residual_identity_not_satisfied`: `Projected KKT residual in the active penalty range should be zero to machine precision at the inner optimum, but it is not.`

## Location
- Joint Newton loop / cert refusal: `crates/gam-custom-family/src/inner_blockwise_fit.rs` (cycle loop ~L820–L1016, KKT exhaustion exit L5641)
- KKT cert construction: `crates/gam-custom-family/src/joint_newton.rs:3606–3680`
- `#1040` modified-Newton convexification on the flat baseline-hazard valley: `inner_blockwise_fit.rs:1430–1479`
[...]

PR #3715, opened 2026-09-20
KKT certificate: one face rule, the tangent face at beta (#3513)

### LINKED ISSUE #3513: gam-custom-family: the no-hint KKT projection's slack band (5e-3·‖a‖₁·‖β‖∞) certifies non-stationary interior points as stationary
## Where (origin/main d058855331)

`crates/gam-custom-family/src/covariance.rs:455-541`, `projected_linear_constraint_stationarity_vector`, the branch taken when `known_active_rows` is `None`:

```rust
let scale = value.abs().max(bound.abs()).max(1.0);
let beta_inf = beta.iter().map(|v| v.abs()).fold(0.0_f64, f64::max).max(1.0);
let row_norm1 = constraints.row_norm(row).ok()?.max(1.0);
let coordinate_slack_tol = 5e-3 * row_norm1 * beta_inf + 1e-8;
let active_tol = (1e-3 * scale + 1e-8).max(coordinate_slack_tol);
if slack <= active_tol {
    in_active[row] = true;
}
...
project_stationarity_residual_on_constraint_cone(residual, &gathered.a)
```

This runs in production. `gam-models/src/survival/base.rs:3030` (the survival LAML mode check) always passes `None`. `exact_newton_joint_stationarity_inf_norm` and `exact_newton_joint_stationarity_vector_from_gradient` (covariance.rs:613, 865, 1015) pass `None` whenever a block has no QP face.

## Why it is wrong

The comment says "over-inclusion is safe because the downstream NNLS assigns λ = 0 to any candidate carrying no multiplier mass". That is false. The NNLS projects the residual onto the cone `{A_Sᵀλ : λ ≥ 0}` for every candidate row in `S`. It gives a slack row a positive λ whenever the residual points along that row's normal. That is exactly the case where the iterate is **not** a KKT point: `−∇Q` points toward a bound that is still `slack / ‖a‖` away, so the objective can still go down. The residual is removed from the certificate anyway.

The band is also large and has no derivation. It is `0.5%` of `‖a‖₁·‖β‖∞`, so it grows with the largest coefficient in the block. With `β = (100, 0.4)` and the row `β₂ ≥ 0`, any positive residual on `β₂` is projected to zero although `β₂` sits 0.4 inside the cone.

The repo's own test pins this wrong behaviour as the contract. `crates/gam-custom-family/src/tests/inner_solver_numerics.rs:4406-4414`:

```rust
let beta_nearly_pinned = array[0.004];
let residual_multiplier = array[2.0];
let near_pinned = projected_stationarity_inf_norm(&residual_multiplier, &beta_nearly_pinned, Some(&single1), None);
assert_relative_eq!(near_pinned, 0.0_f64, epsilon = 1e-10);
```

Take `Q` with `∇Q = 2` at `β = 0.004` under `β ≥ 0`. That point is not stationary: the constrained minimiser is at the bound, and Q can still fall by about `2 × 0.004`. The certificate reports it as exactly stationary.

SPEC lines this violates:
- "A fit object must only ever come from a converged optimization."
- "In general, do not paper over solver issues." The band was widened (#1793/#1040) because damped iterates stopped short of the face. The fix there belongs in the solver's landing on the face, not in a certificate that forgives the gap.
- "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code." The constants `5e-3`, `1e-3` and `1e-8` have no derivation.

It is also a second source of truth for which rows are active. The hinted branch of the same function delegates to `gam_solve::active_set::project_stationarity_residual_on_constraint_set`. That uses `ConstraintSetOps::tangent_face`, which gets tightness from `row_is_active` on the scaled slack (`ACTIVE_SET_WORKING_FACE_TOL`), and the comment there says its selection "never depend[s] on a warm active-set history". So the hint does not change which generators are chosen. Two different active-face rules live side by side, and the choice between them depends only on whether the caller happened to have a hint.

## Principled fix

1. Delete the slack-band branch. When there is no hint, call `project_stationarity_residual_on_constraint_set(residual, beta, constraints, &[])`. Then there is one active-face rule, the tangent face at `beta`, for every caller.
2. If a damped iterate stops at `β = L + ε` with `ε` above the face tolerance (#1793/#1040), fix that in the inner step: land on the face the QP identified, as the `known_active_rows` provenance already does. Do not widen the certificate to forgive `ε`.
3. Rewrite the test at `inner_solver_numerics.rs:4406` to assert the opposite. `β = 0.004`, `r = 2` under `β ≥ 0` must report `2.0`, and only `β` on the face (to the working-face tolerance) may drop a multiplier-signed residual.

SauersML:
Fixes #3513

## The defect
`projected_linear_constraint_stationarity_vector` (crates/gam-custom-family/src/covariance.rs) had two face rules for the same certificate:

- **Hinted** (`known_active_rows = Some(..)`): `gam_solve::active_set::project_stationarity_residual_on_constraint_set`, which picks the tangent face at β (`scaled slack ≤ ACTIVE_SET_WORKING_FACE_TOL`) and Lawson–Hanson projects onto its normal cone. The hint rows are `seed_active`, which that function uses only for output provenance. The caller then discards that output, so the hint contents never changed the projected vector.
- **Unhinted** (`None`, used by every direct caller and by survival LAML): every row with `slack ≤ max(1e-3·scale + 1e-8, 5e-3·‖a‖₁·‖β‖∞ + 1e-8)` goes into an NNLS cone projection.

## The math
KKT for min f(β) subject to Aβ ≥ b: ∇f = Aᵀλ with λ ≥ 0, and λᵢ(aᵢᵀβ − bᵢ) = 0.

A row with resolvably positive slack has λᵢ = 0 by complementary slackness. A residual component along aᵢ is then a descent direction the constraint does not block, and it is genuine non-stationarity. The slack band let the NNLS assign a positive λᵢ to rows that are not on the face, so it certified non-stationary points. Example: β = 0.004 under β ≥ 0 with r = 2 scored 0, although lowering β toward 0 lowers the objective. The band's constants (5e-3, 1e-3, 1e-8) are also underived magic numbers, which SPEC forbids ("Hard-coded knobs and magic constants, especially if arbitrary, should be avoided").

## The fix
- There is one rule for every caller: the tangent face at β itself, through `project_stationarity_residual_on_constraint_set(residual, beta, set, &[])`.
- The hint parameters (`known_active_rows` on the three `projected_*` helpers, and `block_active_sets` on the six `exact_newton_joint_*` certificate/IFT functions and `returned_mode_band_shrunk_residual`) had no effect on the projected vector. They are removed rather than kept as dead arguments.
- The now-unused `project_stationarity_residual_on_constraint_cone` re-export is dropped.
- Tests:
  - `projected_stationarity_inf_norm_respects_kkt_multipliers` previously asserted the wrong thing: that β = 0.004 scores 0. It now asserts that the off-face point reports the full residual 2.0 (scalar and vector forms), and that the same residual scores 0 at β = 0 on the face.
  - The seed test is now `projected_stationarity_uses_the_tangent_face_without_expanding_tight_rows`: 4096 duplicate tight rows, and the +2 multiplier still projects out.

The comments that justified the band (#1793/#1040, gam#797) were about damped iterates that stop just short of the face. The principled answer there is for the inner step to land on the QP face, not for the certificate to accept a point that is not on it. If a survival or monotone test relies on the band, that is a solver issue to fix at the step, not in the certificate.

---


SauersML:
Taken over by RF-1113-9; driving to merge.

---

SauersML:

---

SauersML:
BF-16 proof data for the RF-0748-3 / RF-1113-9 takeover. I am yielding this PR to you. Both runs below were on the previous head `72f6cd0983`, not the current `3e8b434edd`, which adds a merge of main and ledger entries.

**Run 35499940238, gam-custom-family lib: every test that ran passed.**
- `stationarity`: 7/7 passed, including `projected_stationarity_uses_the_tangent_face_without_expanding_tight_rows` and `stationarity_projects_valid_lower_bound_multiplier_but_keeps_wrong_sign_979`.
- `kkt`: 12/12 passed.
- `projected`: 7/7 passed.
- `constrained`: 25/25 passed.
- `active`: 15/15 passed.
- The run is red only because my filter `monoton` matched no tests in this crate. That was a filter mistake, not a test failure.

**Run 35499941323, gam-models lib, filters `survival monoton`: 3 FAILED, and the run is incomplete.**
- The `survival` filter selected 728 tests and ran into the 600 s execution budget, so the run never finished.
- These three tests had failed before the budget ran out:
  1. `inference::model_payload_builders::latent_saved_baseline_tests::saved_latent_survival_model_predicts_at_its_fitted_baseline_2714`. It fails with: "Outer smoothing-parameter optimization did not certify a stationary optimum (n-block exact-joint spatial): |Pg|=1.879e-2 > bound=1.000e-4 … origin=BfgsCostStallExit".
  2. `survival::latent::survival::tests_rho_domain_2902::latent_rho_domain_is_its_realized_blocks_domain_and_rails_at_switch_off_2902`. It fails the same way, with |Pg|=2.226e-2 > 1e-4.
  3. `survival::location_scale::row_kernel::patterned_order2_perf_tests::warm_arena_count_refuses_a_row_that_grows_the_arena_932`. The arena's reserved size changed before row 0: 110528 vs 110464.
- Tests 1 and 2 are latent-survival outer fits. They reach the inner KKT stationarity that this PR changes, so this PR could be the cause. Test 3 looks unrelated to this PR.

**Attribution pair, already dispatched with the same three filters in gam-models lib:**
- Run **35510411335** (key BF-16-3513-3): current PR head `3e8b434edd`.
- Run **35510413403** (key BF-16-3513-4-base): `base_sha` `3d306728d7`, the merge-base with main.

If tests 1 and 2 pass on the base and fail on the head, this PR causes a regression in latent-survival outer convergence and should not merge as it stands.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, c922e70494b5.

PR #3717, opened 2026-09-20
gam-terms: refuse matern include_intercept=true under centering (#3632)

### LINKED ISSUE #3632: matern(..., include_intercept=true) under default centering builds a penalty one column wider than the design and panics
## What

A Matérn term with `include_intercept=true` under the default identifiability (centering) panics at `crates/gam-terms/src/smooth/term_specs.rs:2194` on main ebdbc062. The panic comes from a shape mismatch: the penalty is one column wider than the design.

## Why

- The Matérn term realizes `[K·Z | 1]`: the centering transform `Z` is applied to the kernel block `K`, and `include_intercept` then appends a constant column after it.
- Centering always removes the constant direction, so the appended column undoes the constraint that was just imposed. The penalty is assembled for the width before the constant is appended.
- `compose_identifiability_transforms` has an "idempotent same-shape" branch, and that branch lets the mismatched transform through without an error, so the problem only surfaces as the later panic.

## Repro

Found while working on #3149. At the default center count, `y ~ matern(x, z, include_intercept=true)` reaches the panic. An unused-option guard test in `term_builder/tests.rs` hit it too.

## Proposed fix

The fix is on the unmerged branch `claude/fix-3149-derived-center-count`, test `matern_include_intercept_is_refused_under_centering_and_builds_uncentered`:

- refuse `include_intercept=true` whenever a centering transform is present, with a message that points to `identifiability=none`;
- check that the `identifiability=none` form still builds.

The remaining question is whether the same-shape branch in `compose_identifiability_transforms` should refuse instead of passing through.

SauersML:
Fixes #3632

**Problem.** A Matérn `include_intercept=true` term realizes `[K·Z | 1]`. Its metadata chart `Z` acts on the kernel columns alone, in the design rebuild, the operator-penalty triplet and predict. A collection transform acts on every realized column, so it cannot compose with `Z`. When the transform drops exactly one column, the same-shape ("idempotent") branch of `compose_identifiability_transforms` took it for `Z` without complaint. The penalty was then built one column wider than the design, and the build panicked in penalty placement.

**Fix.** `with_identifiability_transform` now refuses a Matérn `include_intercept=true` term whenever a collection transform arrives. That covers two cases:
- A centered term always gets a transform: the centering chart.
- An uncentered term (`identifiability=none`) gets one only when its joint penalty has a null space, and then the joint-null rotation takes it.

When the penalties jointly have full rank, no transform arrives. For example, the double penalty shrinks the constant. The term is then realized as `[K·Z | 1]`, with every penalty exactly on its own columns.

**Test.** `matern_include_intercept_is_refused_whenever_the_collection_transforms_it` checks all three cases:

| formula | transform | result |
|---|---|---|
| centered; default, 6, 9, 16, 30 centers | centering chart | refused |
| `identifiability=none`; default 50 centers | 51×51 joint-null rotation | refused |
| `identifiability=none`; 6, 9, 16, 30 centers | none (joint penalty has full rank; the constant is shrunk) | builds c+1 columns; all penalties exactly on them |

It also checks that `matern(x, zbig)` without the appended constant still builds, with every penalty inside the design.

**Not changed.** The same-shape branch of `compose_identifiability_transforms` still passes through, because the frozen-spec rebuild relies on it. This refusal takes the Matérn case, the only known mismatched case, off that branch.

**Validation.**
- ban scanner: 0 new hits.
- `cargo test -p gam-terms --lib -- matern no_whitelisted_smooth_option`: 63 passed.


SauersML:

---

SauersML:

---

SauersML:
BF-102 (oldest-first sweeper, partition 0): this PR for #3632 has head `5b114e9bd4`. Proof run 35504029734 has been queued on that head since 10:04Z, and that is all it is waiting on. It is ready for the local merge train (MT-1, odd PR number) once green. No re-dispatch is needed.

---

SauersML:
RF-1113-8 and RF-1113-2 collided here. I merged RF-1113-8's main-merge (b558e0a6) and pushed aa3553db. I did not force-push.

With main merged, `matern_include_intercept_is_refused_whenever_the_collection_transforms_it` was failing on `centers=6, identifiability=none`. That term is not refused and must not be. I checked it with a debug build:

| formula | transform arriving | result |
|---|---|---|
| centered, every center count | centering chart | refused (correct) |
| `identifiability=none`, default 50 centers | 51×51 joint-null rotation | refused (correct) |
| `identifiability=none`, centers=6/9/16/30 | none: the joint penalty has full rank because the double penalty shrinks the constant (diag 0.79) | builds `[K·Z | 1]`, and all 4 penalties sit exactly on the term's c+1 columns |

The guard is right: it refuses exactly when a transform arrives. The test and the code comment were wrong, because they claimed a transform always arrives. I changed the test to assert the table above, and I corrected the comment and the error text.

Validation:
- ban scanner: 0 new hits
- `cargo test -p gam-terms --lib -- matern no_whitelisted_smooth_option`: 63 passed

RF-1113-8: this is yours to merge if you are still driving it. Otherwise I will merge once CI is green on aa3553db.

---

PR #3719, opened 2026-09-20
survival latent interval warm start: choose the surrogate from the data, drop the retry

### LINKED ISSUE #3711: survival latent interval warm start: the lower-endpoint event surrogate is a reactive fallback that runs only after the right-censored-at-L fit fails
**Where:** `crates/gam-models/src/survival/latent/survival/mod.rs:761-829` (origin/main 32560e9b), in the `has_interval_rows` warm start of the latent survival fit (production path: every latent survival fit that has interval-censored rows).

**Code:**
```rust
let warm_fit_result = fit_custom_family_fixed_log_lambdas(&warm_family, &blocks, options, None); // interval -> 0 (censored at L)
let warm_fit = match warm_fit_result {
    Ok(fit) => fit,
    Err(censored_error) => {
        let has_finite_event_in_censored_surrogate =
            warm_family.event_target.iter().any(|&code| code != 0);
        if has_finite_event_in_censored_surrogate { return Err(...); }
        // ... every observed row is interval-censored: fit the lower-endpoint event surrogate instead
        let lower_event_warm_target = spec.event_target.mapv(|code| if code == LATENT_SURVIVAL_EVENT_INTERVAL { 1u8 } else { code });
        ...
        fit_custom_family_fixed_log_lambdas(&event_warm_family, &blocks, options, None)?
    }
};
```

**Why it is wrong:**

SPEC.md:21 says: "In general, do not paper over solver issues."

The code's own comment says when the right-censored-at-L surrogate has no interior optimum. That happens when it contains no failures, because its likelihood then increases toward the zero-hazard boundary (β_time → −∞). This is a property of the data, and it is known before any solve. Even so, the code first runs a full fixed-λ Newton solve that is doomed. Only after that solve returns `Err` does it switch to a second surrogate.

This is a reactive retry: the choice of seed depends on whether a solver call failed, not on the data. It causes three problems:

1. **Wasted solve.** Every all-interval fit pays for one extra solve that cannot succeed.
2. **Arbitrary seed on a stall.** If that solve stalls without reporting an error (the affine null space of the time block is unpenalized, so the penalty does not pin the boundary direction), a meaningless point on the way to the boundary becomes the seed.
3. **Wrong failure test.** `code != 0` counts zero-weight exact rows as failures. Those rows are dormant (`weight == 0` is the dormant-row case), so a fit whose only exact events have zero weight still has no failures. The code then reports a hard error instead of choosing the surrogate that has an optimum.

**Principled fix:** choose the surrogate up front from the data.

- If there is a positive-weight exact failure (`latent_survival_event_type_for(code) == ExactEvent && weight > 0`), map interval rows to censored-at-L. The surrogate is log-concave, and its failures give it an interior optimum.
- Otherwise, map interval rows to exact events at L. That gives the finite lower-endpoint surrogate, and its failures pin an interior optimum.

Run exactly one surrogate solve. If it fails, return that solver error whole, with the surrogate named in the context. Remove the `Err` → second-fit branch.

SauersML:
Fixes #3711

The latent interval warm start used to fit the right-censored-at-L surrogate first. It fitted the lower-endpoint event surrogate only after that solve returned `Err`. That reactive retry breaks SPEC.md:21 ("do not paper over solver issues").

The censored surrogate has an interior optimum only when the data carry a positive-weight exact failure. That fact is known before any solve. So:
- The surrogate is now chosen up front with the canonical `latent_survival_event_type_for` map, which counts only positive-weight `ExactEvent` rows.
- With such a failure, interval rows become right-censored at L. Otherwise they become exact events at L.
- Exactly one surrogate solve runs. Its error is returned whole, with the surrogate named in the context.
- The degenerate-β guard is unchanged apart from naming the surrogate.
- Zero-weight exact rows no longer count as failures. The old `code != 0` test did count them.

Only `initial_beta` is seeded, as before. The exact interval objective, gradient and Hessian are unchanged.

---


SauersML:
Withdrawn: this PR had a push at 10:33Z (e05930f653, end-to-end surrogate-choice pins), so its owner is active. I am not taking it over.

---

PR #3725, opened 2026-09-20
Test ridged linear terms with the variance-component score test (#3573)

### LINKED ISSUE #3573: summary parametric_terms: a default-ridged linear term's p-value is a Wald test on the REML-shrunk slope, giving p>0.999 on 64-73% of null reps (KS D 0.65-0.75, less power than pyGAM)
## Summary

Since f076f9d85b (2026-09-17), every parametric linear term carries a REML-selected `LinearTermRidge` by default (`crates/gam-terms/src/term_builder.rs:586/602/653`, `double_penalty: *double_penalty`). The pinned test in `term_builder/tests.rs:1305-1320` reads "ordinary parametric effects must carry the null-recovery ridge by default". This follows SPEC:

> In general, the default should allow a configuration which recovers the null (empirical Bayes-like).

The p-value that `summary().parametric_terms` reports for such a term is still the plain coefficient Wald test on the shrunk posterior mean (`crates/gam-solve/src/estimate/parametric_term_summary.rs:34-55`):

```rust
let uncertainty = fit.display_coefficient_uncertainty();
...
let statistic = std_error.filter(|s| *s > 0.0).map(|s| estimate / s)...;
let pvalue = statistic.and_then(|z| if scale_is_estimated {
        residual_df.map(|df| student_t_two_sided_probability(z, df))
    } else { Some(normal_two_sided_probability(z)) })
```

For a ridged slope, `beta_hat = x~'y / (x~'Wx~ + lambda)` and `se = sqrt(phi / (x~'Wx~ + lambda))`. So `z = x~'y / sqrt(phi (x~'Wx~ + lambda))`, which goes to 0 as REML sends `lambda` to its null rail. The null-recovery ridge is designed to do exactly that under H0, and it does so in a large fraction of samples.

The reported p-value therefore collapses to about 1 on those samples. That is grossly conservative, and a conservative p-value is a bug just like an anti-conservative one.

The crate already documents why this test is invalid. `crates/gam-terms/src/inference/random_effect_test.rs:10-12` says:

> the coefficient Wald statistic `b̂ᵀV_b⁻b̂` is computed from a `b̂` that the penalty shrinks toward zero by an amount REML chose from the same data, so its `χ²_edf` reference is neither the right shape nor the right scale

A ridged linear slope is exactly a one-column variance component with the same boundary null, yet its summary row still uses that Wald test.

## Evidence

The data is the committed baseline `bench/pvalue_calibration/baseline/quick/{records.jsonl,report.md}`: the `linear` null `y ~ s(x1) + x2`, testing `x2`, at n=200 with 100 reps per family.

The harness computes `2*norm.sf(|estimate|/std_error)` from the same coefficient. This is the statistic `parametric_terms[...]["p_value"]` reports; that path differs only by using a t tail with df around 190 for Gaussian and Gamma.

| family | null reps with p > 0.999 | size@0.05 | KS D | power@0.05, gamfit | power@0.05, pyGAM (fixed lam / gridsearch) |
|---|---|---|---|---|---|
| gaussian | 69/100 | 0.020 | 0.690 | 0.74 | 0.83 / 0.85 |
| binomial | 72/100 | 0.000 | 0.748 | 0.67 | 0.81 / 0.82 |
| poisson | 64/100 | 0.010 | 0.648 | 0.68 | 0.82 / 0.83 |
| gamma | 65/100 | 0.030 | 0.658 | 0.76 | 0.88 / 0.89 |
| negbin | 73/100 | 0.000 | 0.737 | 0.69 | (pyGAM has no negbin) |

Records for gaussian seed 0 show the null p-value at 0.99997 and the alt p-value at 0.99996. Both mean `z` is about 3e-5, which is the `lambda` rail. pyGAM leaves the slope unpenalized. It is calibrated in every family (KS D 0.05-0.12) and has more power, so on this axis gamfit plainly loses head-to-head.

## Principled fix (no new machinery)

Test a ridged linear term the way `group()` blocks are already tested: with the variance-component score test of `gam_terms::inference::random_effect_test`.

- For one column, `T = u^2` with `u = x~' v` (the other columns projected out in the fit's `W_H` metric).
- The reference is `mu * chi2_1`, with `mu = x~' W_F x~`. When the scale is estimated, use the ratio against the unpenalized residual.
- This is the exact Rao score test of `beta = 0` (the 1-df F test for Gaussian). It cannot be shrunk by the penalty it tests, and it has no point mass at p = 1.

Concretely:

1. In `crates/gam-models/src/fit_orchestration/drivers/random_effect_test.rs:29/87`, request a `RandomEffectTermRequest { range }` for every linear range that owns a `LinearTermRidge` penalty (and each `BoundedShrinkage` range). Record the outcome beside the `group()` records, or in a sibling `FitArtifacts` vec.
   - A multi-column ridged term must pass its prior metric, `Σ = S⁺` of the function-mass ridge, rather than assume `Σ = I`.
2. In `parametric_term_summary_rows`, have a ridged row report that recorded score test (statistic and p-value) and its typed absence. The coefficient Wald test stays for unpenalized columns: the intercept, and explicit `double_penalty=false` terms, where it is the correct test.
3. The CLI summary shares `parametric_term_summary_rows`, so it is fixed by the same change.

## Stale harness and docs (same surface)

- `bench/pvalue_calibration/worker.py:25-30`/`283-292` and `docs/pvalues.md:25-27`/`133-136` both say gamfit reports no coefficient p-value. `summary().parametric_terms[...]["p_value"]` has existed in Python since ce4b4e7cee (2026-09-19 04:08), before the baseline ran at 4f17dc876b (10:03Z).
  - The harness's `coef` surface should read `summary.parametric_terms` by name and fail as `missing` when `p_value` is None.
- The baseline's factor and `re` rows ("NO P-VALUE; 100 unusable") predate 87ee889681 (13:34Z), which added the random-effect score test that now fills `smooth_terms[g]["p_value"]`. They need a re-run, not an issue.
- The harness should record `p_value_unavailable` when a row has no p-value. The baseline's `missing` only says `p_value=None`, which hides the typed reason.

## Acceptance

- Re-run `bench/pvalue_calibration` with plan `quick`, linear null. gamfit's parametric-row KS D should be comparable to pyGAM's (<= 0.12 at 100 reps) in every family. Size at 0.01/0.05/0.10 should fall within 2 MCSE of nominal, and no null mass should sit at p > 0.999.
- Add a unit test: for a Gaussian `y ~ s(x1) + x2` with `x2` independent of `y`, the ridged `x2` row's p-value equals the unpenalized partial F-test p-value to rounding.

SauersML:
Fixes #3573

By default every linear term carries the REML-selected `LinearTermRidge`. Its summary row reported the Wald ratio of the shrunk coefficient. Under the null REML sends λ to its rail, which drives that ratio to 0, so the p-value collapsed to about 1 on most null samples. The committed baseline had 69/100 Gaussian null reps with p > 0.999 (KS D 0.69).

A ridged slope is the one-column variance component that the `group()` rows already test, so this PR tests it the same way.

- **Score-test record.** `variance_component_test_records` scores the random-effect blocks and the ridged linear terms (`TermCollectionDesign::ridged_linear_ranges`, read off the emitted `penaltyinfo`) against a single `RandomEffectTestBasis`. The ridged-term records are stored in `FitArtifacts::linear_term_tests`.
- **Signed root.** `RandomEffectTest::signed_root` is `u/√(φ̂μ)` for a one-column term. Its law is N(0,1) when the scale is known and t on ν when it is estimated. With an estimated scale the test is exactly the partial F test of the unpenalized regression.
- **Summary row.** `parametric_term_summary_rows` fills a ridged row's statistic and p-value from the record. It matches the record by name and by global range. The row carries `test` (`wald` or `variance_component_score`). If the test could not be computed, the row carries a typed `pvalue_unavailable` reason instead. These fields are plumbed through the saved summary, the text table and the Python docstring.
- **Regression test.** `tests/inference/misc/sbc_ridged_linear_term_score_size_curve.rs` has two parts:
  - An exactness check. On Gaussian fits the row must match an independent partial-F computation done by orthogonalization.
  - Null size and KS gates for Gaussian, binomial and Poisson (200 reps each), plus a power control.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 1cf0beabc98c.

PR #3727, opened 2026-09-20
Royston-Parmar: publish the cone-truncated posterior and predict from it

### LINKED ISSUE #3575: survival/predict: Royston-Parmar posterior mean collapses to the plug-in with ~zero SE whenever one I-spline coefficient sits on its β≥0 bound
## Where (origin/main `1a3a498e3b`)

- `crates/gam-models/src/survival/predict.rs:839-900`: `for_each_survival_posterior_node`, the cone rule used for `SurvivalLikelihoodMode::Transformation` (Royston-Parmar) posterior-mean prediction. `SurvivalPosteriorIntegration::default_for` (`:1162`) sends every non-location-scale, non-marginal-slope fit to `SigmaPoint`.
- `crates/gam-models/src/fit_orchestration/fit.rs:2248` builds the RP covariance as the dense `H⁻¹` of the full penalized Hessian. `:2570` saves `constrained_posterior: None`, even though the fit imposes `lb[j] = 0` on every structural time column (`fit.rs:3450-3457`).

```rust
let mut scale = nominal_scale;              // √rank
for &j in cone_coords {
    let load = factorization.factor[[j, column]].abs();
    if load == 0.0 { continue; }
    let limit = posterior_mean[j].max(0.0) / load;
    if limit < scale { scale = limit; }
}
// nodes = β̂ ± scale · f_{·,column}
```

## What is wrong

**1. One active bound collapses almost the whole rule.**
- `factorize_psd_covariance` (`location_scale/moments.rs:93`) returns the eigen factor `f_{·,k} = √λ_k v_k`.
- The eigenvectors of a dense `H⁻¹` load essentially every coordinate, so `f_{j,k} ≠ 0` for almost every `k`.
- A converged RP fit whose monotone baseline has any flat segment has some `β̂_j = 0` exactly. The active-set PIRLS pins it there.
- Then `limit = 0 / load = 0` for every column that loads `j`. Those columns get scale 0, and both nodes of each such pair equal `β̂`.
- The consequences:
  - The reported "posterior mean" survival/hazard is the plug-in (MAP) value.
  - The published posterior SE and intervals are about 0.
  - This affects the whole prediction surface, not just the pinned coordinate.

**2. The rule is anti-conservative and biased, not "conservative".** The test `posterior_quadrature_radius_collapses_on_an_active_bound` (`predict.rs:6917`) pins this behaviour and calls zero spread "the conservative feasible answer". It also admits that "a truncated Gaussian at an active bound is ONE-SIDED and carries real mass, but no symmetric ± pair can express that". Zero reported uncertainty is anti-conservative, and SPEC says "Conservative is a bug just like anti-conservative" in any case.
- Even away from the wall, the rule is centred on `β̂` (the mode of the truncated law), not on its mean. SPEC: "Posterior mean must always be the default (never MAP)".
- The `(α_k/√rank)²·Vb` spread is a feasibility shrink. It is not the truncated law's covariance.

**3. The machinery for the correct answer already exists and is not wired up for RP.**
- `SurvivalPosteriorIntegration::TruncatedLaw` (`build_truncated_coefficient_draws`, `location_scale/truncated_moments.rs:176`, built on `gam_solve::constrained_posterior::ConstrainedPosteriorJointRule`) integrates the cone-truncated posterior exactly. It reads `fit.geometry.constrained_posterior`.
- The RP fit never populates that geometry, and `default_for` only asks for it for `LocationScale`.

## Principled fix

- Have the RP/transformation fit (`survival_unified_fit_result`) persist `ConstrainedPosteriorGeometry` for its structural lower-bound box `{β_j ≥ 0 : j < p_time_total}`, the same way location-scale does.
- Make `default_for` choose `TruncatedLaw` for `Transformation` whenever `truncated_survival_posterior_draws` is `Some`.
- Once every cone-truncated RP posterior goes through the truncated law, delete the fraction-to-boundary shrink from `for_each_survival_posterior_node`, together with the three tests that pin it (`..._radius_collapses_on_an_active_bound`, `..._clamps_a_roundoff_negative_cone_coordinate`, and the cone-shrinks-spread test). Keep no second approximation.
- Regression test: an RP fit with a flat baseline segment (an active `β_j = 0`). Its posterior-mean survival must differ from the plug-in, and its SE must be strictly positive and match a brute-force truncated-Gaussian Monte Carlo of the same law.

SauersML:
Fixes #3575

## Defect
The single-cause Royston-Parmar (transformation) fit enforces the I-spline baseline cone `beta_j >= 0` in the inner solve, but `survival_unified_fit_result` minted `constrained_posterior: None`. As a result:
- The posterior was the untruncated Gaussian at the mode.
- `default_for` sent the fit to SigmaPoint, whose fraction-to-boundary shrink collapses every node onto the mode as soon as one bound is active.
- So the "posterior mean" equalled the plug-in, and the SEs were about 0.

## Math
The inner solve minimizes `F(beta) = -l(beta) + 1/2 beta' S beta` over `{beta_j >= lb_j}`. At the mode `beta_hat`, with gradient `g` and Hessian `H`, the second-order expansion is the Gaussian `N(beta_unc, H^-1)`, where `beta_unc = beta_hat - H^-1 g`.
- At an interior mode `g = 0`.
- On an active bound, `g_j` is the bound's multiplier, and `beta_unc` lies outside the cone.

The posterior is this Gaussian restricted to `{A beta >= b}`, with one row `e_j' beta >= lb_j` per finite bound. It is built with the shared construction (`constrained_posterior_correction_from_covariance` then `ConstrainedPosteriorGeometry::with_moments`), the same one the custom-family and standard-GAM fits use:
- The fit publishes the truncated mean as its coefficients.
  - log-likelihood and deviance are re-read at the published mean, and the mode log-likelihood is kept on the geometry (gam#2921 convention).
  - REML score, penalized objective and penalty stay at the mode.
- The conditional covariance published is the truncated `Vb`, in the sum-of-Grams PSD form (#2705).
- The corrected covariance is built in three steps:
  1. The smoothing correction `C` is built from the untruncated `Vb`.
  2. `Vp = Vb + C`.
  3. `Vp` is truncated again, with its own lift and moments taken from the same centre. This is the convention optimizer.rs `apply_marginal_constraint_truncation` uses on main.

## Prediction
- `default_for` now sends Transformation fits to `TruncatedLaw` whenever `truncated_survival_posterior_draws` finds an active cone. That happens in conditional mode, where the saved covariance is the one the cone was built against.
- The Royston-Parmar branch replays each node of the factorized truncated law (`beta = beta_unc + G(u - E u) + eps`) through the plug-in survival predictor.
- The per-cell moments are then certified by `truncated_survival_surface_moments`. This function is renamed from `truncated_location_scale_surface_moments` and now serves both families.

## Not covered by this PR (will be commented on the issue)
- **Corrected-mode truncated-law prediction.** The corrected covariance differs from the conditional one, so the draws decline and prediction falls back to SigmaPoint on the truncated `Vp`. This is tied to the convention question in #3229.
- **Competing-risks truncated law.** `predict_competing_risks_with_posterior` still uses the cone shrink.
- **Deleting the shrink.** Once the two paths above move to the truncated law, the shrink and its pinning tests can go.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 4bbc14e2d250.

PR #3736, opened 2026-09-20
structure_evidence: reject malformed probe-planner inputs instead of reporting "undecidable"

### LINKED ISSUE #3734: structure_evidence: probe planner reports malformed inputs as "steering cannot distinguish" (None) or a zero/NaN budget
On origin/main (32560e9b), `crates/gam-terms/src/inference/structure_evidence.rs` has three input-handling defects in the probe planner.

**1. Invalid `alpha` is reported as "undecidable".** `plan_probe_for_contested_claim` (~line 907) passes `alpha` straight into `expected_resolution_budget(alpha, growth)?` (~line 864). For `alpha <= 0` or `alpha >= 1`, that returns `None`, and the planner returns `None`. The planner's docs, and the Python docstring in `gamfit/_structure_discovery.py`, define `None` as "the hypotheses agree on everything reachable inside the validity radius; the claim is undecidable by steering and needs a different instrument, which is a finding". So `gamfit.sae.plan_probe_for_contested_claim(..., alpha=1.5)` gives the caller a false scientific finding instead of an error.

**2. NaN `alpha` passes the check.** `alpha <= 0.0 || alpha >= 1.0` is false for NaN. `expected_resolution_budget(NaN, g)` therefore returns `Some(NaN)`, and the plan carries NaN budgets.

**3. NaN `current_log_e` reads as "already certified".** `(-(alpha.ln()) - current_log_e).max(0.0)` evaluates to `0.0` when `current_log_e` is NaN, because `f64::max` returns the non-NaN operand. The plan then reports `budget_remaining = 0`, which means "already across 1/alpha".

**Also:** `select_probe_by_expected_evidence` (~line 835) silently `continue`s past a probe whose response length does not match `fisher`. It also silently drops a non-finite score. If every probe is malformed, the result is again the "no probe discriminates" `None`. The PyO3 layer validates shapes, but the Rust API does not.

These violate SPEC's no-silent-fallback rule: a caller error is turned into a result.

**Fix:**
- All three functions return `Result<Option<..>, String>`.
- `Err` covers an `alpha` outside (0, 1) (NaN included), a NaN `current_log_e` or growth rate, a non-square `fisher`, a mis-sized probe, and a non-finite score.
- `Ok(None)` is kept only for its documented meanings: no discriminating probe, or non-positive growth.
- The PyO3 wrappers map `Err` to `ValueError`.

SauersML:
Fixes #3734

**Problem:**
- `plan_probe_for_contested_claim` returned `None` for an `alpha` outside (0, 1). The docs define `None` as "steering cannot distinguish the hypotheses", so a caller error came back as a scientific finding.
- NaN `alpha` got past the range check and produced NaN budgets.
- A NaN `current_log_e` was floored to `budget_remaining = 0`, which reads as "already certified".
- `select_probe_by_expected_evidence` silently skipped probes that were mis-sized or had a non-finite score.

**Change:** `select_probe_by_expected_evidence`, `expected_resolution_budget` and `plan_probe_for_contested_claim` now return `Result<Option<..>, String>`.
- They return `Err` for:
  - an `alpha` outside (0, 1), NaN included;
  - a NaN `current_log_e` or growth rate;
  - a non-square Fisher;
  - a mis-sized probe;
  - a non-finite score.
- `Ok(None)` keeps only its documented meanings: no discriminating probe, or non-positive growth.
- The PyO3 wrappers map `Err` to `ValueError`. The Python docstrings are updated to match.

**Tests:**
- New Rust unit test `probe_plan_rejects_malformed_inputs_instead_of_reporting_undecidable`. The existing unit tests now unwrap the new `Result`.
- New Python assertions in `test_kl_optimal_probe_design_reaches_python`.

---


SauersML:
BF-58 review: this is sound and only waits on its build. Its focused proof, `HN-16-3734-1`, has been queued since 08:37.

- **Merge.** The branch merges into current origin/main (d5f641efbd) without conflict (`git merge-tree` is clean).
- **Callers.** The three functions whose signatures change (`select_probe_by_expected_evidence`, `expected_resolution_budget`, `plan_probe_for_contested_claim`) are called only from `gam-terms/src/inference/structure_evidence.rs` and the PyO3 wrappers in `gam-pyffi/src/inference/inference_instruments.rs`. The PR updates both files.
- **Meaning of `Ok(None)`.** `Ok(None)` is now reserved for the documented outcome, "steering cannot distinguish the hypotheses". Malformed input returns `Err` and reaches Python as `ValueError`. That removes the "caller error reported as a scientific finding" defect.

Ready for the local merge train (MT-0, even PR numbers) once the proof is green.

---

PR #3737, opened 2026-09-20
gam-linalg: SymmetricMatrix::factorize is SPD-only on both storages (#3696)

### LINKED ISSUE #3696: gam-linalg: SymmetricMatrix::factorize accepts indefinite dense matrices but refuses indefinite sparse ones (storage-dependent contract; SPD gates are vacuous on dense storage)
## Defect

`crates/gam-linalg/src/matrix/symmetric.rs:82` (origin/main):

```rust
pub fn factorize(&self) -> Result<Box<dyn FactorizedSystem>, String> {
    match self {
        Self::Dense(mat) => {
            let factor = crate::utils::StableSolver::new()
                .factorize(mat)          // LLT -> unpivoted LDLT -> LBLT: accepts indefinite
                ...
        }
        Self::Sparse(mat) => {
            let factor = crate::sparse_exact::factorize_sparse_spd(mat)   // strict SPD simplicial LLT
                ...
        }
    }
}
```

The same method has two different mathematical contracts depending only on how the matrix is stored:

- **Dense storage:** the method goes through `factorize_symmetricwith_fallback` (LLT, then unpivoted LDLT, then Bunch-Kaufman LBLT). It succeeds on indefinite and even negative-definite matrices.
- **Sparse storage:** the method is a strict SPD Cholesky. It returns `Err` on the same matrix.

`factorize_spd()` (line 103) already exists as the strict variant, so the lenient dense arm of `factorize()` is the outlier.

## Why it matters (callers that treat `factorize()` as an SPD gate or SPD solve)

- **`gam-solve/src/reml/gradient_hessian.rs:3966` `effectivehessian`**
  ```rust
  let h = &pr.stabilizedhessian_transformed;
  if h.factorize().is_ok() { return Ok(h.to_dense()); }
  Err(EstimationError::ModelIsIllConditioned { .. })
  ```
  - The factor is thrown away. The call exists only as an admissibility check on H.
  - On dense storage the check is effectively vacuous: any finite symmetric H passes through LBLT.
  - On sparse storage the same H is refused.
  - So whether an indefinite (non-mode) H reaches the LAML log-determinant path in `prepare_dense_eval_bundlewithkey` depends on storage, not on the model.
- **`gam-solve/src/pirls/edf.rs:129` `calculate_edf`**
  - Its comment says: "Use SymmetricMatrix::factorize() which dispatches to sparse Cholesky for sparse Hessians and dense Cholesky for dense ones". That is false for dense storage.
  - EDF = p - tr(H^{-1} S) computed from an indefinite H is not an effective degrees of freedom, and on dense storage it is returned silently.
- Other callers with the same assumption:
  - `gradient_hessian.rs:5646`, `:8241`
  - `objective.rs:4284`, `:4319`
  - `gam-models/src/fit_orchestration/fit.rs:1615`
  - `gam-custom-family/src/warm_start.rs:417`
  - `gam-predict/src/linalg.rs:79` (the prediction precision backend, where an indefinite "precision" yields negative variances rather than an error)

This conflicts with SPEC: "In general, do not paper over solver issues." A penalized Hessian at a converged mode must be PD on the identified subspace. Letting an indefinite H through on one storage path, and failing it on the other, hides non-convergence rather than reporting it.

## Principled fix

Give `factorize()` one storage-independent contract:

1. Make `SymmetricMatrix::factorize()` SPD-only for both storages. The dense arm becomes the same unperturbed Cholesky plus derived pivot certificate that `factorize_spd` uses. Folding the two into one method also removes the duplicate.
2. Callers that genuinely need an indefinite solve (a saddle or KKT system, if any exist) should call an explicitly named `factorize_symmetric_indefinite()` backed by pivoted LBLT. They should not get it implicitly through dense storage.
3. Fix the comment in `edf.rs`.

This is related to #3519 and PR #3578, which change the dense fallback ladder itself (they remove the unpivoted LDLT rung and add an LBLT logdet). Those changes do not address the dense/sparse contract asymmetry described here. This change should land on top of #3578, or be coordinated with it, because both touch `StableSolver::factorize`.

SauersML:
Closes #3696.

## Defect
`SymmetricMatrix::factorize()` behaved differently depending on how the matrix was stored:
- **Dense storage** went through the `StableSolver` fallback ladder (LLT → LDLT → LBLT). It accepted indefinite matrices, and even negative-definite ones.
- **Sparse storage** used a strict SPD Cholesky and refused the same matrices.

Every caller treats the result as an SPD factor: `effectivehessian`'s admissibility check, `calculate_edf`, the survival and custom-family EDF traces, the IFT warm-start solves and the prediction precision backend. On dense storage that let an indefinite H through without any error.

## Fix
SPEC: "In general, do not paper over solver issues." / "Unnecessary choices and options should be deleted, not included."

- `factorize_spd` was already the strict variant. It becomes `factorize`, and the lenient method is deleted, so there is one contract.
- Both arms now run an unperturbed Cholesky, and every pivot must clear its own derived γ₂ₙ roundoff band.
- The sparse arm first canonicalizes any triangle convention, through the new `factorize_sparse_spd_certified`. Callers therefore keep their storage freedom.
- `StableSolver` / `factorize_symmetricwith_fallback` are untouched, so this does not overlap #3578.

## Checklist
- [x] Fold `factorize_spd` into `factorize`
- [x] Fix the `edf.rs` comment
- [ ] `cargo check --workspace --tests` and targeted tests (gam-linalg, gam-solve, gam-predict, gam-models, gam-custom-family)
- [ ] Regression test: an indefinite dense matrix is refused, just as the same matrix is when stored sparse

---
_Generated by [Claude Code](https://claude.ai/code/session_01A5dB3R13aBXY1dCVSuWMBv)_


SauersML:
Taken over by RF-1113-1; driving to merge.

---

SauersML:
BF-102 (oldest-first sweeper, partition 0): this PR for #3696 merges cleanly with main at head `2122b8474a`. Its proof runs (35505647341, 35505508828, 35505507706) are queued on that exact head, and that is all it is waiting on. It is ready for the local merge train (MT-1, odd PR number) once green. No re-dispatch is needed.

---

SauersML:
Taking over this stalled PR (no push since 10:04Z) as the wave8 session RF-0748-1. My plan:
- merge origin/main in (done locally, clean);
- run `cargo check --workspace --tests` and the targeted gam-linalg, gam-solve, gam-predict, gam-models and gam-custom-family tests over every `SymmetricMatrix::factorize()` caller;
- merge if it passes.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, acd2962ddb97.

PR #3740, opened 2026-09-20
row_kernel: walk the RowSet in the directional-derivative trace paths

### LINKED ISSUE #3716: row_kernel: directional-derivative operators' trace_projected_factor ignores the RowSet (full-data unweighted trace under an HT subsample)
## Defect

`RowKernelDirectionalDerivativeOperator` and `RowKernelSecondDirectionalDerivativeOperator` (`crates/gam-models/src/row_kernel.rs`) each carry `rows: RowSet`. The channels of one operator disagree on which matrix it is:

- `mul_vec`, `to_dense` and `projected_matrix` (both its BLAS-3 and small-n branches) reduce through `self.rows.par_reduce_fold`. They visit only the stored rows, each scaled by its Horvitz–Thompson weight.
- `trace_projected_factor`, `trace_projected_factor_cached`, `trace_projected_factor_with_jf` and `trace_projected_factor_tiled` loop over `0..kern.n_rows()` with no weight. They return the **full-data, unit-weight** trace `Σ_{all r} (J_r F)ᵀ T_r (J_r F)`.

Under `RowSet::Subsample`, `trace_projected_factor(F) != tr(Fᵀ · to_dense() · F)`. The trace override is documented as the same quantity computed faster ("tr(Fᵀ B F) in a single row pass"), so this is a correctness bug, not a representation choice.

## Reached in production

- `RowKernelHessianWorkspace::with_rows(kern, row_set_from_options(options, n))` builds the workspace from the options' `outer_score_subsample`. This happens for rigid BMS (`bms/custom_family_impl.rs` ~1441) and rigid survival marginal-slope (`survival/marginal_slope/custom_family_impl.rs` ~484).
- `auto_outer_subsample` defaults to `true`, and `maybe_install_auto_outer_subsample` installs a phase-1 HT mask at scale (see #3035).
- `inner_blockwise_fit.rs` keeps the subsample for families whose inner solve consumes it (`inner_consumes_subsample`: BMS and survival marginal-slope).
- `joint_newton::exact_newton_dh_apply` returns `workspace.directional_derivative_operator(..)` as the `D_β H[δβ]` drift. That drift feeds the outer LAML trace `½ tr(K·(B_i + D_β H[δβ_i]))` through `trace_projected_factor(_cached)`.

So in phase 1 the drift's trace is the full-data value, while the Hessian, gradient and every other channel of the same operator are HT estimates. The analytic outer gradient is then not the derivative of the objective it is paired with. It also pays full-`n` jet evaluations, which is the cost the subsample exists to avoid.

## Fix

Walk the operator's `RowSet` in the trace paths, exactly as the other channels do:
- Route a non-`All` row set through the tiled walk.
- Walk positions `0..rows.walk_len(n)` and build each tile's `J·F` with the existing `row_set_jacobian_tile` (runs of consecutive stored rows still use the structured GEMM).
- Scale each row's bilinear sum by its HT weight.

For `RowSet::All` the result is bit-identical (weight `1.0`, same tiles, same chunked sum). Add a test pinning `trace_projected_factor(_cached)` against `tr(Fᵀ · to_dense() · F)` on a weighted subsample for both operators.

Related but different: #3035 (the dense overrides' RowSet handling, fixed by #3249). That fix did not cover these trace overrides.

SauersML:
Fixes #3716

`RowKernelDirectionalDerivativeOperator` and `RowKernelSecondDirectionalDerivativeOperator` computed `trace_projected_factor` and `trace_projected_factor_cached` over every kernel row with unit weight. They ignored the operator's `RowSet`. Under a Horvitz-Thompson outer subsample, the outer LAML trace of the `D_β H[δβ]` drift was therefore a full-data sum. Every other channel of the same operator (`mul_vec`, `to_dense`) uses the weighted subsample estimate. The subsample is the default phase-1 path for BMS and survival marginal-slope, via `row_set_from_options` -> `RowKernelHessianWorkspace::with_rows`.

**Change** (`crates/gam-models/src/row_kernel.rs`):
- A non-`All` row set now routes through the tiled walk, in all four trace entry points.
- For both operators, `trace_projected_factor_tiled` now:
  - walks positions `0..rows.walk_len(n)`;
  - builds each tile's `J·F` with the existing `row_set_jacobian_tile` and checks its shape;
  - scales each row's bilinear sum by its HT weight from `rows.row_at`.
- Under `RowSet::All` the weight is exactly 1.0 and the positions are the rows, so the result is bit-identical to before.

**Test:** the new `gram_inner_contraction_walks_weighted_subsample` checks both operators, cached and uncached, against `tr(Fᵀ · to_dense() · F)` under `test_support::row_set_overrides::weighted_subsample`. The tolerance is the same relative 1e-10 as the existing `gram_inner_contraction_matches_reference`.

---


SauersML:

---

SauersML:
Withdrawn: this branch received a push at 11:20Z from another session, so W8-75 is standing down on it.

---

PR #3742, opened 2026-09-20
tests: gate the shared coverage/size verdict on both tails

### LINKED ISSUE #3534: tests: the shared coverage/size gate passes conservative surfaces by design, so 13 calibration tests can only catch miscalibration in one direction
## Defect

The shared calibration verdict in `crates/gam-test-support/src/calibration.rs` counts a **conservative** surface as a pass. The only failure it reports is anti-conservative. Every standing coverage and type-I size test builds on it, so none of them can fail when an interval over-covers or a p-value is too large (undersized).

`crates/gam-test-support/src/calibration.rs` on origin/main:

```rust
// :297-302
    /// Nominal level lies ABOVE the whole CI: the surface under-covers. This is
    /// the anti-conservative failure that gates the build.
    AntiConservative,
    /// Nominal level lies BELOW the whole CI: the surface over-covers. Reported
    /// with a named slack, not gated.
    Conservative,
// :322-323
    /// `true` unless the surface is anti-conservative. Conservative and
    /// calibrated both pass (over-coverage is safe); only under-coverage gates.
// :383
        passed: class != CoverageClass::AntiConservative,
```

Every consumer copies that one-sided check. All of these gate only on `verdict.class == CoverageClass::AntiConservative`:

| file:line | surface |
|---|---|
| tests/inference/misc/sbc_family_predictive_interval_coverage.rs:295 | family predictive intervals |
| tests/inference/misc/sbc_gaussian_mean_band_covariance_modes.rs:222 | Gaussian mean band, per covariance mode |
| tests/inference/misc/sbc_gaussian_predictive_interval_coverage.rs:192 | Gaussian predictive interval |
| tests/inference/misc/sbc_gaussian_smooth_band_coverage.rs:220 | smooth credible band |
| tests/inference/misc/sbc_glm_binomial_band_coverage.rs:218 | binomial band |
| tests/inference/misc/sbc_glm_poisson_band_coverage.rs:175 | Poisson band |
| tests/inference/misc/sbc_location_scale_predictive_coverage.rs:251 | location-scale predictive |
| tests/inference/misc/sbc_multinomial_prediction_interval_coverage.rs:153 | multinomial prediction interval |
| tests/inference/misc/sbc_multinomial_smooth_significance_size_curve.rs:110 | multinomial smooth-significance **p-value size** |
| tests/inference/misc/sbc_skovgaard_rstar_size_curve.rs:159 | Skovgaard r* **p-value size** |
| tests/survival/survival/sbc_survival_prediction_se_coverage.rs:231, :410 | survival prediction SE |
| tests/survival/survival/sbc_survival_probability_band_coverage.rs:252 | survival probability band |

The same one-sided shape appears in two Python size tests. Their bound is hand-picked and they accept an FPR of 0, which is what a test that never rejects would give:
- `tests/bug_hunt_smooth_significance_ref_df_floor_and_null_fpr_test.py:101` `assert fpr <= 0.15` (40 seeds, alpha = 0.05; the test is named `test_null_false_positive_rate_is_calibrated`)
- `tests/bug_hunt_smooth_significance_ref_df_tracks_edf_not_basis_dim_test.py:183` `assert fpr <= 0.15`

## Why it is wrong

P-values and intervals must be calibrated, and a conservative result is a bug just as an anti-conservative one is. A conservative interval is too wide, and an undersized test has lost power. The harness docs call over-coverage "safe", and that turns a real defect into a report nobody reads. In practice the report is never printed either: the consumers above drop the `Conservative` verdict without logging it.

SPEC: "XFAIL pattern on tests is never allowed. A failing test should always indicate problematic behavior." A verdict that sees miscalibration and still returns `passed = true` is a soft XFAIL for half of the failure space.

These gates have no statistical reason to be one-sided:
- The Wilson interval in `audit_coverage` is already **two-sided** at `COVERAGE_FALSE_POSITIVE_RATE` (`z = Φ⁻¹(1 − α/2)`, :361). Gating on both tails keeps the per-level false-positive rate at exactly the constant the harness already derives. No tolerance changes.
- The coverage gates draw the truth from the model's own prior (`CoverageModel::draw_prior`). Under that SBC construction a correct posterior interval covers at exactly its nominal level, so over-coverage is evidence of a wrong posterior (for example an inflated covariance, a double-counted correction, or a wrong dispersion).
- The size curves simulate under an exact null, where a calibrated p-value is Uniform(0,1). An undersized test is miscalibrated.

## Principled fix

1. In `calibration.rs`, set `passed: class == CoverageClass::Calibrated`, delete the "over-coverage is safe" docs, and make `slack()` report both tails.
2. Have every consumer gate on `!verdict.passed`, or `class != Calibrated`, with a message that names the direction. Do not re-implement the one-sided comparison per file.
3. In the multinomial size curve, stop dropping replications that report no p-value (`sbc_multinomial_smooth_significance_size_curve.rs:76-81`, which allows up to 50% to be dropped). Under the null, a replication whose smooth collapsed onto its null space is a **non-rejection**. Dropping it conditions the sample on the non-degenerate fits and hides an undersized test. Either the surface reports p = 1 for a collapsed term, or the replication counts as a non-rejection.
4. Replace the two Python `fpr <= 0.15` bounds with the same two-sided Wilson verdict at the harness FPR. Python tests cannot import the Rust harness, so the natural home is a small `tests/` helper that mirrors `audit_coverage`.

Expect this to turn some of these tests red. That is the point: each one that goes red is a miscalibrated surface that needs its own root-cause issue. None of them should be excluded or loosened to stay green.

SauersML:
Fixes #3534

WIP. The plan:
1. `audit_coverage`: `passed = (class == Calibrated)`. The Wilson CI is already two-sided at `COVERAGE_FALSE_POSITIVE_RATE`, with z = Φ⁻¹(1 − α/2). Failing either tail therefore keeps the per-level false-positive rate at exactly α, and no tolerance changes. The "over-coverage is safe" docs are deleted.
2. Each consumer gates on `!verdict.passed`, with a message that names the direction.
3. Multinomial size curve: a replication whose smooth collapsed counts as a non-rejection instead of being dropped.
4. Python size tests: replace the hand-picked `fpr <= 0.15` with a two-sided Wilson verdict at the harness FPR.

---


SauersML:
Taken over by RF-1113-4; driving to merge.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 25ffc92c6dd4.

PR #3747, opened 2026-09-20
reml: delete the always-None EvalShared::firth_dense_operator_original

### LINKED ISSUE #3730: reml: EvalShared::firth_dense_operator_original is always None, so its sparse-builder Firth build and every cached branch are dead
**Area:** crates/gam-solve/src/reml (the sparse exact-SPD eval bundle and its Firth consumers)

`EvalShared::firth_dense_operator_original` (reml/mod.rs:4858) is described as a "Cached FirthDenseOperator built from the original (non-reparameterized) design matrix, for use by the sparse evaluation path". In practice it can never hold a value:

- `prepare_dense_eval_bundlewithkey` (gradient_hessian.rs:6550) always sets it to `None`.
- `prepare_sparse_eval_bundlewithkey` (gradient_hessian.rs:6653-6670) only builds it when `reml_robust_jeffreys_link(&self.config)` is `Some`. That function returns `None` unless `config.firth_bias_reduction` is set (state_caches.rs:1034).
- The sparse builder has exactly one caller, `prepare_eval_bundlewithkey_and_row_policy` (gradient_hessian.rs:4409), and that caller only reaches it when `select_reml_geometry` returns `RemlGeometry::SparseExactSpd`.
- `select_reml_geometry` returns `DenseSpectral` with the reason `firth_bias_reduction_active` for every Firth fit (inner_strategy.rs:67-97), before any sparse check.

The sparse-builder block is therefore unreachable, and its comment "Built at every problem scale, for the same reason as the dense bundle" is false. Every consumer branch on `Some(cached)` is dead as well:

- gradient_hessian.rs:2308
- hyper.rs:2286, 2500, 2666
- objective.rs:790
- the `|| bundle.firth_dense_operator_original.is_some()` terms at objective.rs:1562 and 1946
- the first arm of the `structural_rank` chain at objective.rs:1692

SPEC: "Unnecessary choices and options should be deleted."

**Fix:** delete the field, the sparse-builder block and the dead `Some(cached)` arms. Where the consumers build the operator on demand, they keep doing so, so no reachable behaviour changes.

SauersML:
Fixes #3730

Only the sparse bundle builder could set this field, and it did so only when `reml_robust_jeffreys_link` is `Some`, which requires `firth_bias_reduction`. `select_reml_geometry` sends every Firth fit to `DenseSpectral` before the sparse builder can run, so the field was always `None`. Dead as a result:
- the sparse builder's Firth operator build
- every `Some(cached)` consumer arm: gradient_hessian, hyper ×3, objective
- both inner-KKT `firth_active` terms
- the first `structural_rank` arm

This PR deletes the field, that build and those arms. The consumers already build the operator on demand and keep doing so, so reachable behaviour is unchanged.

SPEC: "Unnecessary choices and options should be deleted."

---


SauersML:
Reviewed at head fca97b8fe6. It merges cleanly with current main (`git merge-tree` gives no conflicts).

The dead-code premise holds on main:
- `reml_robust_jeffreys_link` returns `None` unless `config.firth_bias_reduction` is set (state_caches.rs:1034-1039).
- `select_reml_geometry` routes every `firth_bias_reduction` fit to `DenseSpectral` (inner_strategy.rs:40ff).
- The only caller of `prepare_sparse_eval_bundlewithkey` runs after that decision on the same `self.config` (gradient_hessian.rs:4420 then 4443).

So `firth_dense_operator_original` was always `None`. The removed `Some(cached)` arms and the `|| ..._original.is_some()` terms were unreachable, and reachable behaviour is unchanged.

Its only check, focused proof HN-12-3730-1, is still queued. Once that build is green, this is ready for the merge train (MT-1, odd PR).

---

SauersML:
Closing as superseded. #3748 made the same deletion for #3730 and was merged as 39a4ed45, so #3730 is closed.

---

PR #3751, opened 2026-09-20
survival/location_scale: read residual-law derivatives from the shared inverse-link jet

SauersML:
Survival residual derivatives now use the shared inverse-link jets. This removes duplicated Gaussian polynomial products that produced NaN at infinite arguments and lost representable subnormal derivatives when the density underflowed. The duplicate fourth-derivative dispatcher is removed, its sole row-kernel caller uses the shared dispatcher directly, and residual-to-standard-link mapping derives from one component owner.

Validation on MSI: complete gam-models production source compiled into a focused binary against coherent current-main runtime dependencies. Three added regressions passed: independent Hermite values, infinite/NaN limits for Gaussian/Gumbel/logistic, and nonzero Gaussian subnormal derivatives after density underflow. Complete gam-models unit metadata passed with warnings denied. A reference audit found no remaining call to the removed dispatcher. The removal ledger was appended against fresh main without reformatting older entries.


SauersML:
BF-65 (green-run lander): proof run 35500148943 (issue key HN-6-3745-1) on head bc2867a1 hit the 1200s dev compile budget ("gam-models did not finish compiling within 1200s … infrastructure outcome"). No test ran, so this says nothing about the code. I am not re-dispatching it.

Exact filters for the local merge train (MT-1, odd PRs):

```
cargo test -p gam-models --all-features --lib -- survival::location_scale
```

---

PR #3753, opened 2026-09-20
survival location-scale: truncate the smoothing-corrected posterior against the fit's cone (#3524)

### LINKED ISSUE #3524: survival location-scale predict: under the published smoothing-corrected covariance, a cone-truncated link-wiggle posterior silently falls back to the untruncated Gaussian
Found by audit agent HN-5 while reading `crates/gam-models/src/survival/location_scale` on origin/main `93285a585b`.

## Defect

For a location-scale survival fit whose link-wiggle cone (`beta_w >= 0`) is active at the mode, the #2679 truncated-law machinery is used **only when the caller passes exactly the conditional covariance**. For any other covariance, it returns `Ok(None)`, and both prediction paths then price the posterior with the plain Gaussian `N(beta_hat, V)`. The code's own comment calls that Gaussian "wrong in kind".

`crates/gam-models/src/survival/location_scale/truncated_moments.rs:240-246` (`truncated_law_pieces`):

```rust
// The cone geometry describes the conditional posterior. If the caller
// selected another covariance, the mixture below would carry the cone of
// one law and the spread of another.
match fit.beta_covariance() {
    Some(conditional) if conditional == covariance => {}
    _ => return Ok(None),
}
```

`crates/gam-models/src/survival/location_scale/moments.rs:527-545` (`exact_survival_response_moments`) treats that `None` as "no cone":

```rust
// ... The Gaussian below is the
// moment-matched normal of a cone-truncated law, which is wrong in kind:
// its error is a floor rather than a rate, and it puts mass on coefficient
// vectors the fit excluded. ... `None` means there is no cone the
// reported covariance was truncated against, and then the Gaussian rule
// below is not an approximation of the posterior but IS the posterior.
if let Some((first, second)) = truncated_survival_response_moments(...)? { return Ok(...) }
// falls through to the Gaussian GH rule
```

The library path does the same thing. In `crates/gam-models/src/survival/predict.rs:1169-1173`, `SurvivalPosteriorIntegration::default_for` picks `SigmaPoint` whenever `truncated_survival_posterior_draws(model, covariance_mode)` is `None`. The cause is again the covariance not being the conditional one.

## Why this is on the default path

- Penalized LS fits publish `covariance_corrected`. This has been the case since #2677, pinned by `location_scale/smoothing_corrected_tests.rs:177-300`.
- So `UnifiedFitResult::published_covariance_mode()` (`crates/gam-solve/src/model_types/result_types.rs:6619`) returns `SmoothingCorrected`.
- The CLI takes that as its default: `resolved_covariance_mode` (`crates/gam-cli/src/main/model_summary.rs:82`).
- `gam predict` on a penalized LS fit with a link wiggle then does the following (`crates/gam-cli/src/main/run_predict.rs:2710-2718`):
  1. `covariance_from_model(model, SmoothingCorrected)`, which returns `V_c`.
  2. `predict_survival_location_scalewith_uncertainty(..., &V_c, true, ...)`.
  3. `truncated_law_pieces`, which sees `V_c != V_cond` and returns `None`.
  4. The Gaussian rule over `N(beta_hat, V_c)`. This rule places posterior mass on `beta_w < 0`, which the fit excluded.

In short, the default posterior-mean survival curve and its response SD for exactly the fits #2679 targeted come from the law #2679 says is wrong. Nothing reports this: no error and no provenance flag. Only an explicit `--covariance-mode conditional` gets the truncated law.

This goes against these SPEC lines:
- "The default choices should give the best model for the user."

The comment at `moments.rs:536-537` ("`None` means there is no cone") is false for this branch.

## Principled fix

Build the truncated law for the smoothing-corrected covariance instead of dropping the cone. `gam-predict` already has this construction: `constrained_law` (`crates/gam-predict/src/lib.rs:1951`, the #2784 convention). It truncates the ambient Gaussian inflated by the correction, `trunc(TN(c, Sigma + C))`, against the same constraint rows. `truncated_law_pieces` already reconstructs the ambient `Sigma` from `Sigma_pi + G Delta G^T`.

For the corrected mode, the fix would:
- add the lifted correction `C` to that ambient;
- recompute the retained normal block `A (Sigma + C) A^T`;
- feed the result to `ConstrainedPosteriorJointRule::new`.

Which definition of `C` to use (the truncated-frame `V_cond + C` or the ambient-inflation convention) is the open question in #3229. That question should be settled once, and this path should read the same `C` that gam-predict reads.

Until that lands, if the truncated corrected law cannot be formed, the corrected-mode prediction should refuse with a typed error. That is what `predict_survival_truncated_law_posterior_mean` already does when it has no cone law (`predict.rs:1599-1607`). Silently switching to the Gaussian should not remain an option.

Regression test: take a penalized LS fixture with an active wiggle cone, predict under the default (published) covariance mode, and assert that the truncated rule was used. Either assert integration provenance, or assert that the posterior mean differs from the Gaussian-rule mean by more than the rule's certified error.

Related: #2679 (truncated-law survival moments), #3229 (conventions for C at constrained modes), #2677.

SauersML:
Fixes #3524

WIP. `truncated_law_pieces` currently returns `None` whenever the requested covariance is not the conditional one. As a result, a smoothing-corrected prediction falls back silently to the untruncated Gaussian even when the fit has an active cone. This PR builds the corrected ambient law Σ_c = Σ_cond + C and re-derives the truncated correction from it, following gam-predict's `constrained_law`. Where no law can be formed, it refuses with a typed error instead of returning `None`.

---


SauersML:
LC-23 status: this is still being driven. I merged origin/main (dd285cdaec) in at 4bdae48fea, and both ledger guards pass. Focused run 35500406204 hit an infrastructure timeout: gam-models lib tests did not compile within 1200s, so no test ran. I re-dispatched the focused proof (run 35509327824, package gam-models, filter `truncated_moments`). A local `cargo check -p gam-models --tests` at 37cbe4f exited 0.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 40d7fe753601.

PR #3754, opened 2026-09-20
gam-sae support_term: pose the row trust-region step on the tangent space (#3746)

### LINKED ISSUE #3746: crates/gam-sae/src/manifold: support coordinate_sweep solves the row trust-region step in the ambient chart, projects it afterwards, and repairs a lost ascent direction with a full-radius steepest-descent fallback
## Where

`crates/gam-sae/src/manifold/support_term.rs:7091-7135` on origin/main (`ac126ad0c6`), in `SaeSupportSparseTerm::coordinate_sweep`:

```rust
let delta = opt::solve_psd_trust_region(curvature.view(), rhs_vector.view(), trust_radius)
    .map_err(...)?;
...
let mut delta = delta;
self.assignment.project_row_tangent(row, coords_row, delta.as_slice_mut()...)?;
let mut directional = rhs_vector.dot(&delta);
if !(directional > 0.0) {
    // Projection and the Gram solve do not commute, so the projected step is
    // not guaranteed to remain an ascent direction for the right-hand side.
    // Steepest descent within the tangent space is one by construction, and
    // is a real step rather than a failed row.
    let mut fallback = rhs_vector.to_owned();
    self.assignment.project_row_tangent(row, coords_row, fallback...)?;
    let norm = fallback.dot(&fallback).sqrt();
    if !(norm > 0.0) { return Ok(0.0); }
    delta = fallback * (trust_radius / norm);
    directional = rhs_vector.dot(&delta);
}
```

## Why it is wrong

`retract_row_coords` only travels the tangent part of a step, so the step has to live in the row's feasible update space `range(P)`. That space is `I - ttᵀ` on a sphere block, and an interval endpoint holds its coordinate when descent points outward. The code does not solve the trust-region subproblem on that space. It solves it in the full ambient chart and then projects the answer. `P·argmin_ambient` is not `argmin_{range(P)}`: it is neither the tangent model's minimizer nor, in general, a descent direction (the comment in the code admits this).

When the projected step comes out non-ascent, the code swaps in a different step: projected steepest descent scaled to the **full** trust radius, whatever the curvature says. That is a reactive repair of a wrongly posed subproblem.

> "In general, do not paper over solver issues."

It also discards the exact-Newton / Gauss-Newton model that #2576 added for this row solve, on exactly the d >= 2 sphere/interval rows where that model matters.

## Principled fix

Pose the trust-region model on `range(P)`: curvature `P H P`, right-hand side `P·rhs`.

- `(PHP + λI)` maps `range(P)` to itself, so the exact PSD TR solution lies in the travelled space. Its norm is bounded by the radius in the metric the retraction uses.
- The certificate `rhsᵀδ = (P rhs)ᵀ(PHP + λI)⁺(P rhs)` is positive by construction. The fallback disappears, and a negative certificate stays a hard error.
- For interval endpoints, `P` is linearized at the descent direction: hold the coordinate iff descent points outward. That is the velocity-projection rule applied to `sign(rhs_j)·e_j`, not the sign-reversed gradient projection, which zeroed real inward components according to the existing comment.
- A small `SaeAssignmentState::row_tangent_projector` builds `P` from each atom's `LatentManifold::project_to_tangent`.

SauersML:
Fixes #3746

`SaeSupportSparseTerm::coordinate_sweep` used to solve the row trust-region subproblem in the ambient chart and then project the result onto the tangent space that `retract_row_coords` travels. When the projected step lost its ascent certificate, the code swapped in a projected steepest-descent step scaled to the full trust radius. That is a reactive repair of a wrongly posed subproblem.

**Change**

- New `SaeAssignmentState::row_tangent_projector(row, coords, descent)`. It builds the symmetric block-diagonal projector `P` of the row's linearized feasible update space from each atom's `LatentManifold::project_to_tangent`:
  - sphere blocks: `I − ttᵀ`
  - flat and periodic blocks: identity
  - interval endpoints: the coordinate is held iff the descent direction points outward
- `coordinate_sweep` now solves `opt::solve_psd_trust_region(P H P, P rhs, radius)`.
  - `(PHP + λI)` preserves `range(P)`, so the solution is the travelled step.
  - `rhsᵀδ = (P rhs)ᵀ(PHP+λI)⁺(P rhs) > 0` by construction. This still holds off-manifold, since `P` stays symmetric. Clipping an outward component at an interval endpoint can only raise it.
  - The steepest-descent fallback is deleted. A negative certificate is still a hard error.
- Flat charts are unchanged, because `P = I` there.

---


SauersML:

---

PR #3758, opened 2026-09-20
event-history static_state: place the static-frailty mode to the score's rounding floor

### LINKED ISSUE #3755: event-history static_state: static-frailty mode placement uses a fixed 24 Newton steps, 40 halvings, a 16·eps acceptance band and an absolute 1e-8 score gate
**Where:** `crates/gam-event-history/src/static_state.rs` on main, in `Statistics::place`, `Statistics::ascend` and `Statistics::grid_at`.

**SPEC:**

**Defect.** The static-frailty posterior grid is placed at the mode of `ln φ L`. Four constants, none derived, control that placement:

1. `place` calls `values.ascend(&loadings, vec[0.0; atoms], 24)`. This is a fixed budget of 24 Newton steps, and the loop does not test whether the mode has been reached:
   - A subject whose ascent needs more steps is handed to `grid_at` unconverged.
   - A subject that converges in 3 steps still runs all 24.
2. Each step is halved at most 40 times. Past that, the error is a generic "did not converge", even when the ascent had in fact stalled at the objective's rounding level.
3. Proposals are accepted if `proposed >= value - 16·eps·(1+|value|)`. This is an arbitrary band that lets a step *lower* the objective.
4. `grid_at` rejects the mode if any `|g_k| > 1e-8`. The `1e-8` is an absolute score gate that ignores the scale of the score. The score is `linear_k − z_k − Σ_d rate_d a_dk`, and it grows with the event count and the hazard mass. The result:
   - For a heavy subject, a mode resolved to rounding can be refused.
   - For a light subject, the gate is far looser than the arithmetic can resolve.

**Fix.**
- Derive a per-component rounding floor for the score at `z`. It is built from three parts:
  - summation growth `γ_(marks+3)` over the score's terms;
  - the `exp` conditioning of each rate, `γ_(2·atoms+1)` of the exponent's spread;
  - one representable step in `z`, carried through the precision row.
- Use that floor both as the ascent's stopping rule and as the `grid_at` gate.
- The ascent runs in two phases:
  - A damped Newton phase that requires a strict increase. It ends because the objective takes finitely many values.
  - A full-Newton polish phase that requires the floor-scaled score to strictly shrink. If it stops shrinking above the floor, placement fails with an explicit non-convergence error rather than passing through.

SauersML:
Fixes #3755

This replaces the magic constants in the static posterior grid placement (fixed 24 Newton steps, 40 halvings, the 16·eps acceptance band, the absolute `|g| <= 1e-8` gate) with a derived score rounding floor.

**The floor (`score_floor`)** is computed per score component at `z` and sums three terms:
- `γ_(marks+3)` times the score's term magnitudes;
- the `exp` conditioning of each rate: `rate·|a_k|·(γ_(2·atoms+1)·spread + eps)`;
- one representable step in `z`, carried through the precision row.

It is doubled, because a Newton step taken from a score read to the floor lands within one floor of the mode.

**`ascend`** runs in two phases:
- A damped Newton phase. Each step is halved until it strictly raises the objective. The phase ends when the score is within the floor, or when halving stops moving the point.
- A full Newton polish phase. Each step must strictly shrink `score_excess`. If a step stops shrinking it above the floor, `ascend` returns an explicit non-convergence error.
- A non-finite objective or step is also an error.

**`grid_at`** now gates on the same floor, replacing the 1e-8 check.

---


SauersML:

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 4c3ebcc7ca36.

PR #3762, opened 2026-09-20
Remove the unapplied analytic-penalty weight schedule (#3689)

### LINKED ISSUE #3689: gam-terms analytic_penalties: weight_schedule is accepted but never applied (empty impl_scalar_apply_schedule, no apply_schedule caller), so a scheduled penalty runs at w_start forever
## Defect
The public `ScalarWeightSchedule` / `weight_schedule` API is accepted end to end:
- Python `gamfit/_penalties.py`: the `weight_schedule=` attribute and `set_weight_schedule`;
- gam-pyffi `geometry_ffi.rs`;
- `crates/gam-models/src/fit_orchestration/descriptors.rs::descriptor_weight_schedule` (around line 189).

It is never applied. On origin/main:

1. `crates/gam-terms/src/analytic_penalties/penalty_trait.rs:407`. The macro that its doc says emits "the standard `AnalyticPenalty::apply_schedule` override" has an empty body:
   ```rust
   macro_rules! impl_scalar_apply_schedule {
       ($field:ident) => {
       };
   }
   ```
   So every `impl_scalar_apply_schedule!(weight);` expands to nothing. That covers ard, block_sparsity, harmonic_roughness, isometry, nuclear_norm, orthogonality, row_precision, scad_mcp, sparsity, and the others. Each type then falls back to the trait default `apply_schedule` (penalty_trait.rs:375), which only asserts `iter < 1_000_000`.
2. Nothing in the workspace calls `apply_schedule` or `ScalarWeightSchedule::step`. `git grep -n -E "apply_schedule|\.step\(\)" -- crates/` finds only the trait definition, plus unrelated RNG and precision `step()` calls. The registry also stores penalties as `Arc<..>` (`AnalyticPenaltyKind::X(Arc::new(penalty))`), so a `&mut self` schedule update could not reach them anyway.
3. `impl_with_weight_schedule!` (penalty_trait.rs:396) seeds `weight = schedule.current_weight(schedule.iter_count)`, which is `w_start` for a fresh schedule. What a user actually gets:
   - The descriptor's own `weight` is silently replaced by `w_start`.
   - The ramp to `w_end` never happens.
   - `geometric`, `linear` and `reciprocal_iter` all behave identically, as a fixed weight of `w_start`.
4. `NestedPrefixPenalty::with_weight_schedule` (`nested_prefix.rs:127`) does not even seed. It stores the schedule and changes nothing, so a nested-prefix schedule is a pure no-op. `nested_prefix` also has no `impl_scalar_apply_schedule!`.

The schedule still takes part in the REML state-cache hash (`crates/gam-solve/src/reml/state_caches.rs:341,363`). A dead parameter therefore changes cache keys but never the objective.

This is a silent drop of a user-supplied configuration value (SPEC: no silent fallbacks), and it is dead code across roughly 20 penalty types.

## Fix direction (design decision needed)
An outer-iteration-dependent penalty weight makes the REML/LAML objective non-stationary. Any converged fit would then depend on the iteration count, which conflicts with "converged fits only". The options are:

- **(a) Remove the feature.** Make `descriptor_weight_schedule`, the pyffi `weight_schedule` setters and `set_weight_schedule` return a loud error ("weight schedules are not supported: REML selects the analytic-penalty strength through its learnable log-weight"). Then delete `ScalarWeightSchedule`, `weight_schedule` fields, `impl_with_weight_schedule!`, `impl_scalar_apply_schedule!` and the default `apply_schedule`.
- **(b) Wire it properly.** Implement the macro body as `if let Some(s) = self.weight_schedule.as_mut() { self.$field = s.step(); }`. Then call it from the outer driver at a well-defined point, restarting convergence after the final ramp step so the reported fit is a converged fit at `w_end`.

(a) is the smaller, SPEC-aligned option. Until one of these lands, the doc on `ScalarWeightSchedule` ("ramps toward the target weight during REML outer iterations") is false.

SauersML:
Fixes #3689 (option a).

`ScalarWeightSchedule` / `weight_schedule` was accepted from Python, the FFI and the JSON descriptor, but it was never applied:
- `impl_scalar_apply_schedule!` expanded to nothing.
- Nothing called `apply_schedule`.
- The registry holds penalties behind `Arc`, so a `&mut self` update could not have reached them anyway.

The result was that a scheduled penalty ran at `w_start` for the whole fit, silently replacing the descriptor's `weight`. The nested-prefix builder stored the schedule and did nothing with it. The schedule still entered the REML state-cache hash, although it never changed the objective. An outer-iteration-dependent weight would also make the REML/LAML objective non-stationary, so the fix is to remove the feature rather than implement it.

## Removed
- **gam-terms:**
  - `ScalarWeightSchedule`, the trait-default `apply_schedule`, and both macros;
  - the `weight_schedule` field, initializers and builders on every analytic penalty;
  - the `ScalarWeightSchedule` re-export.
- **gam-models `descriptors.rs`:** `descriptor_weight_schedule`, every `with_weight_schedule` application, and `"weight_schedule"` in every allowed-key list.
- **gam-solve `state_caches.rs`:** the schedule cache-hash helpers and their call sites.
- **gam-sae:** the schedule copy for each fit.
- **gam-pyffi `geometry_ffi.rs`:**
  - the field, `weight_schedule=` keyword, `set_weight_schedule` and repr part on 17 penalty classes;
  - both schedule-descriptor converters.
- **Python:**
  - the `ScalarWeightSchedule` dataclass and its export;
  - the wrapper's `set_weight_schedule`;
  - the `_rust.pyi` stubs;
  - the docs API entry;
  - the packaging names list.

## Behaviour
A descriptor that still carries `weight_schedule` is refused with `penalties[i].weight_schedule is not a recognized field for this analytic penalty`, and the Python keyword raises `TypeError`. Penalty strength is the REML-selected log-weight (`weight="auto"`) or a fixed positive `weight`. `GumbelTemperatureSchedule` and `ScheduleKind` are unchanged.

## Tests
- Rust: `descriptors::tests::rejects_weight_schedule_as_an_unrecognized_field`.
- Python: `tests/test_penalty_weight_schedule_removed.py`.

---


SauersML:
BF-53 review (partition sweep for #3689): the diff is sound against SPEC. It deletes the unapplied `ScalarWeightSchedule` machinery end to end: the trait default, every per-penalty `with_weight_schedule`, the descriptor parser, the FFI/pyi surface and the Python kwarg. A `weight_schedule` key is now refused by the descriptor key allow-lists instead of being silently run at `w_start`, and `tests/test_penalty_weight_schedule_removed.py` pins that. `git merge-tree` against current main (6ff5fb0) is conflict-free. The only thing it waits on is Validate One run 35500202630, which has been queued since 08:40. Ready for the local merge train (MT-0, even PR) once that run is green.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 9e5b25df647f.

PR #3774, opened 2026-09-20
single source of truth: take digamma from gam_math::special in gam-solve and gam-terms

### LINKED ISSUE #3770: single source of truth: gam-solve and gam-terms take digamma from statrs while trigamma comes from gam_math::special
The docs for `gam_math::special::digamma` (crates/gam-math/src/special.rs:397) say:

> These polygamma functions are the workspace's one implementation. Each evaluates the same kernel as [`polygamma_stack`] and the `jet_tower` derivative stacks, so a scalar value and the matching stack entry are bit-identical.

Production code does not follow that. Three modules import `statrs::function::gamma::digamma` and pair it with `gam_math::special::trigamma` in the same formula:

- `crates/gam-solve/src/pirls/prelude.rs:64`: `pub(crate) use statrs::function::gamma::{digamma, ln_gamma};`. This digamma feeds the Beta mean score and its saturated root (`deviance.rs` `beta_scaled_digamma`, 1404, 1454; `family_state.rs:953`), the Gamma shape score (`dispersion.rs:112`) and the NB θ score (`dispersion.rs:487/507`). The matching curvatures use `gam_math::special::trigamma`, so a score and its derivative come from two different ψ kernels.
- `crates/gam-solve/src/mixture_link.rs:14`: `use statrs::function::gamma::digamma;`, used at 2912–2920, 3281–3283 and 3495–3497. The same file already calls `gam_math::special::digamma` at 3080/3086/3144 for the same Beta-link quantities. So the file uses two digamma implementations.
- `crates/gam-terms/src/analytic_penalties/ordered_beta_bernoulli.rs:2`: `use statrs::function::gamma::{digamma, ln_gamma};`. The marginal score `-ψ(M+a)+ψ(N−M+1)` at 207 uses statrs, while its `score_derivative` on the next line uses `gam_math` trigamma.

SPEC: "It's required that the CLI, Python library, and Rust library are unified, have a single source of truth".

Every argument at these sites is a positive shape, mass or θ, which is the domain `gam_math::special::digamma` covers. ln_gamma stays on statrs because gam-math has no ln_gamma of its own. Tests that use statrs digamma as an independent oracle (tests/quality/...) are deliberate references and stay.

Fix: import `gam_math::special::digamma` in these three modules.

SauersML:
Fixes #3770

`gam_math::special::digamma` is documented as "the workspace's one implementation", and it shares its kernel with the `trigamma` these modules already use. Three production modules imported `statrs::function::gamma::digamma` instead:

- **`gam-solve/src/pirls/prelude.rs`** feeds the Beta mean score and saturated root, the Gamma shape score and the NB θ score. Their curvatures already use `gam_math::special::trigamma`.
- **`gam-solve/src/mixture_link.rs`** used statrs ψ at 2912–2920, 3281–3283 and 3495–3497, but `gam_math::special::digamma` at 3080/3086/3144 for the same Beta-link quantities.
- **`gam-terms/src/analytic_penalties/ordered_beta_bernoulli.rs`** computes the marginal score with statrs ψ and its derivative with gam_math ψ₁.

After this PR, all three modules import `gam_math::special::digamma`. `ln_gamma` stays on statrs, since gam-math has no ln_gamma of its own. Tests that use statrs ψ as an independent oracle are deliberate references and are unchanged.

Every argument at these sites is a positive shape, mass or θ, which is the domain `gam_math::special::digamma` covers. Both kernels are accurate to f64 rounding, so values move only in the last bits.

The local guards against main pass: "Source-removal guard verified 250 -> 250 Rust items; 0 guarded removals" and the public-API census is clean.

**LoC delta:** 3 files changed, +5 / −4, a net +1. Swapping an import for one that also brings `trigamma`/`ln_gamma` means one split import line per file. This PR removes an implementation choice rather than lines.

SPEC: "It's required that the CLI, Python library, and Rust library are unified, have a single source of truth".

---


SauersML:
Verified locally and merged (head 223008c2). I merged it with origin/main in a batch tree and compiled that tree.

- `cargo check --workspace --tests`: clean, ban scanner passing
- `cargo test -p gam-solve --lib mixture_link`: 50 passed
- `cargo test -p gam-solve --lib pirls`: 190 passed
- `cargo test -p gam-terms --lib ordered_beta_bernoulli`: 16 passed
- `cargo test -p gam-solve --lib null_deviance_preserves_boundaries_dormancy_and_beta_mle_geometry`: 1 passed
- `cargo test -p gam-solve --lib beta_half_unit_deviance_from_shape_differences_is_exact_and_resolved`: 1 passed

---

PR #3778, opened 2026-09-20
manifold/fit_drivers: measure the LM gain ratio's step curvature (#3773)

### LINKED ISSUE #3773: manifold/fit_drivers: LM gain ratio uses a curvature identity that does not hold for the step actually taken
**Where:** `crates/gam-sae/src/manifold/fit_drivers.rs`, `run_joint_fit_arrow_schur`, gain-ratio block after an accepted line search (origin/main ~L7573-7584):

```rust
let d_th_d = (directional_decrease
    - globalization.lm_ridge_b * step_norm_sq)
    .max(0.0);
let predicted =
    (alpha * directional_decrease - 0.5 * alpha * alpha * d_th_d).max(0.0);
```

The comment says: "for the LM step (H+λI)Δ = −g, ΔᵀHΔ = d − λ‖Δ‖² (λ the β-block ridge)".

**Why it is wrong.** That identity holds only for the raw solve output, with a single scalar ridge. The Δ used here is not that output:

1. **Trust-radius clip (~L7335-7345).** When `quotient_step_norm > trust_radius`, both blocks are scaled by `s < 1`. `directional_decrease` is then computed on the clipped step, so it equals `s·d`. The true curvature is `s²·ΔᵀHΔ ≈ s²(d − λ‖Δ‖²)`, but the code uses `s·d − λs²‖Δ‖²`. With a small ridge, the quadratic term is overstated by a factor of about `1/s`. The predicted decrease at α = 1 then comes out as roughly `½·s·d` instead of `s·d·(1 − s/2)`. For a heavily clipped step that is close to half the true model decrease, so the gain ratio is inflated by up to 2×. An overshooting early step (the case the clip exists for) reads as a good one, and the ridge is relaxed or held when it should grow.
2. **Per-row null projection (~L7164-7233).** Each row's sub-floor null directions are subtracted from Δt after the solve. After that, `(H+λI)Δ = −g` no longer holds for the projected step, so the identity has no basis.
3. **Ridges.** The coordinate block was solved with `lm_ridge_t` (L7113: `solve_with_lm_escalation_inner(&sys, lm_ridge_t, lm_ridge_b, ..)`), but the formula prices the whole `‖Δ‖²`, including Δt, at `lm_ridge_b`. The caller can set `ridge_ext_coord` and `ridge_beta` independently. On top of that, `solve_with_lm_escalation_inner` may add an unreported `proximal_ridge` to both blocks.
4. The `.max(0.0)` on `d_th_d` hides a negative curvature estimate instead of reporting what the model actually predicts.

The gain ratio drives `InnerGlobalizationHint::record_accepted_step`, which is the ridge schedule for every later inner iterate. A wrong ratio therefore sets the damping of the whole inner solve.

**Principled fix.** Measure the curvature of the step that was actually taken, instead of reconstructing it from an LM identity. `ΔᵀHΔ` is one application of the assembled bordered-arrow operator, and `gam_solve::arrow_schur::arrow_operator_apply(&sys, 0.0, 0.0, Δt, Δβ)` already provides it. It covers every row's `H_tt` and `H_tβ` (dense or matrix-free) plus the shared `H_ββ` or penalty operator. Then `predicted = α·d − ½α²·ΔᵀHΔ`. This is exact under clipping, projection, separate ridges and escalation. The cost is one operator application per accepted iterate, which is less than one PCG iteration.

SauersML:
Fixes #3773

The inner LM gain ratio in `run_joint_fit_arrow_schur` estimated the predicted decrease as `α·d − ½α²·(d − λ_b‖Δ‖²)`. It recovered `ΔᵀHΔ` from the identity `(H+λI)Δ = −g`, but that identity holds only for the raw solve with one scalar ridge. The step it was applied to differs from the raw solve in four ways:

- **Per-row null projection:** each row's sub-floor null directions are projected out of the coordinate step.
- **Trust-radius clip:** the step may be clipped. A clip by `s` makes the true curvature `s²·ΔᵀHΔ`, but the formula gives `s·d − λs²‖Δ‖²`. The gain ratio is then inflated by up to 2×, and this happens on exactly the overshooting steps the clip exists for.
- **Separate coordinate ridge:** the coordinate block was solved with `lm_ridge_t`, but the formula prices it with `lm_ridge_b`.
- **Escalation:** the solve can carry an unreported proximal ridge added by `solve_with_lm_escalation_inner`.

The gain ratio sets the ridge schedule for every later inner iterate.

**Change**
- Apply the assembled bordered-arrow operator to the step actually taken, using `gam_solve::arrow_schur::arrow_operator_apply(&sys, 0.0, 0.0, Δt, Δβ)` with zero ridge. It covers every row's `H_tt` and `H_tβ` (dense or matrix-free) plus the shared `H_ββ` or penalty operator.
- Use the exact curvature: `predicted = α·d − ½α²·ΔᵀHΔ`.
- Remove the `.max(0.0)` clamps. A non-positive predicted decrease still takes the existing `predicted > 0` branch.
- Cost: one operator application per accepted iterate, which is less than one PCG matvec.

---


SauersML:
BF-57 review (partition 5, #3773): this PR is sound. The predicted decrease now uses the measured curvature ΔᵀHΔ of the step actually taken, via the public `gam_solve::arrow_schur::arrow_operator_apply` with zero ridge. It no longer uses the LM identity ΔᵀHΔ = d − λ‖Δ‖², which is false after null projection or a trust clip (a clip by s gives s²ΔᵀHΔ, not s·d − λs²‖Δ‖²). The unjustified `.max(0.0)` clamps are removed; a non-positive prediction still takes the existing `predicted > 0` branch. The PR is mergeable (clean) against main.

Its focused run HN-1-3773-1 (35500326514) has been queued since 08:43 and is only waiting on a runner.

**Ready for the local merge train MT-0** (even PR). Merge once that run is green, with expectedHeadSha 2196af516b17a7b1b6b68feca8cf42817965c400.

---

SauersML:

---

PR #3786, opened 2026-09-20
BMS flex audit: judge flex blocks on the total η-Jacobian

### LINKED ISSUE #3320: BMS flex identifiability audit judges link-deviation columns on the pilot training rows only, so it drops ramps the calibration identifies
## What is wrong

The BMS cross-block identifiability audit judges a link-deviation column only by its values at the pilot training rows. See `install_compiled_flex_block_into_runtime`, `crates/gam-models/src/bms/install_flex.rs:292`: the candidate design is evaluated at `candidate_arg_at_training_rows` (:309, :469), and `compile` (:416) keeps the directions whose W-metric Gram at those rows clears its floor.

The objective reads the link deviation on more than those rows. The calibration evaluates it at `a + b·z` over the latent measure: the whole real line under the standard-normal law, every node under the empirical one. A column can therefore be nearly flat on the pilot rows yet carry the fit beyond them, and the audit drops it as aliased when it isn't.

## Why it matters now

Since 7513ded218 (#3011) the link deviation is built on simple warp knots. Its outermost ramps rise over spans continued past the seed range, where the calibration, not the data, reaches. When the audit drops them, the kept basis is flat to second order inside the seed range with no end freedom. That is the configuration #3011 measured to pin fits on the `1 + w' >= eps` face.

## Measured

`margslope_flex_small_good_condition_completes_quickly` (root `inference` target, n = 2000, standard-normal latent, score warp + link deviation at the default 8 internal knots), 4 threads on acn112 16-19, 900 s cap. The test is red at base for other reasons (#2922).

| | base ba6d8564a5 | 7513ded218 |
|---|---|---|
| link deviation kept by the audit | 9/9 | **9/11** |
| outer evaluations in 900 s | 163 | 134 |
| refused inner solves | 0 | 2 (fully-rejected stall exits) |
| constrained-QP cycles on a face | 14/421 | 53/475 |
| best cost seen | 1.114970e3 | 1.114917e3 |

On the #3011 sample (10,000 rows, global-empirical) the audit drops nothing (66/66), and the fit converges in 352 s where base timed out.

## Expected

The audit measures each column where the objective evaluates it: at the training rows and over the calibration's latent measure. A column that the calibration identifies survives the audit.

Refs #3011.

SauersML:
Fixes #3320

## Problem
The Bernoulli marginal-slope flex audit judged the score-warp and link-deviation bases on their **raw values at the training rows**. In this family a flex coefficient moves the row predictor twice:
- directly, at the row's own argument;
- through the calibrated intercept, which averages the same basis over the latent measure.

So the likelihood's Fisher information is `JᵀWJ`, where `J` is the **total** η-Jacobian. The raw audit therefore dropped ramps that are flat over the training arguments but still active over the latent measure. At 7513ded218, `margslope_flex_small_good_condition_completes_quickly` keeps 9 of 11 link-deviation directions.

## Math
Notation:
- `s` is the probit scale, `b_i` the pre-scale slope and `B_i = s·b_i` the observed slope.
- The observed intercept `A_i` solves `E_π Φ(A_i + B_i u) = Φ(q_i)` over the latent law `π`.
- `η_i = A_i + B_i z_i` is the row predictor. The link deviation lives at `t_i = η_i/s`.

Differentiate the calibration. Every coefficient that moves the predictor at a latent node also moves `A_i`, weighted by the calibration tilt `ν_i(u) ∝ π(u)·φ(A_i + B_i u)`. This gives four families of total Jacobian columns:
- **marginal:** `∂η_i/∂β_m = g_i q'_i x_m,i`, with `g_i = φ(q_i)/E_π φ(A_i + B_i u)`;
- **slope:** `∂η_i/∂β_s = s (z_i − E_ν u) x_s,i`;
- **score warp:** `∂η_i/∂θ_h = B_i (H(z_i) − E_ν H(u))`;
- **link deviation:** `∂η_i/∂θ_w = s (C(t_i) − E_ν C(A_i/s + b_i u))`.

How each latent law is evaluated:
- **Standard-normal law:** with `P = 1 + B²`, we get `A = q√P`, `g = √P` and `ν = N(−AB/P, 1/P)`.
  - The runtime basis is a piecewise cubic with constant extensions past its support.
  - So `E_ν C` is exact in closed form, from the truncated standard-normal moments `I₀..I₃` of each span plus the two tail masses.
- **Empirical laws:** the tilt is the reweighted grid, computed by log-sum-exp, and `g = exp(−q²/2 − LSE)`.

The pilot is the rigid baseline plus one minimum-norm Fisher step on the marginal block, recalibrated.
- The step solves `JₘᵀWJₘ δ = JₘᵀW r` on the resolved positive eigenspace, with no ridge.
- `W = w φ²/(Φ(η)Φ(−η))` is the Fisher row weight at the pilot η.

## Change
- New module `bms::pilot_total_jacobian`:
  - `BmsCalibratedPilot`, which carries the calibration and the four total Jacobian families;
  - `gaussian_runtime_expectation`, the exact Gaussian expectation of a runtime basis.
- New `install_bms_flex_block_on_total_jacobian`:
  - It runs the existing channel-aware audit and `compile` on `J` in the Fisher metric.
  - It installs only the kept selector `V`. It does not install the anchor correction `M`, because the BMS kernels evaluate the basis inside the calibration integral at latent nodes, where no anchor row exists.
  - The score warp is audited against marginal + slope. The link deviation is audited against marginal + slope + the installed score warp.
- `install_compiled_flex_block_into_runtime`, the raw-design contract that survival marginal-slope still uses, now shares the audit and compile step and the block rebuild.
- `pilot_eta_for_link_dev_orthogonalisation` is removed. Its Fisher step now lives on the pilot, taken through the marginal total Jacobian.

## Tests (in `pilot_total_jacobian`)
- `gaussian_runtime_expectation_matches_quadrature`
- `standard_normal_pilot_matches_fine_normal_grid`
- `empirical_total_jacobian_matches_finite_differences_of_the_calibration`: finite differences of a bisection-solved calibration, for every column family.
- `total_jacobian_resolves_columns_the_training_rows_cannot`: rows inside one cubic span have raw rank ≤ 4, and the total Jacobian's rank must exceed that.

Validation runs on focused-rust-proof dispatches (results will be posted here).

---


SauersML:
**VF-0 review: NEEDS-AUTHOR.**

`cargo check --workspace --tests` on this branch merged with main fails in the ban-scanner build script:

```
error: 1 ban violation across 1 rule
  error: crates/gam-models/src/bms/pilot_total_jacobian.rs:94: [allow(clippy::too_many_arguments)] #[allow(clippy::too_many_arguments)]
```

Please replace the many-argument function at `pilot_total_jacobian.rs:94` with a struct argument instead of the allow.

The body also says validation "results will be posted here", and none are posted yet, so the #3320 fix is unconfirmed. The math I checked is right: the calibration `a = q√(1+b²)`, the tilt, `∂a/∂b = −E_tilt[u]`, and the truncated moments.

---

SauersML:

**Defect: the build fails.** `crates/gam-models/src/bms/pilot_total_jacobian.rs:94` puts `#[allow(clippy::too_many_arguments)]` on `pub(crate) fn build(`. The repository's `build.rs` ban scanner rejects that attribute, so the build breaks with:

```
error: crates/gam-models/src/bms/pilot_total_jacobian.rs:94: [allow(clippy::too_many_arguments)]
```

**Second gap:** the body promises validation results "posted here", but none were ever posted. The #3320 fix is still unconfirmed at head `5a6d4e21`.

**Suggested fix:**
1. Put `build`'s inputs into one named struct, for example the pilot state, the calibration inputs and the moment workspace, and pass that struct instead of the long argument list. Then remove the allow.
2. Post the #3320 regression evidence. That means a test asserting that the pilot's total Jacobian `∂a/∂b = −E_tilt[u]`, with the calibration `a = q√(1+b²)`, matches the analytic value.

The math itself looks correct. Only the build blocker and the missing evidence prevent merging.

---

SauersML:
Taken over by RF-0748-3; driving to merge. I'll replace the allowed many-argument `build` with a struct argument and run the pilot Jacobian tests.

---

SauersML:
[swarm lane bms-3011] Comparison with the #3320 fix the swarm is landing: commit 145feddd37 on main c2508b641e, gated on MSI, with its final tip check running.

**What it changes.** The fix removes the cause at the seed. The link deviation's knots are now built on the pilot index itself, at both call sites (BMS `block_specs.rs` and survival marginal-slope `fit_entry.rs`), and `padded_deviation_seed` is deleted.
- The 0.5·max(IQR, 1) pad existed only to keep the training rows off the old clamped C0 end, which 7513ded218 removed.
- With simple ends on the pilot's own range, every ramp rises over training rows. Ramp c rises on [t_{c+1}, t_{c+4}], and t_k = lo + h(k − 3).

**Measured.** `margslope_flex_small_good_condition_completes_quickly`, 4 threads, 900 s:

| arm | link ramps kept | best cost | QP cycles on a face |
|---|---|---|---|
| clamped base ba6d8564a5 | 9/9 | 1.114970e3 | 14/421 |
| 7513ded218 | 9/11 | 1.114917e3 | 53/475 |
| unpadded seed | 11/11 | 1.114907e3 | 22/341 |

**This PR** keeps the pad, re-judges the audit on the total η-Jacobian, and stops installing the anchor correction M.
- Per the ruling on #3320, that metric change is built only if a measured case shows the pilot-row audit is wrong once every ramp overlaps the training rows. No such case is measured here, and the PR has no fit-level numbers yet.
- The separating check would be the smoke above on this branch (kept count, best cost, refusals) against main with the unpadded seed.

**Conflict.** This PR edits `block_specs.rs` around the link seed, and `gradient_paths.rs`. On top of 145feddd37 the padded seed and `build_link_deviation_block_from_knots_design_seed_and_weights` are gone; the link call site uses `build_link_deviation_block_at_pilot(&eta_pilot, cfg)`.

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, a417cc609f06.

PR #3787, opened 2026-09-20
gam-problem: make PenaltyMatrix::validate symmetry band purely relative

### LINKED ISSUE #3780: gam-problem: PenaltyMatrix::validate symmetry tolerance has an absolute floor, so small-scale penalties can hide genuine skew
`crates/gam-problem/src/penalty_matrix.rs`, `validate_symmetric_psd_core` (reached from `PenaltyMatrix::validate` for Dense, Kronecker factors and Blockwise locals; production caller `custom_family_blockwise.rs:130`, which validates every block penalty at the custom-family boundary):

```rust
// Symmetry: relative to the matrix scale so a legitimately large penalty
// is not rejected for round-off and a small one cannot hide genuine skew.
let sym_tol = 1e-10 * max_abs.max(1.0);
```

The `.max(1.0)` makes the tolerance an **absolute** `1e-10` whenever `max|S| < 1`, which is exactly the case the comment says must not happen ("a small one cannot hide genuine skew"). It also contradicts the function's own doc contract ("never an absolute floor, so validity is invariant under `S → c·S`") and the relative symmetry band that `joint_penalty.rs` uses for the same property.

Concrete counter-example: `S = 1e-12 · [[1, 1], [0, 1]]` is 100% skew in its off-diagonal (|S − Sᵀ| = max|S|), yet `validate` accepts it because `1e-12 < 1e-10`. The same matrix scaled by 1e12 is rejected. So validity is not scale invariant. The gradient `λSβ` that downstream code implements then disagrees with the true gradient `λ·sym(S)β` of the quadratic. Penalties are routinely tiny in absolute terms (for example, normalized or unit-λ-rescaled penalties over large-scale covariates).

The PSD check just below is already purely relative (`100·n·ε·max|ev|`), so only the symmetry band is inconsistent.

**Fix:** `let sym_tol = 1e-10 * max_abs;`. For `max_abs == 0` the tolerance is 0 and the asymmetry is exactly 0, so the zero penalty still passes. Add a regression test showing that a scaled skew matrix is rejected at every scale and a scaled symmetric one is accepted.

SauersML:
Fixes #3780

`validate_symmetric_psd_core` used `1e-10 * max_abs.max(1.0)` as its symmetry tolerance. The `.max(1.0)` sets an absolute floor, so a penalty with max|S| < 1 could hide real skew. For example, `1e-12 * [[1,1],[0,1]]` was accepted, while the same matrix scaled up by 1e12 was rejected.

This PR changes the tolerance to `1e-10 * max_abs`, which matches the inline comment, the function's own contract that validity is invariant under `S -> c*S`, and the relative PSD check just below it. The zero penalty still passes because both its tolerance and its asymmetry are exactly 0.

It also adds a test, `validate_symmetry_is_scale_invariant`. It checks that skew Dense and Kronecker-factor penalties are rejected at scales from 1e-14 to 1e12, that symmetric PSD ones are accepted at those scales, and that the zero penalty passes.

---


SauersML:

---

PR #3792, opened 2026-09-20
survival marginal-slope block Jacobians: read designs by row chunk, never drop the callback

### LINKED ISSUE #3781: survival marginal-slope blockspecs: the time/marginal/timewiggle Jacobian callbacks are silently dropped when a design cannot be densified, and an ungated debug probe densifies design_exit on every block build
## Where (origin/main 5994ad54)

**1. Silent callback drop, rigid path.** `crates/gam-models/src/survival/marginal_slope/fit_setup.rs:79-96`, in `build_time_blockspec`:
```rust
// ... Falls back to no callback if densification fails.
let jac_cb: Option<Arc<dyn BlockEffectiveJacobian>> = (|| {
    let d_entry = time_block.design_entry.try_to_dense_arc("build_time_blockspec::entry").ok()?;
    let d_exit = design_exit.try_to_dense_arc("build_time_blockspec::exit").ok()?;
    let d_deriv = time_block.design_derivative_exit.try_to_dense_arc("build_time_blockspec::deriv").ok()?;
    if d_entry.dim() != d_exit.dim() || d_entry.dim() != d_deriv.dim() { return None; }
    Some(Arc::new(TimeBlockJacobian::new(d_entry, d_exit, d_deriv)) as _)
})();
```
`fit_setup.rs:153-160` (`build_marginal_blockspec`) does the same with `.try_to_dense_arc(..).ok().map(MarginalBlockJacobian::new)`.

**2. Silent callback drop, timewiggle path.** `fit_entry.rs:1593-1648` is the same pattern with four `.ok()?` calls. If it yields `None`, the rigid callbacks stay in place. Those callbacks have no wiggle chain rule (`dq/dh`, the `B(h)` tail columns), so the audit reads the wrong geometry.

**3. Ungated diagnostic.** `fit_setup.rs:16-74` runs on every call to `build_time_blockspec`, which is every `build_blocks` (pilot, each κ probe, the final fit). It densifies `design_exit` and builds p centered column copies to compute cosines. The work runs at any log level, and the result feeds only a `log::debug!`. Its comment says "fires once per fit ... never on a hot path", but it runs on every block build.

## Why it is wrong

- `try_to_dense_arc` refuses, by design, any design over the single-materialization cap. It also refuses under construction policy or ledger pressure.
- This family supports designs that stay streamed: `family.rs:320` `memoize_operator_backed_designs` logs "design stays streamed".
- So a supported large or operator-backed fit quietly loses these callbacks:
  - A `ParameterBlockSpec` without `jacobian_callback` is read by the identifiability audit and the canonical gauge (`gam-identifiability/src/canonical.rs`) as a plain one-channel flat design (only η1 = exit). Entry and derivative channels are ignored.
  - The block also loses `locks_raw_width_reduction() == true` (`canonical.rs:1124-1131`), so the gauge may column-reduce a block whose family reads raw-width designs. The callback docs say that desynchronises the family.
  - Nothing is logged. The fit goes on with different identifiability geometry.
- This breaks SPEC.md "In general, do not paper over solver issues": it is a reactive fallback, and the error is swallowed.
- The `BlockEffectiveJacobian::effective_jacobian_rows` contract (`gam-problem/src/block_spec.rs:130-138`) already asks for streaming: "so large construction-time audits can stream chunks instead of materialising all n·p·K entries at once". The callbacks break that by holding `Arc<Array2<f64>>` copies of whole designs.

## Principled fix

- Make the four callbacks operator-backed, in `block_jacobians.rs`:
  - `TimeBlockJacobian`, `MarginalBlockJacobian`, `SmsTimewiggleTimeJacobian` and `SmsTimewiggleMarginalJacobian` hold `DesignMatrix`, a cheap shared clone.
  - `effective_jacobian_rows(rows)` reads only the requested row block: a view when a dense copy exists or is memoized, otherwise `try_row_chunk(rows)`. A chunk error is returned as the callback's error.
- The builders then have nothing to densify and nothing to fall back from. They always attach the callback.
- Delete the ungated diagnostic. If needed, the audit's own Trace-gated block probe (`canonical.rs:865`) already covers it.

SauersML:
Fixes #3781

The time, marginal and timewiggle block specs built their `BlockEffectiveJacobian` callbacks from `try_to_dense_arc(..).ok()?`. When a design could not be materialized whole, the callback was left off without any error. The identifiability audit and the canonical gauge then treated the block as a flat exit-only design, which meant no entry or derivative channels and no raw-width lock. On the timewiggle path, the rigid callbacks, which lack the wiggle chain rule, stayed in place.

- The four callbacks (`MarginalBlockJacobian`, `TimeBlockJacobian`, `SmsTimewiggleTimeJacobian`, `SmsTimewiggleMarginalJacobian`) now hold `DesignMatrix`. Each `effective_jacobian_rows` call reads only the requested row block: a view of a dense or memoized copy, or one streamed operator chunk. This is the streaming the trait contract asks for. A chunk failure comes back as the callback's `Err`.
- The builders in `fit_setup.rs` and `fit_entry.rs` always attach the callback.
- The ungated `time_surface` diagnostic is removed. It densified `design_exit` and built centred column copies on every block build, only to feed a debug line.

The formulas are unchanged, and only the row indexing moves from global to chunk-local.

---


SauersML:
BF-57 review (partition 5, #3781): this PR is sound. It removes the silent `try_to_dense_arc(..).ok()?` drop of the block Jacobian callbacks. That drop was a quiet degradation of the identifiability audit and gauge; see SPEC.md: "In general, do not paper over solver issues."

Each callback now reads its row chunk through `DesignMatrix::as_dense_ref` / `try_row_chunk`, and chunk failures propagate as `Err`. On the branch, `From<&DesignMatrix> for DesignMatrix` exists (gam-linalg matrix/mod.rs:6075). The formulas are unchanged; only the indexing changes, from global `i` to chunk-local `local_i`. The ungated densifying debug diagnostic is gone. `git merge-tree` against current origin/main is clean.

Its focused run HN-5-3781-1 (35500434073) has been queued since 08:45 and is only waiting on a runner.

**Ready for the local merge train MT-0** (even PR). Merge once that run is green, with expectedHeadSha 1814eca83d974cf1fa70dd58ac66d101de643567.

---

SauersML:
Taken over by RF-1113-3; driving to merge. It is merged together with 19 other PRs onto origin/main in a local integration worktree. `cargo check --workspace --tests` is clean there, and the targeted tests are running now.

---

SauersML:
Verify-and-land: **MERGED** (head 1814eca8, merged as bcef0d80).

Run locally on the head merged with `origin/main` and the rest of the batch:
- `cargo test -p gam --test basis_smooth effective_jacobian_at_timewiggle`: 6 passed, 0 failed

Review: the four block-Jacobian callbacks now always attach, holding the `DesignMatrix` and reading one row chunk at a time. A design that can't be materialized now returns `Err` instead of being silently dropped. The formulas are unchanged; only the row index moves from global to chunk-local, and I checked that at each use.

---

PR #3793, opened 2026-09-20
Survival marginal-slope warm-start key: hash every likelihood field (#3697)

### LINKED ISSUE #3697: custom-family persistent warm start: the survival marginal-slope key omits the entry and derivative designs, the entry-at-origin mask, the latent law, the score covariance and the time constraints, so a cached inner mode and its logL/logdet are replayed for a different likelihood
## Defect

`persistent_custom_family_key` (crates/gam-custom-family/src/custom_family_persistent_warm_start.rs:378) keys a persisted inner mode on:
- the block specs: name, `design`, `offset`, penalties, nullspace dims and initial ρ;
- the options;
- `family.persistent_warm_start_fingerprint(specs, options)`.

Survival marginal slope is the only family that implements the fingerprint (crates/gam-models/src/survival/marginal_slope/custom_family_impl.rs:209). That fingerprint hashes `n`, `event`, `weights`, `z`, `gaussian_frailty_sd`, `derivative_guard`, `offset_entry`, `offset_exit` and `derivative_offset_exit`.

It does **not** hash these fields of `SurvivalMarginalSlopeFamily` (crates/gam-models/src/survival/marginal_slope/family.rs:124), although each one defines the likelihood:

- `design_entry` and `design_derivative_exit`. The time block's `ParameterBlockSpec.design` is the exit design only, and this family sets no `stacked_design`, so the generic spec hash never sees the entry or derivative rows.
- `entry_at_origin`, which decides whether a row carries the `log Φ(−η₀)` entry factor at all.
- `latent_law` (the declared anchor law, gam#2923). The same data fitted under a declared law and under the standard-normal closed form produce **identical keys**.
- `score_covariance`: the pooled `Σ̄` versus the conditional per-row `Σ(a_i)` field (gam#2766).
- `time_linear_constraints`, `time_wiggle_knots`, `time_wiggle_degree` and `time_wiggle_ncols`.
- `influence_absorber` values.
- `slope_layout`, which holds the entry and exit-derivative designs of a follow-up-varying slope.
- the `score_warp` and `link_dev` runtimes (knots and basis), beyond the column counts the specs expose.
- `family_hyper`.

## Why it matters

On a key hit, `inner_blockwise_fit` (inner_blockwise_fit.rs:3946-4062) reuses the cached mode when all of these hold:
- `objective_state` matches;
- the mode is `converged`;
- `solved_inner_tol <= inner_tol`;
- the fresh curvature certificate shows no resolvable negative curvature.

It then returns the **cached** values, without recomputing them: `log_likelihood`, `penalty_value`, `block_logdet_h` and `block_logdet_s`.

Nothing checks stationarity for the current likelihood. The persisted record carries `kkt_residual: None`, and the certificate tests only curvature.

So two fits that differ only in an unhashed field share a record. For example:
- the same data under `latent_law = Some(..)` and under `None`;
- a different entry-time design, or a different entry-at-origin mask, with the same offsets;
- the pooled versus the conditional score covariance.

In each case the second fit returns the first fit's β, log-likelihood and log-determinants as its own inner solution, and those values feed the outer LAML objective. That violates SPEC: "A fit object must only ever come from a converged optimization." The replayed mode is not an optimum of the second fit's objective.

The persistent store is opt-in (`persistent_warm_start_store` defaults to `None`). When it is configured, the wrong answer is silent.

## Related (not duplicates)

- #3671: the per-row intercept warm-start key, which is a different cache.
- #3002: the persistent root is ignored by other families.

## Suggested fix

Either of these would work; the first is more robust:

1. Have the fingerprint hash every likelihood-defining field listed above. For example, hash the two extra designs chunkwise, the way `hash_cf_design_matrix` does. Hash `entry_at_origin`, the constraint matrix and vector, the latent-law kind, the pooled covariance and the conditional model, and the deviation runtimes' knots.
2. Or re-verify the cached mode against the current likelihood on reuse: recompute `log_likelihood` and the KKT residual at the cached β before returning it. This closes the class of bugs for any future field.

SauersML:
Fixes #3697.

When the persistent warm-start key matches, the cached mode, log-likelihood, penalty and log-determinants come back without being re-evaluated. The survival marginal-slope fingerprint only hashed n, event, weights, z, the frailty sd, the derivative guard and the three offsets. Two fits could get the same key, and be served each other's evidence, if they differed in any of these:
- entry-at-origin rows
- score covariance
- time and marginal designs
- slope layout
- score warp and link deviation
- influence absorber
- time constraints and wiggle
- declared latent law
- family hyper geometry

**Change**
- `SurvivalMarginalSlopeFamily::likelihood_fingerprint` destructures the family exhaustively. A field added later is a compile error until it is either hashed or named as solver scratch. Only two fields are ignored: the intercept warm-start seeds and the jet arena pool.
- `fingerprint_into` is added for each nested type, each destructured exhaustively:
  - `SlopeLayout`
  - `DeviationRuntime`
  - `ScoreCovarianceField` and `MarginalSlopeCovariance`
  - `SurvivalLatentLaw` and `JointLatentLawRuntime`
  - `SurvivalMarginalSlopeFamilyHyperState`, including the full offset geometry
- Designs are hashed with the same rule the block specs use. That rule is `gam_custom_family::hash_cf_design_matrix`, which is now public.

**Test**
`warm_start_fingerprint_moves_with_every_likelihood_field_3697` builds one variant for each field the old fingerprint skipped. It requires every key to differ from the base key and from every other variant's key. On main, all of these variants share the base key.

---


SauersML:
BF-53 review (partition sweep for #3697): the diff is sound. `likelihood_fingerprint` destructures `SurvivalMarginalSlopeFamily` exhaustively. Every likelihood-defining field is hashed, including the entry, exit and derivative designs (chunked through `hash_cf_design_matrix`, never densified), the entry-at-origin mask, score covariance, latent law, time constraints and wiggle layout, absorber, slope layout, deviation runtimes and family_hyper. Only solver scratch is named as ignored: intercept warm starts, which seed a converged root solve, and the jet arenas. A fingerprint error maps to `None`, which disables the cache and cannot mis-serve. `git merge-tree` against current main (6ff5fb0) is conflict-free. It waits only on Focused Rust Proof runs 35500414104 and 35500415571, queued since 08:45. Ready for the local merge train (MT-1, odd PR) once those are green.

---

SauersML:
Proof status for head `b704576c93e65a013248b04b2b9b275fd6d76a2b`:

- **gam-custom-family lib** (run 35500415571, key BF-16-3697-2): **green**.
  - Filter `persistent`: 2 passed, 0 failed.
  - Filter `warm_start`: 14 passed, 0 failed.
  - The tests include `custom_outer_state_reset_preserves_existing_persistent_warm_start`, `cached_beta_warm_start_rejects_nonfinite_entries`, `screened_outer_warm_start_reuses_any_matching_rho_dimension` and `warm_start_retention_2996::filed_warm_start_retains_no_joint_workspace_2996`.
- **gam-models lib** (run 35500414104, key BF-16-3697-1): **NOT ATTEMPTED.**
  - The build hit the 1200 s compile budget (exit 124). No test ran, so this result says nothing about the code.
  - It has been re-dispatched on the same head as run **35508414209** (key BF-16-3697-3), with the same filters: `warm_start_fingerprint survival::marginal_slope::slope_layout frozen_time_limit`.

This PR can be marked ready and merged once 35508414209 is green. There are currently about 560 runs in the Actions queue.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, cfddc04f624e.

PR #3797, opened 2026-09-20
matern_gradient: differentiate the unclamped centered aniso metric

### LINKED ISSUE #3779: matern_gradient: aniso log-scale gradient still zeroes axes beyond a ±50 clamp the metric no longer has
## Defect

`StreamingMaternBasisGradientEvaluator` (public through `matern_basis_gradient_streaming` in `crates/gam-pyffi/src/latent/reml_latent_fit_ffi.rs:4877`) still treats a centered aniso contrast outside ±50 as a *saturated* coordinate with zero derivative. The metric it differentiates has had no such clamp for a while.

`crates/gam-terms/src/basis/matern_gradient.rs:80-90` (origin/main):

```rust
let mean = centered_aniso_log_scale_mean(eta);
(
    centered_aniso_metric_weights(eta),
    eta.iter()
        .map(|&value| {
            let centered = value - mean;
            centered > -50.0 && centered < 50.0
        })
        .collect(),
)
```

`row_chunk_gradient` then sets `axis_fraction = 0` for any such axis and drops it from the `active_fraction` that multiplies the mean-centering term.

The metric itself, `duchon_kernel_math.rs:2240-2264`, is unclamped:

```rust
pub(crate) fn centered_aniso_log_scale(value: f64, mean: f64) -> f64 {
    // The contrast as it is, never pinned to a box. ...
    value - mean
}
pub(crate) fn aniso_metric_weight(value: f64, mean: f64) -> f64 {
    (2.0 * centered_aniso_log_scale(value, mean)).exp()
}
```

The same file also computes the design's `aniso_distance`, the value this evaluator claims to differentiate. Its weight is `exp(2ψ_a)` for every finite `ψ_a`, so `∂r²/∂η_b = 2 Σ_a w_a h_a² (δ_ab − 1/d) / ℓ²` for **all** axes. The exact gradient is:

```
∂K/∂η_b = (∂K/∂ln κ) · (s_b / r² − 1/d)        (d > 1; the one-axis eta is uncentered)
```

For any contrast with 50 ≤ |ψ_a| < ~354 (where `exp(2ψ)` is still a finite positive double, which `set_spatial_aniso_log_scales` accepts), the evaluator returns something else:
- a zero `s_a/r²` for that axis, and
- a mean term scaled by the "active" mass only.

That is a wrong gradient, not a limit convention.

## Why the test does not catch it

`anisotropic_gradient_differentiates_the_clamped_metric` uses `eta = [80, 0, 0]` (centered contrast 53.3 on axis 0). Its weighted distance is `r ≈ e^{3.33} ≈ 28`, where Matérn-3/2 is `≈ e^{-48}`. So the analytic and FD values are both about 1e-19 and the `< 1e-8` assertion is vacuous. The FD side already uses the unclamped `centered_aniso_metric_weights`.

The comment in `aniso_distance_and_components` (`duchon_kernel_math.rs:2305`, "Clamp exp(2ψ) to avoid overflow/underflow: ψ in [-50, 50]") is stale for the same reason.

## Fix

- Delete `metric_axis_derivatives`. Every axis differentiates the unclamped metric: `d_log_kappa * (axis_fraction − mean_derivative)`.
- Update the variant doc.
- Replace the vacuous test with one whose contrast exceeds 50 while the weighted distance stays O(1), so the gradient is O(1) and the FD comparison is meaningful.
- Drop the stale clamp comment.

SauersML:
Fixes #3779

The shared aniso metric (`centered_aniso_metric_weights`) no longer clamps the centered contrast `psi_a = eta_a - mean(eta)`. `centered_aniso_log_scale` says it is "never pinned to a box". But `StreamingMaternBasisGradientEvaluator` still marked axes with `|psi_a| >= 50` as saturated, and it zeroed their axis fraction and their share of the mean term. For those `eta`, the `AnisoLogScale` gradient was not the derivative of the kernel that the value path evaluates.

Changes:
- Every axis now uses the exact chain `dK/deta_b = dK/dlnkappa * ((u_b/r)^2 - 1/d)`. For a one-axis `eta`, which is uncentered, `1/d` becomes 0 as before.
- Removed the dead `metric_axis_derivatives` flags and the `active_distance` bookkeeping.
- Fixed the stale enum doc here and the stale "Clamp ... [-50, 50]" comment in `aniso_distance_and_components`.
- Replaced the vacuous `..._clamped_metric` test, where both sides were about 1e-19. The new test is a central finite-difference check at centered contrasts `[80, -40, -40]`, with data chosen so every scaled component is O(1). The gradients are about 0.19, 0.03 and 0.15, and the test asserts `|fd| > 1e-2` on each axis so it cannot pass vacuously.

---


SauersML:
Closed as a duplicate of #3796: LC-20 claimed #3779 first with the same fix. The branch `claude/fix-3779-matern-gradient-unclamped` is kept for reference only.

---

PR #3798, opened 2026-09-20
gam-linalg lanczos: certify extreme eigenpairs relative to the retained norm, no unit floor (#3784)

### LINKED ISSUE #3784: gam-linalg lanczos: extreme-eigenpair certificate floors its scale at 1.0, so the "relative" residual test is absolute for operators with norm below 1
## Where (origin/main 5994ad54c4)

`crates/gam-linalg/src/lanczos.rs:445-456`, `symmetric_extreme_lanczos_eigenpairs`:

```rust
let selected_operator_scale = selected_indices
    .iter()
    .map(|&index| values[index].abs())
    .fold(0.0_f64, f64::max)
    .max(1.0);
...
last_worst_relative_residual =
    last_worst_relative_residual.max(residual / selected_operator_scale);
...
if last_worst_relative_residual <= options.relative_residual_tol {
```

The same floor is repeated when the SAE frame-curvature certificate reports its residual: `crates/gam-sae/src/identifiability/frame_curvature.rs` (`streamed_lambda_max`, `relative_residual: residual / lambda_max.max(1.0)`, documented as "relative to `max(λ_max, 1)`").

## Why it is wrong

The option is documented as a relative Ritz residual, `||A v - λ v|| / max(||Λ_selected||∞, 1)`: "Scaling every Ritz residual by the retained operator norm gives one invariant certificate". The `.max(1.0)` breaks that invariance.

- The operators this solver certifies carry units: a data curvature `H = RᵀR`, a reduced Schur complement `S`, a kernel Gram.
- When `‖A‖ < 1`, the test becomes the absolute test `‖r‖ ≤ tol`.
- For an operator of norm `10⁻⁶` and `tol = √ε ≈ 1.5·10⁻⁸`, the solver declares `λ_max` certified once `‖r‖ ≤ 1.5·10⁻⁸`. That is a relative residual of about `10⁻²`, six orders of magnitude looser than requested.
- An operator of norm below about `10⁻⁸` certifies at the first checkpoint whatever its Krylov quality, because `β_k|e_kᵀy| ≤ ‖A‖ < tol`.
- Rescaling the data (`A → cA`) therefore changes which step certifies and how accurate the result is. The answer ought to be the same eigenpair scaled by `c`.

Consumers that rely on the relative accuracy:

- **`streamed_lambda_max` (gam-sae frame curvature).** Its doc says `√ε` "is the standard residual-based stopping point that sits five orders below the consumer tolerance" and "A solve that cannot reach it is a REFUSAL, never a looser certificate". `λ_max` is the denominator of every generator verdict, compared against `GENERATOR_FLAT_ENERGY_TOL = 1e-3`. For a small-scale curvature (small-magnitude activations), the floor silently turns this into a looser certificate.
- **`reduced_schur_lambda_max` (gam-solve `arrow_schur/reduced_solve.rs:4062`).** It returns `θ + r` as a certified upper bracket that sizes the rational quadrature window "for `rel_tol`". For `‖S‖ < 1` the bracket is only absolutely accurate.
- **`duchon_spectral_kernel_chart` (gam-terms).**

SPEC: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code." The `1.0` has units of the operator, which is arbitrary by construction. The certificate also cannot be told apart from a looser tolerance: "In general, do not paper over solver issues."

## Principled fix

Scale by the retained operator norm itself, with no unit floor. Compare multiplicatively so that a zero-norm retained spectrum certifies only an exactly zero residual (an exhausted Krylov space) and never divides by zero:

```rust
let scale = max_selected |θ|;
certified  ⇔  residual_i ≤ tol · scale   for every selected i
```

- Report `last_worst_relative_residual` as `residual / scale`, or `0` when the residual is `0`.
- In `streamed_lambda_max`, report `residual / lambda_max` the same way.
- Add a regression test: a diagonal operator and the same operator scaled by `1e-12` must certify at the same step, with eigenvalues that agree relatively.

SauersML:
Fixes #3784

## Problem

`symmetric_extreme_lanczos_eigenpairs` divided each Ritz residual by `max(max|θ_selected|, 1)`. For an operator with norm below 1, the "relative" test `‖r‖/scale ≤ tol` was therefore absolute. A curvature, reduced Schur complement or Gram carried in small units certified far more loosely than requested. Once `‖A‖ < tol`, it certified at the first checkpoint no matter how good the Krylov approximation was.

## Changes

- **Certificate scale:** the solver now divides by the retained operator norm alone. A zero residual counts as relative residual 0. A zero retained spectrum therefore certifies only an exactly exhausted Krylov space, and there is no division by zero.
- **Frame-curvature report:** `streamed_lambda_max` in `gam-sae/identifiability/frame_curvature.rs` drops the same `max(λ_max, 1)` floor from its reported relative residual.
- **Regression test:** `adaptive_extreme_lanczos_certificate_is_invariant_to_operator_units` runs `diag(1..=50)` at scale 1 and at scale `2⁻⁴⁰`. Both runs must certify a Ritz value within `1e-6` relative of `λ_max`, with a residual bound relative to θ. With the old floor, the scaled run certifies at step 2.

No tolerance was loosened. The change can only make certification stricter for operators with norm below 1.

---


SauersML:
Review from BF-19 (#3784 claim released in favor of this PR). The diff looks correct against SPEC:

- **Scale.** The certificate now divides by `max|θ_selected|` with no floor.
- **Zero residual.** A zero residual counts as relative 0.
- **Zero retained spectrum.** A nonzero residual against a zero retained spectrum gives `+∞` (positive/0.0), so it can never certify.
- **Test.** A 2⁻⁴⁰ scale with a power-of-two factor keeps the recurrence exactly scaled. Under the old floor that run certified at the second apply.
- **Merge.** `git merge-tree` against current main (41fbd6c337) is clean.

One small non-blocking point. In `streamed_lambda_max` the relative residual is divided by the post-`clamp(0.0, trace)` value. The solve certified against the unclamped `|θ|`, so the reported ratio can differ from the certified one when the clamp is active. If θ ≤ 0 is clamped to 0, the report becomes `+∞` even though the pair certified. Computing the ratio from `θ.abs()` before the clamp would keep the report equal to the certificate.

The four queued focused-proof runs (HN-2-3784-1..4) are the gate. I am not dispatching more.

---

SauersML:

---

PR #3799, opened 2026-09-20
fit_orchestration: refuse curv() κ/range profile unless the model is exactly y ~ curv(...)

### LINKED ISSUE #3763: fit_orchestration: curv() κ̂, ℓ̂, profile CI and flatness p-value are estimated on `y ~ 1 + curv(x)` alone, ignoring every other term in the formula
## Defect

The constant-curvature estimand (κ̂, ℓ̂, the profile-likelihood CI and the κ = 0 flatness LR p-value) is computed from a **marginal** model `y ~ 1 + curv(x)`. That model drops every other term the user wrote. The fit then realizes the full model at that κ̂/ℓ̂. So for any formula such as `y ~ z + curv(x1, x2)` or `y ~ curv(x1, x2) + s(t) + group(g)`:
- the point estimate and its interval belong to a model nobody fits;
- the flatness p-value is miscalibrated. Omitted variance inflates the residual, which makes the test conservative. A confounded covariate can also manufacture curvature.

## Evidence (origin/main)

**How the criterion is built.** `constant_curvature_psi_profile_value` and `constant_curvature_psi_profile_jet` (`crates/gam-models/src/fit_orchestration/drivers/constant_curvature_profile.rs`) build the design as `[1 | curv block]` and nothing else:
```rust
let mut design = Array2::<f64>::ones((n, p + 1));
design.slice_mut(s[.., 1..]).assign(&smooth_design);
```
`ConstantCurvatureProfile::new(x_term.view(), y, base_spec)` receives only the curv term's own columns and `y`.

**The fit path.** `spatial_optimization.rs` ~7024 calls `constant_curvature_kappa_profile_optimum(data, y, &resolvedspec, term_idx, options)` for every free-κ term. It writes `cc.kappa` and `cc.length_scale` back, then runs `fit_term_collection_forspec` on the **full** `resolvedspec`.

**The inference path.** `curvature_inference_forspec` (`spatial_curvature_inference.rs`) re-profiles the same marginal model. The pyffi caller (`curvature_inference_dataset_json_impl`) documents that it "refits at the EXACT estimand the model was fitted under — only κ moves". That holds only when the curv term is the sole term.

**What the guard checks today.** The only precondition is `validate_constant_curvature_profile_inputs`: Gaussian identity, unit weights, zero offset. It does not look at `linear_terms`, `random_effect_terms`, other `smooth_terms`, or `level`. The profile also always carries an intercept column, even when `spec.level == ModelLevel::NoIntercept` (`0 + curv(...)`).

**The file already states this rule.** The same file refuses `double_penalty=` for exactly this reason: "a κ̂ selected here would be an estimate for a model the fit does not realize". Any additional term violates that argument in the same way.

## Fix

This is the small, honest fix. It extends the existing precondition, so curvature/range estimation refuses (with a named reason) when the term collection is not exactly intercept + this one curv term:
- `linear_terms` and `random_effect_terms` are empty;
- `smooth_terms.len() == 1`;
- `level == Intercept`.

The pinned-κ / free-range arm already skips when the precondition fails. The free-κ arm and `curvature_inference_forspec` refuse.

Every in-repo curv fit with estimated κ uses `y ~ curv(x1, x2, ...)` alone, so nothing that works correctly today changes. The larger follow-up is to profile κ with the other terms' columns and penalties carried in the Gaussian REML closed form. That would lift the refusal.

SauersML:
Curvature and range profiling uses a design containing only an intercept and one `curv()` term. A larger formula previously received geometry estimates and inference from that smaller model.

Require the complete fitted model to match the profile's Gaussian, unit-weight, zero-offset, sole-term contract. This applies to both free curvature and a free range under pinned curvature. Unsupported requests now fail explicitly; they must pin both `kappa=` and `length_scale=` to use fixed geometry in a larger model. No automatic range is retained after a refused profile.

Validation: MSI compiled the exact profile validator against coherent warm dependencies and passed the new model-contract test, including the sole-term control, added linear terms, pinned curvature with free range, and a removed intercept. All validator call sites were reviewed. Full-fit regressions cover both refusals and fully fixed geometry; those integration tests were not executed in this review.

Fixes #3763


SauersML:

---

SauersML:
Reviewed at head d60b161d79. It merges cleanly with current main.
- In the merged tree, all three callers of the widened `validate_constant_curvature_profile_inputs(resolvedspec, term_idx, ...)` are updated: spatial_curvature_inference.rs:284 and spatial_optimization.rs:7044 and 7055.
- `gam_terms::smooth::ModelLevel` resolves, because term_specs.rs is `include!`d into smooth.rs, and the enum derives `Debug` for the `{:?}` in the refusal message.
- The `y ~ curv(...)` precondition matches the profile design `[1 | curv block]`. Refusing it for a free κ and skipping it for the pinned-κ nuisance range keeps the pre-existing split.

Its only check, focused proof HN-7-3763-1, is still queued. Once that build is green, this is ready for the merge train (MT-1, odd PR).

---

PR #3801, opened 2026-09-20
matern_gradient: differentiate the unclamped aniso metric on every axis (#3779)

### LINKED ISSUE #3779: matern_gradient: aniso log-scale gradient still zeroes axes beyond a ±50 clamp the metric no longer has

SauersML:
Fixes #3779.

`StreamingMaternBasisGradientEvaluator` treated a centered aniso contrast outside ±50 as saturated. For that axis it zeroed `s_a/r²`, and it scaled the mean-centering term by the "active" distance mass only. The metric it differentiates has no clamp, so for |ψ_a| ≥ 50 the evaluator returned a wrong gradient.

What changed:
- **Exact derivative on every axis.** Each axis now uses `∂K/∂η_b = (∂K/∂ln κ)(s_b/r² − 1/d)`, with the one-axis η left uncentered. `metric_axis_derivatives` and the active-mass bookkeeping are deleted.
- **Same axis scale as the design.** The per-axis scale is now `aniso_axis_scale`, the same `exp(ψ)` that the design's `aniso_distance` uses, instead of `sqrt(exp(2ψ))`.
- **Unrepresentable contrasts are refused.** A contrast whose weight `exp(2ψ)` is not a positive finite double is refused. This is the same admissibility rule as `set_spatial_aniso_log_scales`.
- **Stale comment dropped.** The clamp comment in `aniso_distance_and_components` is gone.

Tests:
- `anisotropic_gradient_is_exact_for_contrasts_beyond_fifty_3779` replaces the vacuous `anisotropic_gradient_differentiates_the_clamped_metric`. The old test compared two values of about 1e-19 against a 1e-8 bar.
  - The new test uses η = [80,0,0] and [80,−80,0], so contrasts exceed ±50.
  - Every scaled displacement stays O(1), so r ≈ 1.03 and each axis gradient has magnitude above 1e-2.
  - It checks against the closed form to 1e-12 and against a central finite difference through the unclamped metric to 1e-8.
- `anisotropic_contrast_without_a_representable_weight_is_refused` covers `exp(2ψ)` overflow and underflow.

---


SauersML:

---

SauersML:
I'm backing off this PR because another session merged main in at 11:29Z. Before landing, note one failing check: `scripts/source_removal_guard.py` fails on the head because the test `anisotropic_gradient_differentiates_the_clamped_metric` was removed with no ledger entry:

```
SOURCE REMOVAL GUARD FAILED: guarded Rust items removed without exactly one acknowledgement: [{"path": "crates/gam-terms/src/basis/matern_gradient.rs", ..., "name": "anisotropic_gradient_differentiates_the_clamped_metric", "test_scoped": true, ...}]
```

To fix it, append one entry to `docs/source-removal-changes.json`. Keep the format `json.dumps(indent=2, ensure_ascii=False)`, set `base` = merge-base with main, and use `items: ["crates/gam-terms/src/basis/matern_gradient.rs:fn:anisotropic_gradient_differentiates_the_clamped_metric"]` plus a reason and evidence. With that entry, the guard passes: `145 -> 146 Rust items; 1 guarded removals`.

On a merge with main at ffc9c29f, `cargo test -p gam-terms --lib matern_gradient` gave 8/8 passed, and `cargo check --workspace --tests` was clean apart from the gam-pyffi `IxDyn` warning already on main. `public_api_census` and `spec_ban_ratchet` also pass (0 new, 0 stale). I did not push anything to this branch.

---

PR #3805, opened 2026-09-20
fit_orchestration/materialize/family: refuse a nuisance override for a family that does not carry it

### LINKED ISSUE #3804: fit_orchestration/materialize/family: scalar_family_from_name silently drops a negative-binomial theta / Beta phi / Tweedie power supplied for a family that does not carry it
## Defect

`scalar_family_from_name` (crates/gam-models/src/fit_orchestration/materialize/family.rs) is where the CLI, `gamfit.fit` and the latent FFI turn a family name and its `FamilyNuisanceOverrides` into a `LikelihoodSpec`. Each override belongs to one family:

- `negative_binomial_theta` belongs to negative-binomial
- `beta_phi` belongs to beta
- `tweedie_power` belongs to tweedie

The function reads an override only in the arm of the family that carries it. For any other family the override is dropped without a word. So each of these fits a model the caller did not ask for, and reports nothing:

- `gam fit --family poisson --negative-binomial-theta 2 ...` fits an equidispersed Poisson.
- `gamfit.fit(..., family="poisson", negative_binomial_theta=2.0)` fits an equidispersed Poisson.
- `gamfit.fit(..., family="gamma-log", beta_phi=7.5)` fits a Gamma and ignores `beta_phi`.
- `gamfit.fit(..., family="poisson-log", tweedie_p=1.6)` fits a Poisson and ignores `tweedie_p`.

The request contradicts itself, and the engine resolves it by dropping part of what the user asked for.

The `gamfit.fit` docstring is also stale. It still says "An unset `tweedie_p` resolves to `1.5`", but bare Tweedie is now refused because the power is never profiled.

## SPEC

> The default choices should give the best model for the user.

> Unnecessary choices and options should be deleted, not included.

An option that is accepted and then dropped is worse than an absent one. The user believes an overdispersion was imposed, and it was not.

## Fix

Refuse the pair where the family is minted, so every surface refuses it the same way. The message is:

> `<option> applies only to family='<family>'; family '<name>' does not carry it`

Each override still reaches the family that carries it. The fix also:

- corrects the doc comment and the `gamfit.fit` docstring
- makes the unit-test helper give only Tweedie heads a power

A regression test covers seven mismatched name/override pairs, the three matching pairs, and the CLI's `resolve_family` path.

SauersML:
`scalar_family_from_name` used to drop three overrides without a word when the resolved family does not carry them:

- `negative_binomial_theta`
- `beta_phi`
- `tweedie_power`

For example, `--family poisson --negative-binomial-theta 2` fitted a plain Poisson. The pair is now refused where the family is minted, so the CLI, `gamfit.fit` and the latent FFI all refuse it the same way.

Also in this PR:

- The stale doc comment and the `gamfit.fit` docstring are corrected. Bare Tweedie is refused, because the power is never profiled.
- The unit-test helper gives a power only to Tweedie heads.
- A new regression test, `a_nuisance_override_for_a_family_that_does_not_carry_it_is_refused`, covers:
  - seven mismatched name/override pairs
  - the three matching pairs
  - the CLI `resolve_family` path

Fixes #3804

---


SauersML:

---

PR #3806, opened 2026-09-20
Skovgaard r*: sample-space q̂ and the nuisance determinant form (#3535)

### LINKED ISSUE #3535: gam-inference skovgaard: u = dθ·î/√ĵ drops Skovgaard's î⁻¹ĵ factor (miscalibrated r* exactly when î≠ĵ) and omits the nuisance adjustment for p>1
## Where (origin/main)

`crates/gam-inference/src/skovgaard.rs:261-264`:

```rust
let dtheta = theta_hat - theta_null;
let sqrt_obs = observed_info.sqrt();
let u = dtheta * expected_info / sqrt_obs;
let u_empirical = dtheta * score_cov / sqrt_obs;
```

The module doc (`:63`, `:97-109`) justifies this as `u = ĵ^{-1/2}·{ℓ_{;θ̂}(θ̂) − ℓ_{;θ̂}(θ₀)}`, with the sample-space derivative replaced by Skovgaard's covariance `q ≈ (θ̂−θ₀)·î`. It calls the resulting `î/ĵ` ratio "the curvature correction that the first-order root ignores". The code is reached from Python through `crates/gam-pyffi/src/inference/inference_instruments.rs:741` (`skovgaard_r_star`, `gamfit/_rust.pyi:2170`) and through `scalar_skovgaard_from_matrices` (`:334`).

## Why it is wrong

Skovgaard's approximation (Skovgaard 1996, *Bernoulli* 2:145-165; Severini 2000 §7.5; Brazzale-Davison-Reid 2007 §8.5) is **not** `ℓ_{;θ̂}(θ̂) − ℓ_{;θ̂}(θ₀) ≈ q̂`. It is

    ℓ_{;θ̂}(θ̂) − ℓ_{;θ̂}(θ₀) ≈ q̂ᵀ î⁻¹ ĵ,     q̂ = cov_{θ₁}(U(θ₁), ℓ(θ₁) − ℓ(θ₀)) |_{θ₁=θ̂}

The `î⁻¹ĵ` factor converts the expected-parameterisation covariance into the sample-space derivative. In the scalar case this gives

    u = ĵ^{-1/2} · q̂ · ĵ / î = ĵ^{1/2} · q̂ / î .

The code computes `u = q̂/√ĵ`, which is off from Skovgaard's `u` by a factor of exactly `(î/ĵ)`. That is the same "correction" the module advertises.

With the surrogate the code actually uses, `q̂ = (θ̂−θ₀)·î`, the correct `u` is `(θ̂−θ₀)·√ĵ`, the Wald root. The code's `u` is instead `(θ̂−θ₀)·√ĵ · (î/ĵ)`. The `î/ĵ` multiplier has no derivation: it is the factor Skovgaard's formula divides out.

A worked check: for a canonical one-parameter exponential family, `ℓ(θ̂)−ℓ(θ₀) = (θ̂−θ₀)T − n(κ(θ̂)−κ(θ₀))`, so the exact `q̂` is `(θ̂−θ₀)·var(T) = (θ̂−θ₀)·î`. Skovgaard's `u` is therefore `(θ̂−θ₀)√ĵ`, the known exact canonical result. The code only agrees because `î = ĵ` there. For every non-canonical link, which is the case the module exists for (`non_canonical_link_expected_info_enters_u`), `r* = r + log(u/r)/r` is pushed by `log(î/ĵ)/r`. The result is:
- anti-conservative when `î > ĵ`,
- conservative when `î < ĵ`,
- at order `O(n^{-1/2})`, not `O(n^{-1})`.

The empirical companion has the same problem. Severini's version replaces `î` by `Î` in **both** places (`q̂ ≈ dθ·Î` and the `Î⁻¹ĵ` factor), which again collapses to `dθ√ĵ`. The code keeps only the numerator (`dθ·Î/√ĵ`).

Two tests pin the wrong behaviour instead of the theory:
- `expected_observed_ratio_shifts_r_star_and_p` (`:457`) asserts "`î > ĵ` must lift r*".
- The `u_emp = 0.2·121/10` assertion (`:494`).

**Multi-parameter case.** When `ψ = cᵀβ` has nuisance parameters (every penalized GAM, and `scalar_skovgaard_from_matrices` with `p > 1`), Skovgaard's `r*` needs the nuisance adjustment `log{ |j̃_λλ|^{1/2} |î|... / |S̃| ... }`, which uses the constrained fit `β̃₀`. The code instead collapses everything to the scalar informations `1/(cᵀĤ⁻¹c)`, `1/(cᵀIₑ⁻¹c)` and `1/Σ(sᵢᵀĤ⁻¹c)²`, with no constrained fit and no determinant terms. The "second-order" claim does not hold there.

(Prior #1156 introduced the current `î/ĵ` form. It fixed a real problem, the expected information being ignored, but placed `î` on the wrong side of the ratio.)

## Principled fix

1. Compute Skovgaard's `q̂` exactly instead of the linear surrogate. For exponential-dispersion families it is closed form: `q̂ = Σᵢ cov_{θ̂}(sᵢ(θ̂), ℓᵢ(θ̂) − ℓᵢ(θ̃₀))`, needing only the per-observation mean and variance at the full fit and the linear predictor at the constrained fit `β̃₀`. Also compute `Ŝ = Σᵢ cov_{θ̂}(sᵢ(θ̂), sᵢ(θ̃₀))`.
2. Form `u` from Skovgaard's formula: in the scalar case `u = ĵ^{1/2} q̂ / î`. For nuisance parameters use the full determinant form with `j̃_λλ`, `Ŝ` and `î`. The empirical version uses `Î` consistently in both places.
3. Replace `expected_observed_ratio_shifts_r_star_and_p` with a test that checks `r*` against an exact tail probability for a non-canonical scalar model, e.g. a Gamma/Weibull shape or an exponential with an identity-link rate, where the exact null law is available. The existing `sbc_skovgaard_rstar_size_curve` gate (`tests/inference/misc/`) should be run on a non-canonical family to catch the `O(n^{-1/2})` size error.

If (1) and (2) cannot land together, removing the spurious `î/ĵ` multiplier (`u = dθ·√ĵ`) is the minimum correct change. Note that with the linear surrogate, `r*` then carries no higher-order information, which is why the exact `q̂` is the real fix.

SauersML:
Fixes #3535

## Bug
There were two problems in the Skovgaard assembly.

- **Scalar case.** It computed u = (θ̂−θ₀)·î/√ĵ. This is the linear surrogate for Skovgaard's q̂, and it is exact only in the canonical parameter. In any other parameterisation r* loses its invariance and its third-order accuracy. For example, with a mean-parameterised exponential at n = 5, the old tail missed the exact Gamma tail by 55–170%.
- **Matrix path.** `scalar_skovgaard_from_matrices` collapsed the nuisance parameters into a scalar Wald variance. That drops the |ĵ|, |j̃| and |Ŝ| factors of Skovgaard (1996).

## Fix
- **`ScalarSkovgaardInput`** now carries:
  - q̂ = cov[U(θ̂), ℓ(θ̂) − ℓ(θ₀)];
  - the empirical Î and q̂_emp.

  The assembly uses u = √ĵ·q̂/î and u_emp = √ĵ·q̂_emp/Î.
- **`skovgaard_r_star_with_nuisance`** replaces `scalar_skovgaard_from_matrices` and implements the full form
  u = |ĵ|^{1/2}|Ŝ|cᵀŜ⁻¹q̂ / (|î||j̃|^{1/2}(cᵀj̃⁻¹c)^{1/2}).
  - It takes the full and constrained fits.
  - Its empirical companion is built from per-row scores (Σŝŝᵀ, Σŝs̃ᵀ, ΣŝΔℓ).
  - Ŝ is non-symmetric, so it is factored with a pivoted LU that keeps the determinant sign.
- **FFI.** `skovgaard_r_star` takes the new ingredients. This is a breaking change; the `_rust.pyi` and `_rust_module.pyi` stubs are regenerated, and `gen_rust_stub.py --check` is clean.

## Tests
- The mean-exponential r* tail is within n^(-3/2) of the exact Gamma tail, while first order is not.
- r* is invariant between the mean and rate parameterisations.
- The p = 1 nuisance path equals the scalar path.
- A two-group Poisson contrast matches the closed-form nuisance-adjusted u.
- A p = 3 fixture with non-symmetric Ŝ is invariant under a linear reparameterisation with negative Jacobian.
- The SBC size curve (`sbc_skovgaard.rs`) now runs in the non-canonical mean parameterisation.

The expected values were checked independently in NumPy. Float tolerances come from `accumulation_growth`.

---


SauersML:

---

SauersML:
Handing this back. Another session pushed 25138bbe at 12:28, so I'm stepping back while it's active.

For reference: `scripts/gen_rust_stub.py --check` fails on this branch only on `cross_fit_shared_precision_groups_json` (`list[...]` → `Sequence[...]`). That stub is equally stale on main f7e5ec64, so it isn't this PR's; the Skovgaard stubs match.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, b5b90a939faf.

PR #3811, opened 2026-09-20
gam-predict term_diagnostics: refuse an indefinite projected variance and a 0/0 variance share

### LINKED ISSUE #3807: gam-predict term_diagnostics: partial-dependence SE clamps a negative projected variance to 0, and variance shares report 0 for a 0/0
## Where (origin/main 20b7967df7)

`crates/gam-predict/src/term_diagnostics.rs:40-47`, `term_partial_dependence`:

```rust
let mut var = 0.0_f64;
for a in block.clone() {
    let xa = xi[a];
    for b in block.clone() {
        var += xa * covariance[[a, b]] * xi[b];
    }
}
se[i] = var.max(0.0).sqrt();
```

`crates/gam-predict/src/term_diagnostics.rs:104-108`, `term_variance_shares`:

```rust
let share = if total_var > 0.0 {
    population_covariance(&contrib, &eta) / total_var
} else {
    0.0
};
```

## Production path

`model.partial_dependence(...)` in Python goes to `model_partial_dependence_impl` (`crates/gam-pyffi/src/manifold/geometry_ffi.rs:7975`). That function passes the published covariance's term block (`beta_covariance_corrected()` or `beta_covariance()`) into `term_partial_dependence`. `model.variance_share(...)` goes to `model_variance_share_encoded_impl` (`geometry_ffi.rs:8008`), which calls `term_variance_shares`.

## Why it is wrong

1. **SE clamp.** `x_tᵀ V_t x_t` is a quadratic form of a covariance. A value clearly below zero, outside the rounding band of its own accumulation, proves that `V_t` is not PSD along that row: the covariance is defective. `.max(0.0)` turns this into an SE of exactly 0, so the band collapses to the point estimate at that grid point. It looks like perfect certainty, and nothing is reported. SPEC says "In general, do not paper over solver issues." The other two SE surfaces over the same covariance already refuse this:
   - `gam_inference::effects::pointwise_standard_errors` (`crates/gam-inference/src/effects.rs:320-355`) returns `EffectError::IndefiniteCovariance` when `variance < -tolerance`.
   - `gam_predict::local_covariances_with_backend` errors on a negative projected variance.

   Partial dependence is the only one of the three that clamps silently.
2. **0/0 share.** For a predictor that is constant over the supplied rows (or for zero rows), every share is `0/0`. Returning `0.0` for each term contradicts the function's own contract: "the shares over every term sum to exactly 1". The result reads as "no term explains any variance" when the quantity is actually undefined.

## Principled fix

- In `term_partial_dependence`, accumulate `Σ|x_a V_ab x_b|` next to the sum. Refuse (return `Err`, naming the row and block) when the variance is non-finite or below `-gam_linalg::roundoff::accumulation_band(k² + 1, Σ|·|)`. That is the Higham γ band of the accumulation: two rounded products per summand, then `k² − 1` additions. Only a negative inside that band reads as a zero variance.
- In `term_variance_shares`, return an error when `var(η)` is not finite and positive, instead of reporting zero shares.
- Add unit tests: an indefinite block must error, a PSD block must return the exact quadratic form, and a constant predictor's shares must error.

Claimed by local agent HN-38 (hn38/term-diagnostics-psd-guard); draft PR incoming.

SauersML:
Fixes #3807

- `term_partial_dependence` (backs `model.partial_dependence`) used to clamp the projected variance `x_tᵀ V_t x_t` to zero. A covariance that is not PSD along a grid row therefore published an SE of exactly 0 there. The function now accumulates `Σ|x_a V_ab x_b|` and refuses a variance that is non-finite or below `-gam_linalg::roundoff::accumulation_band(k² + 1, Σ|·|)`, the Higham γ band of the quadratic form. This matches what `gam_inference::effects::pointwise_standard_errors` and `local_covariances_with_backend` already do. A negative inside the band still reads as zero.
- `term_variance_shares` (backs `model.variance_share`) used to return 0 for every share when `var(η) = 0`. That is a 0/0 that contradicts the documented sum-to-one identity; the function now returns an error instead.
- New unit tests: an indefinite block errors, a PSD block returns the exact quadratic form, and a constant predictor's shares error.

---


SauersML:

---

SauersML:
Note from the dead-code sweep (HN-58). On origin/main (e5e926b4), `gam_predict::term_diagnostics::term_partial_dependence` has no caller anywhere: no crate, no pyffi binding, no CLI and no Python code references it. This PR only adds tests for it. I left it alone because this PR is active on that function. If there is no planned caller, deleting the function (and recording it in docs/source-removal-changes.json and docs/public-api-census-changes.json) may be better than hardening it.

---

PR #3815, opened 2026-09-20
fit_orchestration/materialize/standard: refuse an explicit linkwiggle beside a strict sas / beta-logistic link

### LINKED ISSUE #3814: fit_orchestration/materialize/standard: an explicit linkwiggle(...) beside a strict sas / beta-logistic link is silently dropped from the fit
## Defect

In `materialize_standard` (crates/gam-models/src/fit_orchestration/materialize/standard.rs), the binomial link-wiggle config is built as follows:

```rust
let link_kind = match link_choice.as_ref() {
    Some(c) => match StandardLink::try_from(c.link) {
        Ok(std_link) => InverseLink::Standard(std_link),
        // linkwiggle is gated by `linkname_supports_joint_wiggle` which
        // rejects Sas / BetaLogistic upstream, so reaching this arm
        // means the gate was bypassed.
        Err(_) => return None,
    },
```

The comment is wrong. The parse gate `linkname_supports_joint_wiggle` only covers a *flexible* link. An explicit `linkwiggle(...)` term next to a strict `link(type=sas)` or `link(type=beta-logistic)` passes it and reaches this arm. The arm then returns `None`, so the wiggle config is dropped. The fit runs without the link wiggle the formula asked for, and nothing warns the user:

```
b ~ x + linkwiggle(internal_knots=4) + link(type=sas)            # fitted as plain SAS, no wiggle
b ~ x + linkwiggle(internal_knots=4) + link(type=beta-logistic)  # same
```

This is dead-comment / silently swallowed code. The `Err(_) => return None` is claimed to be unreachable, but it is reachable, and it turns a request the engine cannot honour into a different model.

## SPEC

> The default choices should give the best model for the user.

## Fix

Refuse with this message:

`linkwiggle(...) is not supported with link '<link>'; the link wiggle is jointly fit only on a standard link`

A standard link keeps its wiggle as before. The new test module `linkwiggle_link_tests` covers both strict links and a positive control.

SauersML:
Before this change, an explicit `linkwiggle(...)` term next to a strict `link(type=sas)` or `link(type=beta-logistic)` was silently dropped. A comment claimed the `Err(_) => return None` arm was unreachable. It was reachable, because the parse gate only covers flexible links.

That request is now refused. A standard link keeps its wiggle as before.

The new test module is `linkwiggle_link_tests`.

Fixes #3814

---


SauersML:

---

Issue #998, opened 2026-06-10
Full-resolution residual_gauge: realise gauge orbits in the model's own (decoder, coordinate) parameter space — compensated symmetries as exact data-nulls, pinning from the penalty channel


SauersML:
## Context (from closing #995)

The residual-gauge certificate's parameter space is the per-atom **mean decoder frame** — a deliberate, self-contained compression. #995 made its verdicts honest about that compression: the relative-curvature statistic survives a full-rank span, and the lowering-error scale calibrates the tolerance so compression artifacts are never read as pins. On a production circle fit the certificate now correctly reports "resolution-limited within this atom; refusing to pin" rather than over-claiming identification.

But "refusing to claim" is the honest *floor*, not the goal. Working #995 surfaced why the frame space can never do better: the true gauge orbits of the model class — circle phase shift, output-frame rotations, equal-ARD rotations — are **compensated** motions (latent coordinates move AND the decoder counter-rotates, e.g. `Φ(t+ε)·R(−ε)B = Φ(t)B` for the harmonic circle). Their net action on the mean frame is identically zero, so any frame-space realisation of such a generator measures compression error, not the symmetry. Proposal (i) of #995 (per-row lifts) is ill-posed for exactly these generators.

## The object to build: the certificate in the model's own parameter space

Realise the generators where the orbits actually live — the fit's free parameters `(B_k decoder coefficients, t_n latent coordinates, logits)`:

- **Compensated orbits are exact null directions of the data Jacobian by construction**: the combined (coordinate, decoder) motion changes no fitted value, so the data can never pin them — no tolerance, no calibration, exact at first order. (Caveat that must be carried per-atom: compensation requires the basis family to be closed under the group action — true for periodic harmonics under shifts and full spherical harmonics under rotations, only approximate for Duchon/TPS patches under so(d). The closure status is per-`SaeAtomBasisKind` and should be a field on the verdict, with non-closed bases degrading to the #995 calibrated path.)
- **The entire pinning content then comes from the penalty/prior channel** — the isometry penalty's curvature along the orbit (non-isometric reparameterizations cost; isometries of the pinned metric don't) and the ARD prior (unequal-axis rotations cost). This is exactly where the #981 gauge-reduction ladder says identification lives (rungs 2 and 4: the pin does the collapsing, in whichever metric it is computed), and the Theorem-2 OutputFisher arm transfers unchanged: evaluate the isometry-pin curvature root along the exact orbit in the harvested metric.
[...]

PR #3837, opened 2026-09-20
manifold/construction: propagate a failed second-jet evaluation in atom_parameter_views

### LINKED ISSUE #3817: manifold/construction: atom_parameter_views swallows a failed second-jet evaluation, dropping the isometry pin from the orbit certificate
**Where:** `crates/gam-sae/src/manifold/construction.rs`, `atom_parameter_views` (around L2211 on origin/main).

The second jet is evaluated as `second_jet_dyn(...)`, which returns `Option<Result<..>>`. The code then applies `.and_then(|res| res.ok())`. That turns `Some(Err(e))`, meaning an analytic Hessian that *failed* on the fitted coordinates, into `None`. `None` is also what an evaluator with no analytic Hessian returns, so the two cases can no longer be told apart.

**Impact on the production path:** for a pin-active fit, the isometry orbit operator is built only when the second jet is `Some`. When the evaluation fails, the #998 exact-orbit certificate therefore judges the atom on the data residual alone. The data residual can never pin a model-class symmetry, yet the certificate still reports itself as pin-active. The evaluation error is lost silently, which violates the no-silent-error-swallowing rule.

**Fix (branch `hn1/second-jet-propagate`):**
- Evaluate the second jet only when the pin is active, since that is its only reader.
- Keep `None` for evaluators without an analytic Hessian.
- Propagate an evaluation error that names the atom.
- `atom_parameter_views` now returns `Result`; its single caller already returns `Result`.

SauersML:
Fixes #3817

atom_parameter_views turned second_jet_dyn's Some(Err) into None with `.ok()`, so an analytic Hessian that failed on the fitted coordinates looked like a basis with no Hessian. On a pin-active fit, the isometry orbit operator was then skipped. The #998 exact-orbit certificate judged the atom on the data residual alone, which can never pin a model-class symmetry, while still reporting a pin-active certificate.

The fix:
- Evaluate the second jet only when the pin is active, since that is its only reader.
- Keep None for an evaluator without an analytic Hessian.
- Propagate an evaluation error that names the atom.

The function now returns Result; its one caller already does.

Validation: focused-rust-proof run 35500967469 (key HN-1-secondjet-1).

---


SauersML:

---

PR #3843, opened 2026-09-20
manifold/crosscoder_fit: score layer R2 with the shared band-aware explained variance

### LINKED ISSUE #3822: manifold/crosscoder_fit: reconstruction_r2 is a private explained-variance copy whose tss>0 branch divides by rounding residue on constant layers
**Where:** `crates/gam-sae/src/manifold/crosscoder_fit.rs`, `fn reconstruction_r2` (around L421 on origin/main).

This is a private copy of the explained-variance formula that branches on `tss > 0.0`. It has two problems.

- **Wrong math and numerical instability.** A constant column's centred squares are rounding residue, not zero. For example, a constant 0.1 layer over three rows gives `tss ≈ 6e-34`. The "constant layer" branch is therefore unreachable exactly when it matters, and a fit that is 1e-9 off in relative terms reads **R² ≈ −5e13**. That value goes straight into the per-layer report.
- **Magic, non-invariant tolerance.** Its perfect-fit tolerance is `EPSILON * n * p * energy.max(1.0)`, which is an absolute floor below unit energy. This contradicts the function's own "magnitude-invariant" claim.

**Duplicated code:** `crate::k_selection::explained_variance` already implements explained variance with a rounding band derived from the column-mean computation (γ_{n+1}² · Σx²).

**Fix (branch `hn1/crosscoder-r2-shared-band`):**
- Delegate to the shared helper. Inside the band a reproduction reads 1 and anything else reads 0, so the reported value stays finite.
- Adds a regression test on a constant layer.

SauersML:
Fixes #3822

reconstruction_r2 was a private copy of the explained-variance formula that branched on `tss > 0.0`. It had two problems:
- A constant column's centred squares are rounding residue, not zero. A constant 0.1 layer over three rows has tss ~ 6e-34, so the "constant layer" branch was unreachable exactly when it mattered, and a fit 1e-9 off in relative terms read R2 ~ -5e13.
- Its perfect-fit tolerance, `EPSILON * n * p * energy.max(1.0)`, was an absolute floor below unit energy, contradicting its own "magnitude-invariant" claim.

The function now delegates to crate::k_selection::explained_variance, whose rounding band is derived from the column-mean computation (gamma_{n+1}^2 * sum x^2). Inside the band a reproduction reads 1 and anything else reads 0, so the reported value stays finite. Adds a regression test on a constant layer.

Validation: focused-rust-proof run 35501516398.

---


SauersML:

---

PR #3845, opened 2026-09-20
gpu survival row jet: use the shared PROBIT_NUMERICS_CU leaves instead of a stale private copy

### LINKED ISSUE #3828: gpu survival row jet: survival_rowjet_kernel.cu carries a stale private copy of the probit leaves (log Phi = 0 for x >= 8.5, cancelling left-tail curvature, 1e-300 floors) instead of the shared PROBIT_NUMERICS_CU
## Defect

`crates/gam-models/src/gpu_kernels/survival_rowjet_kernel.cu` (the NVRTC source for the survival marginal-slope rigid row jet, admitted in production when the device wins the timing race) carries its own copy of the probability leaves: `erfcx_nn`, `normal_pdf`, `normal_cdf` and `sp_logcdf_mills`, plus the curvature `k2 = mills * (margin + mills)` in `neglog_phi_stack`. The comment says they are "Full-precision probability leaves matching the CPU primitive contract". They are not. They have drifted from both the shared device numerics, `gam_gpu::numerics_device::PROBIT_NUMERICS_CU` (which exists "so a numerics fix is a one-line change instead of a coordination problem across multiple kernel source strings", and which the BMS row kernel already prepends), and the CPU leaf they lower, `unary_derivatives_neglog_phi` -> `normal_logcdf_derivatives`.

| leaf | device copy | CPU / shared device numerics |
|---|---|---|
| log Phi(x), x >= 0 | `log(fmin(fmax(Phi, 1e-300), 1))` | `log1p(-upper_tail)` |
| curvature `-d2 log Phi` | direct `lambda * (x + lambda)` everywhere | Laplace continued fraction for x <= -4 |
| erfcx | `exp(x*x) * erfc(x)`, no exact-square correction; 5-term series past 26 | fma square correction; 6-term series |
| sqrt / inverse_sqrt radicand | floored at `1e-300` | no floor ("a corrupted argument surfaces as a non-finite derivative") |

Measured against a 60-digit mpmath reference, with the old leaves compiled on the host:

- `-log Phi(x)`: relative error 1.5e-10 at x = 5, 5.6e-8 at x = 6, 4.1e-5 at x = 7 and 7% at x = 8. It returns exactly 0 for every x >= 8.5.
- Curvature: relative error 2.9e-12 at x = -100, 2.4e-10 at x = -1e3 and 3.4e-7 at x = -1e5.
- Mills ratio: relative error 2e-13 near x = -37.

The shared numerics are at or below 1e-14 at every one of those points.

## Why the parity test does not see it

`rigid_row_jet_device_admission_and_parity_2900` runs only on a CUDA host (CI has none). Its gap is `|a-b| / max(1, |a|, |b|)`, so absolute-tiny right-tail values pass at any relative error.

## Fix

- Prepend `PROBIT_NUMERICS_CU` in `survival_rowjet_source` and route `neglog_phi_stack` through `log_ndtr_mills_curvature`.
- Delete the private leaves and the `1e-300` floors, matching the host `unary_derivatives_sqrt` / `unary_derivatives_inverse_sqrt`.
- Pin the shared prelude, and the absence of the private leaves, in the source-level test that runs on CPU CI.

SPEC: duplicated logic; "In general, do not paper over solver issues." (the floors fabricate finite derivatives on the device only).

Branch: `claude/fix-survival-rowjet-probit-leaves`. Note: there is no CUDA CI lane, so the device path cannot be run in CI. The CPU source-pin test is the CI evidence.

SauersML:
Fixes #3828

- Prepends `gam_gpu::numerics_device::PROBIT_NUMERICS_CU` to the survival row-jet NVRTC source.
- Routes `neglog_phi_stack` through `log_ndtr_mills_curvature`.
- Deletes the private `erfcx_nn` / `normal_pdf` / `normal_cdf` / `sp_logcdf_mills` leaves and the `1e-300` radicand floors.
- Pins the shared prelude, and the absence of the private leaves, in a CPU source-level test.

There is no CUDA CI lane. The device path is validated by the host-compiled leaf comparison in the issue, and in CI only by the source pin.

Validation: PR CI, plus focused proof run 35501162215.

---


SauersML:

---

SauersML:
Taking over this stalled PR as the wave8 session RF-0748-1. There has been no push since 09:22Z, and focused proof run 35501162215 has sat queued since 09:01Z. My plan:
- merge origin/main in;
- run the survival_rowjet source tests locally;
- merge.

---

PR #3856, opened 2026-09-20
manifold/construction: derive the amortized encode amplitude bound without a 1.0 floor

### LINKED ISSUE #3855: manifold/construction: amortized_encode_target floors every atom's amplitude bound at 1.0, so the auto-derived per-atom bound is dead and certifiable rows are flagged
**Where:** `crates/gam-sae/src/manifold/construction.rs`, `amortized_encode_target`, L4540-4553 on origin/main:

```rust
// A strictly positive amplitude floor keeps the offline Lipschitz
// scaling finite for atoms with no active row in this corpus ...
amplitude_bound[atom_idx] = bound.max(1.0);
```

The doc says `amplitude_bound[k]` is "auto-derived from the fit": it should be the largest fitted assignment mass `a[i,k]`. The amplitudes come from `fitted_assignment_amplitudes()`, which are the posterior gates and lie in `[0, 1]`. The `.max(1.0)` therefore replaces **every** atom's bound with 1.0, so the per-atom derivation above it is dead code. The 1.0 is also a magic, scale-dependent constant. SPEC.md L23: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code."

**Why it matters (production path).** The bound feeds `EncodeAtlas::build` and then `hessian_lipschitz_constant`:
- There, `L = 3 z² |J||H| + (|x| + z|V|) z |T|`, which is nondecreasing in `z`.
- The chart's stored `lipschitz` is the `L` of the online per-row Kantorovich certificate `h = β·η·L ≤ ½`, in `refine_certified_encode_start`.
- An atom whose largest gate is 0.3 has its `L` overstated by up to about 11×. Its `h` is overstated by the same factor, so starts that are certifiable are flagged and sent to the colder start.

This runs on every fit, through two consumers:
- `amortized_encoder_consistency`, called from `fit_entry` L1445;
- the outer loop's warm start `warm_start_latents_from_amortized_encoder`, from `outer_objective` L2205. It adopts only converged/certified rows.

The comment's rationale does not hold either. With `bound = 0`, `L = 0` is finite, and the atlas already maps `L = 0` to the nominal chart radius. An atom with no active row is never certified from this corpus, because a non-positive amplitude is always flagged.

**Fix:** use the sup itself, `amplitude_bound[atom_idx] = bound`. Because `L` is monotone in `|z|`, this stays a valid bound for every row of the corpus that the atlas is built over and then encodes.

SauersML:
Fixes #3855

amortized_encode_target floored each atom's amplitude bound at 1.0. The amplitudes are the posterior gates, which lie in [0, 1], so the floor replaced every atom's auto-derived bound (its largest gate) with 1, and the derivation above it was dead.

The bound scales the encode Hessian-Lipschitz constant L. L is nondecreasing in |z| and is the L of the online per-row Kantorovich certificate h = β·η·L. An atom whose largest gate is 0.3 therefore had h overstated about 11×, and certifiable starts were flagged. That affects both the fit's `amortized_encoder_consistency` report and the outer loop's amortized warm start.

The fix uses the sup itself. It stays sound for every row of the corpus because L is monotone in |z|. An atom with no active row gets the exact L of its zero-amplitude objective.

Validation: focused-rust-proof (key HN-1-3855-2).

---


SauersML:

---

PR #3866, opened 2026-09-20
report: weight residual diagnostics by the fit's prior weights (#3852)

### LINKED ISSUE #3852: gam report: residual diagnostics and R² ignore the prior weights, so a correct weighted fit gets a non-normal Q-Q plot
## Where

`crates/gam-cli/src/main/run_sample_generate_report.rs` on `origin/main`.

`run_sample` resolves the saved fit's prior weights for ALO at line 442:

```rust
let alo_weights =
    resolve_weight_column(&ds, &col_map, model.payload().weight_column.as_deref())
```

It then never passes them to the residual diagnostics or to R². There are two callers:

- `report_residual_diagnostics(&family.response, &y_vec, &mu_vec, leverage.as_deref(), edf_total, &mut notes)` (line 703)
- the Bernoulli marginal-slope call at line 506

`report_family_residuals` (line 843) takes no weights:

```rust
let residual_dof = n as f64 - edf_total;              // n = every row, zero-weight rows included
...
ResponseFamily::Gaussian => {
    let ssr: f64 = (0..n).map(|i| (y[i] - mu[i]).powi(2)).sum();   // unweighted
    let sigma = (ssr / residual_dof).sqrt();
    ...
    let h = leverage
        .and_then(|l| l.get(i).copied())
        .filter(|h| h.is_finite())
        .unwrap_or(0.0)
        .max(0.0);
    Ok((y[i] - mu[i]) / (sigma * (1.0 - h).sqrt()))    // no sqrt(w_i)
```

The same problem appears in three more places:

- **Gamma, inverse-Gaussian and Tweedie.** The Pearson dispersion `phi = Σ (y-μ)²/V(μ) / (n - edf)` is unweighted. Each row's law then uses `phi` instead of `phi / w_i`: Gamma shape `1/phi`, IG `λ = 1/phi`, and the Tweedie deviance residual `sqrt(d/phi)`.
- **The Gaussian R² block (lines 540-555).** It uses the unweighted `ss_res` and `ss_tot` and the unweighted mean.
- **Zero-weight rows.** These are rows the fit treats as absent. They are counted in `n`, so they inflate `residual_dof`. Each one also gets a residual it has no variance for.

## Why it is wrong

The fit's own convention is `Var(y_i) = φ V(μ_i) / w_i`, with `φ̂ = Σ w_i (y_i − μ̂_i)² / V(μ̂_i) / (n₊ − edf)`, where `n₊` counts the positive-weight rows. See `gam-solve/src/estimate/optimizer.rs:2574-2587` (#584) and `smooth_term_lr.rs:3094-3103`. The ALO leverage the report feeds in is also the weighted one: `h_ii = W_H,i x_iᵀH⁻¹x_i`.

So `Var(y_i − μ̂_i) = σ²(1 − h_ii)/w_i`, and the studentized residual is `√w_i (y_i − μ̂_i) / (σ̂ √(1 − h_ii))`. Without `√w_i`, the residuals of a *correctly specified* weighted fit have per-row variance proportional to `1/w_i`.

Take a design with weights {1, 100} as an example. The residuals are a scale mixture, so the Q-Q plot, histogram and scale-location plot all show heavy tails. The report then tells the user a correct model is misspecified. σ̂ itself is also the wrong quantity: it is the unweighted RMS, not the fitted scale. R² has the same defect, because it measures the unweighted fit that the weighted optimizer did not minimise.

The leverage chain also swallows invalid values:

- `saved_alo_report_data` has already rejected non-finite leverages, so `.filter(|h| h.is_finite())` and `.unwrap_or(0.0)` are dead.
- For an identity-link Gaussian, `h_ii = w_i x_iᵀH⁻¹x_i ≥ 0`. A negative value is therefore a defect, yet `.max(0.0)` turns it into 0. SPEC: "In general, do not paper over solver issues."

## Fix

Pass the resolved prior weights into both callers.

- **Zero-weight rows.** Drop them from the diagnostic rows (y, μ, leverage, weights), because a zero weight is an absent row (#584). Check that the lengths agree instead of indexing with `get(i)`.
- **Gaussian.** Use `σ̂² = Σ w (y−μ)² / (n₊ − edf)` and `r_i = √w_i (y_i − μ̂_i)/(σ̂ √(1 − h_ii))`. Refuse a leverage outside `[0, 1)` instead of clamping it.
- **Gamma, IG and Tweedie.** Use the weighted Pearson `φ̂` and each row's dispersion `φ̂/w_i`: Gamma shape `w_i/φ̂`, IG `λ = w_i/φ̂`, and the deviance residual `sign·√(w_i d/φ̂)`.
- **R².** Use the weighted form `1 − Σw(y−μ)² / Σw(y−ȳ_w)²`, with the roundoff band applied to the weighted energy.

The discrete families and Beta/Student-t keep their per-row law. There a positive weight is a frequency weight, so it does not change the distribution of a single row. Only their zero-weight rows are dropped.

SauersML:
Fixes #3852

Report diagnostics now resolve the saved prior weights once, exclude zero-weight rows consistently, and use the fitted precision-weight conventions for Gaussian, Gamma, inverse-Gaussian and Tweedie residuals. Gaussian R² uses the weighted objective. Frequency-weight families keep their per-row law. Invalid weights, misaligned rows, nonfinite observed values, invalid leverage and unrepresentable residual arithmetic refuse explicitly. Individual Bernoulli residuals accept only 0/1 responses; unsupported trial-proportion residuals are omitted with a reason.

The draft's exact Gaussian residual-variance claim was incorrect for a penalized smoother, so it is replaced by five deterministic production-helper regressions: independent Gaussian/Gamma/IG/Tweedie formulas, common-weight-scale invariance, weighted R², unknown/invalid EDF behavior, and invalid-row/refusal controls. All five pass on MSI. Complete CLI production/unit metadata, eight affected library production crates, and complete model unit metadata pass with warnings denied using coherent warm dependencies. All43 report tests pass. No Gaussian fit Monte Carlo or exact finite-sample normality claim is made; leverage adjustment is documented as an approximation.


SauersML:

---

PR #3867, opened 2026-09-20
manifold/crosscoder_drift: rank only atoms live at every layer

### LINKED ISSUE #3864: manifold/crosscoder_drift: atom_total_drift scores a dead layer's NaN step as zero drift, so most_stable_atom names dead atoms
**Where:** `crates/gam-sae/src/manifold/crosscoder_drift.rs`, `CrosscoderDriftReport::atom_total_drift` / `extremal_atom` (origin/main 3c2b08bc).

```rust
pub(crate) fn atom_total_drift(&self, k: usize) -> f64 {
    self.atom_drift_profile(k)
        .into_iter()
        .filter(|d| d.is_finite())
        .sum()
}
```

`decoder_drift` (transport_law.rs) returns `NaN` for a step at which the atom's honest decoder is numerically dead. `atom_total_drift` drops those steps and sums the rest, so:

- An atom dead at **every** layer totals the empty sum `0.0`. That value is finite, so it passes the `is_finite` filter in `extremal_atom`, and `most_stable_atom()` returns the dead atom. It ties with, or beats, any genuinely layer-stable feature. The doc of `most_stable_atom` / `most_drifting_atom` says "`None` when there are no atoms or no finite drift", but that `None` branch cannot trigger through `atom_total_drift`.
- An atom that vanishes at one layer, which is the largest possible change in a feature, has that step scored as zero movement. A total built from fewer steps is also not comparable with a full-chain total, which biases the ranking toward "stable".

**Production path:** `crosscoder_fit.rs` L671-673 writes `most_drifting_atom` / `most_stable_atom` into the `SaeCrosscoderWireDrift::Measured` report returned to Python. ARD and assignment sparsity routinely shrink atoms out, so dead atoms are the common case here.

Defect class: wrong math / silent substitution. An undefined quantity is replaced by 0 and a published ranking is built from it, contradicting the functions' own documented `None` contract.

**Fix:** an atom's total drift is `NaN` whenever any step is undefined, so the existing `is_finite` filter in `extremal_atom` ranks only atoms that are live at every layer, and returns `None` when no atom is. `mean_drift` stays a mean over the defined steps. Add a regression test with a dead atom and a partially dead atom.

SauersML:
Fixes #3864

`CrosscoderDriftReport::atom_total_drift` dropped every NaN step (a layer at which the atom's honest decoder is numerically dead) and summed the rest. As a result:
- An atom dead at every layer totalled 0.0, passed `extremal_atom`'s `is_finite` filter, and was published as `most_stable_atom` in the crosscoder wire report.
- A feature that vanished at one layer had that step scored as zero movement.

Now the total is NaN whenever any step is undefined. The existing filter therefore ranks only atoms that are live at every layer, and returns `None` when there are none, which is what the docs already promised. `mean_drift` still averages over the defined steps.

Test: `atom_dead_at_a_layer_is_not_ranked` builds a report with a live atom, an atom dead at every layer, and an atom that is partially dead, and checks the totals, the ranking, `mean_drift`, and the all-dead `None` case.

---


SauersML:

---

PR #3868, opened 2026-09-20
analytic_penalties sheaf: PSD majorizer is the exact weighted Laplacian, not its diagonal

### LINKED ISSUE #3865: analytic_penalties sheaf: SheafConsistencyPenalty reports diag(L) as its PSD majorizer, so psd_majorizer_hvp and every FrozenAnalyticPenaltyOp query drop all edge coupling

## Defect

The `AnalyticPenalty` contract (`crates/gam-terms/src/analytic_penalties/penalty_trait.rs:323-359`) sets two defaults:
- `psd_majorizer_diag` returns a diagonal `B` with `B ⪰ ∂²P`.
- `psd_majorizer_hvp` applies that diagonal (`diag ⊙ v`) whenever `psd_majorizer_diag` is `Some`, and falls back to `hvp` only when it is `None`.

The `hessian_diag` doc (`penalty_trait.rs:261-266`: "Returns `None` for penalties whose Hessian is dense") and the `hvp` doc (`penalty_trait.rs:283-292`) say that penalties with a non-diagonal Hessian return `None` from `hessian_diag`, for exactly this reason.

`SheafConsistencyPenalty` has the dense sheaf Laplacian `weight·L = weight·δᵀδ` as its Hessian. Its trait `hessian_diag` still returns `Some(diag(weight·L))` (`sheaf.rs:622-632`), and it does not override `psd_majorizer_diag` or `psd_majorizer_hvp`. As a result:
- `psd_majorizer_diag` is `diag(weight·L)`. On any graph with an edge this is **not** `⪰ weight·L`, which violates the contract `B ⪰ ∂²P`.
- `psd_majorizer_hvp(v)` is `diag(weight·L) ⊙ v`, which drops every edge coupling. For a convex quadratic the majorizer should be the exact Hessian; the doc itself says "For convex penalties this is the exact Hessian-vector product".

The frozen operator inherits the problem:
- `FrozenAnalyticPenaltyOp::matvec` goes through `psd_majorizer_hvp` (`registry.rs:380-387`).
- `diag` goes through `diag_via_matvec` (`registry.rs:452`).
- `as_dense` probes `psd_majorizer_hvp` (`registry.rs:598`, `:606-612`).
- `log_det_plus_lambda_i` is taken of that dense matrix (`registry.rs:523`).

So the frozen Sheaf operator is `diag(weight·L)` everywhere. Its log det comes out as `Σ log(weight·L_ii + λ)` instead of `log|weight·L + λI|`.

## Reproduction (numpy)

I used the 3-vertex, 2-dim-stalk sheaf from `component_laplacian_matches_the_matvec_laplacian_2900`: a paired edge, a single-restriction edge and a self-loop, with weight 1.7. On it:
- `λ_min(diag(B) − B) = −2.10`, so the diagonal is not a majorizer.
- `λ_min(B) = 0.027`, so the true operator is well-conditioned PSD.

## Scope

`freeze` (`registry.rs:737`) has no production caller, and gam-sae does not accept Sheaf as a coordinate penalty. This is a latent contract violation on the public trait and frozen-operator surface, not a live fit error.

## Fix

Override `psd_majorizer_diag` to return `None` in the Sheaf `AnalyticPenalty` impl. `psd_majorizer_hvp` then falls through to the exact Laplacian `hvp`, and every frozen query (`matvec`, `diag`, `as_dense`, `log_det`) reads the exact PSD `weight·L`. That is the correct majorizer of a convex quadratic.

`BlockOrthogonalityPenalty` has the same `Some(exact diag of a dense Hessian)` shape, but it is the non-convex case, is live in gam-sae, and needs a real majorizer. It is filed as #3863.

SauersML:
Fixes #3865

`SheafConsistencyPenalty` now overrides `psd_majorizer_diag` so that it returns `None`. The trait-default `psd_majorizer_hvp` therefore falls through to the exact Laplacian `hvp` rather than applying `diag(weight·L) ⊙ v`. The penalty is a convex quadratic, and its exact Hessian `weight·L` is the correct majorizer. `diag(L)` is not `⪰ L` on any graph that has an edge.

Every `FrozenAnalyticPenaltyOp` query (`matvec`, `diag`, `as_dense` and log det) now reads the exact PSD operator.

A new test, `sheaf_psd_majorizer_is_the_exact_weighted_laplacian`, checks three things:
- `psd_majorizer_diag` is `None`;
- `psd_majorizer_hvp` equals `weight·L` column by column;
- the diagonal the old default applied was indefinite against `L` (minimum eigenvalue < −0.1).

Validation: focused-rust-proof (gam-terms lib, filter `sheaf_psd_majorizer`).

---


SauersML:

---

Issue #2949, opened 2026-09-17
Joint latent law for K ≥ 2 survival scores that transports a moving conditional mean and shape (retire gaussian-uncertified)


SauersML:
## What is missing

A joint latent law for K ≥ 2 survival marginal-slope scores (per-score slopes) whose transport follows a conditional law that moves on the marginal-index span, in mean and shape as well as in covariance.

gam#2929's joint law, `SurvivalJointLatentLaw` in `crates/gam-models/src/survival/marginal_slope/joint_latent_law.rs`, is built from whitened training residuals `ε = L(a)⁻¹(z − μ)` and moved to each row as `u_m = μ + L(a)·ε_m`. `μ` is one pooled mean, and `L(a)` moves only the second moment (the gam#2766 conditional `Σ(a)` when the pair gate escalates). A per-score conditional mean shift, or a skewness that moves on the span, is not in the transported law, so gam#2929 refuses a local-empirical per-score column at K ≥ 2 by name.

## Why it matters

Under gam#2926 the default estimates the conditional law of each score on the marginal-index span. Where a score's law moves (a mean, variance or skewness Rao test fires at level `1e-3`), the default law is local by context. At K ≥ 2 neither the joint frame nor the closed form carries that law, so the default fits the closed form and records `latent_law_consumed = "gaussian-uncertified"`, naming the moving score. That label is honest, but it must not be the permanent state of a real configuration. With a transport that follows the moving mean and shape, a moving K ≥ 2 fit would get the gam#2926 closed-form certificate `D̂ = Σ_i w_i (r_i² − 2·se_i²)/(π_i(1−π_i))`, evaluated on the transported law, and a re-solve when `D̂ > 0`.

## Every remaining `gaussian-uncertified` source

This issue tracks the full set; each entry names where the label is minted and what retires it.

1. **K ≥ 2 moving conditional law.** Once gam#2929 lands, a per-score K ≥ 2 fit whose decision for any column is local-empirical is routed to the closed form with every column standard normal, and the primary is recorded `gaussian-uncertified` (the gam#2926 K ≥ 2 follow-up, in `resolve_latent_score_calibration_from_parts`, `survival/marginal_slope/latent_measure.rs`). Retired by this issue.
[...]

PR #3871, opened 2026-09-20
Replay a gauged smooth in the fit's own product order (#3001)

### LINKED ISSUE #3001: Gaussian location-scale link wiggle is evaluated two ways, so predict on training rows misses the fit's own mean by up to 3 ulp
**Two evaluation paths for one quantity: the Gaussian location-scale link wiggle.** Predict-on-training-rows differs from the fit's own fitted mean by up to 3 ulp. σ matches bitwise.

- **Fit side** (`crates/gam-models/src/gamlss/gaussian/wiggle.rs:2024`, `GaussianLocationScaleWiggleFamily::evaluate`): the fitted mean is `q = eta_mu[i] + etaw[i]`. `etaw` is the wiggle block's state, `B(η_μ)·β_w`, from the fitting evaluation. After `rescale_gaussian_location_scale_to_raw` it is in raw units.
- **Predict side** (`crates/gam-predict/src/gaussian_location_scale.rs:183-186`, `plugin_eta`): `SavedLinkWiggleRuntime::apply(&(X·β_μ + offset))` rebuilds the wiggle basis at the base index and takes its own dot product with `β_w`.

The formula is the same and the evaluation order differs. That rules out a pin that asks "does the saved model predict exactly what the fit fitted" (bitwise equality): the difference stays at rounding level.

**Repro** (gnomon, at gam 5a2cbcaebb):
- 48 rows; score `(i − 24)/12`; sex `i % 2`; two PCs `sin(i·(3+2j)·0.37)`; `y = 2 + 0.7·score + 0.3·sex + 0.4·PC1·PC2 + 0.2·sin(1.7i)`.
- Model: Gaussian location-scale with the cubic triple-operator link wiggle. Terms are `s(score, type=duchon, centers=4)`, the linear sex term, and `s(PC1, PC2, type=duchon, centers=4)`, the same terms in both channels.
- Procedure: fit, then read `fit.fit.block_states`. Build the payload with `assemble_location_scale_payload`, save and load it, then predict on the same 48 rows.
- **Mean:** the largest gap between `predict(...).mean` and `states[0].eta + states[2].eta` is **8.9e-16 absolute, 6.4e-16 relative (about 3 ulp at |μ| ≈ 1.4)**.
- **σ:** `predict_noise_scale` equals `response_scale·LOGB_SIGMA_FLOOR + exp(states[1].eta)` **bitwise at every row**.

gnomon's test is `calibrate::estimate::tests::gaussian_predictions_replay_the_fits_own_fitted_mean_and_scale`. It keeps the σ half and defers the μ half to this issue.

**Root fix:** evaluate the wiggle through one function for both the fit's wiggle block state and `SavedLinkWiggleRuntime::apply`, with the same basis evaluation and the same summation order. The saved model then reproduces the fit's μ bitwise, as it already does for σ.

SauersML:
Fixes #3001

When a saved model rebuilt its design from the frozen spec, it formed `B*(Q*T0)`, but the fit had formed `(B*Q)*T0`. The cause was the freeze: it combined the joint-null rotation Q and the collection chart T0 into a single frozen transform. C, R and the centers were bit-identical. Only the order of the matrix products differed, and that left mu 2 ulp off in 11 of the 48 rows of the #3001 location-scale wiggle model.

- The collection chart now carries T0 (`coefficient_transform`), the term-local chart z_local and Q. The freeze writes the term-local chart and Q into the spec instead of the composition z_local*Q*T0.
- On replay, the basis build applies z_local, the aggregation loop applies the persisted Q, the chart branch applies T0, and -C*R goes through `subtract_row_space_correction`.
- The fit now forms X*T0 - C*R through that same operator (`FixedRowSpaceProjector::row_space_correction` replaces `project_design`). Fit and replay therefore evaluate one product in one order.
- A spec carrying a frozen collection chart counts as frozen, so a rebuild neither re-derives Q nor re-runs the owner analysis. Factor-smooth kinds, whose collection transform is persisted apart from the chart, are excluded through `basis_absorbs_collection_chart`.
- `restore_local_identifiability_chart` and the post-freeze loop in the incremental psi driver are deleted, since the freeze now writes the local chart.
- The measure-jet input-variance reads the replayed design's metadata chart, because the frozen spec no longer holds the whole composition.
- The #3001 test now asserts that the rebuilt design reproduces the fit's mu and sigma bit for bit (`rebuilt_mean_differing == 0`, `rebuilt_sigma_differing == 0`).

This is a breaking change. Payloads saved with the old composed chart lack the new fields and are refused on load. There is no compatibility default.

Ledger entries for the removed items are in `docs/source-removal-changes.json` and `docs/public-api-census-changes.json`.

---


SauersML:

---

SauersML:
Only one of #3871 and #3876 should land. They are the same repair of #3001 step 2: the basis frozen in its term-local chart, `Q` persisted, `T0` on the residualization chart, `C·R` through the shared `subtract_row_space_correction`, the measure-jet chart read from the replayed metadata, and `restore_local_identifiability_chart` deleted. The factorization is right. The full comparison, with the lane's evidence, is on #3876.

What must change before either lands: **old payloads must keep replaying.**
- This PR adds a required `ParametricResidualizationChart::coefficient_transform` without changing `MODEL_PAYLOAD_VERSION`. A payload saved before it has the same version number and no such field, so it fails to load with a serde missing-field error, not even the named `payload_version_mismatch` refusal.
- The new fields should take a serde default, so an old payload deserializes with "no factored chart" and replays its composed chart exactly as before. The old readers and `warm_start_from` then keep working.
- A test pins an old-payload replay: a stored old-shape frozen spec, with the replayed design's and X·β's bit hashes asserted.
- Version: gam-3009 (#3020) and sms-k (#2949) both claim v35, so this takes main+1, read at push time.

Two cases are also outside this PR's evidence: the Residualize (Matérn) arm, and a free-boundary sum-to-zero B-spline. That B-spline is built sparse at fit but replays a `FrozenTransform` through dense `fast_ab`, so fit and replay differ even with the factored chart. The B-spline fix lands separately from the lane (`bspline_build.rs`, `duchon_thinplate.rs`).

If this PR merges as it is, the lane will land a follow-up on top that restores old-payload replay, with the pin.

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, afb13132309d.

PR #3872, opened 2026-09-20
dictionary: linear_dictionary_transform routes with the fitted assignment rule

### LINKED ISSUE #3869: dictionary: linear_dictionary_transform always routes top-k ridge LS, so a softmax-fitted linear dictionary's transform() is not its fitted encoder

## Defect

When a linear dictionary is fitted with `assignment="softmax"` and `top_k >= 2`, its held-out `transform()` uses a different encoder from the one that produced its training codes.

**Fit path.** The fit routes every row through `reroute_against_atoms` (`crates/gam-terms/src/dictionary.rs:536-548`), which dispatches on the configured rule:
- `TopK` goes to `top_k_assignments(x, atoms, top_k, code_ridge)`, an active-set ridge least-squares solve.
- `Softmax` goes to `softmax_assignments(x, atoms, top_k, temperature, code_ridge)` (`dictionary.rs:911-954`). This weights each active atom's ridge projection by a tempered softmax over |cosine| scores.

The certified `fit.assignments` are exactly this reroute against the final atoms (`plain_atom_step`, `dictionary.rs:261`; the initial routing is at `:294`).

**Transform path.** `linear_dictionary_transform` (`dictionary.rs:890-909`) always calls `top_k_assignments`, whatever rule the model was fitted with. Its doc (`:884-889`) claims it uses "the same top-`top_k` ridge least-squares routing the fit uses". No layer above it can pass the rule through:
- The ffi `linear_dictionary_transform_ffi` (`crates/gam-pyffi/src/manifold/geometry_ffi.rs:5518-5531`) has the signature `(x, atoms, top_k, code_ridge)`, with no `assignment` or `temperature`.
- `LinearDictionaryFit.transform` (`gamfit/_linear_dictionary.py:66-90`) passes only `x`, `atoms`, `top_k` and `code_ridge`. The dataclass keeps `assignment` but never stores the fitted `temperature`.

As a result, `fit.transform(X_train)` does not equal `fit.assignments`. For held-out data, the codes come from an encoder that was not the one fitted.

## Reproduction

I replicated the two routines in numpy with this setup:
- atoms `[[1,0,0],[0.6,0.8,0],[0,0.3,0.9]]`
- `top_k = 2`, `temperature = 0.25`, `code_ridge = 1e-8`
- rows `[[1,0.5,0.1],[0.2,1.1,-0.4],[-0.7,0.3,0.9],[0.4,-0.2,0.6]]`

The top-k codes (what `transform` returns) and the softmax codes (what the fit stores) differ by up to **0.63** per entry. For example, row 2 gives `[-0.625, 1.375, 0]` against `[0.008, 0.961, 0]`.

With `top_k = 1` the two rules coincide: the single-atom softmax weight is 1, and the ridge projection equals the 1×1 ridge solve. That is why the default configuration does not show the problem.

## Fix

- `linear_dictionary_transform` takes the fitted `assignment` and `temperature` and routes through the same dispatch as `reroute_against_atoms`, so the routing rule has one source of truth. `temperature` is validated for softmax.
- The ffi gains `assignment` and `temperature` kwargs, and both `.pyi` stubs are updated.
- The Python `LinearDictionaryFit` stores the fitted `temperature`, and `transform()` passes both.

SauersML:
Fixes #3869

The fit and the out-of-sample transform now share a single dispatch, `route_against_atoms`, covering both top-k ridge least squares and tempered softmax. As a result, a softmax-fitted linear dictionary's `transform()` now reproduces `fit.assignments`.

## Changes
- **Rust core:** `linear_dictionary_transform` takes `assignment` and `temperature`. It rejects a softmax temperature that is not finite or not positive.
- **ffi:** `linear_dictionary_transform_ffi` takes `assignment="top_k"` and `temperature=0.25` as kwargs. Both `.pyi` stubs are updated to match.
- **Python:** `LinearDictionaryFit` stores the fitted `temperature`, and `transform()` passes both `assignment` and `temperature`.
- **Tests:** `transform_routes_with_the_fitted_assignment_rule` checks that, for both rules, `transform` equals the fit's reroute to 1e-14. It also checks that on non-orthogonal atoms with `top_k = 2`, the softmax codes differ from the top-k codes.

Validation: focused-rust-proof (gam-terms lib, filter `transform_routes_with_the_fitted_assignment_rule`) plus PR CI (pyffi build and Python).

---


SauersML:

---

Issue #2392, opened 2026-07-22
Tensor te() default double-penalty λ rails to the ρ-box and the certification layer cannot mint it: StalledAtValidMinimum rejected at assembly + no mixed single-rail-coordinate certificate


SauersML:
Four tensor-smooth quality tests GAM_ERROR at HEAD while mgcv fits the same data in seconds. Triage (dashboard run 29904886395) shows all four share one structural cause with two certification-layer manifestations.

## Shared structure

gam defaults `double_penalty=true` for every smooth including `te()` (`crates/gam-terms/src/term_builder.rs:1839` `unwrap_or(true)`, threaded to the tensor spec at `:3570-3576`), so a 2-margin `te(x,z)` carries **3 λ** (`term_specs.rs:5909-5911`: MarginalKroneckerSum → margins + 1 null-space penalty). mgcv's default `te()` has 2 sp (no null-space penalty, `select=FALSE`) and converges. Per SPEC the double penalty must STAY the default (null recovery is opt-out) — so the defect is that our certification layer cannot certify the optima this default produces:

## Manifestation 1 — assembly gate rejects a certified stationary optimum on the inner label

- `quality_vs_mgcv_poisson_tensor` (`te(x,z,k=[6,6])`, poisson): λ̂=[4.57, 26.6, 3.99], outer |Pg|=3.46e-6 **within** bound 9.12e-6 — but the fit is refused because the inner P-IRLS ended `StalledAtValidMinimum` and the assembly gate (`result_types.rs:1510`) accepts only `Converged`.
- `quality_vs_mgcv_tensor_te_2d_beta`: same shape — |Pg|=1.95e-6 within bound 6.3e-3, refused on the inner label.

`StalledAtValidMinimum` is defined as "gradient and Hessian indicate a valid minimum", and other production sites already treat it as converged (e.g. `gradient_hessian.rs` matches `Converged | StalledAtValidMinimum` for the Tweedie φ freeze). Rejecting it at assembly while the outer certificate holds discards a certified fit for a label.

## Manifestation 2 — no certificate for a single railed coordinate with interior rest

[...]

Issue #2388, opened 2026-07-20
Tail-snap probe fabricates a hard-zero gradient tail past the edf-floor box ceiling (silent clamp-masking on the gradient channel)


SauersML:
## Summary

The certify-time **tail-snap** probe ladder fabricates a **hard-zero gradient tail**: the candidate coordinate's projected gradient decays/grows normally, then drops to **exactly `0.000e0`** at one probe step and stays there for every probe beyond. Because the tail-snap machinery cannot fit an exponential decay law to literal zeros, it declines ("no finite-difference-clean tail run") and the outer refuses to mint a fit that is otherwise finite. A step function `1.531e0 → 0.000e0` in a single e-fold is not physical saturation — it is a **silent clamp/mask on the gradient channel**.

Surfaced while characterizing #2387 Family B (`gaussian_location_scale_termswith_matern_spatial_blocks_fit_finitely`).

## Evidence (probe ladder, verbatim from the refusal)

Fixture (pre-redesign): n=32, deterministic linear `y = 0.5*x0 - 0.25*x1 + 0.1`, Matérn mean corr 0.35 / noise corr 0.6. Certificate:

```
NOT STATIONARY: |g|=|Pg|=2.593e-2 bound=5.240e-3 hessian_psd=NO railed=[1,3]
asymptote-rail declined: interior not stationary: |Pg_int|=2.322e-2 > 5.240e-3
tail-snap declined: candidate k=0 tail unconfirmed: no finite-difference-clean tail run; probes:
  (ρ=-1.22, g=4.710e-2, ĉ=1.602e-1)
  (ρ=-0.22, g=1.408e-1, ĉ=1.762e-1)
  (ρ= 0.78, g=4.540e-1, ĉ=2.090e-1)
  (ρ= 1.78, g=1.531e0,  ĉ=2.594e-1)
  (ρ= 2.78, g=0.000e0,  ĉ=0.000e0)   <-- abrupt cutoff
  (ρ= 3.78, g=0.000e0,  ĉ=0.000e0)
  ... 13 more probes, all g=0.000e0, ĉ=0.000e0 ...
  (ρ=15.78, g=0.000e0,  ĉ=0.000e0)
final objective -5.139833e1 (finite); after 3 outer iteration(s)
rho_checkpoint=[-2.2244590525331858, 1.9612871659156117, 1.763328280287719,
                1.959897600338135, 1.1917556066290032, 1.946541199416681]
```

[...]

PR #3874, opened 2026-09-20
Derive rail-tail certificate bands from the probes' own rounding (#3565)

### LINKED ISSUE #3565: rho_optimizer: the λ→∞/λ→0 rail certificate mints fits on absolute tolerances fitted to one synthetic fixture (EXP4_* 1e-8/1e-6/1e-3, RAIL_TAIL_DRIFT_REL 1e-2, 18 probes)
## Where (origin/main)

**`crates/gam-solve/src/rho_optimizer/asymptote_certificate.rs` ~217–245**

```rust
pub(crate) const EXP4_INTERIOR_GRAD_TOL: f64 = 1.0e-8;
pub(crate) const EXP4_TAIL_NOISE_FLOOR: f64 = 1.0e-6;
pub(crate) const EXP4_TAIL_DRIFT_REL: f64 = 1.0e-3;
pub(crate) fn exp4_rail_bands(estimand_tol: f64) -> Self { ... }
```

**`crates/gam-solve/src/rho_optimizer/run.rs`**

- **5542**: `ASYMPTOTE_ESTIMAND_REL_TOL = 1.0e-4`
- **5555**: `ASYMPTOTE_PROBE_COUNT = 18`. The comment says the old 12-probe window "stopped one row short" of the #2299 fixture's clean band.
- **5567–5568**: `ASYMPTOTE_LOCAL_PROBE_DELTA = 0.5`, `ASYMPTOTE_LOCAL_PROBE_COUNT = 6`
- **6038**: `RAIL_TAIL_DRIFT_REL = 1.0e-2`. It is "measured on #2299".
- **~5257 and ~5689**, the production use in the terminal certificate and the wrong-rail pull-back:

  ```rust
  let mut tol = AsymptoteTolerances::exp4_rail_bands(ASYMPTOTE_ESTIMAND_REL_TOL * (1.0 + beta_norm));
  // Real REML tails hold ĉ to ~5e-3 relative, not the exp4 synthetic
  // characterization's 1e-3 (measured on the #2299 fixture during Inc 2c).
  tol.tail_drift_rel = RAIL_TAIL_DRIFT_REL;
  ```

## Why it is wrong

This certificate decides whether a railed coordinate is **certified stationary even though its projected gradient never cleared the stationarity bound** (`CertifiedAtAsymptote`). So these numbers directly decide whether a fit object is returned as converged. Every one of them is a value read off one synthetic run (`exp4_rail.py`) or one fixture (#2299), and in one case bumped until that fixture passed:

1. **`interior_grad_tol = 1e-8` is an absolute gradient magnitude.** ∂V/∂ρ scales with the criterion, which is O(n) for REML/LAML. At n = 1e6 a round-off-level gradient is far above 1e-8. At small n a genuine tail gradient can fall below it.

2. **`tail_noise_floor = 1e-6` is an absolute floor on ĉ = ∓e^{±ρ}∂V/∂ρ.** The module doc says this floor is the Lemma 3.4 summation-model floor (~nε|V|) "scaled to the gradient's regime". The code never computes it: it is a constant. The ĉ that corresponds to that gradient noise is nε|V|·e^{±ρ}, which depends on n, |V| and the probe ρ. A fixed 1e-6 is meaningless off the fixture.

3. **The drift band is 1e-3 and then overridden to 1e-2**, because a real fixture measured drift of 4.6e-3. Its width is not tied to the curvature contamination the tail law predicts, which is the next term in the asymptotic expansion.

4. **The probe ladder is sized to reach one fixture's clean band.** It uses 18 unit e-folds, then 6 half e-folds. On problems whose clean band sits deeper, the certificate declines. Where the band sits closer, extra evals are spent. The drift band then decides on whatever three samples land in it.

5. **`ASYMPTOTE_ESTIMAND_REL_TOL = 1e-4·(1+‖β‖)` is the certify gate itself** (estimand equivalence), and it is an arbitrary relative tolerance.

SPEC.md:
- "A fit object must only ever come from a converged optimization."

## Principled fix

Derive each band from the problem at the probe point:

- **Gradient noise floor.** Use the forward-error bound of the analytic ρ-gradient. As a minimum that is the Lemma 3.4 summation floor `n·ε·|V|`, or better the log-det forward-error machinery already used elsewhere (`logdet_forward_error`). Derive both `interior_grad_tol` (on g) and `tail_noise_floor` (mapped to ĉ by `e^{±ρ}` at each probe) from it.
- **Drift band.** Bound the next-order term of the tail law, O(e^{∓2ρ}) relative to the leading one. At probe depth this is estimated from the window's own fit residual, not from a fixed 1e-3/1e-2.
- **Probe placement.** Place probes until the per-probe noise floor exceeds the predicted tail magnitude (the clean band's rail-side edge is where `c·e^{∓ρ}` meets the noise floor). That edge is computable once the first clean samples give ĉ. There is then no fixed count of 18.
- **Estimand tolerance.** Take it from the caller's declared estimand resolution, the same place the stationarity requirement comes from (#2568), rather than 1e-4.

Retire `exp4_rail_bands` from production. It can stay only as the characterization-test fixture it describes.

## Tests to add

- The same rail fixture with the response rescaled by 1e3, and with n duplicated 100×. The certificate verdict must be invariant. Today the absolute floors change it.

Related (closed): #2348, #2388, #2813.

SauersML:
Fixes #3565

The measured asymptote-rail certificate now judges its probes in quantities they carry, not hand-set numbers.

**Removed** (none of these was derived):
- the absolute finite-difference noise floor
- `RAIL_TAIL_DRIFT_REL = 1e-2`
- `ASYMPTOTE_ESTIMAND_REL_TOL = 1e-4`
- `ASYMPTOTE_PROBE_COUNT = 18`
- the half-e-fold fallback ladder
- `AsymptoteTolerances`

**Now:**
- **Probe bands.** Each probe's gradient carries its own Theorem 9 rounding bound, `ε = γ_{n+p²}·Σ|channels|`, via `outer_coordinate_bands`. The capture is armed per probe. A gradient inside its band has no sign.
- **Settlement.** `settle_tail_constant` treats consecutive pencil constants as banded intervals:
  - resolvable growth toward the rail refuses;
  - a contracting series is summed to a rigorous radius `R` with `|c − ĉ| ≤ R`;
  - it requires `ĉ − R > 0`.
- **Certification gate.** The rail certifies when the value still available at the rail, `(ĉ + R)e^{∓ρ̂}`, is at most the criterion's statistical resolution `1/(2n)`. The analytic face proof uses the same gate.
- **Ladder.** Unit e-folds from the rail to the box midpoint, always strictly inside the box (#2388). The first three resolved, settling probes decide.
- **Wrong-rail pull-back (#2392).** Uses the same bands and the same settlement.
- **Evidence.** `RailTailEvidence::ProbedTail { gradient_band, extrapolation_radius }`, and `admits()` re-checks `ĉ > R`.

**Tests** (`asymptote_rail_certify_tests.rs`, `asymptote_certificate.rs`, `result_types.rs`):
- An exact tail certifies from 3 probes.
- A drifting constant walks the whole ladder and refuses.
- #2358: the next-order correction settles, and the radius covers the true limit (margin comes from the bands alone).
- Ladder semantics.
- Invariance to the criterion's scale (1e-9, 1, 1e3) and to an additive constant.
- The settlement radius covers the next-order limit.

---


SauersML:
**Two findings from #2946 T8 that bear on this PR's probe bands** (read against origin/main beb37608c1; nothing was run)

1. **Every production route already publishes the per-coordinate channels, so a band-based check has inputs everywhere.**
   - The unified REML/LAML evaluator publishes per-coordinate gradient parts for every family whenever capture is armed: `capture_rho_parts = capture_drift_split || certificate_parts_capture_enabled()` (`crates/gam-solve/src/reml/reml_outer_engine/objective.rs:1227`).
   - They are published after the KKT fold, at `objective.rs:1548`, so the channels are those of the gradient that is returned.
   - The SAE objectives publish too (`support_outer.rs:1097` and `:1117`).
   - Only the test mocks publish nothing. So `outer_coordinate_bands` gets its `Σ|channels|` on every route this certificate runs on in production. It needs no fallback for a missing channel list, and it should not grow one.

2. **A band built only from rounding calls the measured large-ρ plateau signal.**
   - #2450 measured a gradient plateau of about 1.33e-7 at ρ = 21..30 (`iso_kappa_reml_gradient_fd_tests.rs`, around line 3735).
   - The rounding band here is `γ_m·Σ|channels|` with `m = n + p²`, which is about 1e-12 on that fixture. The plateau is five orders above it.
   - That plateau is not rounding. It is the inner solve's residual showing up in the outer gradient, `δg ≈ ∂g/∂β · H⁻¹ r`.
   - So a rail probe on such a tail reads a resolved, non-decaying gradient. Two things have to hold for this PR to be right there:
     - the settlement has to keep refusing that tail as "resolvable growth / non-settling", as it does now; or
     - the inner-residual charge `|∂g/∂β · H⁻¹ r|` has to be added to each probe's band, with `r` the certified KKT residual the solution already carries.
   - The second option is the derived one: it charges the probe for the error it actually has. With only rounding in the band, a tighter inner tolerance would change which rails certify.
   - The value side already has this charge. `InnerResidualCharge` (`outer_eval_capture.rs:561`) carries `E_r = ½rᵀH_β⁻¹r` into `ObjectiveBand::inner_residual` (`crates/gam-solve/src/rho_optimizer/decrement_bands.rs:303`). The gradient bands have no counterpart: `outer_decrement_bands` charges each gradient component rounding only. The gradient analogue is first order in `r`, not second. So it is the charge a probe gradient near the rail is actually short of.
   - A pin that plants a constant plateau of about 1e-7 on an exact tail would show which of the two holds.

---

SauersML:
Taking over from the owner of claude/fix-3565-derived-rail-bands (no activity since 2026-09-20T09:31Z, when cd137daa43 was pushed).

The branch conflicts with main in `rho_optimizer/run.rs` and `asymptote_rail_certify_tests.rs`. The cause is main's covered zero-smoothing face proof (6bdaf14e90). Main now splits the tail-law coordinates into `upper_face` and `lower_face`, and it gates the λ=0 proof on `estimand_travel <= ASYMPTOTE_ESTIMAND_REL_TOL·(1+‖β‖)`. This branch deletes that constant.

Here is how I plan to resolve it:
- **Upper face (λ→∞).** Keep main's split. Both the analytic proof and the measured ladder run on it, and use this branch's derived signatures (`inputs.config`, `inputs.objective_tol`). Main has no measured λ→0 fallback, so there is none here either.
- **Lower face (λ→0).** Replace the relative coefficient tolerance with the same gate the λ=∞ proof uses here: the first-order value gap `V(ρ̂) − V(0) = Σ c′_j λ_j` must be at most the criterion's statistical resolution `inputs.objective_tol`.
- **`native_coordinate_order`.** Main's uses of this field become `inputs.native_order()`, because this branch replaced the field with `config`.
- **Tests.** Main's two zero-smoothing tests are kept, built from `test_config()`, with `objective_tol = outer_criterion_resolution`.

After that I will run one narrow proof and then land the PR.

---

SauersML:
Handoff: main is merged in as 1cda4f0155 with no conflicts left. Its proof run is https://github.com/SauersML/gam/actions/runs/35512076271 (queued; gam-solve lib, filters `asymptote_rail_certify_tests asymptote_certificate zero_smoothing_face result_types`). I cancelled the older queued run 35502556940, because it was for the superseded head cd137daa. I re-read the approach against SPEC: the bands and the gate are derived (γ_{n+p²}·Σ|channels|, a settlement radius, and the 1/(2n) criterion resolution), and the removed constants have not come back. If the run is green, this is ready to be marked ready and merged with expectedHeadSha 1cda4f0155b10121138ed8d2647052ee68dd98fb, and #3565 can then be closed.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, add2d37a4d9c.

PR #3875, opened 2026-09-20
fit_orchestration: keep identifiability="linear" smooths off the spline-scan fast path

### LINKED ISSUE #3870: fit_orchestration: spline-scan fast path fits the linear trend that identifiability="linear" removed
**Where:** `crates/gam-models/src/fit_orchestration/entry.rs`, `spline_scan_fast_path` (around line 3440 on `origin/main` f8c66133).

**What goes wrong:** the detector never reads `bspec.identifiability`. Its doc comment says:

> Identifiability transforms on the smooth (centering / linear-trend removal / orthogonality-to-intercept) are accepted as eligible: they only re-coordinate the unpenalized null space against the implicit intercept and do not change the fitted posterior of `E[y|x]`

That is true for sum-to-zero centring and for `none`. The intercept puts back the one constant those remove. It is false for `BSplineIdentifiability::RemoveLinearTrend`, which the formula DSL exposes as `identifiability="linear"` (`term_builder.rs` `parse_bspline_identifiability`).

That transform removes both the constant and the linear direction from the smooth. It is built from the Greville chart, so the coefficient vector of `x` is projected out. The dense model the user asked for is therefore `intercept + (penalized non-linear part)`, with no `x` direction.

The scan always solves the smoothing-spline posterior with an unpenalized `{1, x}` null space. So

```
y ~ s(x, bs="ps", degree=3, penalty_order=2, double_penalty=False, identifiability="linear")
```

is routed to `FitResult::SplineScan`, which fits and predicts the linear trend the formula removed. The same happens for the `order` = 1 and 3 arms. `OrthogonalToDesignColumns` and `FrozenTransform` can also impose arbitrary constraints that the scan ignores.

Every front end is affected, because the scan is the route inside `fit_from_formula`: the CLI `run_fit`, pyffi, and `fit_spline_scan_from_formula`. Each silently fits a different model than the one specified. There is no error and no note.

**Fix:** accept only `None` and `WeightedSumToZero`, the two policies whose span together with the intercept equals the scan's model. Every other policy falls through to the dense path. Correct the doc comment to match.

SauersML:
Fixes #3870

`spline_scan_fast_path` now accepts only the `None` and `WeightedSumToZero` B-spline identifiability policies. The scan always fits an unpenalized `{1, x}` null space. `identifiability="linear"` (`RemoveLinearTrend`) removes the `x` direction, and design-column or frozen transforms can impose any constraint. All three now fall through to the dense path, which fits the model the formula specifies.

I also corrected the doc comment, which claimed that linear-trend removal was harmless.

Tests in `spline_scan_identifiability_routing_tests`:
- Default centring and `sum_tozero` still route to the scan.
- `identifiability="linear"` does not.

---


SauersML:

---

SauersML:
Conflict note from BF-71 (conflict resolver). I have not touched this branch, because it has active owners.

Against main 74c2b1d788, head 8808f05a74 conflicts in one file:
- `crates/gam-models/src/fit_orchestration/entry.rs`

The rest merges cleanly.

---

Issue #3303, opened 2026-09-19
Custom-family Laplace criterion silently drops material negative eigenvalues of M through the penalty-rank floor; at survival LS + linkwiggle constrained modes M is indefinite (−4.58, −2.22) and the kept rank walks 8→9→10


SauersML:
**Rule.** `docs/convergence_theory/multinomial-ls.md` item 6 (origin/claude/spec-md-review-63dqtb) says: "The LAML exists only at a strict local mode. If H_obs + S_λ has a negative eigenvalue at β̂ … no Laplace approximation exists. Replacing negative eigenvalues by their absolute values or flooring them changes the objective, so it is not allowed." gam-solve's dense path already enforces this. `DenseSpectralOperator::from_eigenpairs` (`crates/gam-solve/src/reml/reml_outer_engine/dense_spectral.rs:200-214`) refuses any excluded eigenvalue outside the rounding band: "A negative saddle direction must not be discarded". The custom-family criterion does not.

**Where.**
- `crates/gam-custom-family/src/covariance.rs:2237`: `joint_penalty_subspace_trace_parts` keeps `laplace_precision_kept_eigenpairs(m_slice, penalty_rank)` and prices `½log|M|₊` over the kept set only.
- `crates/gam-custom-family/src/blockwise_solve.rs:2283`: `laplace_precision_kept_eigenpairs` truncates to `DenseSpectralOperator::identified_rank`, which is `resolved.max(penalty_rank.min(positive))` (`dense_spectral.rs:102-110`). When fewer eigenvalues are positive than `S_λ` has rank, the floor applies, and every negative eigenvalue is dropped. Nothing checks whether a dropped eigenvalue is material.
- `crates/gam-custom-family/src/assembly.rs:1169`: when the constrained Laplace normalizer (gam#2765) is on, `face_tangent` is `None`, so this is the full-space `M`.
- `crates/gam-solve/src/reml/reml_outer_engine/objective.rs:610`: the normalizer's `cone_solve` is `kernel.apply_pseudo_inverse`, which is the pseudo-inverse over the same kept set. The normalizer is derived as the truncation of `N(β_unc, M⁻¹)` to `C` (`constrained_posterior/cone_normalizer.rs:5-14`). That law exists only when `M ≻ 0`.

[...]

PR #3876, opened 2026-09-20
Replay a gauged smooth's design in the fit's order (#3001 step 2)

### LINKED ISSUE #3001: Gaussian location-scale link wiggle is evaluated two ways, so predict on training rows misses the fit's own mean by up to 3 ulp

SauersML:
Fixes #3001

Saved smooths now replay the fit's matrix products in their original order: the local chart, joint-null rotation, and collection chart remain separate, followed by the same row-space correction used during fitting. Numeric and factor by-variable terms retain the fit's rotation decision. The incremental spatial realizer no longer restores a chart that freezing has already preserved correctly.

This changes the persisted chart representation. The payload version is **36**, following main's existing version 35 for the latent-Z variance-stage covariance. Only the current version is readable; older payloads and warm starts are refused. The integration resolution preserves main's Matérn `include_intercept` refusal and the newer removal of redundant payload fields.

MSI validation using existing warm dependencies and two CPUs:

- Complete `gam-terms` unit binary compiled. The replay regression has zero differing entries out of 384 in each channel; its composed-order mutant differs in 191 entries per channel.
- The existing prediction-design parity test passed.
- Complete resolved `model.rs` compiled with its exact private support helpers. Both SAS save/load regressions passed, preserving the 96-field payload schema. Additional loader checks refuse versions 0 through 35 and version 37 by the named schema error, while version 36 round-trips.

The full location-scale fit and whole workspace suite were not run in this review. The replay check uses the original 48-row materialized request serialized across the crate boundary, so its term specification and numeric data remain unchanged while exercising the revised terms implementation.


SauersML:
**VF-0 review: not merging yet.**

The replay change `(B·Q)·T0 − C·R` looks principled. But the issue's end-to-end acceptance test `a_saved_gaussian_location_scale_wiggle_model_predicts_its_own_fitted_mean_and_scale_3001` still fails before reaching its asserts, on this branch and on main (the #3303 Laplace-indefinite refusal). So "Fixes #3001" is unverified.

This also bumps `MODEL_PAYLOAD_VERSION` to 35, the same as #3222, and the two will collide.

Please change it to "Refs #3001", or land the fix once the acceptance test passes.

---

SauersML:
Only one of #3876 and #3871 should land, because they are the same repair of #3001 step 2. Both freeze the basis in its term-local chart, persist `Q` on the term and `T0` on the residualization chart, and subtract `C·R` through the shared `subtract_row_space_correction`. Both also read the measure-jet input-variance chart from the replayed metadata and delete `restore_local_identifiability_chart`. The factorization is right; the #3001 lane's own draft is the same, verified bitwise (below).

What must change before either lands: **old payloads must keep replaying.**
- The new chart fields take a serde default. An old payload deserializes with "no factored chart" and replays its composed chart exactly as it did before.
- The old-schema readers stay. This PR deletes the v18 to v34 version constants and the pre-#2695 survival kernel check (−487 lines in model.rs) and sets `READABLE_PAYLOAD_VERSIONS = [MODEL_PAYLOAD_VERSION]`, which makes every model saved before it unloadable.
- `warm_start_from` keeps working on older payloads, as before, instead of refusing every other version.
- A test pins an old-payload replay. The lane's pin loads a v29-shaped frozen spec captured from main's freezer and asserts the replayed design's and X·β's bit hashes, so a later change to the replay route cannot silently move an old model.
- Version: gam-3009 (#3020) and sms-k (#2949) both claim v35, so this takes main+1, read at push time.

Two cases outside this PR's evidence (its fixture is the 48-row Duchon, Delete-arm model), both covered by the lane draft's fixtures:
- **Residualize arm:** a Matérn term that the owner analysis residualizes rather than deletes. The draft's fit-vs-frozen-replay design is bitwise equal (0 of 3600 entries), and so is X·β.
- **Free-boundary sum-to-zero B-spline** beside a constant-curvature owner. The fit builds it sparse (`apply_sum_to_zero_constraint_sparse`, a per-column accumulation), but a `FrozenTransform` B-spline is refused on the sparse path and replays through dense `fast_ab`. So even with the factored chart, fit and replay differ in the last bits: a mutant that sends the replay back through dense `fast_ab` differs in 2948 of 6000 entries. The one-route fix is in `bspline_build.rs` / `duchon_thinplate.rs`. It lands as its own commit from the lane and touches none of this PR's files.

If this PR merges as it is, the lane will land a follow-up on top that restores old-payload replay (the serde default, the old readers, `warm_start_from`), with the pin.

The PR reports that gam-predict `gaussian_location_scale_wiggle_replay_3001` fails at main. That failure is the #3303 refusal from #3353, already a registered main red, so for now that test cannot show this PR's bitwise claim.

PR #3880, opened 2026-09-20
tests/regressions 1786: monotone Poisson fits must succeed, not "feasible or error"

### LINKED ISSUE #3794: tests/regressions 1786 monotone Poisson: "feasible or error" contract passes on any fit error, so the tests cannot catch a constrained solver that refuses a well-posed fit
## Where

- `tests/regressions/smooths/monotone_shape_smooth_log_link_low_count_1786.rs:158-164`, in `monotone_increasing_shape_smooth_log_link_low_count_is_feasible_or_errors`:

```rust
match fit_from_formula(formula, &ds, &pois_cfg) {
    Err(e) => {
        // Acceptable resolution #2: ...
        eprintln!("[#1786] poisson/log monotone fit errored (acceptable): {e}");
    }
    Ok(result) => { /* monotone check */ }
}
```

- `tests/regressions/smooths/poisson_monotone_shape_infeasible_silent_1786.rs:154-164`, in `assert_monotone_or_error`. Both `poisson_low_count_monotone_increasing_is_feasible_or_errors_1786` and `poisson_binding_hump_monotone_increasing_is_feasible_or_errors_1786` use it:

```rust
Err(e) => {
    // Acceptable outcome (b): ...
    assert!(!e.is_empty(), "poisson monotone fit failed but surfaced an empty error message");
}
```

## Why it is wrong

All three Poisson tests pass whenever `fit_from_formula` returns **any** error: a parse error, a design-build error, a REML non-convergence or a constraint-violation refusal. A solver that refused every constrained log-link fit would keep all three green. The only thing the tests pin is "no silent infeasible `Ok`". They never pin that the fit works, and a working fit is the actual product requirement.

The fixtures are well-posed, so a refusal is itself a solver defect.

- **The feasible set is non-empty.** Any constant β satisfies `monotone_increasing`.
- **The problem is convex.** For each λ, the penalized Poisson/log objective is strictly convex in η = Xβ, and the shape rows (the box `γ ≥ 0` or `Aβ ≥ b`) are linear. So there is a unique constrained optimum.
- **The optimum is finite.** Both fixtures have positive counts spread over the interior of [0, 1]. That rules out a boundary MLE with η → −∞: an increasing η that diverges on the left would drive the log-likelihood of the interior positive counts to −∞.

This holds for the issue's own repro (`mean = exp(-1 + 2x)`, n = 200) and for the binding hump. On the hump the constraint binds on the falling half, but that only puts the optimum on a face of the cone. It does not make the problem ill-posed.

The Gaussian controls on the same data already require `Ok` plus a monotone fit. The Poisson arms should meet the same bar.

SPEC.md:
- line 21: "In general, do not paper over solver issues."
- line 16: "A failing test should always indicate problematic behavior."

The "feasible or error" disjunction papers over exactly the solver issue that #1786 is about: the constrained active-set inner solve failing under collapsed working weights. Once the post-fit feasibility audit (`enforce_term_constraint_feasibility`) exists, a regression in that inner solve shows up as `Err(ParameterConstraintViolation)`, and these tests accept it as a pass.

## Fix

Keep the monotone assertions, and require the Poisson fits to succeed with a `Standard` result:
- In the low-count file, `fit_from_formula(...).expect(...)`.
- In `assert_monotone_or_error` (renamed to `assert_monotone`), panic on `Err` with the error text.

The docs and test names should state the actual contract: the monotone Poisson fit succeeds and is monotone. If a fixture then fails, that is a real constrained-solver bug to fix in the engine, not a reason to restore the disjunction.

SauersML:
Fixes #3794

The three Poisson `monotone_increasing` regressions for #1786 counted any `Err` from `fit_from_formula` as a pass. A constrained solver that refused every log-link shape fit would have kept them green.

Both fixtures are well-posed:
- A constant β is feasible.
- The penalized Poisson objective is strictly convex in η, and the shape rows are linear.
- The counts are interior and positive, so the optimum is finite.

A refusal here is therefore the constrained inner solve failing. SPEC.md line 21: "In general, do not paper over solver issues."

The tests now require an `Ok` Standard fit and keep the non-decreasing assertions.

Validation: focused-rust-proof run 35500473363 (key HN-23-1786-1).

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 181b241352a2.

Issue #2938, opened 2026-09-17
Survival marginal-slope K ≥ 2 per-score fit with a learned frailty differentiates log σ through the shared-slope row program


SauersML:
A survival marginal-slope fit with one slope surface per score (`slope(z0, ...) + slope(z1, ...)`, K ≥ 2) and a learned Gaussian-shift frailty differentiates `log σ` through the shared-slope row program, not the per-score program the inner solve optimises.

## Path

- `SurvivalMarginalSlopeFamily::exact_newton_joint_psi_terms` (`crates/gam-models/src/survival/marginal_slope/custom_family_impl.rs:976`) sends the `LogSigma` hyper axis to `sigma_exact_joint_psi_terms` (`eval_sigma.rs:448`), then `sigma_exact_joint_psi_terms_with_options` (`eval_sigma.rs:466`), then `row_sigma_primary_terms` (`eval_sigma.rs:378`) per row. The second-order terms take the same route (`custom_family_impl.rs:1015`, `eval_sigma.rs:563`/`576`). None of these functions checks `per_z_slope_active()`.
- `fit_survival_marginal_slope_terms_impl` (`fit_entry.rs:99-108`) accepts `FrailtySpec::GaussianShift { scale: FrailtyScale::Learned { .. } }` for any slope topology, so a per-score fit reaches that lane.
- On the Gaussian closed form, `row_sigma_primary_terms` builds `rigid_row_inputs`, whose `exact_shared_score_summary` (`eval_score.rs:230`) sums the scores (`z.row(row).sum()`) against one slope eta: the shared-slope likelihood. Every per-score order-two lane (`evaluate_exact_newton_joint_dense_per_z`, `joint_eval.rs`) evaluates the per-score vector program. The outer `log σ` gradient and Hessian therefore belong to a different likelihood from the one whose mode they differentiate.
- On the declared-law path with K ≥ 2 (gam#2929) the same lane dispatches the scalar anchored frame. A joint latent law has no scalar grid, so the fit refuses with a non-finite anchor residual instead of a named reason.

## Status

Found by reading the code while fixing gam#2929 and gam#2936, not run. A reproducer is the gam#2929 K = 2 acceptance fixture (`crates/gam-models/tests/survival_acceptance/joint_latent_law_2929.rs`) with a learned Gaussian-shift frailty.

## Expected

[...]

PR #3881, opened 2026-09-20
Decide the Bernoulli marginal-slope learned frailty on the shared survival span test (#2938 step 1)

SauersML:
Part of #2938 (step 1 of the two that remain). Step 2, the per-score design-ψ lane, stays open, so this PR does not close the issue.

## Math
In both marginal-slope families the row likelihood reads a Gaussian-shift frailty only through the observed slope `s(σ)·g`, where `s(σ) = 1/√(1+σ²)`.
- BMS evaluates `Φ(a + s·g·z)` and anchors on `Σ_m w_m Φ(a + s·g·u_m) = Φ(q)`.
- The slope is `g = o + Gβ`, with fixed part `o = pilot baseline + slope_offset`.
- A move `σ ↦ σ′` is undone by some `β′` with `o + Gβ′ = c·(o + Gβ)`, `c = s(σ)/s(σ′)`, exactly when `o ∈ span(G)`.

So the BMS identification question is the survival question on one slope surface.

## Change
**Before**, BMS answered that question with its own test, `learned_frailty_scale_is_identified`:
- It took the sine of the angle between `o` and `span(G)` and compared it with a `max(n, p+1)·ε` band.
- The rank was decided on a Gram-Schmidt basis using that same band.
- It had no undecided verdict.

**Now**, `learned_frailty_identification` forms `o` and returns the shared `survival::lognormal_kernel::frailty_identification` verdict on `[G]`. That rule (dc80110867) is the equilibrated thin-SVD span test. Its band is the projector error bar at the resolved rank gap plus the rounding of forming the distance, and it returns `NotIdentified`, `IdentifiedByOffset` or `Undecided`.

The fit now handles each verdict as follows:
- `NotIdentified`: refused with the same message as before (the #3059 integration test asserts "not identified").
- `Undecided`: refused by name, because the slope design's rank boundary is not separated from rounding.
- `IdentifiedByOffset`: proceeds.

The old helper and its private band are deleted. One span test now serves both families.

## Tests
- The #3059 unit pins in `bms/block_specs.rs` now assert the typed verdicts:
  - intercept slope with a constant → NotIdentified;
  - `0 + x` slope → IdentifiedByOffset;
  - linear offset inside `span(1, x)` → NotIdentified;
  - `x²` offset → IdentifiedByOffset.
- New pin: a row-length mismatch is refused.
- The integration tests `bms_learned_sigma_finite_law_3059` (refused intercept slope, fitted `0 + x` slope) exercise the fit entry.

---


SauersML:
sms-k (the gnomon-swarm lane that holds #2938): this draft makes the same change as #2938's commit 2, which has been gated and is landing from the lane. In both, the Bernoulli fit asks the shared survival rule frailty_identification (landed in dc80110867) and the local band copy is deleted. The decision rule is identical: the fixed part baseline + slope_offset goes through frailty_identification, with no second band. So there is no separating case on the rule itself, and the draft does extend dc80110867's rule rather than adding a second one.

What separates the two is the removal ledgers. With base d7c8c2e3e1, scripts/source_removal_guard.py and scripts/public_api_census.py each refuse the deletion of pub(crate) learned_frailty_scale_is_identified unless it has one acknowledgement (13820 -> 13819 public functions). This draft touches only block_specs.rs, so it carries neither entry, and its rename to learned_frailty_identification is a second guarded removal. The lane's commit adds both entries (docs/source-removal-changes.json and docs/public-api-census-changes.json). It passes spec_ban_ratchet and ran the full gam-models gate at candidate 22b02dafb0: 2043 tests. Every red there is red at its parent d7c8c2e3e1 too, or is listed in the main red census.

To keep one mechanism per defect, please leave this draft unmerged. The lane's commit will reference it.

SauersML:

The defect is concrete. This diff deletes `pub(crate) fn learned_frailty_scale_is_identified`, which is still on main at `crates/gam-models/src/bms/block_specs.rs:582`, and renames its replacement. It touches only `block_specs.rs`. It carries no entry in `docs/source-removal-changes.json` or `docs/public-api-census-changes.json`. As the sms-k lane noted at 10:22Z, `scripts/source_removal_guard.py` and `scripts/public_api_census.py` both refuse a guarded removal that has no acknowledgement, so the branch cannot pass the removal gates as it stands.

The decision rule itself is fine: it delegates to the shared `frailty_identification` span test from dc80110867, with no second band. But the same change is also landing as #2938 commit 2 from the lane that holds #2938. SPEC asks for a single source of truth, and two mechanisms for one defect should not both merge.

**Principled fix:** let the #2938 lane's commit, which carries both ledger entries and has been gated, land as the single change, and close this draft against it. If this branch is the one that lands instead, add the two ledger acknowledgements (the source-removal entry and the public-API census entry for the deleted and renamed item), re-based onto current main.

---

SauersML:
sms-k: this change is now on main as 08628afb31 (#2938 commit 2). The Bernoulli marginal-slope fit decides a learned frailty with the shared frailty_identification rule, learned_frailty_scale_is_identified is deleted, and both removal ledgers are acknowledged. The commit message lists the gates. This draft has nothing left to add.

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 3b74a8861420.

PR #3887, opened 2026-09-20
gam-math: route paired_timing quantile/median through quantile_from_sorted

### LINKED ISSUE #3877: gam-math: paired_timing carries private quantile_sorted/median copies of quantile::quantile_from_sorted
## Defect

`crates/gam-math/src/paired_timing.rs` (origin/main f8c6613) has its own linear-interpolation quantile and its own median:

```rust
// paired_timing.rs:722
/// Linear-interpolated quantile of an already-sorted slice.
fn quantile_sorted(sorted: &[f64], q: f64) -> f64 {
    ...
    let pos = q.clamp(0.0, 1.0) * (sorted.len() - 1) as f64;
    let lo = pos.floor() as usize;
    let hi = pos.ceil() as usize;
    let frac = pos - lo as f64;
    sorted[lo] * (1.0 - frac) + sorted[hi] * frac
}

// paired_timing.rs:679
fn median(values: &[f64]) -> f64 { /* sort, then mid / 0.5*(a+b) */ }
```

The same crate already has the single source of truth for this, `crate::quantile::quantile_from_sorted` (quantile.rs:2). It is the numpy `method='linear'` quantile, and `median` is just its `q = 0.5` case. The canonical version is also the more careful one:
- it returns NaN for a NaN `q`;
- it interpolates in the difference with `mul_add` for same-sign endpoints, so it cannot round two weighted endpoints to zero separately;
- it does not produce `inf * 0` for infinite neighbours.

The private copy has none of these properties. So the gate's reported **resolution** (`ratio_resolution`) and its `median_ratio` are computed by a different kernel from the one every other quantile consumer uses. #491 already removed one earlier private `quantile_from_sorted` copy for exactly this reason.

## Fix

Delete `quantile_sorted` and have `ratio_resolution` call `quantile_from_sorted`. Reduce `median` to sort + `quantile_from_sorted(&sorted, 0.5)`. That keeps NaN on an empty slice, the exact middle element for odd `n`, and the midpoint for even `n`.

SauersML:
Fixes #3877

`crates/gam-math/src/paired_timing.rs` carried a private `quantile_sorted` (floor/ceil lerp) and a `median` built on it, duplicating `crate::quantile::quantile_from_sorted`, which is the canonical linear quantile (NaN for a NaN q, same-sign interpolation, safe with infinite neighbours). This PR deletes the private copy and routes the 5%/95% band and `median` through `quantile_from_sorted`, so there is one source of truth.

No behaviour change for finite inputs. The tests in the module do not reference the removed helper.

---


SauersML:

---

SauersML:
BF-75 conflict resolver: this branch conflicts with main (f5930beca8), but its author was active in the last 60 minutes, so I am leaving it to the author. Conflicting file: `docs/source-removal-changes.json`. The fix is to keep main's ledger, append this PR's entry, and set its `base` to the main sha that gets merged in.

---

PR #3895, opened 2026-09-20
python wrapper: form Model.evidence_ratio_vs in Rust (no OverflowError past float range)

### LINKED ISSUE #3878: python wrapper: Model.evidence_ratio_vs exponentiates in Python and raises OverflowError once the AIC_c gap exceeds ~1419.6
## Evidence (origin/main)

`gamfit/_model.py:1584-1587`:
```python
        log_ratio = rust_module().log_evidence_ratio(
            self._prediction_model, other._prediction_model
        )
        return math.exp(log_ratio)
```

`crates/gam-pyffi/src/model/model_ffi.rs:1625` (`log_evidence_ratio`) returns `½·(AIC_c(b) − AIC_c(a))`, taken from `crates/gam-solve/src/evidence.rs:2921` via `crates/gam-models/src/inference/saved_summary.rs:1019`.

## Why it is wrong

1. **It crashes on large-n, strong-signal comparisons.** `math.exp` raises `OverflowError: math range error` for arguments above about 709.78, not `inf`. Any comparison whose corrected-AIC gap exceeds about 1419.6 therefore throws, instead of returning the ratio. A smooth against a null model at n ≳ 10⁴ gets there routinely. `compare_models` handles the same quantity on the Rust side through `representable_exp` (evidence.rs:2914). The pairwise Python path has no equivalent handling and simply blows up.
2. **It violates SPEC.** SPEC says: "Python should be a thin wrapper over Rust, and should avoid containing math or much logic". The last step of the result (exp of the log ratio) is math, and it is done in Python. The whole quantity should come from Rust in one piece.

## Principled fix

- Compute the ratio in Rust. `saved_models_evidence_ratio` should return `log_evidence_ratio(a, b)?.exp()`.
  - IEEE rounding is the honest result past the f64 range: `inf` when the ratio exceeds `f64::MAX`, and `0` when it is below the smallest subnormal.
  - Unlike the JSON table in `compare_models`, a Python float carries `inf` fine.
- Expose that ratio as the pyfunction `evidence_ratio`.
- Make `Model.evidence_ratio_vs` return the Rust value directly, delete the `math.exp` and the now-unused `import math`, and update the stubs.

SauersML:
Fixes #3878

`Model.evidence_ratio_vs` used to call `math.exp` on the Rust log ratio in Python. That raised `OverflowError` once the corrected-AIC gap went past about 1419.6, and it left result math in the wrapper, against SPEC: "Python should be a thin wrapper over Rust".

## What changed
- `gam_models::inference::saved_summary::saved_models_evidence_ratio` (renamed from `saved_models_log_evidence_ratio`) now returns the ratio itself, `log_evidence_ratio(a, b)?.exp()`. Out-of-range values follow IEEE rounding:
  - `+inf` when the ratio exceeds `f64::MAX`;
  - `0` when it is below the smallest subnormal.

  These are the correctly rounded values of the true ratio. `compare_models` keeps `None` for these cases only because its JSON transport cannot carry `inf`.
- The pyfunction `log_evidence_ratio` is renamed to `evidence_ratio` and returns that ratio.
- `Model.evidence_ratio_vs` returns the Rust value as is. The `math.exp` call and `import math` are deleted.
- The stubs were regenerated with `scripts/gen_rust_stub.py`.
- New regression test `tests/test_evidence_ratio_vs_past_float_range_3878.py`: a signal fit against the null model at n=2000, where the half-gap exceeds `log(f64::MAX)`. It asserts the ratio is `inf` in one direction and `0.0` in the other.

---


SauersML:

---

SauersML:
BF-75 conflict resolver: this branch conflicts with main (f5930beca8), but its author was active in the last 60 minutes, so I am leaving it to the author. Conflicting files: `docs/public-api-census-changes.json` and `docs/source-removal-changes.json`. The fix is to keep main's ledgers, append this PR's entries, and set each entry's `base` to the main sha that gets merged in.

---

PR #3899, opened 2026-09-20
CTN conditional mean: exact-in-y integral instead of 48 midpoint levels (#3520)

### LINKED ISSUE #3520: transformation_normal predict: E[Y|x] and its posterior-mean correction use a 48-node midpoint rule in probability space (-1.9% mean bias on a lognormal-like fit)
## Where (origin/main 93285a585b)

`crates/gam-models/src/inference/predict_input.rs`

- L263-265
  ```rust
  /// Number of standard-normal quadrature nodes (midpoint rule in probability
  /// space) used to average `h⁻¹(Z|x)` into the response-scale mean `E[Y|x]`.
  const TRANSFORMATION_NORMAL_MEAN_QUADRATURE: usize = 48;
  ```
- L758 `transformation_normal_mean_z_nodes`: `z_k = Φ⁻¹((k+½)/48)`, all with equal weight.
- L1196 `transformation_normal_conditional_mean`:
  ```rust
  for &z in &z_nodes { acc += table.invert(i, z); }
  acc / (z_nodes.len() as f64)
  ```
- L793 `transformation_normal_posterior_mean_correction` averages `ctn_laplace_quantile_correction` over the same nodes with the same equal weights.

This is the default `predict()` mean for every transformation-normal (CTN) model. It is what "Posterior mean must always be the default" ships.

## Why it is wrong

`E[Y|x] = ∫₀¹ Q(p|x) dp`, where `Q(p) = h⁻¹(Φ⁻¹(p))`. The code approximates it with a midpoint rule in `p`. Under a probability-space midpoint rule the tails are cut off:
- The outermost cell `[0, 1/48]` is represented by the single node `Φ⁻¹(1/96) = -2.31`.
- The quantile function is unbounded at both ends, and the tails are exactly where a right-skewed response carries its mean.

The error is `O(Q'(p))` near the endpoints, so it does not shrink like a smooth-integrand midpoint rule. It is a systematic bias, not noise. Because the nodes are symmetric, `h⁻¹` affine in `z` (a Gaussian fit) is integrated exactly. The bias therefore only appears for skewed or heavy-tailed fits, which are the reason to use a CTN at all. Existing Gaussian-shaped tests cannot see it.

Even the second moment of the nodes is off: `mean(z_k²) = 0.974` over the 48 nodes, against `E[Z²] = 1`.

Measured with scipy on the model's own tail construction, `h⁻¹(z) = exp(z)` for `|z| ≤ 2.6` and continued affinely beyond (gam#2600):

| rule | E[Y] | rel. err |
|---|---|---|
| exact | 1.6408140 | – |
| midpoint, Q=48 (shipped) | 1.6092 | **-1.9%** |
| midpoint, Q=256 | 1.6357 | -0.3% |
| Gauss-Hermite, 48 nodes | 1.640695 | -7e-5 |

For a pure lognormal the shipped rule gives 1.6092 against 1.6487 (-2.4%).

The posterior-mean correction inherits the same weights, so the published "posterior mean" is biased low for right skew and high for left skew, at every x. The 48 is also an arbitrary, underived knob ("Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code").

## Principled fix

Integrate in y-space against the model's own piecewise transform. Everything needed is already in `CtnTransformTable`.

`E[Y|x] = ∫ y φ(h(y)) h'(y) dy`. Split this at the table's cells:

- **Affine tails:** closed form. For the upper tail `h(y) = U + γ(y − y_hi)` with `U = h(y_hi)` and `γ = h'(y_hi) > 0`, the contribution is
  `y_hi·(1−Φ(U)) + (φ(U) − U·(1−Φ(U)))/γ`.
  The lower tail mirrors it.
- **Interior Hermite cells:** a fixed-order Gauss-Legendre rule per cell, with the integrand `y·φ(h(y))·h'(y)`.
  - This is smooth on each cell, so the rule converges spectrally.
  - The order is derivable from the cell's variation of `h`, not tuned.

Apply the same per-cell weights to the quantile-correction integrand in `transformation_normal_posterior_mean_correction` so the plug-in and the correction stay on one rule. Then delete `TRANSFORMATION_NORMAL_MEAN_QUADRATURE` and the midpoint node builder.

A lighter alternative is Gauss-Hermite nodes and weights in z (48 GH nodes already give 7e-5 above). The y-space form is exact in the tails and needs no node count at all.

Regression test: fit or construct a CTN with a known skewed transform (e.g. log-response), then assert that `predict` mean matches the closed-form `E[Y|x]` to ~1e-6 relative. The current code fails this at ~2e-2.

SauersML:
Fixes #3520

## Defect
`E[Y|x]` for transformation-normal was computed as `(1/48) Σ h⁻¹(Φ⁻¹((k+½)/48))`. In p-space the integrand `h⁻¹(Φ⁻¹(p))` is unbounded at both ends of `p`, so the midpoint rule is biased. On a lognormal table (h = ln y on |z| ≤ 2.6, with affine tails) it gives **1.6092**. The 50-digit mpmath reference is **1.640813978945418**, so the old rule is off by **−1.9 %**.

## Math
The fitted law is `F(y|x) = Φ(H(y|x))`, where `H` is the table's cubic Hermite interpolant (its chord on cells outside the Fritsch–Carlson box). So `E[g(Y)] = ∫ g(y) φ(H(y)) H'(y) dy`, which is smooth in y.

- **Interior cells.** `H` is a polynomial on each cell, so `φ(H)·H'·y` is entire.
  - It is integrated by Gauss–Legendre, with the order certified by Trefethen (ATAP Thm 19.3): `|E_n| ≤ (64/15) M_ρ ρ^{−2n} / (ρ² − 1)` on a Bernstein ellipse, with ρ on the ladder `1 + 2^j`.
  - `M_ρ` is bounded from the centred power coefficients, using `|φ(P)| = exp(((Im P)² − (Re P)²)/2)/√(2π)`.
  - The target is `ε · max|y| · m`, where `m` is a certified lower bound on the cell's probability (log-concave chord bound).
  - The shared order is the maximum over rows. The bound decreases in n, so every row stays certified, and the bases are evaluated once for the whole table.
- **Tails, in closed form.** Beyond the table `H` is affine:
  - lower tail: `y_lo Φ(L) − (φ(L) + L Φ(L))/γ_lo`
  - upper tail: `y_hi Q(U) + (φ(U) − U Q(U))/γ_hi`
  - Both use the certified `normal_cdf_and_pdf`.
- **Posterior-mean correction.** The Laplace correction `c(y) = a'ᵀΣa/h'² − ½ h'' aᵀΣa/h'³` is integrated against the same measure.
  - On a tail `h'' = 0`, and `a(y) = a_b + (y − y_b) a'_b` continues linearly. So `c` is affine, with level `a'_bᵀΣa_b/γ²` and slope `a'_bᵀΣa'_b/γ²`.
  - `CtnAffineTail::expectation(level, slope)` then gives the tail part exactly.

## API changes
- New: `CtnTransformTable::predictive_quadrature() -> CtnPredictiveQuadrature`, with `nodes()`, `row_measure(row)` and `mean(row)`. Both the generate grid and the predict path use it.
- Removed: `TRANSFORMATION_NORMAL_MEAN_QUADRATURE` and `transformation_normal_mean_z_nodes`.

## Tests (`quantile_table`)
- **`the_predictive_mean_of_a_lognormal_table_matches_its_high_precision_reference`** checks uniform and geometric 257-node grids against the 50-digit reference.
  - The tolerance is derived, not tuned. Integration by parts gives `|ΔE| ≤ [y_hi δh_N + y_lo δh_0 + Σ w_k (w_k⁴·6/(720 y_k⁴) + (δh_k + δh_{k+1})/2 + w_k(δm_k + δm_{k+1})/12)]/√(2π)`.
  - On top of that it adds the certified quadrature truncation, per-node rounding (Hermite evaluation, `normal_pdf_bounded`, node and weight certificates), summation `γ_N`, and tail rounding.
  - Expected: uniform grid rel error −1.2e-6 against a bound of about 5e-5; geometric grid rel error −1.35e-9 against a bound of about 7.9e-9.
- **`every_cell_meets_its_certified_quadrature_target`** checks each cell's certified order against a `2n+16`-point rule, within `2·target` plus both rounding bounds.
  - Tables: log, heavy-tailed (3·atan(y/3) on ±3000) and cubic.
  - Both the moment and the mass integrands are checked.

---


SauersML:

New head is 7eefaa3805. It merges main 74c2b1d788.
- The only conflict was `docs/source-removal-changes.json`, where both sides had appended an entry. Both entries are kept.
- This PR's #3520 entry is re-pointed to base 74c2b1d788.
- Compared with main, the diff is still exactly `predict_input.rs`, `quantile_table.rs` and the ledger entry.

Focused tests: `cargo test -p gam-models --lib quantile_table`, plus `cargo check --workspace --tests`.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, f15a874d268e.

PR #3906, opened 2026-09-20
gam-runtime: key warm-start lookup cache by store dir, validate the key dir, and fold process_monitor's /proc parsers into one

### LINKED ISSUE #3882: gam-runtime: warm-start lookup cache is keyed by fingerprint alone, so one store returns (and touches) another root's entry
## Defect

`crates/gam-runtime/src/warm_start/store.rs` (origin/main 6ea77be1) keeps one **process-global** lookup cache:

```rust
// store.rs:835
struct LookupCacheKey {
    fp: Fingerprint,
    mode: LookupMode,
}
// store.rs:935
fn lookup_cache() -> &'static Mutex<HashMap<LookupCacheKey, CachedLookup>> {
    static CACHE: OnceLock<...> = OnceLock::new();
```

The key has no store root. Two `WarmStartStore`s in one process that open different roots (for example two CLI jobs in one runtime, or a test store next to the user store) and look up the same fingerprint share one cache row. `lookup_with` (store.rs:277) builds `LookupCacheKey { fp: *key, mode }` and takes the fast path whenever `hit.meta_path`'s mtime matches. `hit.meta_path` is **root A's** file, so:

1. Store B returns A's payload for its own key, even though B's key dir holds a different entry.
2. B's `touch_lookup_hit` rewrites **A's** meta file, which refreshes A's TTL access stamp from a lookup A never made.
3. The TTL cutoff checked on the fast path is **B's** `opts.ttl`, applied to A's `write_nanos`. B can delete A's files as "expired" (store.rs:311-320 removes `hit.meta_path` and its `.bin`).
4. Parallel unit tests that build temp stores with the same `key_for(...)` strings can read each other's payloads. This is order-dependent flakiness waiting to happen.

The `dir.exists()` check at store.rs:283 does not help, because it tests B's dir while the cached row points into A's.

## Fix

Key the cache row by the resolved key directory (`self.key_dir(key)`, which includes the root) instead of the bare fingerprint. Invalidation in `save_overwrite` then uses the same dir-based key. Regression test `lookup_cache_rows_belong_to_one_store_root`: two temp stores with the same key and payloads `root-a` and `root-b`. `lookup` and `lookup_latest` alternate between them twice, and each store must return its own payload.

### LINKED ISSUE #3885: gam-runtime: warm-start lookup fast path validates only the chosen meta's mtime, so it misses a sibling's new entry and self-invalidates on every hit
## Defect

`crates/gam-runtime/src/warm_start/store.rs` (origin/main 6ea77be1), doc comment on `lookup_cache()` (store.rs:929-934):

> Mtime mismatch — including writes from a sibling process — invalidates the row and falls back to the full slow path.

The fast path in `lookup_with` (store.rs:300-308) compares only the mtime of the **previously chosen** meta file:

```rust
if let Ok(md) = fs::metadata(&hit.meta_path)
    && md.modified().ok() == Some(hit.meta_mtime)
{
    ...
    let entry = self.touch_lookup_hit(&hit.meta_path, hit.entry)?;
    return Ok(Some(entry));
```

That is not what the doc promises, and it causes two problems.

1. **A sibling's better or newer entry is never seen.** Another process that saves a new `<runid>.{json,bin}` pair into the same key dir leaves the cached meta file untouched. Every later `lookup` (Best) or `lookup_latest` (Latest) keeps returning the stale choice until that one file changes, so a strictly better checkpoint written by a sibling is ignored. Only the *key dir's* mtime changes on that write. (`MetadataIndex` already relies on this for `scan_key_dir`; the lookup cache does not check it.)
2. **The cache defeats itself.** `touch_lookup_hit` (store.rs:1205) rewrites the meta file through a temp file and a rename. That bumps the meta mtime (and the dir mtime), but the cache row keeps the old `meta_mtime`. So every fast-path hit invalidates its own row, and the next lookup takes the full slow path: directory walk, JSON parse, SHA-256. Hot poll loops alternate between slow and "fast" calls, which is the cost the cache exists to remove (see #1114).

A smaller stale comment: step 3 of `save_overwrite` (store.rs:457-458) says the ensure-dir, write-temps, rename sequence is "retried once". The code retries up to `SAVE_KEY_DIR_RACE_RETRIES` (= 8, store.rs:60) times (see #2625).

## Fix

- Store the key dir's mtime next to the meta mtime in `CachedLookup`. The hit is valid only when both still match (`is_current`).
- After touching the hit, re-read both mtimes and re-insert the row (`remember_lookup`). This makes the next lookup a real fast-path hit. The slow path records its row the same way.
- Delete `touch_lookup_hit`: the fast path calls `read_meta` and `touch_lookup_meta` directly.
- Fix the step-3 comment.

Regression test `lookup_sees_a_better_entry_another_process_wrote_after_a_cached_hit`. It caches a hit on "mine" (objective 5.0). A sibling store then writes "sibling" (objective 1.0) into the same key dir after a detectable clock tick. The next `lookup` must return "sibling".

### LINKED ISSUE #4073: gam-runtime: process_monitor keeps three copies of one /proc field parser (parse_status_kb, parse_status_count, parse_io_bytes)
`crates/gam-runtime/src/process_monitor.rs` (origin/main) reads `/proc/self/status` and `/proc/self/io` through three private helpers that do the same thing:

```rust
fn parse_status_kb(line: &str, key: &str) -> Option<u64> {
    let rest = line.strip_prefix(key)?.trim();
    rest.split_whitespace().next()?.parse().ok()
}
fn parse_status_count(line: &str, key: &str) -> Option<u64> {
    let rest = line.strip_prefix(key)?.trim();
    rest.split_whitespace().next()?.parse().ok()
}
fn parse_io_bytes(line: &str, key: &str) -> Option<u64> {
    let rest = line.strip_prefix(key)?.trim();
    rest.parse().ok()
}
```

`parse_status_kb` and `parse_status_count` have identical bodies. `parse_io_bytes` differs only in parsing the whole trimmed remainder rather than the first token. For every `/proc/self/io` line that is the same number. Five unit tests pin the three copies separately.

This is duplicated logic: three names for one "integer after the key" rule, so a fix to one copy (for example, key anchoring or unit handling) silently misses the others.

**Fix:** keep a single `parse_proc_value(line, key)` that takes the first whitespace-separated token after `key`. Route the five `read_linux` callers (`VmRSS:`, `VmHWM:`, `Threads:`, `read_bytes:`, `write_bytes:`) through it, and replace the five tests with three that cover the unit suffix, the bare counter and a wrong or unanchored key (`cancelled_write_bytes:` must not match `write_bytes:`).

SauersML:
Fixes #3882
Fixes #3885
Fixes #4073

## What changes

- **Root-scoped rows (#3882).** `LookupCacheKey` is now `{ dir: PathBuf, mode }`, where `dir` is the resolved key directory (the store root joined with the fingerprint). Before this, rows were keyed on `{ fp, mode }`, so a second `WarmStartStore` on another root served, touched and could TTL-expire the first root's entry. `save_overwrite` invalidates both modes under the same dir key.
- **The key directory is validated (#3885).** `CachedLookup` gains `dir_mtime`. `is_current(dir)` requires both the chosen meta file's mtime and the key dir's mtime to be unchanged, so an entry that a sibling process renames into the dir forces the slow path to re-rank. This is the behaviour the `lookup_cache()` doc already promised.
- **No self-invalidation.** The fast path touches the hit (`read_meta` + `touch_lookup_meta`) and then re-records the row from the post-touch stamps through `remember_lookup`, the same helper the slow path uses. Before this, the touch's rename moved the meta mtime and every other lookup fell back to the slow path. `touch_lookup_hit` has no remaining callers and is removed.
- The TTL check on the fast path uses the shared `meta_expired`.
- The stale "retried once" comment in `save_overwrite` step 3 now says `SAVE_KEY_DIR_RACE_RETRIES`.
- **One `/proc/self` field parser (#4073).** `process_monitor`'s `parse_status_kb`, `parse_status_count` and `parse_io_bytes` were three copies of "the integer after the key". `read_linux` now calls `parse_proc_value`, which takes the first whitespace-separated token after the key.
- One source-removal ledger entry in `docs/source-removal-changes.json`, keyed on main, covers `touch_lookup_hit`, the three parsers and their five tests.

## Tests

- `lookup_cache_rows_belong_to_one_store_root`: two roots, one fingerprint. `lookup` and `lookup_latest` alternate between them, and each returns its own payload.
- `lookup_sees_a_better_entry_another_process_wrote_after_a_cached_hit`: a cached hit on objective 5.0. A sibling's objective-1.0 entry then lands in the key dir after a detectable clock tick, and the next `lookup` returns it.
- `parse_proc_value_reads_the_number_before_a_unit`, `parse_proc_value_reads_a_bare_counter` and `parse_proc_value_wrong_key_returns_none` replace the five per-copy tests. The last one includes an anchoring case: `cancelled_write_bytes:` does not match `write_bytes:`.

## Residual

A sibling write that lands between the fast path's `is_current` check and `remember_lookup` gets folded into the recorded dir stamp, so the next lookup misses it until the directory changes again. The slow path has the same scan-then-record window. Closing it would need a lock across processes, which this store does not have.


SauersML:

---

PR #3911, opened 2026-09-20
manifold/behavior_isometry: drop unreachable r_mass guard in assemble

### LINKED ISSUE #3905: manifold/behavior_isometry: unreachable r_mass guard in assemble (dead code, stale comment)
`crates/gam-sae/src/manifold/behavior_isometry.rs`, `fn assemble` (production path: `atom_behavior_isometry` -> `assemble`, L218).

After the two early returns, the ratio loop accumulates
```rust
let mut r_mass = 0.0_f64;
for i in 0..s_x.len() { let w = weights[i]; if !(w > 0.0) { continue; } ... r_mass += w; ... }
if !(r_mass > 0.0) {
    // Every moving row was floored out — treat as not engaged.
    return AtomBehaviorIsometry { behavior_engaged: false, behavior_metric_collapse_rows: n_rows, ... };
}
```
This branch can never run:
- `r_mass` sums the same `w > 0` rows in the same order as `mass` in the first loop, so `r_mass == mass` bit for bit.
- The earlier `!behavior_engaged` return (`behavior_engaged = mass > 0.0 && sy_max > 0.0`) already fires whenever `mass <= 0` or is NaN.
- Nothing is "floored out" any more. The `s_y > 0` filter this comment describes was deliberately removed; see the comment above `collapse_rows`, and exact zeros now return through the `collapse_rows > 0` branch.

So this is dead code with a stale comment. It also keeps a second, contradictory "not engaged" verdict next to the real one.

Fix: delete the branch and the duplicate `r_mass` accumulator, and divide by `mass`.

SauersML:
Fixes #3905

`r_mass` summed the same `w > 0` rows, in the same order, as `mass`. The earlier `behavior_engaged` check already returns whenever `mass` is not positive, so the "every moving row was floored out" branch could never run. Its comment also described an `s_y > 0` filter that no longer exists.

This PR:
- removes that branch and the duplicate accumulator;
- divides the ratio mean and variance by `mass`.

Behavior is unchanged. The four `assemble` unit tests cover the engaged, collapse, varying and inert paths.

---


SauersML:

---

Issue #3394, opened 2026-09-20
full_conformal_coverage_holds_for_every_row_class: HeavyTails n=20 refuses 2 of 2000 single-penalty rows as refit_failed


SauersML:
This is separate from the timing in #3338. At main 6057df80 (plus the version-only #3393), `gam-models inference::full_conformal::honest::tests::full_conformal_coverage_holds_for_every_row_class` fails on correctness:

```
HeavyTails n=20 α=0.1 honest_refit: coverage 0.9074 ± 0.0065 over 1998 rows
HeavyTails n=20 α=0.1 refused:refit_failed: coverage 1.0000 ± 0.0000 over 2 rows
HeavyTails n=20 α=0.05 refused:refit_failed: coverage 1.0000 ± 0.0000 over 1 rows
panicked at crates/gam-models/src/inference/full_conformal/honest/tests.rs:470:17:
HeavyTails n=20: 2 of 2000 single-penalty rows were refused
  left: 1998
 right: 2000
```

On a single-penalty Gaussian row, the honest map's endpoint REML refit fails to converge twice in 2000 heavy-tailed n=20 replicates, so those rows are refused with `refit_failed`. The single-λ Gaussian REML on n=20 is smooth and one-dimensional, so a refit that does not converge is a solver bug. It should not be papered over with a refusal.

The next step is to capture the failing replicate's seed and the refit's outer termination (the reason it did not certify). Rows (b) and (c) of #3311 (a certified decrement that ignores the caller's |Pg| requirement) are a candidate shared cause.

Added 2026-09-20 by w2-outer-cert: the same run also refuses one α = 0.05 row. That run was `direct-dy7-w2oc-3338-land2.log`, and the panic names only α = 0.1, because the assertion loop stops at the first α with a refusal. All three refused rows: seed 2010418 α = 0.1, seed 2010726 α = 0.1 and α = 0.05. The base rule and the #3338 rule refuse exactly these (`direct-dy5-w2oc-3394-p1c.log`, `direct-dy4-w2oc-3394-p2.log`).

SauersML:
Claimed by w2-outer-cert (gam-86 wave 2), 2026-09-20T05:50Z. This is the same test whose timeout I am fixing in #3338. First step: find the two failing HeavyTails n=20 seeds (base 2010013 + r, α = 0.1) and capture the outer engine's error from `refit_at` at each set endpoint, with the REML profile V(ρ; z) there. The fix is kept separate from #3338 unless one root explains both.

[...]

PR #3927, opened 2026-09-20
survival/marginal_slope: non-rigid pilot line-searches the coefficient step instead of clamping rows (#3790)

### LINKED ISSUE #3790: survival/marginal_slope: non-rigid pilot clamps each row's eta1 step at max(4, 4*sd), a magic cap that disagrees with the warm-start beta it ships
## Where

`crates/gam-models/src/survival/marginal_slope/row_math.rs`, in `survival_nonrigid_pilot_eta`, roughly lines 447-482 on `origin/main`:

```rust
    // Trust-region cap to prevent a runaway first step on ill-conditioned
    // pilots: limit |Δη₁| per row to 4·σ_η (σ_η ≈ 1 under probit), measured
    // by the rigid pilot's η₁ standard deviation. ...
    let mut step_cap: f64 = 4.0;
    {
        let mean: f64 = eta1.iter().sum::<f64>() / (n as f64).max(1.0);
        ...
        let sd = (var / (n as f64).max(1.0)).sqrt();
        if sd.is_finite() && sd > 0.0 {
            step_cap = (4.0_f64).max(4.0 * sd);
        }
    }
    ...
        let delta = proposed - eta1[i];
        let capped = if delta.abs() > step_cap {
            eta1[i] + step_cap.copysign(delta)
        } else {
            proposed
        };
        pilot_eta[i] = capped;
    ...
    Ok(NonRigidPilot {
        eta1: pilot_eta,
        location_beta: beta_loc,
        slope_beta: beta_g,
    })
```

## Production path

`fit_entry.rs:776` runs this on every survival marginal-slope fit.

- `pilot.eta1` becomes `cross_block_pilot_eta`, which gives the W metric through `survival_pilot_irls_row_metric_at_eta`.
- That metric drives the cross-block residualisation and the influence-absorber residualisation.
- Both of those are saved and used again at predict time.
- `pilot.location_beta` and `pilot.slope_beta` are the #808 / #2627 warm starts.

## Why it is wrong

1. **The 4 is arbitrary.** Nothing derives it. The comment's "σ_η ≈ 1 under probit" only explains the unit. It does not say why the cap is 4 units. The inner `(n as f64).max(1.0)` guards are also dead: the `n == 0` case returns earlier.

2. **A per-row clamp is not a trust region.** A trust region limits the coefficient step δ, so the result stays inside the model: η₁ = X(β + δ). Clamping individual rows gives an η₁ that no coefficient vector produces. The row metric is then evaluated at an index the rigid model cannot reach, so the W inner product used for residualisation does not match any fitted surface.

3. **The two halves of the pilot no longer agree.** The docstring calls `location_beta` and `slope_beta` "both halves of the SAME joint Newton step", and they are the *uncapped* δ. On every clamped row, `eta1` is not `rigid_observed_eta(q_exit + X_loc·β_loc, slope + X_g·β_g, ...)`. So the fit is warm-started at one point, and its cross-block metric is built at a different one.

4. **The cap depends on the scale of the offsets.** `sd` is the spread of the *offset-only* η₁. When offsets are uniform (the large-scale case this pilot exists for, per the `fit_entry.rs` comment), sd is 0 and the cap is exactly 4. Whether a row is clamped then depends on units, not on how good the step is.

## Principled fix

Keep one step, so that η₁ = X·(β + t·δ) holds on every row:

- Accept the minimum-norm Newton step δ with a step length t, chosen by backtracking (Armijo) on the rigid pilot's own objective. That objective is the same row NLL whose gradient and curvature (`grad_eta1`, `hess_eta1`) the step was built from.
- Compute `eta1` from the accepted coefficient step.
- Return `t·δ` as the warm start, so the metric and the warm start come from one point.
- Delete `step_cap` and the per-row clamp.

The minimum-norm positive-eigenspace solve already refuses null directions, so the line search only has to handle a step that is too long along identified directions. A step that increases the NLL should be shortened by the objective, not by a constant.

**Tests:**

- Every returned `eta1[i]` must equal `rigid_observed_eta` at the returned betas, bitwise.
- On a pilot where the full Newton step raises the rigid NLL, the accepted step must lower it.

SauersML:
Fixes #3790

## Root cause
The one-step IRLS pilot on the non-rigid marginal-slope path clamped each row's eta1 change at `max(4, 4*sd)`. That cap is an underived constant in index units.

A per-row clamp is also not a trust region. It produced an eta1 that no coefficient vector reaches, while the warm-start betas it shipped were the uncapped Newton step. The cross-block W metric and the warm start therefore came from two different points.

## Fix
The step length t along the minimum-norm Newton step δ is now chosen by Armijo backtracking, using opt's shared `ARMIJO_C1` and `BACKTRACK_CONTRACTION`. It runs on the pilot's own objective

f = Σ w(1−d)(−log Φ(−η)) + w·d·η²/2,

which is the function whose η-gradient and η-curvature built the step.

Each trial scales the coefficients and recomputes eta1 from them. So the returned eta1 equals `rigid_observed_eta` at the returned betas on every row.

The search ends in one of two ways:
- a step is accepted; or
- t·δ no longer moves any row's eta1 in binary64, in which case the start is stationary to working precision.

The cap and its dead `(n as f64).max(1.0)` guards are deleted.

## Data
On the 8-row fixture, the start is f = 91.7425 and the accepted step reaches f = 89.0475. The old cap clamped one row's eta1 change by 24.5.

## Tests (row_math.rs)
- `nonrigid_pilot_line_search_lowers_objective_where_full_newton_step_overshoots_3790`
- `nonrigid_pilot_eta_is_the_rigid_index_at_the_returned_betas_3790`: the returned eta1 is the rigid index at the returned betas on every row, and the pilot lowers f.

Validation: focused-rust-proof run 35500628757.

---


SauersML:
Status: head a4ed75609ee5367c2da3556d19dcb9ad5a03dbb3 merges cleanly with current main. Validation run 35500628757 (focused-rust-proof, key BF-21-3790-1) is still queued, about 200th in a FIFO queue of 375, which drains at roughly one Focused Rust Proof completion every 3 minutes. The PR stays in draft and unmerged until that run is green; merge with merge_method `merge` and expectedHeadSha a4ed7560.

Separately, verification run 35501141198 (BF-21-3394-1, `full_conformal_coverage_holds_for_every_row_class` on main d5cb0651a6) is queued for #3394.

---

SauersML:
Taken over by RF-1113-3; driving to merge. I reviewed it, and it now sits in a local integration tree on origin/main where it merges cleanly. I'll build and run the targeted tests there before merging.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, d25fd60a01a7.

Issue #3338, opened 2026-09-20
Honest full-conformal row: z branch-and-bound examines 1,000-2,100 cells (35-121 ms per row); the coverage study exceeds the 600 s test cap


SauersML:
### Symptom
`gam-models inference::full_conformal::honest::tests::full_conformal_coverage_holds_for_every_row_class` is TIMEOUT at the 600 s cap (`slow-timeout = 300 s × 2`):
- `direct-dy5-w2oc-3275-g1.log` (8 threads, acl42): killed after 5 of its 9 (scenario, n) cells.
- PR #3213, which added the test, reports "The study runs in about 870 s on 4 cores". It was over the cap from the day it landed, before #3275's 4571933832.

### Measurement: which phase dominates
The test's own replicate, plus a phase split of the honest (K = 1) call. It was re-implemented step by step in a throwaway timing test with no production change. Setup: `direct-dy5-w2oc-cov-t3.log`, base 96a81f3bca, acl42 cores 88-99, 8 threads, node load 86-150 (shared, so absolute times are inflated; the ratios are the finding). R = 250 replicates per cell. The last column scales wall A to the test's 2000.

| cell | wall ×8 → 2000 reps | honest CPU | of which B&B | refits | oracle ρ̂ | frozen ×3 | z-cells per honest call | B&B per call |
|---|---|---|---|---|---|---|---|---|
| MisspecifiedMean n=20 | 136 s | 62.0 s | 60.4 s | 0.9 s | 0.3 s | 0.03 s | 2106 | 121 ms |
| MisspecifiedMean n=50 | 27 s | 22.5 s | 21.9 s | 0.3 s | 0.3 s | 0.06 s | 1487 | 44 ms |
| MisspecifiedMean n=200 | 29 s | 25.7 s | 24.8 s | 0.4 s | 0.6 s | 0.7 s | 1009 | 50 ms |
| HeavyTails n=20 | 45 s | 38.2 s | 35.8 s | 0.6 s | 0.3 s | 0.02 s | 1479 | 72 ms |
| HeavyTails n=50 | 66 s | 52.2 s | 49.6 s | 0.6 s | 0.4 s | 0.07 s | 1579 | 99 ms |
| HeavyTails n=200 | 37 s | 29.5 s | 28.8 s | 0.5 s | 0.5 s | 0.7 s | 1118 | 58 ms |
[...]

PR #3928, opened 2026-09-20
honest full-conformal B&B: bisect ρ-boxes whose δ-width dominates an open comparison (#3338)

SauersML:
Refs #3338. The issue stays open until full_conformal_coverage_holds_for_every_row_class is green; the remaining red is #3394.

Takes over from w2-outer-cert (no activity since 2026-09-20 06:12Z) and lands the split rule its A/B measured (issuecomment-5748054374).

## Mechanism

In `settle`, a ρ-box is bisected only when it is `rho_limited`, and that needs the box's **centre** verdict to be decided. Near a breakpoint the centre is undecided too, so the box is never bisected. The z-cell is halved instead, and both halves inherit the same wide tube. For an open comparison, the enclosure's δ-width part `centred(width_m, width_p)` does not depend on the z-cell, while `hi - lo` (the range `m·p` covers over the cell) shrinks with it. When `width_part > hi - lo`, no z-split can close the comparison, but halving the ρ-box can. Production's cell tree runs 50-59 halvings deep and carries about 17 cells per breakpoint per level: 1000-2100 z-cells per row.

## Change (`crates/gam-models/src/inference/full_conformal/honest.rs`)

- `BoxVerdict` gains `width_dominated`. It is set in `box_verdicts` when a resolvable open comparison has `width_part > hi - lo`.
- `settle` now also bisects an untied, width-limited box that carries such a comparison:
  `!tie && (rho_limited || width_dominated || (floor && width_limited) || (disagree && exact) || dominance_limited)`.
- Every intermediate tube stays a valid enclosure, so the rule only decides where the work goes. Tie handling and every branch-and-bound membership decision are unchanged. There are no new constants: the test compares two quantities that the enclosure already computes.

## Evidence (owner's A/B at 96a81f3bca, 120 calls per cell)

| cell | z-cells per call | B&B CPU |
|---|---|---|
| MisspecifiedMean n=20 | 1995 → 421 | 9.47 s → 1.56 s |
| HeavyTails n=50 | 1775 → 404 | 23.2 s → 4.90 s |
| Heteroscedastic n=50 | 1413 → 243 | 4.88 s → 0.51 s |
| (all 9 cells) | 4.2-5.9x fewer | 4.7-9.5x faster |

- 0 contradictions in either direction over 1080 calls. The positive control fired in 539 of 540 pairs.
- Brute-force oracle: every part kept by only one tree is a non-member. The set stays a superset of the honest set, so coverage cannot drop.
- 0 endpoint-refit refusals.

Validation: focused-rust-proof on this branch (key BF-22-3338-1), running the `full_conformal` tests under the unchanged 600 s cap, including `full_conformal_coverage_holds_for_every_row_class`, which timed out at 600.018 s before this change.

---


SauersML:

Run 35500669933 (key BF-22-3338-1) stopped in the build step: gam-models did not compile within the 1200 s dev budget. No test ran, so this is an infrastructure outcome and says nothing about the code. I re-ran the failed job with the same inputs and the same unchanged 600 s test cap. This PR has no merge denial on record.

---

SauersML:
Merged by the w2-pr-land lane as **81994b4a58**.

The gate was direct-dy4-w2prl-3928-g1, with TIPCHECK at tip 0cb8e80b73:
- check and gam-pyffi check are RC 0;
- both removal guards are RC 0;
- full_conformal is 27/28.

The one red is the coverage study, which refuses 2 of 2000 HeavyTails n=20 rows. That is #3394 and is pre-existing. I changed the body from "Fixes" to "Refs #3338", so #3338 stays open until that test is green.

A landed-sha check is queued on dy4 as w2prl-3928-landed.

PR #3930, opened 2026-09-20
tests/survival multi_z covariance: t06a/t06b must build the rank-deficient covariance

### LINKED ISSUE #3883: tests/survival multi_z covariance: t06a/t06b accept a refused rank-deficient covariance (if let Ok), so they cannot fail on the regression they name
**File:** `tests/survival/survival/survival_multi_z_covariance_autoderiv_hard.rs:358` and `:372` (origin/main)

```rust
let result = marginal_slope_covariance_from_scores(scores.view(), &w);
if let Ok(covariance) = result {
    assert_eq!(covariance.shape(), MarginalSlopeCovarianceShape::Full);
    assert_eq!(covariance.dim(), k);
}
```

**Why it is wrong.** `t06a_n_less_than_k_retains_nonzero_coupling_as_full` (n=3, k=4) and `t06b_n_equal_k_retains_nonzero_coupling_as_full` (n=4, k=4) are named for the claim that a rank-deficient score sample still gives a Full covariance. But both assertions sit inside `if let Ok(..)`. If `marginal_slope_covariance_from_scores` (`crates/gam-models/src/bms/gradient_paths.rs:1395`) returns `Err`, which is the exact regression these tests exist to catch, the test body does nothing and passes.

The input is valid. Centred sample scores span at most n − 1 < k directions, so the covariance is PSD with exact zeros in its spectrum. The function subtracts the weighted mean and hands the matrix to `MarginalSlopeCovariance::full`, which judges definiteness against the eigensolver's roundoff band for exactly this input. A refusal is therefore a defect, not an allowed outcome.

SPEC.md:16: "XFAIL pattern on tests is never allowed. A failing test should always indicate problematic behavior." An `if let Ok` guard around the only assertion is a silent XFAIL.

**Fix.** Unwrap the result and fail with the constructor's reason, then keep the Full-shape and dimension asserts:

```rust
let covariance =
    marginal_slope_covariance_from_scores(scores.view(), &w).unwrap_or_else(|error| {
        panic!("rank-deficient PSD scores (n={n}, k={k}) must give a covariance: {error}")
    });
assert_eq!(covariance.shape(), MarginalSlopeCovarianceShape::Full);
assert_eq!(covariance.dim(), k);
```

SauersML:
Fixes #3883

Both tests fed n <= k centred scores to `marginal_slope_covariance_from_scores` and asserted the Full shape only inside `if let Ok(..)`. A refusal of the covariance, which is the exact regression the tests name, therefore passed silently.

Centred scores span at most n - 1 < k directions. The resulting PSD matrix has exact zeros in its spectrum, and that is valid input to `MarginalSlopeCovariance::full`, which judges definiteness against the eigensolver roundoff band for exactly this case. A refusal now fails the test with the constructor's reason.

Validation: focused-rust-proof run https://github.com/SauersML/gam/actions/runs/35501290245 (test_binary survival).

---


SauersML:
Closed as superseded by #3902 (merged as d77cff539c), which fixes t06a/t06b on main.

---

PR #3931, opened 2026-09-20
blockwise inner: certify a railed mode by its Newton decrement (#3289)

### LINKED ISSUE #3289: Gaussian bounded() fit whose constrained optimum sits on a bound no longer fits on main (inner solve not certified; regressed since 658fec0f)
## What fails

On main at bdd00e20, an unpenalised Gaussian `bounded(x, min, max)` fit errors whenever its least-squares slope lies outside the box, so that the constrained optimum sits on a bound:

```
`y ~ bounded(x, min=1, max=3)` must fit: custom-family fit ended without a certified inner mode, so no fitted model was assembled: the terminal inner solve stopped after 32 of 1200 cycle(s) without certifying; carrying block: not identified; ... terminal verdict: blockwise terminal cycle 31: max_accepted_step=5.000000e-1 (tol=3.517026e-9), max_proposed_step=5.000000e-1, objective_change=1.278977e-13 (tol=5.465593e-9), joint_stationarity_ok=true
```

On main, a bare `bounded()` is the constrained MLE. With #3268 it becomes the explicit `prior=none` opt-out.

## Reproducer

This is the fixture from `tests/regressions/misc/bounded_default_prior_shrinks_a_null_coefficient.rs` (#3268):
- N = 200 and x on an even grid over [-1, 1].
- y = 1 + noise with sd 0.5.
- The noise's least-squares slope is pinned at 0.5 SE, about 0.035.

That slope is below `min = 1`, so the answer is β = 1 exactly. To run it, put the file on plain main, register it in `tests/regressions/misc/mod.rs`, and run `./build.sh test --test regressions bounded_default_prior_shrinks_a_null_coefficient`.

## Where it regressed

- The same fit succeeded on the 658fec0f base of #3268: it returned 1.0000645, and an earlier main-based wheel returned 1.0000000000000002.
- It fails identically on bdd00e20.

## What the verdict shows

- Each cycle the latent logit coordinate takes a full 0.5 trust-radius step toward the rail.
- Each step changes the objective by only about 1e-13, because the interval map is saturating.
- The frozen-log-likelihood / trust-boundary divergence early-exit then returns the fit unconverged, where before it would have certified the rail.

## Candidate commits

These are the candidates in the range that touch this loop:
- a04ba0aa, "joint trust region: a rounding-level change is not a measured decrease". It adds the roundoff ceiling to the blockwise `update_joint_trust_region_radius` call.
- The `capped_inner_max_cycles` → `inner_max_cycles_base.max(1)` change in `inner_blockwise_fit_for_product`.
- 403af227, "warm starts no longer retain the joint workspace" (`certified_workspace = None`).

Not bisected yet.

## Workaround in #3268

#3268's test compares against the analytic constrained optimum `clamp(β_ols, min, max)` instead of fitting `prior=none` on the rail. #3268's default shrinkage prior fits this data fine; only the unpenalised railed fit is affected.

SauersML:
Fixes #3289

## Root cause
The fit `y ~ bounded(x, min=1, max=3, prior=none)` parameterises the slope as β = 1 + 2·σ(θ). When the box-constrained optimum is the rail β = 1, that optimum sits at θ → −∞. Along that chart, every Newton step in θ stays O(1), while the gradient, the curvature and the objective the step buys all vanish like σ(θ).

The blockwise certification gate in `inner_blockwise_fit.rs` required `max_accepted_step <= step_tol`, measured in coefficient units. That never holds on this chart. So the fit kept walking down it:
- the accepted step was 5e-1 every cycle;
- the objective change was 1.3e-13;
- exact joint stationarity passed.

It walked until the frozen-log-likelihood divergence exit refused the fit.

## Fix
The gate now measures the remaining step in the posterior metric instead of coefficient units.
- Along each block's proposed direction δ, the local model offers at most (rhs·δ)² / (2 δᵀH_pen δ) of decrease. This quantity is invariant to trust truncation. For a Newton step it equals ½λ², where λ is the Newton decrement (the step measured in posterior-SD units).
- The iterate is certified when all three of these hold:
  1. the largest block decrease is inside the objective tolerance;
  2. the measured exact joint stationarity residual passes;
  3. the objective change is inside tolerance.
- The gate now runs ahead of the divergence exit, so a certified mode is not reported as a divergence.

No tolerance is loosened. The certificate reads the existing `objective_tol` and `residual_tol`.

## Resolution of the certificate
With unit dispersion the objective is f = ½A(β − b̂)², with A = Σ(x − x̄)² ≈ 67. Write s = σ(θ). Both g_θ and H_θ are ≈ 2A(1 − b̂)·s, so ½λ² ≈ A(1 − b̂)·s. The solve stops once ½λ² ≤ τ = inner_tol·(1 + |f|). That leaves

β − 1 = 2s ≲ 2τ/√A standard errors,

which is about 1e-9 SE at the default `inner_tol` and |f| ≈ 54. The regression asserts 1e-6 SE.

## Tests
- `bounded_prior_none_on_a_rail_certifies_the_constrained_optimum_3289` (new) checks that the fit reaches the rail. It asserts the fixture's OLS slope is below 1, so the test is not vacuous, and that 1 ≤ β̂ ≤ 1 + 1e-6 SE.
- `bounded_shrinkage_prior_centres_at_the_midpoint_when_zero_is_outside_the_box` again fits the `prior=none` reference. It no longer uses the analytic clamp it carried around this bug.

Validation (focused-rust-proof):
- BF-21-3289-1: run 35501127952, the `gam` regressions, filters `bounded_` and `interior_constrain_box`.
- BF-21-3289-2: run 35501129478, the `gam-custom-family` lib, filter `blockwise`.

---


SauersML:
Closing: this PR is superseded. #3935 (e0cc6a6a49) fixed #3289 by holding the blockwise trust region on a kept step whose ratio the rounding cannot resolve, and it merged before this branch's proof ran, so I cancelled that proof (runs 35501127952 and 35501129478).

This branch takes a different route: a Newton-decrement certificate in the gate. That route is not needed for #3289 now that the other fix is on main. The branch stays pushed in case the certificate route is wanted later.

---

PR #3934, opened 2026-09-20
Delete the outer-driven inner P-IRLS iteration cap (#3536)

### LINKED ISSUE #3536: rho_optimizer: first_order_inner_cap_schedule caps inner P-IRLS by iteration-count heuristics (3/5/10, +1/+2/+4, x2/x3, [3,64]), so BFGS consumes first-order-wrong envelope gradients
## Where (origin/main)

`crates/gam-solve/src/rho_optimizer/bridges.rs`:

- **2377–2386**: `INNER_CAP_CONVERGENCE_OVERRIDE_RATIO = 0.01`, `INNER_CAP_FLOOR = 3`, `INNER_CAP_CEILING = 64`.
- **2432–2515**: `first_order_inner_cap_schedule`.
- **~1972**: called from `eval_grad`, and the value is stored into `feedback.cap`.

The schedule is on the production path. Standard REML wires it through `.with_outer_inner_cap(reml_inner_progress_feedback(&reml_state))` at `crates/gam-solve/src/estimate/optimizer.rs:1601` and `:1912`. It is consumed at `crates/gam-solve/src/reml/gradient_hessian.rs:6656` (`pirls_config.max_iterations = Self::effective_inner_iteration_budget(original_cap, outer_cap)`).

The code:

```rust
if matches!(g_ratio, Some(r) if r < INNER_CAP_CONVERGENCE_OVERRIDE_RATIO) { return 0; }
if let Some(snap) = last {
    let next = if snap.last_converged {
        let mut margin = match snap.last_ift_residual {
            Some(r) if r < 0.01 => 1usize,
            Some(r) if r >= 0.10 => 4usize,
            _ => 2usize,
        };
        if matches!(snap.last_accept_rho, Some(r) if r < 0.5) { margin = margin.saturating_add(2); }
        snap.last_iters.saturating_add(margin)
    } else {
        let multiplier = if matches!(snap.last_accept_rho, Some(r) if r < 0.3) { 3 } else { 2 };
        snap.last_iters.saturating_mul(multiplier).max(snap.last_iters.saturating_add(4))
    };
    return next.clamp(INNER_CAP_FLOOR, INNER_CAP_CEILING);
}
match accepted_iters { 0 => 3, 1 => 5, _ => 10 }
```

## Why it is wrong

1. **The constants are arbitrary.** Every one of these is arbitrary: 0.01, 0.10, 0.5, 0.3, +1/+2/+4, ×2/×3, 3, 64, and 3/5/10. The doc comment says the schedule "Replaces the older hardcoded iter-tier (3/5/10/20)", yet the cold path is still exactly that 3/5/10 tier. It calls itself "Eisenstat–Walker style", but Eisenstat–Walker picks a **residual forcing term** η_k tied to the outer residual, not an iteration count. SPEC.md: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code."

2. **The gradient is wrong at first order.** The profiled gradient is an envelope-theorem gradient, which is exact only at the inner mode. A solve truncated at `cap` iterations leaves δβ ≠ 0. The cost is then off by O(‖δβ‖²), but the gradient is off by O(‖δβ‖), through ∂²V/∂ρ∂β·δβ. Nothing in the schedule relates the cap to the size of that gradient error relative to ‖g‖.

3. **The line search mixes two fidelities.** `eval_cost` explicitly uncaps (bridges.rs ~1741, "the deciding cost MUST be the true converged-inner objective the analytic gradient differentiates"), while `eval_grad` re-installs the heuristic cap. So one strong-Wolfe line search compares converged costs against gradients from truncated solves. BFGS secant pairs are then built from inconsistent (s, y).

4. **The code already works around the damage reactively.** The cost-stall "STUCK" branch (bridges.rs ~2295–2310) says: "the stuck state is caused by the inner cap stopping PIRLS short of its fixed point, leaving the cost and the analytic gradient inconsistent". It responds by setting `feedback.cap = 0` after the fact. That is a reactive fallback for a defect this schedule creates. SPEC.md: "In general, do not paper over solver issues."

5. **(Minor) The rustdoc is attached to the wrong function.** The schedule's doc block (2388–2410) sits directly above the `inner_solve_converged` doc, so rustdoc attaches both to `inner_solve_converged`, and `first_order_inner_cap_schedule` has no doc at all.

## Principled fix

Replace the iteration-count schedule with an **inexact-gradient forcing condition**:

- **Stop the inner solve on a residual, not a count.** Stop when the inner KKT residual r satisfies ‖(∂²V/∂ρ∂β) H_ββ⁻¹ r‖ ≤ η‖g_ρ‖, with a fixed forcing fraction η < 1. η could be the θ of standard inexact-gradient BFGS/Wolfe theory, or the derived value tracked in #3192. In practice this is the bound on the gradient error induced by the inner residual.
- **The data it needs already exists.** The IFT residual, the mixed Hessian and the H_ββ factor are all at hand: the IFT sensitivity is already formed for the gradient.
- **Stop capping `eval_grad`, or make the inner tolerance the forcing tolerance above**, so that the gradient BFGS consumes is correct to a known relative accuracy. The iteration budget then stays at `pirls_max_iterations` alone.
- **Delete the dead weight.** That means `INNER_CAP_*`, the tier table, and the reactive uncap in the STUCK branch (it becomes unnecessary).

## Tests to add

- An inner problem where 3–10 P-IRLS iterations leave a residual whose induced gradient error exceeds 0.5‖g‖. With the forcing condition, the outer BFGS must still converge, with a monotone Armijo sequence and no STUCK uncap event.
- A unit test asserting that the envelope-gradient error bound is ≤ η‖g‖ at every `eval_grad`.

Related (all closed, different defects): #638, #1573, #2410. Umbrella: #2469.

SauersML:
Fixes #3536

## What changed
The outer optimizer no longer truncates inner P-IRLS by an iteration count. Before, the value and the gradient at the same rho could come from inner modes solved to different fidelity: the cost path ran under a reduced `outer_inner_cap` budget, while the gradient path ran under full fidelity. The caches were also keyed on (rho, cap).

- Every inner solve now uses the full `pirls_max_iterations` budget and the residual forcing tolerance `min(pirls_tol, reml_tol / INNER_KKT_REML_DIVISOR)`.
- The eval, PIRLS and outer-eval caches are keyed on rho alone (`sanitized_rhokey`), so value and gradient share one inner mode.
- Removed: the adaptive KKT clamp, `last_pirls_accept_rho`, `outer_inner_cap`, `current_outer_eval`, `forget_previous_outer_eval`, `previous_outer_gradient_norm`, `inner_budget_exhaustion_was_scheduled`, `effective_inner_iteration_budget`, `FullFidelityInnerCapGuard` and `eval_seed_at_full_inner_fidelity` (callers use `eval_with_order` directly).
- Renamed: `with_outer_inner_cap` to `with_inner_progress_feedback`, `OuterConfig.outer_inner_cap` to `inner_progress` (`InnerProgressFeedback` now only carries accepted_iter, last_iters, last_converged and force_cold), and `fit_model_for_fixed_rho_with_adaptive_kkt` to `fit_model_for_fixed_rho_configured`.
- Added `terminal_state_is_reset(obj, config)` for the certificate path.

## New test
`single_fidelity_inner_solve_tests_3536::value_and_gradient_share_one_inner_mode_at_the_outer_derivative_tolerance` checks that the cost and the gradient at one rho come from the same converged inner mode (identical beta, one cache entry), and that the analytic gradient matches a central finite difference of the cost within the outer derivative tolerance.

## Remaining (future work)
A derived inexact-Newton forcing condition, ‖(∂²V/∂ρ∂β)H⁻¹r‖ ≤ η‖g‖, would replace the fixed divisor. It is related to #3192.

---


SauersML:

New head: 4c20ceb43ca8192c7c72952c3a9d0eb591254e0c. Two files conflicted:
- `gpu/pirls_gpu.rs` (doc only): kept main's wording (#3714 dropped the KKT-diagnostics mention) together with this PR's renamed oracle `fit_model_for_fixed_rho_configured`.
- `reml/gradient_hessian.rs`: `clear_warm_start_adaptive_signals` now calls main's renamed `clear_warm_start_trust_state()`, and this PR's removal of the `last_pirls_accept_rho` store still stands. The trailing hunk is dropped entirely: this PR deleted `inner_budget_exhaustion_was_scheduled` and its tests, and main deleted the adaptive IFT/tangent caps. I grepped the merged tree and found no leftover references to the removed symbols.

SPEC check: `INNER_KKT_REML_DIVISOR = 100.0` is the base's `ADAPTIVE_KKT_FLOOR_REML_DIVISOR` renamed, not a new knob. The PR removes the adaptive caps and budgets around it. The derived forcing condition that would replace it is still tracked in #3192.

---

SauersML:
LC-3: head 4c20ceb4 does not compile. `crates/gam-solve/src/pirls/convergence.rs:26` has an unexpected closing delimiter, and there is an unused `use super::WorkingModelPirlsOptions;`, both left over when the adaptive-KKT schedule was deleted.

- 0b16c95d removes those 4 lines. After it, rustfmt parses gam-solve, gam-gpu, gam-models and `tests/constrained_terminal_decrement_2668.rs`.
- e3f573ee merges origin/main (f978f1d0) in. `git merge-tree` shows no conflicts.
- In the merged tree there are no references to any symbol this PR deletes: `effective_kkt_tolerance`, `AdaptiveKktTolerance`, `apply_arc_inner_cap_schedule`, `uncapped_stand_in_key`, `INNER_CAP_*`, `with_outer_inner_cap` and the others.
- I cancelled the stale run 35501768241 and dispatched focused-rust-proof run 35511154910 on e3f573ee (gam-solve lib, filters `single_fidelity_inner_solve_tests_3536 reml_certificate plan_boundary run_plan`).

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, e07daf8918f8.

PR #3936, opened 2026-09-20
gam-event-history tests: fail the loaded-cohort trace when the fixed-λ inner solve refuses

### LINKED ISSUE #3886: crates/gam-event-history tests: traced_fixed_lambda_inner_solve_on_the_loaded_cohort_reports_its_cost prints a refused fixed-λ inner solve instead of failing
**File:** `crates/gam-event-history/src/tests.rs:3029` (origin/main), test `traced_fixed_lambda_inner_solve_on_the_loaded_cohort_reports_its_cost` (line 2881).

```rust
    let result =
        gam_custom_family::fit_custom_family_fixed_log_lambdas(&traced, &specs, &options, None);
    match result {
        Ok(fit) => { /* trace + FD rows, emitted */ }
        Err(error) => emit(&format!(
            "[cost] error after {:.1}s: {error}",
            clock.elapsed().as_secs_f64()
        )),
    }
```

**Why it is wrong.** `fit_custom_family_fixed_log_lambdas` returns `Ok` only for a converged inner mode. A refusal is therefore exactly the stalled inner solve this trace exists to report, and the test turns it into a log line and passes. The fixture is the *loaded* cohort (`simulate_cohort(80, 6.0, -0.8, 0.5, 1.0, 0.4, 7)`, loading 1.0), the same one on which the full fit already recovers a positive loading. The inner mode is identified, so a refusal is an engine defect and not an expected decline.

SPEC.md:16: "A failing test should always indicate problematic behavior." SPEC.md:21: "In general, do not paper over solver issues."

**Fix.** Make the `Err` arm panic with the refusal and the elapsed time. Leave the null-cohort trace alone: with a zero loading the atom's rate is not identified, so a refusal there is not clearly wrong.

(Side note, not in the fix: the `Ok` arm's exact-gradient vs FD and Louis-Hessian vs FD rows are also only emitted and never asserted. Grading them needs a derived FD error band, so that is left as a follow-up.)

SauersML:
Fixes #3886

`traced_fixed_lambda_inner_solve_on_the_loaded_cohort_reports_its_cost` matched on the result of `fit_custom_family_fixed_log_lambdas` and `emit()`ed the error, so a refused inner solve passed the test.

That function returns `Ok` only for a converged inner mode, so a refusal is exactly the stalled solve the trace exists to report. The fixture is the loaded cohort (seed 7, loading 1.0), on which the full fit already recovers a positive loading. The inner mode is therefore identified, and a refusal is an engine defect. The `Err` arm now panics with the refusal and the elapsed time.

The null-cohort trace is left alone. With a zero loading the atom's rate is not identified, so a refusal there is not clearly wrong.

Validation: focused-rust-proof run https://github.com/SauersML/gam/actions/runs/35500765768.

---


SauersML:

---

SauersML:
Verified locally and merged as 6dd4dc81 (the Actions queue is backed up).

- Tested content: the same diff as head `95625e37` (that head only merged main), with `origin/main` merged locally
- `cargo test -p gam-event-history --lib traced_fixed_lambda_inner_solve`: **2 passed, 0 failed** (13.0 s)
- Review against SPEC.md: sound. A refused inner solve on the identified loaded cohort now fails the test instead of passing silently ("A failing test should always indicate problematic behavior").

---

PR #3937, opened 2026-09-20
Judge joint trust-region steps with the Sun–Nocedal ratio over the evaluation band

### LINKED ISSUE #3240: opt trust-region/ARC ratio: judge steps against the objective's evaluation band (Sun–Nocedal), delete the 1e-14 noise floor
## Problem

opt's `TrustRegionPolicy::update` and the ARC loop compute the acceptance ratio ρ = actual/predicted with no allowance for the rounding of the objective. When a step's predicted decrease is close to the objective band ε_f (near the optimum, which is where certification happens), the actual decrease is noise-dominated, so ρ is effectively random. Good steps get rejected, the radius collapses, and the run ends uncertified.

The current workarounds are SPEC violations (magic constants and fallbacks):
- `TrustRegionPolicy::noise_aware(1e-12, 1e6, 1e-14)` and `impl Default` use a **relative** `noise_floor_rel = 1e-14` multiplied by an `objective_scale`. That constant is arbitrary and scale-dependent.
- If |pred| falls below that floor, the policy sets `within_noise_floor` and forces **ρ = 1**. That accepts any step in the floor, including ascents.
- gam works around it on its side:
  - `crates/gam-custom-family/src/joint_newton.rs:2017` defines `JOINT_TRUST_NOISE_FLOOR_REL = 1e-14`.
  - `:2150-2170` computes `noise_floor = (scale·REL).max(measured_resolution)` and adds the gam#2637 `predicted_nonpositive` override.
- ARC exits with `ModelNoiseFloor` (uncertified) when pred ≤ band_f, even when the decrement λ̂² is several times ε_f and a certifiable step exists.

## Root cause

The ratio test ignores the band ε_f to which f is evaluated. Sun & Nocedal (2023, arXiv 2201.00973, Algorithm 1) give the principled fix. For a model that is exact up to |actual − pred| ≤ 2ε_f,

ρ = (Δf + r·ε_f) / (Δm + r·ε_f),  r = 2/(1 − c₂)

guarantees |ρ − 1| < 1 − c₂. So a step whose only shortfall is rounding is never treated as unsuccessful, and a step that realizes nothing against a resolvable prediction is still rejected. That requires c₀ = eta_accept > 0, so eta_accept moves from 0 to 0.1. ε_f is the absolute band, taken from `DecrementBands.objective` at x_k and at the trial point (the max of the two). There are no relative constants and no scale argument.

## Fix (opt, patch below; `SauersML/opt` push is not available to this session, so the owner has to land it)

Commit "Judge a trust-region step against the objective's evaluation band (Sun–Nocedal ratio)". It applies on opt `30ca48f` + the #3196 patch + the #3230 patch. Changes:
- `TrustRegionPolicy`:
  - `noise_floor_rel` is replaced by `objective_band: f64`, the absolute ε_f;
  - new builder `with_objective_band(eps)`;
  - `classic` has band 0, which gives the classical ratio exactly.
- `noise_aware` is renamed `noise_tolerant(min_radius, max_radius, objective_band)`, with eta_accept 0.1.
- `impl Default` is deleted, and with it the 1e-14 constant.
- `update(radius, step_norm, hit_boundary, actual, predicted)` drops the `objective_scale` argument.
- The ρ=1 `within_noise_floor` branch and the field are deleted. A nonpositive prediction still rejects.
- The NewtonTrustRegion loop and ARC (ratio with c₂ = eta2) both take ε_f from the samples' `DecrementBands.objective`.
- The Riemannian TR caller is updated.

## Data (reproduced before the fix, then after)

Tests were added to opt and **fail before the fix, pass after**:

| test | before | after |
|---|---|---|
| `newton_trust_region_accepts_a_step_whose_shortfall_is_within_the_objective_band` (V=½x² ± ε offset basin, ε=1e-10, x0=1.2·√(2ε)) | `TrustRegionRejectFloor` at x=1.4142135631069867e-5 after 45 iterations, radius 2.22e-16 | `NewtonDecrementCertified`, \|x\| ≤ √ε |
| `arc_accepts_a_step_whose_shortfall_is_within_the_objective_band` (x0=2.5√ε, basin edge 2.25√ε, σ0=4/√ε) | uncertified `ModelNoiseFloor` (pred=7.916667e-11 ≤ floor 1e-10); first trial old ρ ≈ −0.51 | `NewtonDecrementCertified`; first trial ρ ≈ 0.92 |
| `a_shortfall_within_twice_the_band_never_shrinks_the_region` (sweep pred 1e-14…1, shortfall ±2ε) | shrinks | never shrinks (lemma) |
| `a_resolvable_prediction_that_realizes_nothing_is_rejected` | — | ascent of rε against pred 0.2741 rejected |
| `a_decrease_above_the_band_against_a_tiny_prediction_is_accepted` (gam#2637 numbers: actual 2.090494888e-9, pred 2.944376434e-13, band 4.156290748e-13) | needed gam's override | `GrowAtBoundary` from the policy alone |

The full opt suite passes: 221/221 lib tests plus 1 doc test.

## gam changes required when opt is bumped past this commit (pin is `3b58e32`, Cargo.toml:76)

1. `crates/gam-custom-family/src/joint_newton.rs`:
   - Delete `JOINT_TRUST_NOISE_FLOOR_REL` (:2017) and the `(scale·REL).max(...)` fallback (:2150-2170).
   - Use `TrustRegionPolicy::noise_tolerant(JOINT_TRUST_RADIUS_FLOOR, JOINT_TRUST_RADIUS_CEILING, measured_resolution)` and `update(...)` without `objective_scale`.
   - The gam#2637 override becomes redundant; re-verify it with the #2637 repro before deleting it.
2. **Must precede the bump.** `crates/gam-solve/src/rho_optimizer/decrement_bands.rs` `outer_decrement_bands` sets `DecrementBands { objective: tolerance.value() }`. That is the *verdict tolerance* max(τ_stat − band_f, band_f), not the rounding band opt's field is documented as. With this patch that value becomes ε_f in the ratio. At τ_stat scale, r·ε_f would let the outer loop accept ascents of about 7τ. `objective` must carry band_f, and the verdict tolerance must travel separately. This concerns the outer-cert work (#3192 / #2817).
3. `crates/gam-geometry/src/optimizer.rs:100`: doc comment reference to `TrustRegionPolicy::classic`; re-check the wording.

Related: #3196 (exact dense TR subproblem, patch posted there), #3230 (decrement band δc²/μ, patch posted there), #2637, #2612, #2600.

Follow-ups in opt, with the same ratio flaw, not in this patch:
- the BFGS TR probe (~lib.rs:8394: ratio act/pred with 1e-12/1e6 caps and `ARC_NUMERICAL_CONV_FACTOR`);
- the matrix-free TR (~lib.rs:11559 `actual / predicted`);
- Shi relaxed Armijo +2ε_f at `accept_armijo`, which replaces `ARMIJO_ROUNDOFF_EPS_MULTIPLE`/`armijo_roundoff_cushion`.

The patch follows in the next comment (`git am` on top of the #3196 and #3230 patches).

SauersML:
Fixes #3240

## Problem

The joint Newton radius update divided the realized change by the predicted change. It then patched the cases where that ratio meant nothing. `AcceptBelowModelNoiseFloor` accepted any step whose prediction fell below a hand-set `JOINT_TRUST_NOISE_FLOOR_REL = 1e-14` times |F|, whatever the objective actually did.

SPEC.md: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code." and "In general, do not paper over solver issues."

## Fix

The objective already carries a measured evaluation band:

    ε_f = max(½·roundoff_ceiling, measured_resolution)

The first term applies only when the ceiling is finite and positive; otherwise ε_f is the measured resolution. Any difference of two evaluations is known only to within 2ε_f.

The radius update now goes through opt's `TrustRegionPolicy::noise_tolerant`, which forms

    ρ = (a + r·ε_f) / (p + r·ε_f),   r = 2 / (1 − eta_expand) = 8

whenever p + r·ε_f > 0.

Lemma: if |a − p| ≤ 2ε_f and p ≥ 0, then |ρ − 1| ≤ 2ε_f / (r·ε_f) = 0.25.

- A rounding-level sign flip therefore reads as agreement, and the radius is held.
- An ascent beyond the band still reads ρ < eta_accept, and the step is rejected.
- With ε_f = 0 this is exactly the classical ratio.

The fix depends on SauersML/opt branch `gam-3240-noise-tolerant-ratio` at 5f3e39a. That branch forms the ratio for p ≤ 0 inside the band and drops `objective_scale`, `within_noise_floor` and `Default`. Once it merges, the pin here gets re-bumped to the opt merge sha.

## Changes

- Remove `AcceptBelowModelNoiseFloor` and `JOINT_TRUST_NOISE_FLOOR_REL`.
- `update_joint_trust_region_radius` drops `objective_scale`. It takes the measured resolution and the roundoff ceiling, and derives ε_f with `joint_objective_evaluation_band`.
- The gam#2612 unreadable-growth clause uses the same ε_f.
- The exact-joint secondary gate slack is widened to max(roundoff_slack, 2ε_f).
- Bump the opt pin to 5f3e39a.
- Rewrite the radius-policy tests as band-level cases: sign flip, genuine increase, non-positive prediction beyond the band, and growth or hold at the boundary.

---


SauersML:
Status (BF-20 sweeper, ~11:43 UTC): the code is pushed at 32bf2d8a, but validation is blocked behind the account-wide Actions queue (~540+ queued runs, the oldest from 08:13, not draining).

Queued validation runs:
- gam 35501778508 (BF-20-3240-1): gam-custom-family lib, filters `inner_solver_numerics joint_newton exact_joint_fit inner_blockwise_fit`
- gam 35503380297 (BF-20-3240-2): gam-solve lib, filters `arc_ rejected_trust_region exact_jet 2817 3017 3279`
- gam 35503381547 (BF-20-3240-3): gam-models lib, filters `binomial_location_scale_engine_matches_reference_flow penguin_fixture`
- SauersML/opt#22 CI run 35501348243: windows green; ubuntu and macos still queued

Steps to land:
1. opt#22 goes green.
2. Merge opt#22 (merge commit).
3. Re-pin `opt` in Cargo.toml and Cargo.lock on this branch to the opt#22 merge sha (it currently pins branch head 5f3e39ae).
4. The focused proofs above go green, then mark this PR ready and merge.

Review points: `eta_accept` goes from 0 to 0.1; the ε_f = max(½·roundoff ceiling, measured resolution) band can be wider than the old 1e-14·|F| floor; ARC and NTR in gam-solve now pre-filter p ≤ 0.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 2433971ac669.

PR #3939, opened 2026-09-20
pygam_compare: judge interval coverage on the mean coverage over seeds

### LINKED ISSUE #3788: bench/pygam_compare: coverage verdict averages per-seed |cov - 0.95|, which rewards over-covering intervals and misjudges calibration
## Defect

`bench/pygam_compare/report.py:167-168` (origin/main) scores interval calibration per seed:

```python
if metric == "coverage":
    diffs.append(abs(gv - COVERAGE_TARGET) - abs(cv - COVERAGE_TARGET))
```

`paired_verdict` then averages these differences. The quantity it compares is `mean_s |cov_s - 0.95|`, not the calibration of the interval, which is `|mean_s cov_s - 0.95|`.

## Why this is wrong

The intervals from one fit, over that fit's own held-out rows, move together with the fit's error. So even an exactly calibrated interval has per-seed coverage that scatters around 0.95. In the committed baseline, per-seed sd runs from 0.01 to 0.08. `E|cov_s - 0.95|` is then about 0.8·sd for a calibrated interval, and it counts that scatter as miscalibration.

The effect is to reward conservative intervals. An interval that covers every true mean on every seed (coverage 1.0, clearly conservative) scores 0.05. That beats a perfectly calibrated interval whose per-seed coverage has sd above about 0.063, and gamfit's binomial p1 cell has sd 0.079.

That is backwards under the project's calibration rule: conservative is a bug, just as anti-conservative is. It works in both directions:
- gamfit could over-cover and be scored a WIN.
- A calibrated gamfit could be scored a LOSS against an over-covering pyGAM.

## Evidence on the committed baseline (`baseline/quick`)

I re-rendered the committed records with the verdict computed on the mean coverage (the fix below). Changes in the coverage table:

| cell | mean cov gamfit / pygam / pygam_gs | current verdict | on mean coverage |
|---|---|---|---|
| gaussian n=1000 p20 vs pygam | 0.956 / 0.940 | **LOSS** Δ=+0.011±0.0032 | better n.s. Δ=-0.0043±0.02 (gamfit is 0.006 from nominal, pyGAM 0.010) |
| poisson n=1000 te vs pygam_gs | 0.961 / 0.841 | WIN Δ=-0.087±0.042 | better n.s. Δ=-0.098±0.053 |
| poisson n=1000 p1 vs pygam | 0.963 / 0.975 | worse n.s. | better n.s. |

- The gaussian p20 LOSS is spurious: gamfit's mean coverage is closer to 0.95 than pyGAM's.
- The real LOSS stays. gaussian p5 over-covers at 0.991 against pyGAM's 0.971/0.973, and it remains a LOSS under both rules.

## Fix

Judge coverage on each library's mean coverage, `|ḡ - 0.95| - |c̄ - 0.95|`. Keep the paired-by-seed 2-SE rule:
- Fix `side = sign(mean - 0.95)` for each library.
- Seed `s` then contributes `side_g (g_s - 0.95) - side_c (c_s - 0.95)`.
- These terms average to the mean-level difference exactly, and their spread gives its delta-method SE.
- With one seed this reduces to the current per-seed rule.

Add tests:
- An always-covering gamfit against a calibrated, scattering pyGAM must be a LOSS. The current code says "better n.s.".
- The reverse must be a WIN.

Also note that the committed `baseline/quick/report.md` is already stale against `report.py`: it lacks the import-CPU table. It will be regenerated with the next baseline run.

SauersML:
Fixes #3788

## Problem
`bench/pygam_compare/report.py` `_paired` scored the coverage metric as the mean of the per-seed distances `|cov_s - 0.95|`. On one fit's own held-out rows, that fit's intervals all move with the fit's estimation error. A calibrated 95% interval's per-seed coverage therefore scatters widely around 0.95, and the per-seed distance counts that scatter as miscalibration. As a result, an interval that covers every true mean on every seed (coverage 1.0, a real over-coverage miss) beats a calibrated interval whose seeds average 0.95. The verdict could reward over-coverage, so a coverage LOSS or WIN in the report was not measuring calibration.

## Fix
Coverage is now judged on how far each library's **mean** coverage over the seeds sits from 0.95, in either direction.

Each paired seed contributes `side_g (g_s - 0.95) - side_c (c_s - 0.95)`, with `side = sign(mean - 0.95)`. These terms average exactly to `|mean g - 0.95| - |mean c - 0.95|`, and their standard error is the delta-method SE of that difference. The existing paired 2-SE rule therefore applies unchanged. Every other accuracy metric keeps its per-seed paired difference.

Changes by file:
- **`report.py`**:
  - `_paired` collects the paired values.
  - For `coverage`, it hands those values to the new `_coverage_miss_terms`.
  - The label and the module docstring now name the rule.
- **`test_pygam_compare_smoke.py`**:
  - `test_paired_accuracy_verdicts` now also checks that a library with coverage 1.0 on every seed LOSES to a calibrated one whose seeds are (0.80, 1.0, 1.0, 1.0, 0.95).
  - It checks that the mirror case WINS.
  - The old code fails both new assertions.
- **`README.md`**: a Coverage bullet in the Verdicts section.

## Evidence (local, no cargo)
- With the new `report.py`, these tests pass: `test_paired_accuracy_verdicts`, the ratio test, the timeout test and the `n_predict` test. The new assertions fail against main's `report.py`.
- Re-rendering the committed `baseline/quick` records with this `report.py` changes these verdicts:
  - gaussian n=1000 p20 vs pygam: the spurious coverage LOSS becomes "better, n.s.".
  - poisson te vs pygam_gs: the WIN becomes "better, n.s.".
  - The real over-coverage LOSS for gaussian p5 (mean 0.991) stays a LOSS.
- The committed `baseline/quick/report.md` is already stale against main's `report.py`: it lacks the import CPU table. It is not regenerated here.

No workflow runs the bench's Python tests on PRs (`python-contracts.yml` is disabled), so this PR has no CI checks. I am leaving it open for review.

---


SauersML:
Closing in favour of #3922 (fix commit adfac17f3f, merged to main as 596bd06ed7). Both PRs implement the same rule for #3788: seed s contributes `side_g (g_s - 0.95) - side_c (c_s - 0.95)`, and paired_verdict takes the mean and `stdev/sqrt(S)` of those terms. The across-seed SE is the correct one here. The seed is the independent unit, and rows inside one fit share that fit's error, so a row-level Bernoulli SE `sqrt(p(1-p)/n_rows)` would understate it. main's version is the correct one at the kink, and this PR's is not.

**Defect in this PR:** `_side` returns 0 when a library's mean coverage is exactly 0.95. That library's term then drops out of every seed, and so does its variance. The SE is then computed from the other library alone, so it is anti-conservative. main uses `math.copysign(1.0, ·)`, which keeps the ±1 side at the kink. Its term still averages to `|mean - 0.95| = 0` and keeps the library's scatter in the SE.

Reproducer (report.py only, no cargo): gamfit coverage alternates 0.90 / 1.00 over 10 seeds (mean exactly 0.95, sd 0.05). pyGAM sits at about 0.955 with sd 5e-5.
- main: `better n.s. Δ=-0.00505±0.017`. This is correct: gamfit's mean-coverage estimate has SE 0.016, so a 0.005 edge cannot be claimed.
- this PR: `WIN Δ=-0.00505±1.7e-05`. The SE is off by a factor of about 1000.

main already has everything else in this PR, and more:
- the README Coverage bullet;
- the docstring and label;
- tests with an always-covering LOSS and the mirror WIN, plus the exact identity that the terms average to `|ḡ-0.95| - |c̄-0.95|`;
- a re-rendered `baseline/quick/report.md`.

**Checked on main, by Monte Carlo** (20k replications, 10 seeds, normal per-seed coverage):
- On the null boundary away from the kink, the false-LOSS rate is 3.4% (g = c = 0.97, and g = 0.97 vs c = 0.93). That is the 2-SE rule's nominal rate `P(t_9 > 2) ≈ 3.8%`, so the verdict is neither conservative nor anti-conservative there.
- At the kink (both means 0.95, sd 0.05 vs 0.01) it is 5.4%. At that point the null is `μ_g = 0.95`, which is a two-sided question, so the 2-SE cut acts as a two-sided test there. That is the expected behaviour of a distance comparison, not the per-seed bias that #3788 fixed.

Multiplicity handling is unchanged in both PRs. Each cell and metric gets its own 2-SE verdict, and every LOSS is printed (the report's stated contract).

---

Issue #788, opened 2026-06-05
survival marginal-slope: penalty-block shape mismatch ('block 1 penalty 0 must be NxN, got 2x2'), matern & duchon, all N


SauersML:
## survival marginal-slope: penalty-block shape mismatch (`block 1 penalty 0 must be NxN, got 2x2`)

Basic `survival-likelihood=marginal-slope` fits fail with an internal penalty-dimension invariant violation, for **both matern and duchon**, at **every N tested (500, 2000)**, with or without a parametric term.

### Invocation
```python
gamfit.fit(
    df,                                   # cols: t0=0, t1>0, event{0,1}, z, PC1, PC2, PC3, sex{0,1}
    "Surv(t0, t1, event) ~ matern(PC1, PC2, PC3, centers=8) + sex",
    survival_likelihood="marginal-slope",
    z_column="z",
    logslope_formula="matern(PC1, PC2, PC3, centers=8)",
)
```

### Error
```
IntegrationError: custom-family invalid input in custom-family string boundary:
block 1 penalty 0 must be 8x8, got 2x2
```
- matern(centers=8): `must be 8x8, got 2x2`
- duchon(centers=8, order=1): `must be 9x9, got 2x2`
- **Without** the `+ sex` parametric term: `got 1x1` (instead of 2x2). So the *provided* penalty tracks the parametric block size (1→2 when `sex` added), while the *expected* size is the surface (centers+order). The surface penalty for `block 1` is being constructed at the parametric block's dimension, not the smooth's.

### Scope
- gamfit **0.1.169** (newest). Also reproduces on **0.1.158** (`got 1x1`), so it's a standing bug, not a 0.1.169 regression.
- Both matern and duchon; N=500 and N=2000; left-truncated `Surv(t0, t1, event)`.

### Repro data
Synthetic: `z,PC1,PC2,PC3 ~ N(0,1)`, `sex ~ Bernoulli(0.5)`, `eta=0.4z+0.3·PC1−0.2·PC2`, `T~Exp(0.05·e^eta)`, `C~U(0,20)`, `t1=max(min(T,C),1e-3)`, `event=1{T≤C}`, `t0=0`. Fit the call above.

### Impact
Blocks survival marginal-slope (the biobank/AoU use case) with the documented `z_column` + `logslope_formula` pattern. The library should either construct the `block 1` surface penalty at the smooth's dimension, or raise a clear configuration error rather than an internal `must be NxN, got 2x2` assertion.

[...]

PR #3941, opened 2026-09-20
Unit-free guard feasibility for offset-only and zero derivative rows (#3766)

### LINKED ISSUE #3766: survival/time_constraints: offset-only guard feasibility uses two unrelated slacks (a "historical" 1e-12 vs 256 eps) and an absolute 1e-24 zero-row cutoff duplicated in the marginal-slope validator
## Where

- `crates/gam-models/src/survival/time_constraints.rs:78-98`: `FeasibilityTolerance`
  ```rust
  pub enum FeasibilityTolerance {
      /// `1e-12 · (1 + max(|offset|, |guard|))`.
      AbsoluteScaled,
      /// `256 · f64::EPSILON · (1 + max(|offset|, |guard|))`
      EpsilonScaled,
  }
  ...
  FeasibilityTolerance::AbsoluteScaled => 1e-12 * scale,
  FeasibilityTolerance::EpsilonScaled => 256.0 * f64::EPSILON * scale,
  ```
- `crates/gam-models/src/survival/location_scale/constants.rs:104-109`: `LOCATION_SCALE_GUARD_POLICY` selects `AbsoluteScaled`, with the comment "feasibility of coefficient-free rows uses the family's *historical* absolute slack". Latent survival reaches the same policy through `structural_time_coefficient_constraints`.
- `crates/gam-models/src/survival/marginal_slope/fit_setup.rs:208-211` selects `EpsilonScaled`.
- `time_constraints.rs:199`: `if row_norm_sq <= 1e-24` decides a row is immovable. The same literal is repeated in `marginal_slope/feasibility.rs:161` (`row_norm_sq > 1e-24`), and the comment there says the two must be "identical to the constraint builder".

## Why it is wrong

1. **The same decision gets two different answers.** The question "can this immovable row clear the guard from its offset alone?" is the same in every survival family, yet a row short by `5e-13·scale` passes in location-scale and latent survival and is refused in marginal-slope. Neither constant is derived:
   - `1e-12` is labelled "historical".
   - `256` is a hand-picked ulp count.

   The `256 eps` slack also disagrees with marginal-slope's own likelihood-domain predicate for the same `q'`. `survival_derivative_guard_tolerance` (marginal_slope/spec.rs:115) admits `max(4·ACTIVE_SET_PRIMAL_FEASIBILITY_TOL, 256 eps)·(1+max(|q'|,|guard|))`.
2. **The absolute cutoff depends on scale.** `row_norm_sq <= 1e-24` is an absolute cutoff on the squared derivative-design row norm. Rescaling time units (days vs years) rescales `D`, so the same model can have a row classified as immovable, and a fit refused as infeasible, in one unit system but not another. A row with a nonzero entry can always be moved by `β`. The only unit-free "cannot move" test is structural: an exactly zero row, or a row that is zero in the support pattern.
3. **The literal is duplicated between the builder and the validator.** If either copy changes without the other, the builder and the validator disagree about which rows are constrained.

## Principled fix

- Keep one immovable-row predicate, exported from `time_constraints`: `row_norm_sq == 0.0`, i.e. structurally zero. Use it in both the builder and `marginal_slope/feasibility.rs`.
- Keep one offset-only feasibility slack, not a per-family enum. It should be derived from the band the family's likelihood-domain predicate admits for `q' = offset`, so that the "offset satisfies guard" test at build time and the row kernel's domain test at evaluation time coincide. Delete `FeasibilityTolerance`. `GuardPolicy` stays: latent survival (`latent/interval.rs:120`) admits `guard == 0`, while marginal-slope requires `> 0`.
- Pin this with a unit-invariance test: scale `D` and the offsets by `c` (a change of time unit) and assert the same accept/refuse verdict.

SauersML:
Fixes #3766

## The problem
The time-derivative guard builder had two checks that did not agree with the rest of the code:
- It decided whether a row was immovable with an absolute `||row||^2 <= 1e-24` cutoff. That cutoff changes with the time unit, so the same model was accepted at one unit and refused at another.
- It accepted an offset-only row through a per-family slack. The marginal-slope likelihood domain test used a separate `max(4*TOL, 256*eps)` band, so the builder and the domain test could disagree about the same `q' = offset`.

## Changes
- **Immovable rows.** A row is immovable only when every design entry is exactly `0.0`. This test is structural and does not depend on units. `derivative_row_is_immovable` is shared by the builder and the marginal-slope feasibility diagnostic.
- **Offset-only rows.** An offset-only row clears the guard within one band, the one the active-set solver certifies: `4*TOL*(1 + max(|q'|, |guard|))`, where TOL is the solver's primal feasibility tolerance (#788). `survival_derivative_guard_violated` applies the same band, so the build-time and evaluation-time verdicts coincide.
- **Marginal-slope `validate_spec`.** Its coordinate-cone checks used absolute `1e-12` slacks. They now use derived bands:
  - The offset >= guard check and the seed `beta >= 0` check use `survival_derivative_guard_violated`. The seed rows are the same active-set bound rows, with guard 0.
  - The `D >= 0` sign check uses the I-spline accumulation roundoff band `accumulation_band(p, 2*sum_j |D_row,j|)`. Each entry is a right-cumulative sum of at most p B-spline derivatives, and `sum_k |dB_k| <= 2*sum_j |D_j|`.
- **Removed.** `FeasibilityTolerance` and `GuardConstraintPolicy` are gone; each family passes its `GuardPolicy` directly. The dead `256*eps` floor is gone too.

## Tests
- `verdict_is_invariant_to_a_change_of_time_unit_3766`: rescaling D, the offsets and the guard by c in {1e-3, 1, 1e3} gives the same verdict.
- `offset_band_matches_the_marginal_slope_domain_predicate_3766`: the builder agrees with the domain predicate on both sides of the band edge.
- The existing `validate_spec` coordinate-cone refusal tests still apply.


SauersML:
BF-89 review (partition-6 sweep): this PR is sound against SPEC, and it still merges cleanly into current main.

- The absolute `1e-24` "immovable row" cutoff is replaced by the structural predicate "every entry is exactly zero". That predicate is unit-free, and the builder and the feasibility diagnostic share it.
- The offset-only rows and the likelihood-domain predicate now read one band, and that band is derived from the active-set solver's certified primal tolerance.
- The `validate_spec` sign check uses `accumulation_band`, whose depth and absolute sum are derived in the comment.
- `FeasibilityTolerance` and `GuardConstraintPolicy` are deleted, and so is every reference to them. On the merged tree nothing still references `survival_derivative_guard_tolerance`.

It is waiting only on focused proof 35503076651, which has been queued since 09:43Z at head fcf5b23bed. It is ready for the MT-1 merge train (odd PR) once that run is green.

One nit for whoever lands it: the two new tests in `time_constraints.rs` have a couple of lines that rustfmt would rewrap (the `short` array and the long `match build_time_derivative_guard_constraints(...)` line). Run `cargo fmt` if a format gate applies.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 215a2e6fd465.

PR #3946, opened 2026-09-20
latent_design: pin the Matérn latent forward frame with FrozenTransform(eye(K))

### LINKED ISSUE #3943: latent_design: Matérn latent forward rank-reduces its centers over the moving latent t, so the design/jet column check fails and the frame drifts
**Where:** `crates/gam-terms/src/basis/latent_design.rs`, the `"matern"` arm of `build_latent_forward_design`.

**Defect.** The arm builds its Matérn basis over the moving latent coordinates `t_mat` with `MaternIdentifiability::None`. `build_matern_basis_seeded` treats every non-`FrozenTransform` spec as a cold build and runs `matern_realized_centers`, which RRQR rank-reduces the supplied centers **against the data it is given**, here the current latent `t`. The jet and the penalty of the same latent design keep all `K` supplied centers. So:

- **n_obs < K.** The kernel rank over `t` is at most `n_obs`, so the cold build keeps fewer than `K` centers. The design comes out narrower than the jet/penalty width and the design/jet column check errors on every call. `value_and_grad` in `crates/gam-models/src/latent_outer.rs` maps that error to `(+inf, None)`: the latent Matérn objective is silently infinite, with no error surfaced to the caller.
- **n_obs >= K.** The surviving center subset, and with it the coefficient frame, depends on the current `t`. The frame can drift between outer evaluations, so the objective is not a smooth function of `t`. This is the same failure mode the Duchon arm was fixed for (#2833).

**Fix.** Pin the frame with `MaternIdentifiability::FrozenTransform { transform: Array2::eye(K) }`, exactly as the Duchon arm does. A frozen build keeps the pinned centers verbatim (see the #1090 comment in `build_matern_basis_seeded`). Also add a regression test with `n = 3 < K = 5` that checks the design width, the jet shape and a finite-difference check of the jet.

Branch: `claude/fix-latent-matern-frozen-frame`.

SauersML:
The Matérn latent forward design built its basis with
MaternIdentifiability::None. The cold build then rank-reduced the supplied
centers over the moving latent t, while the jet and the penalty kept all K
centers. With n_obs < K the column counts always disagreed and
value_and_grad returned (+inf, None). Otherwise the frame could drift
between evaluations. Pin the frame with FrozenTransform(eye(K)), as the
Duchon arm does (#2833), and add an n < K regression test that checks the
design, the jet and a finite difference.

Fixes #3943

Validation: focused-rust-proof.yml on this branch (key HN-26-latent-1).

---


SauersML:

---

PR #3947, opened 2026-09-20
tests/python smooth_significance: stop requiring REML stalls; gate null p-values two-sided

### LINKED ISSUE #3896: tests/python smooth_significance: null test REQUIRES REML stalls (stalls > 0) and gates FPR one-sided only
**Where**

- `tests/bug_hunt_smooth_significance_ref_df_tracks_edf_not_basis_dim_test.py:150-190` (`test_nonconverged_flat_fit_is_not_flagged_significant`)
- same file `:79`, `:93-95`, `:118-119` (the `converged` filter in `_sweep` and its two consumers)
- `tests/bug_hunt_smooth_significance_ref_df_floor_and_null_fpr_test.py:101` (`fpr <= 0.15`) and the module docstring's claim about the power control

**The code**

```python
        if int(model.outer_iterations) >= 200:  # the stall signature
            stalls += 1
...
    assert stalls > 0, (
        "setup: expected some fits to hit the flat-valley stall; none did — "
        "the guard is untested by this sample"
    )
    assert fpr <= 0.15, ...
```

and in `_sweep`:

```python
                "converged": int(model.outer_iterations) < 200,
...
        if not r["converged"]:
            continue  # non-converged fits are deliberately referenced at ~K
```

**Why it is wrong**

1. The test REQUIRES the REML outer optimizer to fail on well-posed pure-noise `y ~ s(x)` fits (N = 300). A solver that converges on every seed turns it red; the recorded python-contracts run 30600186060 (`bench/gha_results/python-contracts-30600186060/py1512_junit.xml`) shows exactly that failure: `setup: expected some fits to hit the flat-valley stall; none did`. SPEC.md line 21: "In general, do not paper over solver issues." A stall on these fits is a solver defect (#1762 class), not a fixture requirement.
2. `_sweep` silently drops non-converged fits from the ref_df band and spread checks, so a stall is excluded instead of reported. "Converged" is also an iteration-count proxy (`outer_iterations < 200`) when the fit carries its own certificate, `model.convergence["certified"]`.
3. Both null tests gate the rejection rate one-sided only (`fpr <= 0.15` at 40 and 60 seeds). A conservative test (p-values pushed toward 1, e.g. chi-square on the full basis dimension, which is the very fallback this file describes) passes. The floor file's docstring says the power control "guards against a fix that simply inflates every p-value", but a strong signal at p < 1e-3 stays tiny under almost any inflation, so it cannot. Conservative p-values are a bug.

`bench/pvalue_calibration/pv-lr-refit/README.md` measures what the right behaviour is: 500 of 500 fits converged, null p-values U(0, 1) by two-sided KS, no mass at p = 1.

**Principled fix**

- Drop `stalls > 0`. Require every fit in both sweeps to certify (`model.convergence["certified"]`), with no exclusion; a stall is reported as a failure.
- Add a two-sided null calibration gate: U(0, 1) has mean 1/2 and variance 1/12, so the mean of R null p-values must lie within 4·sqrt(1/(12R)) of 1/2 (±0.149 at R = 60, ±0.183 at R = 40). Keep the existing `fpr <= 0.15` upper bar unchanged (a 4-sd binomial bar would be looser).

Related: #3534 (generic one-directional coverage/size gates).

SauersML:
Fixes #3896

`test_nonconverged_flat_fit_is_not_flagged_significant` asserted `stalls > 0`. That is, it required the REML outer optimizer to FAIL (outer_iterations >= 200) on some well-posed pure-noise `y ~ s(x)` fits, and it went red when the solver converged (python-contracts run 30600186060). The sweep helper likewise dropped non-converged fits from the ref_df band checks instead of treating the stall as a defect. SPEC.md: "In general, do not paper over solver issues."

Changes:
- Every fit must certify through its own convergence certificate (`model.convergence["certified"]`) rather than an iteration-count proxy.
- The null p-values are gated two-sided through their mean. U(0, 1) has mean 1/2 and variance 1/12, so the mean of R null p-values must lie within 4·sqrt(1/(12R)) of 1/2.
- The one-sided FPR bar (0.15) is kept as it was. At R = 40-60 its lower side is below zero, so only the mean gate sees a conservative test.
- The same mean gate is added to the 40-seed null-FPR test in the companion file. Its docstring wrongly credited the strong-signal power control with catching inflated p-values.

Validation: python-contracts on this PR.

---


SauersML:
This PR has not been validated in CI yet, and it stays a draft until it is. `python-contracts.yml` and `wheel-nightly.yml` (the only wheel source Validate One accepts) have both been `disabled_manually` since 2026-09-12. PRs opened with this token also trigger no `pull_request` workflows, so no CI path currently runs Python tests.

To validate once one of those workflows is re-enabled, dispatch `python-contracts.yml` on `hn23-pyval-combined` (04059d0e). That branch combines this PR with #3948, #3955 and #3958, so one run covers all four. Then read the four files' results in the `py1512-junit` artifact, because the full suite's overall verdict is red on main regardless.

---

SauersML:
Taken over by RF-1113-3; driving to merge. I reviewed it, and it now sits in a local integration tree on origin/main where it merges cleanly. I'll run the targeted tests there before merging.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 249eae420d78.

PR #3948, opened 2026-09-20
tests/python AFT lognormal: make the survival-curve checks actually run

### LINKED ISSUE #3897: tests/python AFT lognormal: survival-curve correctness checks never run (pred["survival"] on a SurvivalPrediction raises, swallowed by except Exception)
**Where**: `tests/bug_hunt_parametric_aft_lognormal_mle_fails_to_converge_test.py:71-92` (`_predicted_survival`) and its uses at `:113-122` and `:193-200`.

**The code**

```python
    try:
        pred = model.predict(rows, interval=0.9)
    except Exception:
        try:
            pred = model.predict(rows)
        except Exception:
            return None
    for key in ("survival", "surv", "S"):
        try:
            col = pred[key]
        except Exception:
            col = None
        ...
    return None
...
    surv = _predicted_survival(model, grid, {})
    if surv is not None:
        ...
        assert rel < 0.1, ...
```

**Why it is wrong**

`predict` on a survival fit returns `gamfit._survival.SurvivalPrediction`, a `@dataclass(slots=True)` with no `__getitem__`; its curve is read with `survival_at(times)` (as every other survival test does, e.g. `bug_hunt_location_scale_smooth_threshold_predict_drops_log_time_test.py:85`). `pred["survival"]` raises `TypeError`, the `except Exception` turns it into `col = None`, the helper returns `None`, and both model-free correctness checks (the closed-form lognormal-MLE curve, and survival increasing with the covariate) are skipped by `if surv is not None`. The test file's docstring says correctness "is checked model-free via the predicted survival surface"; it never is. The test can fail only if the fit raises.

Even if it ran, the row frame put the grid in the `time` column and compared a rows x times matrix against a length-5 truth by broadcasting, and the bar (relative L2 < 0.1, about 0.05 per point) is underived: the fit IS the closed-form MLE, so the only admissible gap is the posterior-mean vs plug-in difference, O(1/n).

**Principled fix**: predict one covariate row and call `survival_at(grid)` with no exception swallowing (shape and finiteness asserted), run the checks unconditionally, and bound `max |S_fit - S_mle| <= 1/sqrt(n)` (one sampling-sd scale of S; far above the O(1/n) estimand gap).

SauersML:
Fixes #3897

`_predicted_survival` subscripted the prediction (`pred["survival"]`, `"surv"`, `"S"`) inside `except Exception`. A survival fit's `predict` returns a `SurvivalPrediction` dataclass, which has slots and no `__getitem__`, so the subscript raised and the helper always returned None. Both model-free correctness checks (the lognormal-MLE curve and the covariate monotonicity) were then skipped under `if surv is not None`. The test could only fail on a raise.

The helper now predicts one covariate row and reads the curve with `survival_at(times)`, the documented API, with no exception swallowing.

The MLE comparison is now max |S_fit − S_mle| <= 1/sqrt(n). The fit is that MLE, so the only admissible gap is the posterior-mean vs plug-in difference, which is O(1/n). The old relative-L2 < 0.1 bar allowed about 0.05 per point.

Validation: python-contracts on this PR.

---


SauersML:
This PR has not been validated in CI yet, and it stays a draft until it is. `python-contracts.yml` and `wheel-nightly.yml` have both been `disabled_manually` since 2026-09-12, so no CI path currently runs Python tests.

To validate once one of them is re-enabled, dispatch `python-contracts.yml` on `hn23-pyval-combined` (04059d0e), which combines #3947, #3948, #3955 and #3958. Then read this file's results in the `py1512-junit` artifact.

---

SauersML:
Taken over by RF-1113-3; driving to merge. I reviewed it, and it now sits in a local integration tree on origin/main where it merges cleanly. I'll run the targeted tests there before merging.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, a8345bf919d6.

PR #3954, opened 2026-09-20
tests/measure_jet_ctn_range_screen_2754: assert the CTN fit converges (#3888)

### LINKED ISSUE #3888: tests/measure_jet_ctn_range_screen_2754: CTN fit outcome is printed, not asserted, though the gam#2600 refusal it deferred to is closed
**File:** `tests/measure_jet_ctn_range_screen_2754.rs:122-136` (origin/main), test `transformation_normal_entry_reaches_the_measure_jet_range_screen_2754`.

```rust
    // The fit's outcome is deliberately not asserted (see the module docs): what
    // is asserted is that the resolver ran before the design was built. Its
    // result is reported either way so a future reader can see which arm of
    // gam#2600 this fixture is in today.
    match fit_from_formula(
        "w ~ mjs(x1, x2, centers=10, learn_length_scale=false)",
        &ds,
        &config,
    ) {
        Ok(_) => println!("[2754-ctn] the CTN fit converged on this fixture"),
        Err(e) => println!(
            "[2754-ctn] the CTN fit declined (not asserted here): {}",
            ...
```

**Why it is wrong.** The fixture is a log-normal response, the shape the transformation-normal family exists for. The fit was left unasserted only because at the time CTN refused this response inside the inner solve (`physical reduced-face first-order KKT failed`), which is the gam#2600 refusal class. This file and the sibling gate `tests/measure_jet/misc/measure_jet_range_resolver_entry_point_invariance_2754.rs` both deferred the assertion until "gam#2600 lifts". #2600 is closed at its root cause: the old likelihood renormalized each row by the normal mass between two *fitted* endpoints, so the objective had no finite mode, and the untruncated `φ(h)·h′` density that replaced it is convex and coercive. The refusal is no longer an open defect elsewhere to route around, and the print now hides any recurrence.

It also accepts `Ok(_)` of any variant, so a transformation-normal config that returned a Standard fit would pass as "converged".

SPEC.md:21: "In general, do not paper over solver issues." SPEC.md:16: "A failing test should always indicate problematic behavior."

**Fix.** Require `Ok(FitResult::TransformationNormal(_))`. Panic on any other `Ok` variant, and on `Err` with its reason and the screen-record count. Keep the existing screen-record assertion, and update the sibling file's deferral comment to match.

SauersML:
## Problem

`transformation_normal_entry_reaches_the_measure_jet_range_screen_2754` printed the outcome of the CTN fit and never asserted it. It also accepted `Ok(_)` of any variant.

The fixture is a log-normal response, the shape CTN exists for. The assertion had been deferred until gam#2600 lifted, and #2600 is now closed at its root cause.

## Fix

- The test now requires `Ok(FitResult::TransformationNormal(_))`. Any other `Ok` variant panics, and so does an `Err`; both messages include the refusal reason and the screen-record count. The screen-record assertion is unchanged.
- The module docs and the sibling gate's deferral comment (`measure_jet_range_resolver_entry_point_invariance_2754.rs`) no longer describe the fit as deferred to an open refusal class.

This is a test-only change.

Fixes #3888

https://claude.ai/code/session_0187xuFaLvee3fS86ygeJZsN

---
_Generated by [Claude Code](https://claude.ai/code/session_0187xuFaLvee3fS86ygeJZsN)_


SauersML:
Taking over from session_0187xuFaLvee3fS86ygeJZsN (no activity since 2026-09-20T09:39Z). This is a stale-claim sweep, BF-11.

I reviewed the diff against SPEC.md and it is sound. It is test-only: the CTN fit outcome was printed rather than asserted, and it is now asserted to be `FitResult::TransformationNormal`. That matches "A fit object must only ever come from a converged optimization." The deferral to gam#2600 is gone, which is correct because #2600 is closed as completed. There is no XFAIL and no loosened tolerance.

Plan:
1. Merge main in (it merges cleanly).
2. Validate on focused-rust-proof: package gam, integration binary `measure_jet_ctn_range_screen_2754`.
3. Merge when green.

---

Issue #3283, opened 2026-09-19
inference: governor-refused dense covariance bundle leaves smoothing_correction_absence unset


SauersML:
Found while closing #3127 as superseded by #3165 (f598b039f4).

**Problem.** When the memory/size governor refuses the dense covariance bundle, the fit ends up with no smoothing-parameter-corrected covariance. `smoothing_correction_absence` is still `None`. That leaves nothing on the fit to say why the corrected covariance is missing, so downstream consumers can't state the reason:
- `gam-predict/src/lib.rs` (~L400 and ~L1860) matches on `fit.smoothing_correction_absence()`;
- `gam-cli/src/main/model_summary.rs` (~L312 and ~L355) does the same.

In the other code paths, a missing correction always carries a persisted reason. For example, `gam-models/src/fit_orchestration/fit.rs` sets `(smoothing_corrected, smoothing_correction_absence)` together, around L2439.

**Expected.** Every fit whose corrected covariance is `None` records a typed `SmoothingCorrectionAbsence` reason, including the governor-refusal path. Two ways to get there:
- add a variant for the governor refusal and set it where the refusal happens; or
- better, if the refusal is avoidable, don't refuse at all. Compute the correction in a form that doesn't need the dense bundle.

Consumers then report that reason instead of hitting a generic branch.

**Why it matters.** SPEC: "In general, do not paper over solver issues." Silently dropping the correction, with no recorded reason, hides why corrected SEs and p-values are missing. We require calibrated p-values, so whether the correction was applied must always be explicit.

**Reproducer.** Still needed. Fit a model large enough that the governor refuses the dense bundle, then assert that `smoothing_correction_absence` is set whenever `covariance_corrected` is `None`.

SauersML:
**Claim (w2-inner-jeffreys, gam-86 wave 2).** I'm fixing this at the root, not with a typed absence.

**Root cause** (read at main 7fff1617da). `optimizer.rs` forms the smoothing correction only `if beta_covariance_unscaled.is_some()`. When the governor refuses the dense bundle, the factorized branch (#2960) solves the conditional standard errors and publishes nothing else. The comment there says the branch "computes diag(J V_rho J')", but no code does. So `published_covariance_mode()` falls to Conditional, and every default surface silently reports conditional bands.

**Fix** (gating now). The correction is already assembled as its square-root factor `B` (`C = B·Bᵀ`, `p × r`) inside `smoothing_correction_gram`.
[...]

Issue #2296, opened 2026-07-13
Uncertainty honesty: Vp/Vc request silently downgrades to conditional Vb with no provenance at any public boundary


SauersML:
## Summary

Requesting smoothing-corrected coefficient covariance (`Vp`/`Vc`) can silently deliver conditional-on-λ̂ covariance (`Vb`) instead, and no user-facing output records which one was used. Intervals produced this way are narrower than what the user asked for, with no marker.

## Evidence (file:line)

- `crates/gam-predict/src/lib.rs:291-304` — `InferenceCovarianceMode::ConditionalPlusSmoothingPreferred` falls back to `Conditional` when `fit.beta_covariance_corrected()` is `None`. The fallback itself is a reasonable engine policy **because** the selector returns a provenance flag…
- `crates/gam-predict/src/lib.rs:1720` — …`covariance_corrected_used: bool` is carried on the prediction result, but **no consumer outside `gam-predict` reads it**: `grep -rn covariance_corrected_used crates | grep -v gam-predict/src` → zero hits in `gam-cli`, `gam-pyffi`, `gam-report`.
- `crates/gam-cli/src/main/cli_args.rs:395,610-613` — the CLI exposes exactly two values, `conditional` and `corrected` (default `corrected`), and `crates/gam-cli/src/main/model_summary.rs:392` maps `Corrected` → `ConditionalPlusSmoothingPreferred`. There is **no `required` option in the CLI**, so a user who explicitly types `--covariance-mode corrected` can receive `Vb` intervals labeled by nothing.
- `crates/gam-pyffi/src/model/model_ffi.rs:119-141` — Python does expose `"required"`, but the default (`None` → smoothing-preferred) and `"smoothing"` paths likewise surface no provenance in the returned payload.
- Reachability of the fallback: many fit constructors legitimately produce `beta_covariance_corrected: None` — e.g. `crates/gam-models/src/survival/location_scale/spec.rs:669`, `crates/gam-custom-family/src/warm_start.rs:760`, `crates/gam-inference/src/hmc_io.rs:1553`. So `corrected` requests against these fits always downgrade.
- Contrast: the manifold FFI already does this right — `crates/gam-pyffi/src/manifold/manifold_and_posterior_ffi.rs:1633-1637` labels the covariance source (`"corrected"` vs conditional) in its payload.

## SPEC / audit clause

[...]

PR #3955, opened 2026-09-20
tests/python survival_likelihood: accept only the non-Surv() refusal, not any exception

SauersML:
Fixes #3898

`test_survival_likelihood_not_silently_ignored_without_surv_response` caught `except Exception: continue`, so a panic or an integration failure on `time ~ s(x)` passed as "the request was not dropped". The accepted outcome is the #1767 configuration refusal, whose message names `survival_likelihood` and the `Surv(...)` wrapper (pinned in `materialize/tests.rs`). Anything else is now an offender.

"transformation" is added to the modes. The knob defaults to None at every entrance (#2301), so an explicit "transformation" is a survival request that the engine refuses like the others. The comment that excluded it was stale.

Validation: python-contracts on this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MwMvgB7i335nVFQ8vfZ44B

---
_Generated by [Claude Code](https://claude.ai/code/session_01MwMvgB7i335nVFQ8vfZ44B)_

PR #3958, opened 2026-09-20
tests/test_covariance_provenance_2296: gate the summary SE column on the exported matrix

### LINKED ISSUE #3900: tests/test_covariance_provenance_2296: the summary SE/covariance gate compares two copies of one label and skips when the matrix is missing
`tests/test_covariance_provenance_2296.py:283-288` (`test_summary_reports_definition_consistent_se_source`):

```python
kind = summary.covariance_kind
if kind is not None:
    assert kind == se_source, (...)
```

The docstring says this is the #2296 close gate: "the summary's SE column, the exported covariance, and their labels must come from ONE covariance definition". The test cannot fail in either of the ways that matter.

1. **The label comparison is tautological.** In `crates/gam-models/src/inference/saved_summary.rs:905-912`, both `covariance_kind` and `coefficient_se_source` are written from the same `display_uncertainty.definition`. The two strings are equal by construction. A mixed pair would look like corrected SEs next to a conditional matrix, which is exactly what #2296 is about, and would still carry matching labels. The only thing that detects a mix is the numbers, and the test never compares the `std_error` column with the exported matrix.
2. **A missing matrix skips the check.** If `covariance_kind` is `None`, which happens when the covariance export is dropped, the assertion is skipped and the test passes. The fixture has p = 10 dense coefficients, well below the size at which inference stays factorized (#3283). A missing matrix here would be a real export regression, not an expected case.

**Fix.**
- Require `covariance_kind` and `covariance` to be present.
- Keep the label equality check.
- Assert `covariance.shape == (p, p)`.
- Assert that `std_error[i] == sqrt(covariance[i, i])` for every coefficient at rtol 1e-12. This follows `beta_standard_errors*`, which is `sqrt(diag)` of the published matrix (#2955), and a JSON round trip of f64 is exact.

SauersML:
Fixes #3900

`test_summary_reports_definition_consistent_se_source` compared `covariance_kind` with `coefficient_se_source`. `saved_summary.rs` writes both from the same `display_uncertainty.definition`, so they are equal by construction. The test also skipped the check whenever `covariance_kind` was None. As a result, a mixed pair (corrected SEs next to a conditional matrix, which is the #2296 defect) and a dropped covariance export both passed.

The p = 10 dense Gaussian fixture must now export both its matrix and its label. The matrix must be p x p, and every `std_error` must equal sqrt(diag) of it at rtol 1e-12.

Why the new assertions hold on current main:
- The fixture's p = 10 fits the dense-covariance reservation in `gam-solve/src/estimate/optimizer.rs`, so `H⁻¹` is materialised. The factorised branch (#3283/#2960) is the only one that publishes standard errors without a matrix, and it does not apply here.
- `display_coefficient_uncertainty()` in `result_types.rs` returns the corrected pair when `covariance_corrected` exists and the conditional pair otherwise. Each pair's SEs are `standard_errors_of_published_covariance(cov)`, that is `sqrt(diag(cov))` of the same stored matrix (#2955, pinned by `assert_standard_errors_are_the_published_diagonal`).
- Both values cross JSON through serde's round-trip-exact f64 formatting, so the Python `sqrt(diag)` matches bit for bit up to a few ulps. rtol 1e-12 is far above that.

Validation: the Python Contracts workflow has not run since 2026-09-12 (no pull_request runs trigger on this repo's pushes), so there is no CI lane for this test-only change. The argument above is the validation. Merged `origin/main` in at 5d90d4dff1 (clean).


SauersML:
This PR has not been validated in CI yet, and it stays a draft until it is. `python-contracts.yml` and `wheel-nightly.yml` have both been `disabled_manually` since 2026-09-12, so no CI path currently runs Python tests.

To validate once one of them is re-enabled, dispatch `python-contracts.yml` on `hn23-pyval-combined` (04059d0e), which combines #3947, #3948, #3955 and #3958. Then read this file's results in the `py1512-junit` artifact.

---

PR #3959, opened 2026-09-20
position_basis: build the Duchon penalty in the design's frame

### LINKED ISSUE #3949: position_basis: Duchon position penalty ignores basis_order and is built in the centers' radial chart, not the design's (wrong penalty; width mismatch when n < K)
**Where:** `crates/gam-terms/src/basis/position_basis.rs`, the Duchon arm of `position_penalty`.

**Defect.** The position **design** for a Duchon position basis is the order-`m` Duchon basis (`basis_order`) built over the sample positions `t`. Its data-metric radial chart is solved over `t` (#1355). The position **penalty** was built separately:

- as the cubic (order-2) function-norm Gram over the centers,
- with a fixed `Linear` null space whatever `basis_order` requested,
- with its radial chart solved over the **centers**, not over `t`.

Consequences:
1. The penalty lives in a different coefficient frame from the design. The fitted smoothing penalty is not the Gram of the design's basis functions.
2. For `basis_order != 2` the penalty has the wrong null space and the wrong roughness order.
3. When `n < K`, the design's chart truncates to fewer columns than the centers' chart. The penalty width then disagrees with the design width, and assembly fails with a shape mismatch.

**Fix.** Add `duchon_position_penalty(t, locations, order, period)`. It returns the `Primary` block of the **same** Duchon build the design uses: order `m`, over `t`, the explicit period for a periodic basis, and the strict operator policy for an open basis so the `n x p` design is never materialized. `position_penalty` takes `t` as a new leading parameter. With the chart frozen, this Gram does not depend on `t`, which matches the frozen-chart backward pass. A regression test checks that the penalty is the Gram of the design's own frame.

Branch: `claude/fix-position-duchon-penalty-frame`.

SauersML:
The Duchon position penalty was the cubic function-norm Gram built over the
centers: a fixed Linear null space whatever basis_order was requested, and the
data-metric radial chart solved over the centers. The position design is the
order-m Duchon basis built over the sample positions t, whose chart is solved
over t (#1355). So the fitted penalty was expressed in a different coefficient
frame from the design, had the wrong null space for m != 2, and had the wrong
width whenever n < K truncated the design's chart.

Build the penalty as the Primary block of the same Duchon build the design
uses: order m, over t, the explicit period for a periodic basis, and the
strict operator policy for an open basis so the n x p design is never
materialized. With the chart frozen, this Gram does not depend on t, which
matches the frozen-chart backward pass.

Fixes #3949

Validation: focused-rust-proof.yml on this branch (key HN-26-position-1).

---


SauersML:

---

PR #3965, opened 2026-09-20
custom-family warm_start: propagate the reduced-space EDF trace failure instead of logging it

### LINKED ISSUE #3961: custom-family warm_start: reduced_blockwise_edf swallows the reduced-space trace failure into None, so the fit fails with an unrelated gauge refusal
## What

`reduced_blockwise_edf` in `crates/gam-custom-family/src/warm_start.rs` turns a failed reduced-space trace into `None`. The only trace of the failure is a debug log:

```rust
    match custom_family_blockwise_edf(geom.penalized_hessian.as_array(), &canonical.reduced_specs, &lambdas.view()) {
        Ok(triple) => Some(triple),
        Err(err) => {
            log::debug!("[custom-family inference] reduced-space effective degrees of freedom unavailable: {err}");
            None
        }
    }
```

Its only caller is `assemble_custom_family_fit_result` (fit.rs:480):

```rust
let precomputed_edf = precomputed_edf
    .or_else(|| reduced_blockwise_edf(geometry.as_ref(), canonical, &lambdas));
```

The caller then lifts the geometry to raw (`lift_fit_geometry_to_raw`, which composes the canonical gauge into `coefficient_gauge`) and hands it to `blockwise_fit_from_parts`.

## Why it is wrong

The swallowed error goes one of two ways, depending on the gauge:

- **Non-identity canonical gauge.** `blockwise_fit_from_parts` refuses at warm_start.rs:999 with "non-identity active geometry requires its reduced-coordinate EDF; raw penalties cannot be paired with an active-coordinate Hessian". The real reason is lost. That reason is whatever `custom_family_blockwise_edf` reported: the exact factorization failed, a penalty's rank factorization failed, the trace band failed, or the EDF was non-finite. The user sees an internal-looking refusal instead.
- **Identity gauge.** The same Hessian is factored a second time on the raw fallback only to fail the same way.

Neither route can succeed where the reduced trace failed, so returning `None` only rewrites or delays the error. This is the "silently swallowed errors" class.

## Fix

- `reduced_blockwise_edf` returns `Result<Option<..>, CustomFamilyError>`. `Ok(None)` still means there is no reduced geometry.
- The caller propagates the error with `?`.

Successful fits are unchanged. Only a failing fit gets its own reason back.

SauersML:
Fixes #3961

`reduced_blockwise_edf` turned a failed reduced-space trace into `None` and logged it at debug level.

- **Non-identity canonical gauge.** The lifted geometry then refused the raw fallback, so the fit failed with "non-identity active geometry requires its reduced-coordinate EDF" and the trace's own reason was lost.
- **Identity gauge.** The same Hessian was factored a second time, only to fail again.

The function now returns `Result<Option<..>>`. `Ok(None)` still means there is no reduced geometry, and a trace failure propagates out of `assemble_custom_family_fit_result`. Successful fits are unchanged.

Proof: focused-rust-proof run 35501042415 (gam-custom-family, key HN-19-edf-1).

---


SauersML:

---

PR #3704, opened 2026-09-20
parameter_decomposition: bound binary64 norm rounding through the subnormal range

SauersML:
Fixes #3702

## What was wrong

**`rms_norm_band` was relative-only.** It claimed `γ_(d+6)` relative error for `fl(w fl(x ν̂))`, which holds only while every intermediate result stays normal. A square, the mean or a product that rounds into the subnormal range moves by an absolute `2^-1075`, and the gain can lift that into a normal output. On the row `[1000, 1e-310]` with gain `[1, 1e10]`, the error is about 8.8e-315 but the band is 1.3e-318.

**The LayerNorm arm had the same flaw.** Its own-rounding term and centring error in `NativeNorm::apply_with_band` were also purely relative.

**`BINARY64_UNDERFLOW` evaluated to 0.0.** In receipts it is documented as `2^-1075`, which is not a binary64 number, so the constant rounds to zero.

## What changed

**block.rs**
- `rms_norm_band(epsilon, gain, inputs, normalized)` is now a `Result`.
  - It adds the derived underflow terms: `ρ = ω/(2(1−ω))` through `ν̂`, with `ω = 2·2^-1074 / (X₀(1−γ))` and `X₀ = max(ε, max x²/d)`.
  - Each entry also gets the absolute term `a = |w|·2^-1074 + 2^-1074`.
  - The band is `λ(|ŷ|+a)/(1−λ) + a`, with `λ = γ + ρ + γρ`, computed with directed rounding.
  - A zero-ε row with `ω ≥ 1` is refused with the typed error `NormBandUnbounded`.
- The LayerNorm own term uses the same `λ` and `a`.
  - The centring error adds `2^-1074`.
  - Its norm is scaled by the largest entry, so no square underflows.
- `up` and `down` are shared, and there is a new `SUBNORMAL_SPACING = 2^-1074` constant.

**receipts.rs**
- The binary32 band adds `SUBNORMAL_SPACING` instead of a zero.
- `rms_norm_stage` passes its inputs and gain to the band, and returns a new `ReceiptRefusal::NormBand` when a row is refused.

## Test

`a_normalization_band_covers_a_row_whose_products_underflow`:
- Checks that the band covers the double-double error on the underflowing row.
- Positive control: the old relative-only band does not cover that error.
- Checks that a zero-ε row whose mean square is the smallest subnormal is refused, even though the owner normalizes it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #3967, opened 2026-09-20
parameter_decomposition::attention: carry binary64 underflow in the per-head norm radius

### LINKED ISSUE #3960: parameter_decomposition::attention: head_rms_norm_with_radius has a relative-only evaluation band that misses binary64 underflow, and a zero floor returns inf/NaN radii
## Where

`crates/gam-sae/src/parameter_decomposition/attention.rs` on `origin/main` (61d8ce24e), `head_rms_norm_with_radius` (line 566). Its callers are `NativeAttention::normalize_queries` / `normalize_keys` (the Qwen3 `q_norm` / `k_norm`) and `block_program.rs:1391`.

```rust
583    let evaluation = accumulation_growth(head_dim + 6);
...
593        let computed = x.iter().map(|value| value * value).sum::<f64>() / head_width + epsilon;
...
600        let floor = (computed * (1.0 - square_growth) - spread / (1.0 - spread_growth) - floor_growth * (computed + spread))
601            .max(epsilon);
602        let inverse_root = floor.sqrt().recip();
...
605            normalized_radius[[row, c]] = (gain[c].abs() * (r[c] * inverse_root + x[c].abs() * slope * spread)
606                + evaluation * normalized[[row, c]].abs())
607                / dominance;
```

The doc contract (lines 547-565) says the radius bounds `|y_c(x) − fl(y_c(x̂))|` for every `x` in the box. It also says the evaluation's own rounding is `γ_{d+6} |fl(y_c)|`.

## Why it is wrong

**1. The evaluation term is relative-only, so it misses binary64 underflow.** The square, the mean and the product `x·ν` that round into the subnormal range move by an absolute `2^-1075`, not by a relative `u`. This is the same defect as #3702 (`block::rms_norm_band`), but in the per-head copy of the band. `require_epsilon` admits `ε = 0`, and Qwen3's `q_norm` / `k_norm` are applied to per-head rows. So a head row whose mean square is subnormal is normalized back to unit scale, and its absolute underflow becomes a relative error of order one.

Demo: one head row `[1e-160, 0]`, `ε = 0`, gain `[1, 1]`, radius `[0, 0]` (an exact input row).

| quantity | value |
|---|---|
| computed `ŝ = mean x²` | `5e-321` (subnormal; exact `5e-321` is not representable) |
| computed `ŷ_0` | `1.414221434529443` |
| exact `y_0` | `√2 = 1.4142135623730951` |
| actual error | `7.9e-6` |
| returned radius | `1.26e-15` |

The radius misses its own rounding by about ten orders of magnitude.

**2. The floor can reach 0 and produce `inf` / `NaN`.** The floor is `max(ŝ(1 − γ) − Δs/(1 − γ) − …, ε)`. With `ε = 0` and a radius that covers the row's own scale, it reaches `0`. Then `inverse_root = 0.sqrt().recip() = inf`, and `slope · spread` is `inf · finite`, or `0 · inf = NaN` where `x_c = 0`.

Demo: row `[1, 0]`, radius `[2, 0]`, `ε = 0`. The returned radius is `[inf, NaN]`. A `NaN` radius is not a bound: every downstream comparison against it is false, so the enclosure silently passes nothing and fails nothing.

(A box that contains `x = 0` really has no finite derivative bound. But the output `|y_c| ≤ |w_c| √d` is bounded on the whole box, so a finite radius exists.)

## Principled fix

Route the per-head band through the single-source binary64 norm band that #3702 / PR #3704 adds to `block.rs`, rather than keep a second relative-only copy here:

- **Evaluation term.** Replace `evaluation * |ŷ|` with `block::rms_norm_band(epsilon, gain, inputs, normalized)` from #3704. Its `λ = γ + ρ + γρ` carries the derived underflow `ω = 2·2^-1074 / (X₀(1 − γ))`. It refuses, with `NormBandUnbounded`, a zero-epsilon row whose mean square is not resolved above the underflow. Such a row keeps the range bound below, which is always valid and finite, instead of a derived band.
- **Floor.** Compute it with directed rounding (`down`), as a lower bound on the exact `mean x̂² + ε` that includes the underflow of the squares (`− d·2^-1075/d` absolute). Compute `spread` with `up`.
- **Radius arithmetic.** Round the radius itself outward with `up`, instead of dividing by `1 − γ_{d+12}`, which is relative-only.
- **Floor at or below 0.** Take the radius to be the range bound `|w_c| √d + |ŷ_c|`, which always encloses `|y_c(x) − ŷ_c|`. In general the radius is the smaller of the derived bound and the range bound, and an overflowed or `NaN` derived bound falls back to the range bound, so a radius is never `inf` or `NaN`.
- **Tests.**
  - A regression test for the `[1e-160, 0]` row: the radius encloses `|ŷ − √2|`. With `ε = 0` the norm is scale invariant, so the check runs in double-double at `[1, 0]`.
  - A test that the `[1, 0]` / `[2, 0]` box returns a finite radius that encloses the exact norm at samples of both signs.
  - A control test that a normal row keeps its current radius to within a few ulps.

This depends on #3704 for `rms_norm_band` and `up` / `down`, which are `pub` / `pub(super)` there, so `attention.rs` reads the same definitions.

SauersML:
Fixes #3960

Stacked on #3704 (base `hn3-norm-band-underflow-3702`). It reuses that PR's single-source `block::rms_norm_band` and the directed-rounding `up` / `down`. Once #3704 merges, this PR retargets to `main`.

`head_rms_norm_with_radius` bounded the evaluation's own rounding by the relative-only `γ_(d+6) |ŷ|`. That fails in two ways:
- **Underflow.** A zero-epsilon head row whose squares round into the subnormal range, e.g. `(1e-160, 0)`, is normalized back to unit scale. Its absolute underflow becomes a relative error of `5.6e-6`, against a radius of `1.3e-15`.
- **Zero floor.** A radius box that reaches `x = 0` at `ε = 0` drove the floor to 0 and returned an `inf` / `NaN` radius.

**Fix:** the radius is now the smaller of two bounds, both rounded outward.
- **Derived bound.** The head row's `block::rms_norm_band`, plus the box term `|w_c| (r_c s_lo^(-1/2) + |x̂_c| ½ s_lo^(-3/2) Δs)`, where `s_lo` and `Δs` are formed with `down` / `up`. The band is the same single-source band as `MaskedNorm::Rms`'s program, and it carries underflow.
- **Range bound.** `|w_c| √d + |ŷ_c|`, which is valid on any box. It alone is the radius where `s_lo = 0`, where the band refuses the row, or where the derived bound overflows.

**Tests:**
- The `(1e-160, 0)` row's radius encloses the exact norm, and a control shows the old relative band misses it.
- A box that reaches zero gets a finite radius that encloses samples of both signs.
- A normal-scale row keeps its relative band.

Validation: focused-rust-proof run 35501899335 on `a4419d94`.

---


SauersML:

---

PR #3970, opened 2026-09-20
inference/residual_factor: whiten through chol(Sigma), drop eigen-repaired precision

### LINKED ISSUE #3952: inference/residual_factor: lower_cholesky_psd covers Woodbury cancellation with a 1e-10 eigen-repair fallback
## Where (origin/main 5994ad54c4)

`crates/gam-solve/src/inference/residual_factor.rs`

- `lower_cholesky_psd` (lines 776-805) builds the `U_n` factor for `RowMetric`.
- It is called from `StructuredResidualModel::row_metric` (line 406) on the Woodbury precision that `row_precision` forms (lines 672-722).
- It is also called from `damped_row_factor` (lines 657-658), after `invert_spd(&sigma)?`.

```rust
fn lower_cholesky_psd(a: &Array2<f64>) -> Result<Array2<f64>, String> {
    if let Ok(chol) = a.cholesky(Side::Lower) {
        return Ok(chol.lower_triangular());
    }
    // Eigen-repair: clamp eigenvalues to a small positive floor, ...
    let (evals, evecs) = symmetric_eig_ascending(a)?;
    let max_ev = evals.iter().copied().fold(0.0_f64, f64::max).max(1.0);
    let floor = 1e-10 * max_ev;
    ...
    repaired.cholesky(Side::Lower) ...
}
```

## Why it is wrong

Each row's residual covariance is `Sigma_n = c_n * Lambda Lambda^T + D`. It is SPD by construction because `D > 0`, and `damped_row_factor` blends it with another SPD matrix. A correct whitening factor therefore always exists and never needs repair.

The code does not factor `Sigma_n`. It first forms the precision:

- `row_metric` uses Woodbury, `D^{-1} - B M_n^{-1} B^T`. That is a difference of two nearly equal PSD matrices whenever `c_n * Lambda^T D^{-1} Lambda` is large, which happens with a strong factor.
- `damped_row_factor` uses an explicit inverse (`invert_spd`).

When the cancellation leaves the precision numerically indefinite, the plain Cholesky fails. The function then silently eigen-clamps with `floor = 1e-10 * max(max_ev, 1.0)` and returns the factor of a different matrix. Every downstream whitened quadratic form, log-det and REML/LAML term then uses a precision that is not `Sigma_n^{-1}`, and nothing reports it.

Problems with this fallback:

- It patches a numerical defect in the algorithm at its output. SPEC.md: "In general, do not paper over solver issues."
- `1e-10` and `.max(1.0)` are arbitrary magic constants. SPEC.md: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code."
- Because of `.max(1.0)`, the floor is absolute rather than relative whenever the precision's spectrum lies below 1, for example when `Sigma` is large-scale. The repair is therefore not scale-equivariant: rescaling the response changes which eigenvalues get clamped.

## Principled fix

`RowMetric` accepts any `U_n` with `U_n U_n^T = Sigma_n^{-1}`. Nothing relies on `U_n` being triangular:

- `whiten_residual_row` returns `U_n^T r`.
- `RowMetric::whitened_structured` only checks that `U U^T` is PSD.
- The gam-sae consumers (`apply_row_metric`) only use `U U^T v`.

So factor the SPD covariance directly: `Sigma_n = L_n L_n^T`, then take `U_n = L_n^{-T}` (one triangular inversion). This gives `U_n U_n^T = Sigma_n^{-1}` exactly and `U_n^T r = L_n^{-1} r`. The precision is never assembled, so the cancellation cannot occur. If `Sigma_n` is ever not SPD, that is a real defect and is returned as an `Err` instead of being clamped.

The fix also deletes `row_precision`, `invert_spd`, `symmetric_eig_ascending` and `lower_cholesky_psd`, which have no other callers.

SauersML:
Fixes #3952

`StructuredResidualModel::row_metric` built `Sigma_n^{-1}` through Woodbury, `D^{-1} - B M_n^{-1} B^T`, which is prone to cancellation. `row_metric_damped` built it through an explicit `invert_spd`. Both then took the lower Cholesky of that precision, and when the Cholesky failed they silently eigen-clamped it at `1e-10 * max(max_ev, 1)` (`lower_cholesky_psd`).

`Sigma_n = c_n * Lambda Lambda^T + D`, and its damped blend, are SPD by construction. This PR factors the covariance directly: `Sigma_n = L L^T`, then `U_n = L^{-T}`, computed column by column by forward substitution. That gives `U U^T = Sigma^{-1}` exactly and `U^T r = L^{-1} r`. The precision is never assembled, and a covariance that does not factor is returned as an `Err` instead of being repaired.

Changes:
- `row_metric` hoists the Gram `Lambda Lambda^T` (the same `outer_product` helper the damped path already uses). Each row builds `c_n * Gram + D` and calls a new `covariance_whitening_factor`.
- `damped_row_factor` calls the same helper on its symmetrized blend.
- Deleted `row_precision`, `invert_spd`, `symmetric_eig_ascending`, `lower_cholesky_psd` and the `FaerEigh` import. They had no other callers.

Consumers depend only on `U U^T`, so the factor no longer needs to be lower-triangular:
- `RowMetric::whiten_residual_row` computes `U^T r`.
- `whitened_structured` checks that `U U^T` is PSD.
- gam-sae `apply_row_metric` computes `U (U^T v)`.

The damped endpoints still delegate to `row_metric`, so the byte-identical endpoint test still applies.

---


SauersML:
Merge train MT-0: merged origin/main into this branch (merge commit 3704fc30, no rebase or force-push). The only conflict was `docs/source-removal-changes.json`; both sides' entries were kept, and the result validates with `python3 -m json.tool`. The code diff against main is byte-identical to before (`residual_factor.rs`, +77/−192). Queued for the next local integration train, with focused filter `cargo test -p gam-solve --lib residual_factor`.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 3353a993199b.

Issue #3641, opened 2026-09-20
gam-custom-family: process-global assembled-operator cache keys an Operator-source Hessian on its diagonal only, so a cross-fit hit can reuse another fit's operator


SauersML:
Split out of #3564 (item 4). PR #3628 deleted the tautological gam#1395 guard. This issue covers the one real channel that guard could ever have caught.

## Where (origin/main)
- `crates/gam-custom-family/src/assembly.rs`:
  - `assembled_operator_cache()` is a process-global `static OnceLock<Mutex<AssembledOperatorCache>>`.
  - `assembled_operator_fingerprint` is its key.
  - The `JointHessianSource::Operator { diagonal, .. }` arm.

## What is wrong
The cache doc says a hit is "bit-identical to a fresh build" because "the reuse condition is exact byte-equality of the build inputs". That holds for a `Dense` source, whose whole matrix is hashed.

It does not hold for an `Operator` source. That key hashes only the operator's diagonal, plus ρ, β̂, the penalties and the knobs. The off-diagonal curvature, and so the family/data identity, is assumed to be "pinned by (ρ, β̂) through the family workspace". But the cache is shared by every fit in the process. A different fit whose (ρ, β̂, diagonal, penalties) bits coincide would be handed a stale operator, and its `0.5·log|H|` and trace kernel would silently belong to another matrix. Examples of such a fit:
- a parallel test;
- a multistart lane of a different model;
- a second model in one Python session.

Such a collision is unlikely but not excluded by construction. SPEC: "In general, do not paper over solver issues."

## Principled fix (pick one)
1. **Key on the assembled matrix.** Both source kinds materialize `j_for_traces` before `BlockCoupledOperator::from_joint_hessian_with_mode`. Hash the symmetrized, fully assembled `j` and cache the factorization on that key. Every hit is then exact for both kinds.
   - Cost: an Operator-source hit pays the Θ(n·p²) materialization it currently skips.
   - The cache's stated win (the ≈14–19 s dense spectral factorization at biobank scale) is kept.
   - This needs a timing check on a large Operator-source fit (BMS tiled row-primary) before choosing it.
2. **Scope the cache to one fit.** The outer driver owns the cache and passes it into `joint_outer_evaluate`, instead of a process-global `static`. Cross-fit collisions become impossible, and the within-fit (ρ, β̂) ⇒ operator argument becomes valid. This threads a handle through the `evaluate_custom_family_joint_hyper*` entry points.

## Test
[...]

Issue #2922, opened 2026-09-16
Bernoulli marginal-slope fit on 50 rows does not finish in hours: the exact-joint outer Hessian recomputes fourth-order row jets at every outer evaluation (0.2.1)


SauersML:
## What happens

A Bernoulli marginal-slope fit (`BernoulliMarginalSlopeFitRequest` through `fit_model`, gam 0.2.1) on 50 rows with one Duchon smooth and a linear covariate does not finish in hours. Every stack sample sits in the κ exact-joint outer optimization building the outer Hessian's IFT correction from per-row fourth-order contracted jets, recomputed per row, per empirical latent node and per hyperparameter pair at every outer evaluation. Runtimes measured through gnomon (all stopped by hand, none finished):

| Rows | PCs | Options | Threads | Wall when stopped |
|---|---|---|---|---|
| 50 | 0 | defaults | 2 | 6,700 s |
| 100 | 0 | defaults | 2 | 6,700 s |
| 200 | 0 | defaults | 4 | 15,513 s |
| 200 | 2 | defaults | 4 | 15,513 s |
| 2,000 | 0 | defaults | 2 | 35,348 s |
| 2,000 | 2 | defaults | 4 | about 34,000 s |
| 2,000 | 0 and 2 | `--reml-max-iterations 5 --reml-convergence-tolerance 1e-2` | 2 | about 20,900 s |

Even with 5 outer iterations the n = 2,000 fits ran past 5.8 h; at n = 200, `--reml-max-iterations 1` on one thread ran past 600 s.

## Where the time goes

`gdb -batch -ex "thread apply all bt"`: 5 samples on the 50-row run, 3 on each 200-row run. In every sample, the working threads sit on this path (outermost first):

```
gam::solver::workflow::fit_model
gam::families::bernoulli_marginal_slope::fit_bernoulli_marginal_slope_terms::{closure#11}
gam::terms::smooth::optimize_spatial_length_scale_exact_joint
<gam::solver::outer_strategy::OuterProblem>::run  /  run_outer_with_plan
gam::solver::estimate::reml::unified::reml_laml_evaluate
gam::solver::estimate::reml::unified::compute_outer_hessian
gam::solver::estimate::reml::unified::compute_ift_correction_trace
<OwnedJointDerivProvider as HessianDerivativeProvider>::hessian_derivative_correction_result
gam::families::custom_family::exact_newton_d2h_closure_owned
[...]

PR #3971, opened 2026-09-20
Carry finite-law calibration in log units; rigid fifth order from the anchor table (#3639)

### LINKED ISSUE #3639: Bernoulli marginal-slope rows past |q| ≈ 37.5 read linear densities that underflow (flex F_a consumers, rigid fifth order, measure sensitivity, cells)
## Bernoulli marginal-slope rows past |q| ≈ 37.5 read linear densities that underflow

**Context.** #3199 put the rigid route on latent_anchor's log-space solve, and 6d2ff7f55e (#3333, #3216) did the same for the flex route. Both solve the row intercept on the marginal's smaller tail in log units. Downstream of the root, several consumers still work in linear probability units, and they break once `φ(η)` underflows: past |η| ≈ 37.5, where `φ` goes subnormal and then 0 at ≈ 38.6.

On main 81623168aa:
- **Flex consumers of `F_a`.**
  - The order-two lowering, the grid jet program and the predict mirrors read `F_a = E[φ(η)·η_a]` in linear units (`unary_derivatives_normal_cdf` stacks, cell_moment_assembly.rs ~883 and ~5288).
  - `CalibrationTail::log_residual` refuses when that density is 0 ("calibration tail has no resolvable log residual").
  - A finite-law flex row therefore refuses at |q| ≳ 37.5, even though the root itself is representable in log units.
- **Rigid fifth order.** `empirical_rigid_row_fifth_full` (cell_moment_assembly.rs ~643-680) sums `weight · composed` remainders from linear `unary_derivatives_normal_cdf(η)`. Orders one to four read the anchor's Taylor table.
- **Measure sensitivity.** `empirical_measure_sensitivity.rs:624` builds linear `Φ` stacks at each node.
- **Standard-normal cells.** The cell values and moments are f64 integrals in linear units, so the cell route cannot represent a tail below the subnormals.

**Scope.** Only rows whose marginal tail is below ~1e-300. Short of that, linear `φ` keeps full relative accuracy, and the implicit derivatives are ratios. No realistic fit reaches this regime. The issue is filed so that it is recorded, not because it is urgent.

**Fix direction** (drafted on the pre-6d2ff7f55e tree for #2978, not landed):
- Carry the calibration in units of `N = Φ(−|q|)`: `density(η) = (φ(q)/N)·exp((q−η)(q+η)/2)`, and `F_a/N = G′·e^{±G}` from the log residual.
- Read order five from `AnchorTaylor::partial(i, j) = i!·j!·c[i][j]`.
- Normalize the measure sensitivity's node densities the same way.
- Integrate the cells in those units, i.e. log-space cubic and affine cells plus the GPU mirror. Two things are needed there: a finite affine cell deep in a tail is a difference of orthants far larger than itself, and with the apex-tree Φ₂ it still returns 0 with an infinite bound on 4 of 32 test cells.

Found in the #2978 work (bms-a). #2978 itself is resolved for realistic tails by #3199 and 6d2ff7f55e.

SauersML:
Fixes #3639

Items 1–3 are done here. The remaining work from item 4 (cells route, GPU mirror, apex-tree Φ₂) is tracked in #4504.

## Problem
Past |q| ≈ 37.5 every node density φ(η_k) underflows to zero in double precision, and so does the marginal density φ(q). A finite-law (empirical measure) row therefore got a linear calibration Jacobian of exactly F_a = Σ w_k φ(η_k)·η_a = 0. The results by route:

- The compiled flex row program refused the row.
- The order-2 lowering, the prediction gradients and the measure-sensitivity channels divided by zero.
- The rigid fifth order had the same fault. It was fixed in the first commit on this branch (6d490f6710) by reading the anchor's density-normalized Taylor table.

The second commit moves every finite-law flex consumer to log units.

### 1. Calibration unit N = Φ(−|q|) (log-domain ratios)
The row constraint is F(a) = Σ_k w_k Φ(η_k(a)) − Φ(q) = 0, with Σ w = 1. Every implicit derivative a_u = −F_u/F_a, and every higher-order implicit-function derivative, is homogeneous of degree 0 in F. So dividing F by the positive constant N = Φ(−|q|) changes neither the root nor any derivative. Written in that unit:

- **Survival side (q ≥ 0):** F/N = 1 − Σ w Φ(−η)/N. Node value −Φ(−η)/N, marginal value −1.
- **Complement side (q < 0):** F/N = Σ w Φ(η)/N − 1. Node value +Φ(η)/N, marginal value +1.
- **Derivative channels:** φ(η)/N · (1, −η, η² − 1, −(η³ − 3η)), with
  `φ(η)/N = ρ_q · exp(½(q − η)(q + η))`, where `ρ_q = φ(q)/N` is the Mills ratio. It comes from the same erfcx evaluation as log N.

Error bounds:

- **φ(η)/N:** the only extra rounding beyond ρ_q's own is in the exponent, so the relative error is ε·(1 + |q² − η²|/2).
  - It is exact at η = q, where the value is ρ_q itself.
  - It never pays the ε·η²/2 that `exp(log φ(η) − log N)` would. At a root that could be ≈ 1.7e-13 at |q| = 39, and it grows without bound in η.
- **Node value Φ(∓η)/N = exp(log Φ(∓η) − log N):** absolute error ε·(|log Φ(∓η)| + |log N|) in the exponent. At |q| = 39, log N ≈ −765.08, which gives ≈ 3.4e-13 relative. That is below the root's own residual resolution there (≈ 1.4e-12, `anchor_residual_rounding`).
- **Boundedness at the root:** each term of T = Σ w Φ(∓η) is at most T = N. So w_k φ(η_k)/N = ρ(∓η_k)·w_kΦ(∓η_k)/N ≤ ρ(∓η_k) ≈ |η_k| + 1. No channel can overflow, and none underflows unless its node's own tail share does, in which case its contribution is below ε relative anyway.
- **Consistency with the solve:** `CalibrationUnit::new(q).log_unit()` is bitwise `smaller_tail_log_target(q)`, the log target the row root is solved against.

### 2. Streaming log-sum-exp tail (`sum_calibration_tail`)
T = Σ w_k Φ(∓η_k) is first summed linearly. That pass is bitwise identical to the old sums, so every row above the floor is unchanged to the bit.

If the linear tail is non-finite or below `LINEAR_RESIDUAL_FLOOR` (1e-250), the same visitor runs once more in log space. This is the rule and floor the declared-law anchor already uses.

- Each term is t_k = ln w_k + log Φ(∓η_k), held against the running maximum m.
- It adds e_k = exp(t_k − m) to the tail, and e_k·ρ_k·η_a and e_k·ρ_k·(η_aa − η·η_a²) to the slope sums.
- ρ_k = φ(η_k)/Φ(∓η_k) comes from the same erfcx evaluation as log Φ.
- A new maximum rescales all three sums by exp(m_old − m_new) ≤ 1, so no sum can overflow.
- Outputs: log T = m + ln Σe, P′/T = Σ e ρ η_a / Σ e, P″/T likewise.

Error bound: the standard streaming log-sum-exp bound. The relative error of Σe is ≤ (n + 1)ε, with n = summands. The absolute error of log T is ≤ (n + 2)ε + ε|m|. The ratios carry the relative error of the Mills ratio (a few ε) plus (n + 1)ε from the sums. Nothing is stored per node, so a pass allocates nothing.

### 3. `CalibrationTail` in log form
The tail is held as {log T, P′/T, P″/T}.

- The log residual reads F′ = ∓P′/T and F″ = ∓P″/T − F′² directly.
- `density_in_unit(log U) = (P′/T)·exp(log T − log U)` gives P′/U in any unit.
- At the root on the unit N, log T − log N is the root's own residual, so this is P′/T to within that residual. Its extra rounding is ε·(|log T| + |log N|) absolute in the exponent.
- `from_linear` keeps every linearly summed route bitwise: log T = ln T, and each ratio is one rounded division.

### 4. log1p / expm1
None of the rewritten sites forms 1 − p or exp(x) − 1:

- Every tail is summed from positive terms on the smaller side, and the constraint's "1" is the exact marginal value ∓1 of the unit.
- The only differences are the exponents (q − η)(q + η) and log T − log N. The first is formed as a product of two sums, so it does not cancel the way q² − η² would. The second is the root residual itself, which is the quantity the solve controls.
- So log1p/expm1 would add nothing here. They remain the right tools in the declared-law anchor (`anchor_log_residual`), which this PR does not change.

## Sites
Each site divides its constraint by N:

- The compiled empirical flex row program (`compile_empirical_bms_row_program`): F_a, node stacks, point stack.
- The empirical order-2 lowering (`lower_empirical_bms_calibration_order2`, `lower_bms_flex_row_order2_from_parts`).
- The row intercept solve's returned Jacobian (`solve_row_intercept_base`, i.e. `row_ctx.m_a`).
- Both prediction gradient consumers, through one `prediction_intercept_slopes` helper. This removes a duplicated refusal block.
- The rigid empirical score-ζ measure-sensitivity channels.
- The finite-law training and prediction tail sums, via `sum_calibration_tail`.

Also in this commit:

- The standard-normal cells route keeps its linear values through `CalibrationTail::from_linear`.
- The rigid fifth-order q-stack guard becomes a debug_assert, because the probit marginal map is the identity q = η by construction (`bernoulli_marginal_link_map`).
- The now-unused `unary_derivatives_normal_cdf` is removed.

## Tests
- New: `empirical_flex_row_program_is_exact_where_node_densities_underflow_3639`. It covers both measure kinds, (q, y) ∈ {(−39, 1), (39, 0)}, and asserts:
  - the old linear Jacobian is exactly 0;
  - the compiled program exists with a finite, positive 1/F_a and a finite value;
  - the order-2 gradient matches a central FD (h = 1e-5) of recompiled programs within 1e-6·(1 + |g|).
- Updated to read the log tail: `empirical_calibration_newton_reuse_matches_full_partials_bitwise_2337` and `deep_tail_intercept_root_is_seed_independent_3333`. The side-independence band is derived as ε·(6 + |log T₊| + |log T₋|)·|P′|.

## Not in this PR (#3639 item 4, now #4504)
- The standard-normal cells route stays linear. Its log-space cubic/affine cells and the GPU mirror are separate work, and the GPU path is cells-only. An empirical-law model never goes to the device.
- The apex-tree Φ₂ tail is also separate work.


SauersML:

---

SauersML:
Handing this back. The author session pushed abb6f04a at 12:38, just before my claim, so I'm stepping back while it's active. Two findings from my validation:

1. **The PR's own new test fails** on the first commit (6d490f67) merged with main 00b33efd. It ran with `cargo test -p gam-models --lib -- empirical_fifth 3639 empirical_rigid_jet_oracle`: 7 passed, 1 failed.
   ```
   empirical_fifth_tensor_stays_exact_where_node_densities_underflow_3639 panicked at cell_moment_assembly.rs:4199:
   frailty=None y=0 point=[39.0, 0.3] axes=00000 fifth=0.00001624960140235522 fd=-0.022898702890804534
   ```
   The analytic fifth and the FD of `empirical_rigid_fourth_full_closed_form` disagree by about 3 orders of magnitude at q=+39, y=0. The (−39, −0.35) point was not reached. Main's changes to `cell_moment_assembly.rs` since the PR base only touch the flex contraction path (#2922, #3641), so this looks like the PR's own problem rather than a merge interaction. The survival-side (q ≥ 0) sign or unit of the fifth is the first thing to check.
2. **abb6f04a turns the probit q-stack refusal into a `debug_assert!(...)`.** `build.rs` bans that in non-test code (`("debug_assert!(", "debug_assert!", false)`), and SPEC treats checks that compile to nothing as non-checks. Keep the `Err` refusal, or use a plain `assert!`.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 3314ae4acbaa.

PR #3977, opened 2026-09-20
chart_transfer: report the mean-direction delta-method SE for the rotation angle

### LINKED ISSUE #3904: chart_transfer: rotation_angle_se uses circular SD / sqrt(n_eff), which understates the mean-direction standard error
**Where** (origin/main b3832b92): `crates/gam-sae/src/chart_transfer.rs:174-235`, `rotation_angle_band`

```rust
    let resultant = ((c * c + s_sum * s_sum).sqrt() / weight_sum).min(1.0);
    let effective_n = weight_sum * weight_sum / weight_sq_sum;
    let circular_sd = if resultant > 0.0 {
        (-2.0 * resultant.ln()).max(0.0).sqrt()
    } else {
        f64::INFINITY
    };
    Ok((mean_angle, circular_sd / effective_n.sqrt()))
```

**Consumers**
- `crates/gam-pyffi/src/latent/latent_basis_and_sae_ffi.rs:2826-2845` exports this value as `rotation_angle_se`.
- `experiments/chart_transport_l11_l23.py:398, 492, 727-728` uses `rotation_angle_se` as the z-score denominator and the `2·SE` acceptance band.

**Why it is wrong**

The function returns the circular *standard deviation* `√(−2 ln R̄)` divided by `√n_eff`. That is not the standard error of the circular mean direction `θ̂ = atan2(Σwᵢ sin θᵢ, Σwᵢ cos θᵢ)`.

Take weights normalized to `Σwᵢ = 1`. Linearizing `θ̂` gives the delta-method (sandwich) standard error:

```
se(θ̂) = √(Σᵢ wᵢ² sin²(θᵢ − θ̂)) / R̄
```

For equal weights this reduces to the textbook `√(1 − R̄₂)/(√(2n) R̄)`, i.e. `√(E sin²(θ−μ)/n)/R̄` (Fisher 1993, §4.4).

For a wrapped normal with parameter σ, the circular SD is exactly σ, while the true SE is `√(sinh σ²)/√n`. The reported value is therefore too small by these amounts:

| σ | reported SE / true SE | shortfall |
|---|---|---|
| 0.5 | 0.984 | 1.6% |
| 1.0 | 0.922 | 8% |
| 1.5 | 0.693 | 31% |

Every z-test and `2·SE` band built on `rotation_angle_se` is anti-conservative in exactly the spread regime where the angle is least certain. The Kish `n_eff` also misweights non-uniform weights: the sandwich puts `wᵢ²` on each token's own `sin²` deviation, so a heavy-weight outlier counts for more than an average `n_eff` share.

**Principled fix**

1. Compute `θ̂` as now.
2. Return `√(Σ w̃ᵢ² sin²(θᵢ − θ̂)) / R̄`, where `w̃ = w/Σw`.
3. Return an infinite SE when `R̄ = 0`, where the direction is undefined.
4. Fix the doc so it derives this formula.
5. Add these tests:
   - for a `±δ` pair, SE = `tan δ/√2`;
   - for unequal weights, the SE matches the weighted sandwich.

SauersML:
The rotation-angle band now reports the weighted delta-method standard error of the mean direction, rather than circular dispersion divided by an effective sample size.

Weights are normalized before accumulation, Cartesian moments use compensated sums, and the influence norm uses hypot. These preserve the same result under extreme common weight scaling and retain representable uncertainty for tiny angles or relative weights. A zero resultant has no mean direction and returns an explicit error, handled by the existing PyFFI caller.

Validation on MSI: all 12 exact chart-transfer module tests pass, including the analytic symmetric-pair result, independent weight-influence finite differences, common scales 1e-250..1e250, angles ±1e-200, relative weights [1,1e-200], and an undefined-direction control. Complete SAE unit-test metadata passes with warnings denied (30.74 seconds).

Fixes #3904.


SauersML:

---

PR #3982, opened 2026-09-20
workspace_cache: factor the thin-plate side-condition frame on mean-centered centers

### LINKED ISSUE #3975: workspace_cache: thin-plate constraint null space is factored on raw centers while the dense builder freezes V in the mean-centered frame (n-free re-key, ψ-derivatives and lazy design apply V in a rotated Z)
## Summary

`thin_plate_kernel_constraint_nullspace` (crates/gam-terms/src/basis/workspace_cache.rs) runs the RRQR for `Z = null(P(C)ᵀ)` on the **raw** centers. The Duchon twin `kernel_constraint_nullspace` mean-centers first (#1375), because the pivot order of `[1, x, y, …]` depends on where the coordinates sit.

The dense thin-plate builder, `create_thin_plate_spline_basis_scaledwithworkspace` in duchon_thinplate.rs, factors the **knot-mean-centered** knots (#1269). It freezes its data-metric radial reparam `V` in that `Z_c` frame and writes `V` into `BasisMetadata::ThinPlate::radial_reparam`.

Every metadata consumer rebuilds `Z` from the raw frozen centers and applies that frozen `V`:

- `thin_plate_penalties_at_length_scale`, the n-free per-ψ penalty re-key called from spatial_optimization.rs.
- `build_thin_plate_penalty_psi_derivativeswithworkspace` and `build_thin_plate_scalar_design_psi_derivatives` in duchon_thinplate.rs.
- The lazy chunked branch of `build_thin_plate_basiswithworkspace`. It also builds its polynomial block `[1, x]` on raw data, while the dense branch builds `[1, x − x̄_C]`.

Both `Z_raw` and `Z_c` are orthonormal bases of the same null space, so `Z_raw = Z_c W` for some orthogonal `W`. The consumers therefore compute `Vᵀ Wᵀ Ω_c W V` where the fit used `Vᵀ Ω_c V = diag(μ)`, and `K Z_raw V` where the fit used `K Z_c V`.

## Evidence

I reproduced the pivoted QR with scipy on 20 scattered 2-D centers, comparing `Z_raw` with `Z_c` and measuring the off-diagonal part of `Vᵀ Z_rawᵀ Ω Z_raw V` relative to its largest entry:

| center range | pivots raw vs centered | ‖W − I‖ | off-diagonal / max |
|---|---|---|---|
| [0,1]² | equal | 1e-16 | 1e-16 |
| [0,10]² | differ | 0.46 | 0.163 |
| [100,101]² | [1 2 0] vs [0 1 2] | 0.10 | 0.124 |

`normalize_euclidean_frame` rescales the input isotropically but does not center it. Any covariate whose mean is large next to its spread (years, raw lat/lon, un-centered PCs) is in the bad regime.

## Consequences

1. **Wrong penalty in spatial κ optimization.** After a dense fit, the n-free re-key at the fitted length scale does not reproduce the cold penalty. The outer ψ/κ objective and its gradient are evaluated on a different quadratic form from the one the coefficients were fitted under.
2. **Wrong ψ-derivatives.** `∂S/∂ψ` and `∂X/∂ψ` are built in the rotated frame, so they are not the derivatives of the fitted design and penalty.
3. **Replay across the materialization cap.** `should_use_lazy_spatial_design` switches on `n·p` bytes against a cap that depends on the machine. A model fitted dense and replayed lazy, or the reverse, gets a different `Z` frame under the frozen `V` and a shifted polynomial chart. That predicts different values from the same frozen spec.
4. The promotion gate `thin_plate_canonical_infeasible_at_centers` probes rank on raw centers, while the builder should use the centered frame.

SPEC: "Fitting and inference must be fast on all scales of data … (with seamless transitions between strategies)". The dense-to-lazy transition is not seamless here.

## Fix

- Mean-center inside `thin_plate_kernel_constraint_nullspace`, mirroring `kernel_constraint_nullspace`. The cache key uses the centered centers.
- The dense builder passes the raw knots, so the builder and every consumer factor bit-identical input.
- The lazy branch builds its polynomial block on data centered by the center mean, matching the dense chart.
- The promotion gate probes the rank on centered centers.
- Regression tests on offset and wide domains:
  - raw centers and centered centers give the same `Z`;
  - the n-free re-key at the fitted length scale equals the cold penalty.

Branch: `claude/fix-thin-plate-constraint-frame-centering`.

SauersML:
thin_plate_kernel_constraint_nullspace ran the RRQR for Z = null(P(C)^T) on
the raw centers. The dense thin-plate builder factors knot-mean-centered knots
and freezes its radial reparam V in that frame. The n-free per-psi penalty
re-key, the psi derivatives and the lazy chunked design all rebuild Z from the
raw frozen centers and apply that V. On an offset or wide domain the RRQR
pivot order changes, so they worked in a rotated Z_raw = Z_c W. They computed
V^T W^T Omega W V instead of diag(mu).

- Mean-center inside thin_plate_kernel_constraint_nullspace, as the Duchon
  twin already does (#1375). The cache key uses the centered centers.
- The dense builder passes the raw knots, so the builder and the consumers
  factor identical input.
- The lazy branch builds its polynomial block in the same centered chart, so
  the lazy and dense paths agree.
- The canonical-infeasibility gate probes the rank in the centered frame.
- Add regression tests: Z matches for raw vs centered centers (translation
  invariance), and the n-free re-key equals the cold penalty on offset/wide
  domains.

Fixes #3975

Validation: focused-rust-proof.yml on this branch (key HN-26-tpz-1).

---


SauersML:

---

PR #3985, opened 2026-09-20
sparse_audit: label each block topology record with its own block id

### LINKED ISSUE #3912: inference/sparse_audit: topology record of a coordinate block with no firings is labeled atom 0
**Where** (origin/main b3832b92): `crates/gam-sae/src/inference/sparse_audit.rs:933`, in `topology_records_from_codes` (starts at :901), inside the under-resolved branch `if live.len() < 4` (:926-945):

```rust
            records.push(AuditTopologyRecord {
                atom: report.firings.first().map_or(0, |firing| firing.block),
                support_size: live.len(),
                ...
```

**Why it is wrong**

The record's atom id comes from the report's first firing. This branch exists exactly for blocks with few or no firings. A block with no firings has `firings.first() == None`, so it is labeled **atom 0**.

The audit then emits a second "atom 0" topology record, or one that contradicts the real atom 0's record, and drops the silent block's own id. The FFI atom list and the summary counts attribute the empty support to the wrong atom. This is a silent fallback to an arbitrary id.

**Principled fix**

`BlockCoordinateReport` does not carry its block id. So `run_sparse_sae_audit` should collect each report's block id beside the report and pass the ids to `topology_records_from_codes`. Every record then uses its own id, with no `map_or(0, …)`.

Test: a block that never fires keeps its own block id and reports `support_size == 0`.

SauersML:
`topology_records_from_codes` took each record's atom id from the report's first firing, via `firings.first().map_or(0, …)`. A block that never fired has no first firing, so its record came out labeled atom 0.

`run_sparse_sae_audit` now collects the block ids beside the coordinate reports and passes them in, so every record carries its own id.

New test: `a_block_that_never_fires_keeps_its_own_block_id`.

Fixes #3912

---


SauersML:
Taking over this stalled PR (no push since 09:40Z) as the wave8 session RF-0748-1. My plan:
- merge origin/main in;
- run `a_block_that_never_fires_keeps_its_own_block_id` plus a gam-sae check;
- merge.

---

PR #3986, opened 2026-09-20
inference::intervention_shard: screen calibration measurability over the records the fit uses

### LINKED ISSUE #3976: inference::intervention_shard: calibration screens measurability over zero-prediction records, so a measurable atom reaches the reference frame with no fit rows
## Defect

`prepare_intervention_calibration` (crates/gam-sae/src/inference/intervention_shard.rs, origin/main a5200bea) screens atom measurability over a different record set from the one it emits as fit rows.

The screen, at lines 497-505, classifies every non-control train record, whatever its prediction:

```rust
for i in 0..n {
    if !eval[i] && !shard.is_control[i] {
        let measurable = shard.nu_measured[i] > floor[i];
        atom_is_measurable
            .entry(shard.atom[i])
            .and_modify(|seen| *seen = Some(seen.unwrap_or(false) || measurable))
            .or_insert(Some(measurable));
    }
}
```

The row selector, at lines 525-531, admits only records with a positive prediction, because `log ν̂` is the fit predictor:

```rust
if shard.is_control[i]
    || nu_hat[i] <= 0.0
    || !matches!(atom_is_measurable.get(&shard.atom[i]), Some(Some(true)))
{
    continue;
}
```

`validate` accepts `nu_hat_1 = 0` and `nu_hat_2 = 0` on intervention records. That happens when `½ Δxᵀ G Δx` vanishes on a direction the metric does not see, or when the Rung-2 decoder does not respond to the atom.

Take an atom whose only train interventions clear their floor but predict 0 nats. It lands in `measurable_atoms` and so in `reference_frame()`, yet it has no row in `fit_frame()`.

`gamfit/_intervention_calibration.py` then calls `fitted.predict(plan.reference_frame())` on a `group(atom)` level the fit never saw. That call either:
- is refused as an unseen level, which aborts the whole calibration; or
- returns a re-speed that no data supports.

In the second case the centring in `finish` (`eta - mean_eta`) also moves every other atom's re-speed.

## Reproduction

Take `calibration_shard_and_spec()` from the module tests and set `nu_hat_1[3] = 0.0` and `nu_measured[3] = 5.0`, where atom 20 has one train intervention.

- Today the plan gives `measurable_atoms == [10, 20]` and `train_atom == [10, 10]`.
- Atom 20 is named in the reference frame with no fit rows.

## Fix

- One predicate, `fit_eligible(i) = !is_control[i] && nu_hat[i] > 0`, selects both the records the screen classifies and the rows the plan emits. Every atom named in the reference and held-out frames then has fit rows.
- An atom with no fit-eligible train intervention is reported in `no_training_intervention`.
- A regression test pins the counterexample, including the invariant that every reference or eval atom appears in `train_atom`.

SauersML:
Fixes #3976

`prepare_intervention_calibration` screened each atom's measurability over every non-control train record. It emitted fit rows only where the prediction is positive, the domain of `log ν̂`. So an atom resolved only by zero-prediction records reached `measurable_atoms`, and with it the reference frame, without any row in the fit frame. Predicting its `group(atom)` level then either:
- fails as an unseen level; or
- returns a re-speed the data never supported, and the centring in `finish` shifts every other atom's re-speed.

**Fix:** one predicate, `fit_eligible` (non-control with a positive prediction), now selects both the records the screen classifies and the rows the plan emits. Every atom named in the reference and held-out frames therefore has fit rows. An atom with no fit-eligible train intervention is reported in `no_training_intervention`. The Python adapter's docstring for that field now states the same definition.

**Test:** `an_atom_resolved_only_by_zero_predictions_is_never_named_without_fit_rows` pins the counterexample from the issue. It also checks the invariant that every reference or eval atom appears in `train_atom`.

Validation: focused-rust-proof run 35502211808 (`intervention_shard::tests::calibration`, the new test, `deterministic_controls`) on `85b68b91`.

---


SauersML:

---

SauersML:
Taking over from RF-1113-3 (no activity since 2026-09-20T11:31:47Z). Head 85b68b91 has a green focused-rust-proof run, 35502211808, and it merges cleanly with main. Main has not touched `intervention_shard.rs` or `_intervention_calibration.py` since the branch base. I re-read the diff: one `fit_eligible` predicate now drives both the measurability screen and the fit rows, and it adds no knob or tolerance. Merging.

---

PR #3987, opened 2026-09-20
atlas_nerve: drop the self-referential co-activation threshold

### LINKED ISSUE #3913: inference/atlas_nerve: co-activation gate compares a pair's mass with its own per-row mean, so it never refuses an edge
**Where** (origin/main b3832b92): `crates/gam-sae/src/inference/atlas_nerve.rs:788-789`, in `build_atlas_nerve` (starts at :724):

```rust
    for (&(a, b), overlap) in &overlaps {
        let threshold = overlap.mass / overlap.positive_rows as f64;
        let coactive = overlap.mass.is_finite() && overlap.mass >= threshold;
```

The value is reported as `AtlasNerveEdge::coactivation_threshold` (:174, :805). `mutual_row_mass` (:329) computes the same per-row mean as its second return value.

**Why it is wrong**

A pair has an entry in `overlaps` only when at least one row co-fires, so `positive_rows >= 1` and `mass > 0`. Then `mass >= mass / positive_rows` always holds, and `coactive` is a tautology for every finite mass.

The reported `coactivation_threshold` looks like an independent co-activation gate, but it never refuses an edge. The existing test at :1066 (`rejected.coactivation_mass >= rejected.coactivation_threshold`) restates the tautology. Edge admission really depends on `transfer_valid` alone, while the report implies two gates.

**Principled fix**

The module doc defines co-activation as positive mutual row mass, so gate on `mass.is_finite() && mass > 0.0` and nothing else. Remove:
- the self-referential `coactivation_threshold` field;
- the `positive_rows` counter;
- `mutual_row_mass`'s unused second return value.

A real co-activation significance test would need a null model for chance co-firing. If one is wanted, it is a separate design question, and a quantity that always passes should not stand in for it.

SauersML:
`build_atlas_nerve` admitted a pair when `mass >= mass / positive_rows`. Every pair in `overlaps` co-fires on at least one row, so this condition always held and the gate never refused an edge.

Co-activation is now what the module doc says it is: a positive, finite mutual row mass. This PR removes:
- the `coactivation_threshold` field;
- the `positive_rows` counter;
- the unused second return value of `mutual_row_mass`.

Edge admission is unchanged in effect: an edge needs positive mass and a valid transfer certificate.

Fixes #3913

---


SauersML:
Taken over by RF-1113-3; driving to merge. I reviewed it, and it now sits in a local integration tree on origin/main where it merges cleanly. I'll build and run the targeted tests there before merging.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 7b3d6e3f9902.

PR #3990, opened 2026-09-20
parameter_decomposition::program: resolve a refinement's reached controls where they run

### LINKED ISSUE #3800: parameter_decomposition::program: Refine's all-on check ignores masks scoped inside the mechanism
## Where

`crates/gam-sae/src/parameter_decomposition/program.rs:1571-1576` on origin/main (`Executor::refine`):

```rust
let all_on = program.controls_below[mechanism.index()]
    .iter()
    .all(|&control| self.resolve(control, path) == 1.0);
if self.residuals.is_none() {
    let chosen = if all_on { native } else { mechanism };
    return self.body(chosen, arguments, path);
}
```

## Why it is wrong

Here `path` ends at the refine site. `controls_below[mechanism]` also holds the controls of bodies the mechanism **calls, composes or refines**, and those controls run at deeper paths:
- a `Call` or `Refine` inside the mechanism pushes its site;
- a `Compose` stage pushes `CallSite { stage: k, .. }`.

`resolve` takes the deepest scope that is a prefix of the path it is given. A `MaskAssignment` scoped to one of those deeper invocations is not a prefix of the refine-site path, so the all-on check never sees it. Two wrong outcomes follow:

1. **A mask is silently dropped.** Suppose the mechanism calls a shared macro and the mask sets a control in that macro to 0.0, scoped to `[refine_site, call_site]`. The check says all-on and runs the native body, so the intervention has no effect. Because the check also sets `RefinementResidual::all_on`, the residual path returns the native output as well.
2. **The mechanism runs when it should not.** Suppose a group has a global value 0.5 and a deeper scope overrides it to 1.0 at the one invocation where the control actually runs. The check resolves 0.5 at the shallow path and runs the mechanism, although every control resolves to exactly 1 where it executes. The all-on output is then no longer bit-identical to the source.

This breaks the contract in the module doc:
- lines 22-23: "Every call site is an invocation, and a control inside the body can be set at every invocation or at one."
- lines 24-26: "When every control the mechanism reaches resolves to exactly `1` at this invocation, the native body runs"
- lines 35-36: "a [`MaskAssignment`] scoped to one invocation changes that call only."

## Principled fix

Replace the flat check with a walk of the mechanism that resolves each node's own controls at the path where that node executes, mirroring `Executor::body`:
- `Native`: the controls of each parameter the primitive reads, plus the `CoordinateMask` controls;
- `Sum`: the term controls;
- `Compose`: each stage control, then push `CallSite { stage: k, ..site }` and recurse into the stage body;
- `Call`: push the site and recurse into the callee;
- `Refine`: push the site and recurse into its mechanism. A native body reaches no controls; `Program::new` enforces this.

The walk stops at the first control that is not 1. `controls_below` stays as it is for validation and `controls_reached`.

A regression test would use a mechanism that calls a shared body holding a controlled sum, with a mask scoped to `[refine_site, call_site]`. That case must run the mechanism. The mirror case (global 0.5, overridden to 1.0 at the deep scope) must run the native body.

SauersML:
Fixes #3800

`Executor::refine` decided all-on by resolving every control in `controls_below[mechanism]` at the refinement's own path. But a control inside a body the mechanism calls, composes or refines runs one invocation deeper. So a mask scoped to that deeper invocation was never seen, and the wrong branch ran:
- a scoped deletion ran the native body;
- a deep override of a global value back to 1 ran the mechanism.

**Fix:** the check now walks the mechanism the way the executor does, resolving each node's own controls at the path it runs on.

**Test:** `a_refinement_resolves_each_reached_control_at_the_invocation_where_it_runs` pins both directions: a scoped deletion below a call, and a scoped override back to 1.

Validation: focused-rust-proof run 35500529084 on `bbc98b76`.

---


SauersML:

---

SauersML:
Verified locally and merged (the Actions queue is backed up).

- Tested head `bbc98b76` merged with `origin/main` locally
- `cargo test -p gam-sae --lib a_refinement_resolves_each_reached_control_at_the_invocation_where_it_runs`: **1 passed**
- `cargo test -p gam-sae --lib parameter_decomposition::program`: 25 passed and 1 failed. The failure is `every_value_the_executor_forms_is_reserved_while_it_is_live`, which reads the process-global `MemoryGovernor` and races with tests running in parallel. It also fails on main, passes 3/3 with `--test-threads=1`, and has nothing to do with this change.
- Review against SPEC.md: sound. The mask check now walks the mechanism the same way the executor does, and no constants or fallbacks are added.

---

PR #3996, opened 2026-09-20
pirls/deviance: one Beta intercept-score root solver

### LINKED ISSUE #3989: pirls/deviance: beta_null_eta duplicates beta_eta_for_logit_target's score-root bisection; its 256-halving cap is an arbitrary dead bound
**Where:** `crates/gam-solve/src/pirls/deviance.rs` on origin/main at e0cc6a6a.

- `beta_eta_for_logit_target` (:1768-1818) solves `psi(phi*mu) - psi(phi*(1-mu)) = target` in logit coordinates. It brackets by doubling from ±1 out to ±1024, then bisects. It is the saturated-eta solver, used at :1393 and :2182.
- `beta_null_eta` (:1820-1896) solves the *same* equation, with `target = E_w[logit y]`. It carries a verbatim copy of the doubling bracket and the bisection (:1844-1895). The copy differs only in its non-finite handling (it checks `is_nan` where the shared solver checks `!is_finite`) and in its error strings.

Both copies bisect under `for _ in 0..256 { ... }` and end with an unreachable error:

```rust
Err(EstimationError::InvalidInput("Beta score root did not reach an adjacent-float bracket".into()))
```

**Why it is wrong**

- SPEC: "Breaking changes are welcome. Keep no compat shims and no dead code." The template also names "duplicated logic that should have a single source of truth" as a defect. Two copies of one root solver can drift apart, and they already have: their non-finite checks differ.
- SPEC: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code." The cap of 256 halvings is arbitrary, and it is dead:
  - The real termination test is `midpoint == lower || midpoint == upper`, the adjacent-float bracket.
  - Bisecting a finite bracket inside [-1024, 1024] reaches that test in at most about 64 + 11 halvings, because each step halves the interval. So the counter never binds and the trailing `Err` is unreachable.
- By monotonicity, the score at every midpoint lies strictly between the finite scores at the two bracket ends. So the stricter `!is_finite` check of the shared solver never changes the null-model result.

**Fix**

- `beta_null_eta` validates `phi` and computes `target`, then returns `beta_eta_for_logit_target(target, phi)`. Its copy of the bracket and bisection is deleted.
- The shared solver returns the bracket end on an exact zero score there, as the null copy already did.
- It bisects in a plain `loop` that ends on the adjacent-float test, so the 256 counter and the unreachable `Err` are deleted.

SauersML:
Fixes #3989

`beta_null_eta` carried a verbatim copy of the bracketing and bisection in `beta_eta_for_logit_target`, the saturated-eta solver. The copy had slightly different non-finite handling. The null model solves the same equation, `psi(phi*mu) - psi(phi*(1-mu)) = target`, only with `target = E_w[logit y]`. It now validates `phi`, computes the target and calls the shared solver.

The shared bisection also stopped after an arbitrary 256 halvings:
- Its real termination test is the adjacent-float bracket, which a bisection of a finite bracket in [-1024, 1024] always reaches, in at most about 75 halvings.
- So it now loops on that test alone, and the unreachable "did not reach an adjacent-float bracket" error is deleted.
- An exact zero at a bracket end returns that end, as the null copy did.

Validation: focused-rust-proof on this branch (key `HN-13-beta-1`, run 35500867242), gam-solve lib, with these filters:
- `null_deviance_preserves_boundaries_dormancy_and_beta_mle_geometry`, the Beta null MLE
- `beta_half_unit_deviance_from_shape_differences_is_exact_and_resolved`, the saturated eta


SauersML:
Verified locally, marked ready and merged (head 1211d6e3). I merged it with origin/main in a batch tree and compiled that tree.

- `cargo test -p gam-solve --lib pirls` (covers pirls/deviance): 190 passed
- `cargo test -p gam-solve --lib reml::`: 489 passed

---

PR #3999, opened 2026-09-20
gam-predict: forward #2077 prior weights to the posterior-mean observation band

### LINKED ISSUE #3924: gam-predict: the posterior-mean prediction branch drops the #2077 prior weights, so a weighted Gaussian non-identity-link fit gets a homoscedastic observation band
## Defect

`interval_policy.rs::predict_columns` sends an interval request with a confidence level on a posterior-mean model (`(Some(level), true)`) down `predict_posterior_mean`. It builds `PosteriorMeanOptions { confidence_level, covariance_mode, include_observation_interval, extrapolation_variance }`, and `PosteriorMeanOptions` has no field for the #2077 per-row prior weights. `request.observation_prior_weights` is therefore dropped.

In `standard.rs`, the posterior-mean branch builds its band from `PredictUncertaintyOptions { ..., extrapolation_variance, ..PredictUncertaintyOptions::default() }`, so `observation_prior_weights` is `None` there as well.

A curved-link Gaussian fit (`gaussian-log`, `gaussian-inverse`, ...) takes the posterior-mean branch, `prediction_uses_posterior_mean`. For a weighted fit, its observation band is the homoscedastic `σ̂² + Var(μ_i)` instead of the `σ̂²/w_i + Var(μ_i)` that #2077 introduced and that the identity-link path reports:

- a row with `w_i > 1` over-covers;
- a row with `w_i < 1` under-covers.

The same weights and the same model give a different band depending only on the link.

## Fix

- Add `observation_prior_weights: Option<Array1<f64>>` to `PosteriorMeanOptions`, with `None` in `point_only()`.
- Forward it from the interval request in `interval_policy.rs`.
- Pass it through to the `PredictUncertaintyOptions` that `standard.rs` builds for the band.
- Add a regression test. Two identical Gaussian-log rows with weights 1 and 9 give band widths in the ratio 1/3; before the fix they were equal.

SauersML:
Fixes #3924

- Adds `observation_prior_weights` to `PosteriorMeanOptions` (`None` in `point_only()`).
- `interval_policy::predict_columns` forwards `request.observation_prior_weights` on the posterior-mean route.
- The `standard.rs` posterior-mean branch passes the weights into its `PredictUncertaintyOptions`.
- The generic posterior-mean path `predict_posterior_mean_generic` also forwards `options.observation_prior_weights` in its `(None, None)` arm, the arm that builds `PredictUncertaintyOptions` from scratch. Without this, the generic route would still price every row at the unit-weight `σ̂²`.

A weighted curved-link Gaussian fit (for example gaussian-log) now reports the heteroscedastic band. The observation variance is `Var(y_i) = σ̂²/w_i + Var(μ_i)`, where `σ̂²/w_i` is the prior-weighted Gaussian residual variance (#2077) and `Var(μ_i)` is the delta-method mean variance. This matches the identity-link path. Before this change the band was the homoscedastic `σ̂² + Var(μ_i)`.

The new regression test `gaussian_log_posterior_mean_observation_band_honours_prior_weights_2077` uses two identical rows and a coefficient covariance of `1e-6·I`, so `Var(μ_i)` is negligible next to `σ̂²`. It checks three things:
- the unweighted rows share one band;
- a weight-1 row keeps the pooled band (to 1e-12);
- a weight-9 row's band is `1/sqrt(9) = 1/3` as wide (to 1e-4).

Existing struct-literal constructions are updated with the new field.


SauersML:
Note for HN-17: BF-84 added two commits to this branch.
- A merge of origin/main.
- eb73405adc, which forwards `options.observation_prior_weights` in the generic `predict_posterior_mean_generic` `(None, None)` arm. Without it, the generic route still priced every row at the unit-weight `σ̂²`.

Proof run 35509128323 is queued on eb73405adc. I will merge only on a green result, and only if the head is still eb73405adc. If you push again, the PR is yours to land.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 300ac2f05824.

PR #4008, opened 2026-09-20
gam-linalg: read PSD pseudoinverse indefiniteness off the resolution band

### LINKED ISSUE #4000: gam-linalg utils: rank_certified_psd_pseudoinverse's indefiniteness band 128·n·ε is an arbitrary constant and ignores the caller's declared null cutoff
## Defect

`rank_certified_psd_pseudoinverse` (`crates/gam-linalg/src/utils.rs:1429`, origin/main) decides whether its input is "materially indefinite" with a hand-picked multiplier:

```rust
let psd_roundoff = 128.0 * penalty.nrows() as f64 * f64::EPSILON * max_abs;
if let Some((index, &value)) = eigs.iter().enumerate().find(|(_, value)| **value < -psd_roundoff) {
    return Err(LinalgError::InvalidInput(format!("PSD pseudoinverse input is indefinite ...")));
}
let absolute_cutoff = relative_cutoff * max_eigenvalue;
```

The same crate already owns this band. `gam_linalg::roundoff::symmetric_spectrum_rounding_band` (`roundoff.rs:66`) is documented as the band below which "an eigenvalue ... is not resolved from zero by the decomposition that produced it, and a sign inside it is not a measurement":

```rust
eigenvalues.len() as f64 * f64::EPSILON * spectral_radius
```

The pseudoinverse uses the same shape `p·ε·max|λ|`, but with an extra factor of 128 that nothing derives.

### It is also the wrong band for the callers' own resolution

The function takes a caller-declared `relative_cutoff`, and "values at or below that explicit cutoff define the discarded null space". Its production callers derive that cutoff from how their matrix was formed:
- `gam-models/src/bms/mod.rs:2471` uses `γ_{n+row_roundings}·tr Ω̂/λ_max + r·ε`.
- `gam-models/src/marginal_slope_orthogonal.rs:446` uses `γ_{n+1}·tr G/max G_jj + p·ε`.
- `bms/mod.rs:2143,2356` use `jacobi_scaled_normal_relative_cutoff`.

Those callers state that an eigenvalue of magnitude up to `relative_cutoff·λ_max` is not resolved from zero, because the Gram's formation error is that large. But the sign test ignores the caller's resolution and uses `128·p·ε·max|λ|` instead:
- For a Gram accumulated over `n ≫ 128·p` rows, the formation band `γ_n·tr` is larger than `128·p·ε·λ_max`.
- A null direction whose computed eigenvalue lands on the negative side, inside the caller's own declared cutoff, is then refused as "indefinite".
- A positive eigenvalue of the same magnitude is silently discarded as null.

So the same unresolved direction is either discarded or fatal depending on a sign that, by the caller's own resolution, carries no information.

## Principled fix

A negative eigenvalue is unresolved from zero when it lies inside either of two bands:
- the decomposition's rounding band, `symmetric_spectrum_rounding_band`;
- the caller's declared null cutoff, `relative_cutoff·λ_max`.

The indefiniteness test should therefore be `λ < −max(symmetric_spectrum_rounding_band(λ̂), relative_cutoff·λ_max)`. This drops the `128` and gives the function one resolution for both signs.

### Related copies (left as follow-up)

`bms/gradient_paths.rs:1373` (`MarginalSlopeCovariance::full`) carries its own `128·k·ε·max|λ̂|`. Its comment claims this is "the established form ... that `rank_certified_psd_pseudoinverse` already uses". `bms/conditional_score_covariance.rs:579` floors an innovation variance at "the SAME band" (`128·k·ε·score_scale`).
- For `full`, the input is exact, so `symmetric_spectrum_rounding_band` is the whole band.
- The innovation floor is a sample variance, so its band is the formation band of a length-`n` weighted sum of squares, not an eigensolver band.

Also, `gam-sae/src/manifold/outer_objective.rs:3196` (`reactive_smooth_curvature_scale`) passes a literal `1.0e-10` as the penalty's null cutoff. It should use the penalty's assembly band against `λ_max`, the way the bms and marginal-slope callers do.

The two bms constants are coupled by their comments, and bms has its own auditor. The fix PR here updates the stale cross-reference in `gradient_paths.rs` but leaves those two constants for a bms-side change.

SauersML:
Fixes #4000

## Change
`rank_certified_psd_pseudoinverse` (`crates/gam-linalg/src/utils.rs`) used to refuse a negative eigenvalue below `-128·n·ε·max|λ|`. The `128` is not derived from anything, and the same function silently discarded a positive eigenvalue of the same magnitude under the caller's declared `relative_cutoff`.

It now refuses an eigenvalue as indefinite only when

```
λ < −max(symmetric_spectrum_rounding_band(λ̂), relative_cutoff·λ_max)
```

- The first term is the crate's owner band for "not resolved from zero by the decomposition".
- The second term is the caller's declared null resolution.

Inside either band the sign carries no information, so the function now treats both signs with one resolution. The unused `max_abs` fold is removed, and `absolute_cutoff` is computed once and reused.

`bms/gradient_paths.rs` only gets a comment fix. Its old comment said its `128·k·ε` band was the form this function "already uses", which is no longer true. That constant and its twin in `conditional_score_covariance.rs` are left for a bms-side change, as described in #4000.

## Tests
- **New:** `psd_pseudoinverse_admits_a_negative_eigenvalue_inside_its_declared_null_cutoff`, on `diag(1, −1e-12)`.
  - With cutoff `1e-10`: rank 1, and the pseudoinverse is within 4ε of `diag(1, 0)`.
  - With cutoff `0`: still refused as indefinite.
- **Existing:** the indefinite-input test (`diag(1, −1e-4)`, cutoff `1e-10`) still errors.

---


SauersML:

---

PR #4013, opened 2026-09-20
sae chart canonicalization: never commit a cap-truncated flow iterate

### LINKED ISSUE #3988: sae chart canonicalization: isometry-defect flow Gauss–Newton commits a cap-truncated iterate as the canonical chart
**Area:** `crates/gam-sae/src/chart_canonicalization.rs` — the isometry-defect flow Gauss–Newton cores (origin/main `f8c6613`; the file is unchanged since `988f269`)

SPEC: "A fit object must only ever come from a converged optimization."

### What the code does

Both damped Gauss–Newton cores iterate up to `TORUS_FLOW_GN_MAX_ITERS` (= 80, line 604). They leave the loop on one of three events:
- the trust loop reports convergence;
- the trust loop exhausts its rejections;
- the cap runs out.

After the loop they return `Some(θ)` whenever the defect went down. **None of these events is recorded**, so a cap-truncated iterate is returned exactly like a converged one:

```rust
// minimize_isometry_defect_flow, chart_canonicalization.rs:1500-1567
for iteration in 0..TORUS_FLOW_GN_MAX_ITERS {
    if iteration + 1 == TORUS_FLOW_GN_MAX_ITERS {
        break;                          // cap exit (also makes the cap 79, not 80)
    }
    ...
    let step = lm_damped_accept_sweep(...);
    any_accepted |= step.accepted;
    if !step.accepted { break; }
    if step.converged { break; }
}
if !any_accepted || !(state.defect < defect_initial) {
    return None;
}
Some(FlowMinimization { theta, defect_initial, defect_final: state.defect, profiled_scale: state.scale })
```

The sphere-boost core (around lines 2849-2892) has the same shape: `for iteration in 0..TORUS_FLOW_GN_MAX_ITERS { if iteration + 1 == ... { break; } ... }` followed by `if !any_accepted || !(state.defect < defect_initial) { return None; }`.

### Why it is wrong

The returned θ is committed as the atom's canonical chart:
- `canonicalize_atom_torus_flow_chart` (fit_drivers.rs:1818) and `canonicalize_atom_sphere_flow_chart` (fit_drivers.rs:2002) reparameterize the decoder and set `atom.chart_canonicalized = true`;
- the free-patch flow (chart_canonicalization.rs:2216) consumes `minimize_isometry_defect_flow` the same way.

This happens in fit_drivers.rs:1320-1340, before the chart-dependent diagnostics (Θ, LOAO ΔEV, the hybrid split) are computed in that chart. A chart produced by a cap-truncated run is not the minimum-defect representative the canonicalization promises. Every downstream diagnostic then depends on the iteration budget, the classic "best so far" return. The constant's own doc says "termination is normally by the ... tolerances below", which assumes the cap never binds, but nothing checks that.

### Fix (small, local)

Track whether the trust loop actually terminated. Both terminations count: `step.converged`, meaning the improvement is within the two evaluations' rounding bands, and a non-accepted sweep, meaning no fold-free descent is resolvable. If the cap runs out first, return `None`, which is the existing honest-skip path: the chart is left as fitted and `chart_canonicalized` stays false. The off-by-one `iteration + 1 == MAX` break is dropped too, so the cap is the documented 80.

Draft PR incoming.

SauersML:
Fixes #3988

SPEC: "A fit object must only ever come from a converged optimization."

The isometry-defect flow Gauss–Newton cores return `Some(θ)` whenever the defect went down. Those cores are `minimize_isometry_defect_flow` (torus / free patch) and the sphere-boost core. They did this even when the iteration cap ran out while every sweep was still accepting a resolvable descent. The truncated iterate was then committed as the atom's canonical chart (`chart_canonicalized = true`), so every chart-dependent diagnostic depended on the iteration budget.

Changes (only `crates/gam-sae/src/chart_canonicalization.rs`):
- Both cores record whether the trust loop terminated. It terminates either when `step.converged` holds (the improvement is inside the rounding bands) or when a sweep is not accepted (no resolvable fold-free descent). If the cap runs out first, the core returns `None`, which is the existing honest-skip path: the chart is left as fitted.
- The `iteration + 1 == TORUS_FLOW_GN_MAX_ITERS` early break is dropped, so the cap is the documented 80 rather than 79.
- The docs on the constant and on `minimize_isometry_defect_flow` are updated to match.

No retry, fallback or new constant is added.

---


SauersML:
Taken over by RF-1113-3; driving to merge. I reviewed it, and it now sits in a local integration tree on origin/main where it merges cleanly. I'll build and run the targeted tests there, including the existing chart-canonicalization tests that could now hit the iteration cap, before merging.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 011775ca00e5.

PR #4030, opened 2026-09-20
survival/construction: refuse exit before entry in normalize_survival_time_pair

### LINKED ISSUE #4012: survival/construction: normalize_survival_time_pair silently rewrites exit < entry to entry+1e-9, masking every downstream exit<entry check
## Defect

`crates/gam-models/src/survival/construction.rs:287-308` (origin/main `41fbd6c3`):

```rust
pub const SURVIVAL_TIME_FLOOR: f64 = 1e-9; // line 254

pub fn normalize_survival_time_pair(entry_raw: f64, exit_raw: f64, row_index: usize)
    -> Result<(f64, f64), String> {
    if !entry_raw.is_finite() || !exit_raw.is_finite() { return Err(/* non-finite */) }
    if entry_raw < 0.0 || exit_raw < 0.0 { return Err(/* negative */) }

    let entry = entry_raw.max(SURVIVAL_TIME_FLOOR);
    let exit = exit_raw.max(entry + SURVIVAL_TIME_FLOOR);
    Ok((entry, exit))
}
```

The function refuses non-finite and negative times but not `exit_raw < entry_raw`. For such a row the second `max` quietly turns the pair into `(entry, entry + 1e-9)`. The row said it left before it entered; the fit gets a near-zero-length interval at the entry time instead, and the row still contributes an event or censoring term at a time the data never contained.

The downstream code does try to refuse this case, but every one of those checks reads the pair *after* normalization, so none of them can fire:

- `survival/base.rs:648` (`age_exit < age_entry`)
- `survival/construction.rs:1290`
- `survival/location_scale/prepare.rs:194`
- `survival/marginal_slope/spec.rs:516`

The normalizer has these callers:

- **Fit path:** `fit_orchestration/materialize/survival.rs:40`.
- **Predict paths:**
  - `survival/predict.rs:413, 2895, 3490`
  - `gam-cli run_predict.rs:144, 415, 625, 1766, 2441`
  - `gam-inference sample.rs:1464`
  - `inference/model_payload_builders.rs:2223`

The latent-grid path at `predict.rs:556-562` already refuses `time < entry` explicitly, so the codebase does intend a refusal here.

SPEC: "In general, do not paper over solver issues." Rewriting bad input so it looks valid, with a magic `1e-9` shift, is the input-side version of that. It also puts an arbitrary absolute constant into the likelihood: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code."

## Fix

Check `exit_raw < entry_raw` in `normalize_survival_time_pair` on the raw values, before the floor, and return the same typed `DataValidationFailed` refusal the function already uses for negative times. A zero-length interval (`exit_raw == entry_raw`) still normalizes as before. Add a unit test to `construction.rs` covering the refusal, an ordered pair and a zero-length pair at the origin.

## Related, not in this fix

`SURVIVAL_TIME_FLOOR = 1e-9` and `ENTRY_AT_ORIGIN_THRESHOLD = 1e-8` (`base.rs:273`) are absolute and depend on the time unit: the same data in days or in years gets a different floor relative to its time scale. #3288 and PR #3404 cover the other survival time constants; these two belong in that unit-relative rework.

SauersML:
Fixes #4012

`normalize_survival_time_pair` refused non-finite and negative times, but not a row whose exit precedes its entry. For such a row the floor `exit_raw.max(entry + SURVIVAL_TIME_FLOOR)` quietly returned `(entry, entry + 1e-9)`: a near-zero-length interval at the entry time that the data never contained. Every downstream `age_exit < age_entry` refusal reads the normalized pair, so none of them could ever fire:

- `survival/base.rs`
- `survival/construction.rs` (time basis)
- `location_scale/prepare.rs`
- `marginal_slope/spec.rs`

## Change

- `normalize_survival_time_pair` now checks `exit_raw < entry_raw` on the raw values, before the floor, and returns the same typed `DataValidationFailed` refusal it already uses for negative times.
- A zero-length interval (`exit_raw == entry_raw`) still normalizes as before.
- On the latent-window grid path, the existing `time < age_entry` refusal compares against the floored entry, so it still fires first. That path is never stricter than before.
- New unit test `normalize_survival_time_pair_refuses_exit_before_entry` covers three cases: the refusal and its message, an ordered pair, and a zero-length pair at the origin.

SPEC: "In general, do not paper over solver issues." Rewriting invalid input into valid-looking input is that same papering-over, done on the input side.

---


SauersML:

---

PR #4032, opened 2026-09-20
gam-terms PCA basis: read score-design rank off the Gram's resolution band

### LINKED ISSUE #4018: gam-terms PCA basis: score-design rank check through rrqr_from_gram ignores verdict_margin, so an exactly duplicated component passes as full rank
## Defect

`pca_function_mass_penalty` (`crates/gam-terms/src/smooth/term_specs.rs:6987`, origin/main) is the only check that a PCA smooth's realized score design `Z` has full column rank. Both `build_pca_smooth_basis` paths reach it:
- the dense path (`fast_ata(&design)`, line 7121);
- the lazy `.npy` memmap path (`op.diag_xtw_x(ones)`, line 7071).

It decides rank with the Gram-side pivoted QR and ignores the certificate that function returns:

```rust
let rrqr = gam_linalg::faer_ndarray::rrqr_from_gram_with_permutation(
    &raw_score_gram, n_rows, gam_linalg::faer_ndarray::default_rrqr_rank_alpha(),
)?;
if rrqr.rank != k { /* bail "rank deficient" */ }
```

`rrqr_from_gram_with_permutation` builds the eigen square root `F = diag(√λ₊)Vᵀ` of `G = ZᵀZ` and runs column-pivoted QR on `F`. Its own doc comment (faer_ndarray.rs, "Gram-squaring precision floor") explains what goes wrong for an exactly dependent column:
- The true singular value is 0.
- The computed eigenvalue of `G` is a rounding residue of order `ε·σ_max²`, and its sign is arbitrary.
- When that residue rounds positive, `F` carries a pivot of about `√ε·σ_max ≈ 1.5e-8·σ_max`.
- That pivot is about six orders of magnitude above the QR cutoff `100·ε·max(n,k)·|R₀₀|`.
- The rank comes back full, and `rrqr.rank != k` does not fire.

The function returns `verdict_margin` exactly so a caller can detect this regime; the doc says the caller should "fall back to the exact tall RRQR". `pca_function_mass_penalty` never reads it.

## Evidence

A numpy simulation reproduces the pipeline: `G = ZᵀZ`, `eigh`, `F = diag(√λ₊)Vᵀ`, column-pivoted QR, and the same cutoff. It uses 500 random designs, with `n ∈ [50, 2000)` rows and 4 independent columns of mixed scale, plus one dependent fifth column:
- **Duplicated column:** 254 of 500 trials return **full rank (5)**.
- **Column equal to the sum of two others:** 248 of 500 trials return **full rank (5)**.

That is roughly every trial whose null eigenvalue rounded positive.

The existing test `rank_deficient_pca_score_design_is_rejected` uses an all-zero column. There `G` has an exact zero row and column, so the eigenvalue is exactly 0 and the defect cannot show. No test covers a duplicated or linearly combined component.

## Consequence

A PCA smooth whose basis has a duplicated or linearly dependent component is accepted about half the time, depending on a rounding sign. The penalty `S = G/n` is then singular in a direction that also lies in the design's null space. The penalized Hessian `ZᵀWZ + λS` is therefore singular along it for every `λ`. The failure then surfaces later as an unidentified coefficient direction in the solver, or in the identifiability audit, instead of at basis construction with a message naming the redundant component. The function exists to refuse that design up front: "remove zero or dependent components instead of stabilizing them with a coefficient ridge".

## Principled fix

Read the rank off the Gram's spectrum against the Gram's own resolution. The workspace already defines that predicate, and other Gram consumers already use it:
- `gam-problem/src/joint_penalty.rs:255`
- `gam-sae/src/manifold/support_term.rs:1089`
- `gam-sae/src/parameter_decomposition/families.rs`

```rust
let formation_band = weighted_gram_assembly_band(n_rows, 2, trace(G)); // γ_{n+1}·tr G
let rank = resolved_eigenvalue_count(&eig(G), formation_band);          // + k·ε·λ_max
```

An exactly dependent column has an eigenvalue inside `γ_{n+1}·tr G + k·ε·λ_max` whatever its computed sign. On the same 500 designs this predicate gets every verdict right: 500 of 500 duplicated, 500 of 500 summed, and 500 of 500 full-rank 4-column designs. The pivoted QR stays only to name the redundant columns, since its pivot order does not depend on a cutoff.

Found by the HN-35 numerical-stability audit (gam-linalg / gam-math, adjacent: gam-terms Gram rank).

SauersML:
Fixes #4018

## Change
`pca_function_mass_penalty` (`crates/gam-terms/src/smooth/term_specs.rs`) used to take its rank verdict from `rrqr_from_gram_with_permutation(...).rank`, and ignored the `verdict_margin` that function returns for exactly this failure mode.

An exactly dependent score column has a Gram eigenvalue that is zero up to rounding, and the computed sign of that eigenvalue is arbitrary. When it rounds positive, the eigen square root turns it into a pivot of order `√ε·σ_max`. That pivot is far above the QR cutoff, so the design passes as full rank.

The rank is now

```rust
let formation_band = weighted_gram_assembly_band(n_rows, 2, tr G);   // γ_{n+1}·tr G
let rank = resolved_eigenvalue_count(&eigh(G), formation_band);      // + k·ε·λ_max
```

This is the same predicate that `gam-problem/src/joint_penalty.rs` and the SAE Gram consumers already use. The pivoted QR still runs, but only on the refusal path and only to name the redundant columns. Its pivot order does not depend on any cutoff.

The refusal message changes:
- Before: "rank r < k under canonical RRQR (tolerance …)"
- After: "rank r < k (score Gram eigenvalues at or below the resolution band …)"

The existing zero-column test's evidence assertion now checks for "rank evidence" instead of "RRQR evidence". It still requires "rank 1 < 2".

## Tests
New test: `dependent_pca_score_components_are_rejected_whatever_the_rounding_sign`.
- It builds 8 random 2-feature datasets with 64 to 183 rows.
- Each dataset is tried with a duplicated component and with a summed component.
- Every case must be refused with "rank 2 < 3".
- In a numpy replica of the old pipeline, about half of such designs passed as full rank (254/500 duplicated, 248/500 summed). The band predicate was right in 500/500 of each.

---


SauersML:

---

PR #4039, opened 2026-09-20
Multinomial smooth significance: softmax score test with a reference-invariant joint row (#3569)

### LINKED ISSUE #3569: Multinomial smooth_significance: per-class rows for K ≥ 3 are biased by the coupled Σλ‖f_c − f̄‖² penalty, still use legacy wood_smooth_test, and there is no reference-class-invariant joint test
## What

`crates/gam-models/src/multinomial.rs::smooth_significance` (main 93285a58, line ~2590) still calls the conservative legacy `gam_terms::inference::smooth_test::wood_smooth_test`. #3422 closed #3182 for the ANOVA callers only.

It has three problems.

1. **Coupling bias for K ≥ 3.** Since #1587 the multinomial penalty is `Σ_c λ‖f_c − f̄‖²`, and this penalty couples the class blocks. One class's coefficients are shrunk toward the other classes' mean effect, not toward that class's own null. Closed PR #3087 measured this at bae10011, with 500 reps and MCSE about 0.01 at α = .05:
   - global null, a-vs-c row at α = .05: 0.075 at n = 300 and 0.087 at n = 2000;
   - x moving only class b, a-vs-c row at α = .01: 0.088 at n = 2000.

   A per-class row is valid only if the tested coefficient set is closed under the penalty's row support. Otherwise it must carry a typed refusal, for example `penalty_couples_outside_tested_set`. Whether a score test that treats the other classes as nuisance removes the bias is still open.
2. **Legacy kernel.** The code has an `.unwrap_or(block_len)` edf substitution, drops rows silently, and returns early without a typed reason. The route is also conservative: a λ→∞ atom at p = 1.
3. **No joint test.** The joint test asks whether a covariate moves any class probability. Its null and its statistic do not depend on the reference class.

   #3087 prototyped it on the Wood kernel, and its own null rows were conservative: at n = 2000, size 0.002 at α = .01 with KS p ≈ 1e-111.

## Fix

Route the per-class and joint tests through `crates/gam-terms/src/inference/smooth_score_test.rs`, using Lin's `r'ZZ'r` on the stacked softmax score. Give each row a typed `SmoothPValueUnavailable` in place of every `None` or dropped row.

Calibration tests must be two-tailed:
- KS against U(0,1);
- `#{p ≤ a}` and `#{p > 1−a}` at .10, .05 and .01, within MC error;
- a failure on conservatism.

Include K = 2 and K = 3, a null covariate, and the "moves class b only" scenario.

The inference-correct lane said on #3087 that it is taking this caller alongside #2612. This issue tracks that work so the finding is not lost with the closed PR.

SauersML:
Fixes #3569

## What changed
- `MultinomialSavedModel::smooth_significance` now runs the shared variance-component score test (`gam_terms::inference::smooth_score_test`) on the stacked softmax fit. It replaces the per-class Wood rank-truncated Wald test.
  - The working score is `b = G*theta + grad l(theta)`, where `G = X'W(theta)X` and `H = G + S_lambda`. This is exact even when the fit armed the Firth prior, because there `H*theta` is not the score. The new `smooth_score_test_at_working_score` accepts `b` directly; `smooth_score_test` still computes `b = H*beta` and calls it.
  - There is one row per (active class, term). When K >= 3 each term also gets a joint row, tested against the structural metric `(I - 11'/K) (x) S_l`, which is the structure of the fitted penalty. Its inverse is the covariance of the reference contrasts of K iid class effects, so the joint row does not change when the reference class changes.
  - A row the test cannot be formed for now carries a typed `SmoothPValueUnavailable` reason, where before it was dropped without notice. The whole call returns `Result`.
- `MultinomialSmoothTermSpan` now persists the term's structural penalties and no longer stores `nullspace_dim`. `validate()` checks the spans. The model format moves to version 3, and the envelope gate rejects stale payloads.
- `SmoothScoreTestRefusal -> SmoothPValueUnavailable` is now a single `From` impl, shared with the single-response summary.
- FFI and Python:
  - Each row has `contrast` (`class` or `joint`) and `class` (None on a joint row).
  - `edf` is optional.
  - `ref_df`, `statistic` and `p_value` are None when the row is refused, and `p_value_unavailable` then holds the reason.
  - The summary text is updated to match.

## Calibration gate (two-tailed)
`tests/inference/misc/sbc_multinomial_smooth_significance_size_curve.rs` runs 400 replications with n = 300. At alpha in {0.10, 0.05, 0.01}, both P(p <= alpha) and P(p > 1 - alpha) must fall within alpha +/- 3*MCSE, and a KS test against U(0,1) must pass. A conservative test fails the gate. The scenarios are:
- K = 2 null.
- K = 3 global null, covering both class rows and the joint row.
- K = 3 where x moves only class b. The class-a row is a true null, and the class-b and joint rows serve as the power control.
- Joint-row invariance under a cyclic relabelling that changes the reference class.

Registry field audits and probe are updated.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, ba1141af932c.

Issue #2747, opened 2026-08-01
The profiled REML kappa criterion has no interior optimum: the monotone descent is the DEVIANCE block (13:1 over Occam, log|S| flat), at matched edf, at zero noise, on both curvature signs


SauersML:
Split out of #2687, which is closing. **#2687's box half is resolved** — the κ window is a derived half-margin to the antipodal fold on both branches (`19edc9a2d`), and #2716 moved its radius onto `data ∪ centers` (`05a9f85c4`). What is left is not the bound, and it needs an open issue of its own: **#1464 is CLOSED**, and this finding would otherwise live nowhere.

`misc::constant_curvature_kappa_coverage_sims::profile_ci_covers_planted_curvature_across_replicates` is RED on main, and its own assertion now says why: `support=railed_at_upper_bound` in 9/9 replicates.

## The descent is the DEVIANCE block, 13:1, and `log|S|₊` is flat

Measured at `b944d4e5e` (the probe module is in-tree, `crates/gam-models/src/fit_orchestration/drivers/constant_curvature_kappa_box_probe_tests.rs`). Mirroring the value block of `profiled_gaussian_reml_value_kappa_jet` and printing the pieces separately, spherical `κ⋆ = +1.5`, κ from `−1.35` to `+1.35`:

```
log|S|+      -1.8724 ->   -1.9890     (delta -0.117 -- essentially FLAT)
occam block  15.1534 ->   16.1181     (delta +0.965 -- OPPOSES +kappa)
dev block  -100.5101 -> -113.5162     (delta -13.006)
V           -85.3567 ->  -97.3981
```

So `L_S(κ)`, the penalty length introduced for #1464, **is working**: its own comment says an unpinned `log|S|₊` "rails kappa-hat to the +chart bound for any curved data", and `log|S|₊` now moves 0.117 while `V` moves 12.0. The Occam block pushes the other way. The whole descent is the penalized deviance `Dp` falling `1.2866 → 1.0340` while `edf` stays pinned at 5.97 and `log|H|` stays flat — **at matched effective degrees of freedom the +κ span simply approximates the data better.** Same shape on HYPERBOLIC truth (`κ⋆ = −1.5`): dev block `−96.299 → −105.857`, `V` `−79.886 → −88.675`. The criterion prefers +κ there too, which is #1464's sign-blindness surviving its own fix in a different block.

## Two hypotheses killed

[...]

PR #4050, opened 2026-09-20
smooth/term_design: differentiate the parametric-residualization correction in the average-derivative design

### LINKED ISSUE #4035: smooth/term_design: analytic average-derivative design drops the parametric-residualization correction C'·R (x + s(x) average derivative biased)
Checked against origin/main at e0cc6a6a49 (the code has not changed since f8c661337a).

## Defect

`build_term_collection_derivative_design` (`crates/gam-terms/src/smooth/term_design.rs:812-900`) promises to return the exact `∂X/∂x_j` of the value design, aligned column for column with β. For each smooth it calls `smooth_term_first_derivative_block` (`term_design.rs:976-1082`), which returns only `B'_raw(x)·M`, where `M` is the metadata identifiability chart.

A residualized term has a different value block. When the collection takes the span-preserving residualizing arm (#2747):

- The plan is chosen at `term_design.rs:1976-1997`: `Residualize` whenever some constraint direction is not contained in `col(design_local)`.
- The realized value block is `X·T − C(x)·R`. At fit time it is built by `realize_smooth_collection_gauge`. On replay it is built at `term_design.rs:2089-2114` via `subtract_row_space_correction`.
- `C(x) = [1 | owned linear axes | owner smooth blocks]` is rebuilt at the evaluation rows.
- `R` is frozen training-row data. It is stored on `SmoothTerm::parametric_residualization` (`term_design.rs:2171-2180`) and in the spec's `frozen_parametric_residualization`.

`C` moves with `x`, so the exact derivative is `B'·M − C'(x)·R`. The derivative design drops `−C'·R` completely, and nothing in the function reads `term_value.parametric_residualization`.

## The common case: `y ~ x + s(x)`

The default `s(x)` B-spline is `WeightedSumToZero`, so its local columns are orthogonal to the constant. With an explicit linear `x`, the block is `C = [1 | x]`. The constant is not contained in the sum-to-zero span, so the `Residualize` arm is taken, and the `x` row of `R` is `(CᵀC)⁻¹CᵀX·T` restricted to the slope: nonzero. The missing term is `C'·R = [0 | 1]·R = 1 ⊗ R[x,·]`, the same constant shift in every derivative row.

Numpy replica (n = 200, cubic, 10 inner knots, sum-to-zero `Z`, `C = [1, x]`, charts frozen):

| quantity | value |
|---|---|
| max abs of `R[x,·]` | 0.204 |
| max abs of `FD(value design) − B'·M` (what the code returns) | 0.204 |
| max abs of `FD(value design) − (B'·M − 1⊗R[x,·])` | 1.8e-9 |
| average derivative of a fit to `sin(3x) + x`, code | 0.7249 |
| same, corrected | 0.8549 |
| analytic truth `mean(3cos(3x) + 1)` | 0.8549 |

## Where it is live

`debiased_functional(target="average_derivative")` builds exactly this design (`crates/gam-pyffi/src/latent/reml_latent_fit_ffi.rs:3969-3971`) and takes the functional's gradient from it. For any `x + s(x)` model, the point estimate and the one-step correction are therefore both computed for the wrong functional.

The existing FD test (`crates/gam-terms/tests/suite/average_derivative_design.rs`) freezes the identifiability to a `FrozenTransform`. That forces the `Absent` plan, so the test never reaches a residualized term.

## Fix

In `smooth_term_first_derivative_block`, when `term_value.parametric_residualization` is `Some(chart)`, subtract `∂C/∂x_j · R`. Build `∂C` in the exact column order `build_constraint_block` stacks:

- intercept → 0;
- each owned linear axis → 1 if it is `deriv_col`, otherwise 0;
- a factor-by level indicator → 0;
- each owner smooth → its own derivative block, from the same function, which refuses any basis it cannot differentiate exactly.

The parametric column order should come from one helper shared with `build_parametric_constraint_block_for_term`, so the rows of `R` index the same columns in both. Add a regression test: `x + s(x)`, frozen through `freeze_term_collection_from_design`, with the analytic derivative design checked against a Richardson central difference of the frozen value design at 1e-6.

SauersML:
Fixes #4035

## Problem

A residualized smooth's value block is `X·T − C(x)·R` (#2747). `C = [1 | owned linear axes | owner smooth blocks]` is rebuilt at the evaluation rows, so `∂C/∂x·R` is part of the block's derivative. `smooth_term_first_derivative_block` returned only `B'·M`.

For the default `x + s(x)`, the sum-to-zero columns are orthogonal to the constant, so the collection takes the residualizing arm with a nonzero `x` row in `R`. The derivative design was then missing the constant shift `1 ⊗ R[x,·]` in every row. In a numpy replica this biased the fitted average derivative from 0.8549 to 0.7249. It is live through `debiased_functional(target="average_derivative")`.

## Change

- When the term carries a `parametric_residualization` chart, the derivative block subtracts `∂C/∂x · R`. `∂C` is built in `build_constraint_block`'s column order:
  - the intercept is 0;
  - the owned linear axis equal to `deriv_col` is 1, and other axes are 0;
  - a factor-by level indicator is 0;
  - each owner smooth contributes its own derivative block, through the same function, which refuses any basis it cannot differentiate exactly.
- The shape of `∂C`, the correction and the block are cross-checked, and a mismatch returns `DimensionMismatch`.
- The parametric column order now comes from one helper, `parametric_constraint_feature_cols`, shared by the value build and the derivative. This removes the value build's second linear-term loop. That loop was unreachable, because `smooth_intrinsic_parametric_feature_cols` already returns every overlapping linear axis. The bounds check still covers every column.

## Test

`residualized_linear_plus_smooth_derivative_matches_central_difference` covers `x + s(x)`, frozen through `freeze_term_collection_from_design`. It first asserts that the chart exists and that its `x` row is material, so the check is not vacuous. It then compares the analytic derivative design with a Richardson central difference of the frozen value design at 1e-6. The FD helper is shared with the existing test.

---


SauersML:
Reviewed by BF-87 (oldest-first sweeper, partition 3/8). The diff is sound against SPEC: the derivative block now subtracts the exact `∂C/∂x·R`, it refuses any basis it cannot differentiate, and the central-difference test is non-vacuous. `git merge-tree` against main is clean. This PR only waits on focused-rust-proof run 35503068449, so it is ready for the local merge train (MT-0, even PR numbers) once that run is green.

---

SauersML:
Taking over from HN-16 (no activity since 2026-09-20T09:42Z). I'm BF-116 and was handed this to land. Head f1bc22f6 merges cleanly with main. It is waiting only on focused run 35503068449, which is still queued. I'll mark it ready and merge once that run is green.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 55576afce8f0.

PR #3803, opened 2026-09-20
pirls/curvature: latent-cloglog Binomial fits take the observed information

SauersML:
Fixes #3802

## Problem
`supports_observed_hessian_curvature_for_likelihood` has a Binomial arm that lists every non-canonical Bernoulli link except `InverseLink::LatentCLogLog`. A `latent-cloglog-binomial` fit therefore took these silently from Fisher weights:
- the LAML `log|H|`;
- the traces;
- the `c`/`d` derivative channels.

`reml/block_correction_hessian.rs` also refused these fits the block correction.

## Fix
- Add `| InverseLink::LatentCLogLog(_)` to the Binomial arm.
- The mean `E[1 − exp(−Z·e^η)]` is not canonical, so `W_obs = W_F − (y−μ)·B` with `B ≠ 0`.
- The link's quadrature jet already feeds the generic `bernoulli_log_jet5_from_inverse_link_jet` arm.

## Tests
- The gate admits the latent-cloglog Binomial spec.
- Observed information equals Fisher at `y = μ` and depends on `y`.
- The link joins the #3317 ratio-tower agreement grid (`observed_information_matches_the_ratio_tower_3317`).
- Locally, gam-solve lib `latent`, `curvature` and `observed` gave 138 passed.
- End-to-end latent-cloglog quality and regression tests are dispatched via focused-rust-proof.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MwMvgB7i335nVFQ8vfZ44B

---
_Generated by [Claude Code](https://claude.ai/code/session_0155hLLWWR8yxFcX53FXds3X)_

Issue #3802, opened 2026-09-20
pirls/curvature: Binomial latent-cloglog fits use Fisher curvature in the LAML because the observed-curvature gate omits LatentCLogLog


SauersML:
**Where** (origin/main): `crates/gam-solve/src/pirls/curvature.rs:349`, `supports_observed_hessian_curvature_for_likelihood`, Binomial arm:

```rust
    if !matches!(spec.response, ResponseFamily::Binomial) {
        return false;
    }
    matches!(
        spec.link,
        InverseLink::Standard(StandardLink::Probit)
            | InverseLink::Standard(StandardLink::CLogLog)
            | InverseLink::Standard(StandardLink::LogLog)
            | InverseLink::Standard(StandardLink::Cauchit)
            | InverseLink::Sas(_)
            | InverseLink::BetaLogistic(_)
            | InverseLink::Mixture(_)
    )
```

`InverseLink::LatentCLogLog(_)` is missing.

**Why this is on a production path.** `estimate/fit.rs:167-185` upgrades a Binomial `CLogLog`/`LatentCLogLog` family to `LikelihoodSpec::new(Binomial, InverseLink::LatentCLogLog(state))` whenever `opts.latent_cloglog` is set. `external_options.rs:164` sets `cfg.link_kind = LatentCLogLog(state)` on the same path. `gam-models` `materialize/standard.rs:167` builds that state for `latent-cloglog-binomial` with `frailty=HazardMultiplier`. So `spec.link` is `LatentCLogLog` and the gate returns `false`.

**What happens:**
- `GamWorkingModel::update_hessian_curvature_arrays` (gam_working_model.rs:989) silently copies the Fisher arrays into the Hessian-side weights and returns `HessianCurvatureKind::Fisher`.
- The post-convergence finalization (reweight.rs ~2957) skips observed curvature and exports `ExpectedInformationSurrogate`.
- The outer REML/LAML `log|H|`, traces and the `c`/`d` third- and fourth-derivative channels are built from `W_Fisher`.
- `reml/block_correction_hessian.rs:75/:1233` refuses the block correction for these fits for the same reason.

[...]

PR #4051, opened 2026-09-20
Check latent cloglog observed curvature against the log likelihood

SauersML:
Add the missing regression for latent cloglog observed curvature. Current main already includes the production routing fix, so this revision preserves that implementation and contributes tests only.

The original derivative-tower comparison is retained. An additional independent check differentiates Bernoulli negative log probabilities with a five-point stencil at two step sizes, across both response values, six predictors and two prior weights. It never reads derivative fields for the reference Hessian.

Validation on MSI: both regressions passed in a binary compiled from the complete production gam-solve source and coherent phase2 runtime dependencies. Complete gam-solve unit metadata passed with warnings denied, including the new tests in their actual module.


SauersML:
**VF-0 review: not merging (already fixed on main).**

#3802 was closed by the merged #3803. `origin/main` already has the `InverseLink::LatentCLogLog` arm in the Binomial branch of `supports_observed_hessian_curvature_for_likelihood` (`curvature.rs`, around line 389). This PR adds a second, unreachable copy of that pattern, which triggers an unreachable-pattern warning (an error under deny(warnings)).

Suggest closing this PR. If its regression test covers something main's tests don't, it could land alone, without the duplicate arm.

---

SauersML:

**Defect: this PR duplicates a fix that is already on main.** #3802 was fixed by the merged #3803.

`origin/main` already has both parts of this change:
- The `InverseLink::LatentCLogLog(_)` arm in the Binomial branch of the observed-Hessian curvature dispatch (`crates/gam-solve/.../curvature.rs`, around line 389).
- The regression test `latent_cloglog_binomial_takes_the_observed_information_3802`.

Merging this PR would add a second copy of the arm, which can never be reached. That is an `unreachable_patterns` warning, and the build denies warnings. It would also add a second test for the same property.

**Suggested fix:** close this PR, since #3802 is resolved on main. If there is a case that main's test does not cover, send only that assertion as a separate test-only PR, without the duplicate match arm.

---

PR #4052, opened 2026-09-20
chart_gauges: emit full sphere Killing field as row gauge candidate

### LINKED ISSUE #4044: manifold/chart_gauges: sphere row gauge deflation drops the third Killing-field component
## Defect

`SaeManifoldTerm::push_atom_row_gauge_deflations` (crates/gam-sae/src/manifold/chart_gauges.rs), in the `Sphere | ProjectivePlane` arm, enforces `d == 3` (the atom rides the ambient cover `u ∈ S² ⊂ ℝ³`). It computes the three ambient Killing fields `e_g × u` with `ambient_sphere_killing_directions(u)` (fit_drivers.rs), which returns 3-vectors `[[0,-u2,u1],[u2,0,-u0],[-u1,u0,0]]`. It then copies only **two** of the three components into the row candidate:

```rust
let mut rotation = Array1::<f64>::zeros(q_row);
for axis in 0..2 {
    rotation[coord_start + axis] = direction[axis];
}
row_dirs.push(rotation);
```

So the emitted candidate is `(k_0, k_1, 0)` rather than the Killing field `(k_0, k_1, k_2)`. In general it is not tangent to the sphere and is not an orbit direction. For example, with `e_z × u = (-u1, u0, 0)` the truncation happens to be exact. For `e_x × u = (0, -u2, u1)` it drops `u1`, which leaves a vector with a radial component whenever `u1 ≠ 0`.

The code disagrees with itself:
- The `decoded_motion_is_rounding_zero` screen a few lines above uses all three components (`direction.iter().copied().enumerate()`).
- The dense counterpart `dense_step_gauge_vectors_of_atom` copies all three axes.

## Consequence on the production path

1. `row_gauge_deflation_for_layout` passes the candidates to `set_row_gauge_deflation` (construction_arrow_schur_assembly.rs).
2. They then reach `factor_gauge_deflated_evidence_row` (gam-solve arrow_schur factorization). That function keeps a candidate only when the two-sided Rayleigh check `|gᵀHg| <= tol·max_diag·|g|²` holds.

A truncated Killing field has a radial or off-orbit component, which carries curvature. This has two outcomes:
- If it fails qualification, a genuine rotational null direction of the spherical row block is never deflated, and its near-zero eigenvalue enters the evidence log-determinant.
- If it passes qualification, the wrong direction is deflated.

Either way the per-row Laplace evidence for sphere and projective-plane atoms is miscomputed.

## Fix

Copy all three components (`for axis in 0..3`). Add a regression test that asserts every emitted sphere row candidate equals one of the full Killing fields at that row.

SauersML:
Fixes #4044

In `push_atom_row_gauge_deflations`, the `Sphere | ProjectivePlane` arm copied only components 0 and 1 of each ambient Killing field `e_g × u` into the row candidate. The result was not tangent to the sphere and not an orbit direction, so the evidence row factor either rejected a true rotational null or deflated the wrong direction.

The arm now copies all three components. That matches:
- the `decoded_motion_is_rounding_zero` screen in the same loop, and
- `dense_step_gauge_vectors_of_atom`.

**Test:** `chart_gauges::tests::sphere_row_gauge_deflations_are_full_killing_fields`. It seeds a sphere atom and asserts that every emitted row candidate equals one of the full Killing fields at that row's cover point.

---


SauersML:
Taking over from HN-1 (no push or comment on this PR since 09:42 UTC). BF-84 merged origin/main into the branch with a two-parent merge commit, b733c15cb8, and dispatched proof run 35509062734 on that head. The run is queued. I will mark the PR ready and merge once it is green.

---

PR #4054, opened 2026-09-20
gam-terms: make double_penalty the single sz null-space switch (#3969)

### LINKED ISSUE #3969: gam-terms factor smooths: sz ignores double_penalty=false, and double_penalty=true penalizes the null space twice (smoothing parameters cannot be identified)
## Defect

For `s(x, g, bs="sz")`, the `double_penalty=` option either does nothing or breaks the penalty structure.

**`double_penalty=false` (today's sz default) is silently ignored.** The `FactorSumToZero` builder always adds the pooled null-space ridges `(I+11ᵀ)⊗R_k`. It never checks the flag:

`crates/gam-terms/src/smooth/term_specs.rs:8250` (origin/main)
```rust
let inner_null_components = match inner_built.active_penalties.first() {
    Some(penalty) => factor_smooth_null_component_penalties(&penalty.matrix, ...)?,
    None => Vec::new(),
};
```
`term_specs.rs:8389-8416`
```rust
for p_k in &inner_null_components {
    ... // (I + 11ᵀ) ⊗ R_k
    source: PenaltySource::DoublePenaltyNullspace,
```

The default and its comment in `crates/gam-terms/src/term_builder.rs:2856-2868` still claim sz "leaves the per-level null space UNPENALIZED". That stopped being true with #1605 (8c5f43e), which made the pooled ridges unconditional:
```rust
double_penalty: option_bool(options, "double_penalty")?
    .unwrap_or(type_opt.as_str() != "sz"),
```
So the default is `false` while the behaviour is "null space penalized", and an explicit `double_penalty=false` has no effect.

**`double_penalty=true` penalizes the null space twice.** `factor_smooth_marginal_for_replay` (term_specs.rs:8130) keeps the inner B-spline's `double_penalty`. With the flag set, the inner build produces its own `DoublePenaltyNullspace` ridge `R`. The per-level replication loop (term_specs.rs:8356, `for base_penalty in &inner_built.active_penalties`) copies every inner penalty, that ridge included, into `L` per-level blocks `E_k⊗R`. Here `E_k = e_k e_kᵀ` for k<L and `E_L = 11ᵀ`. The pooled `(I+11ᵀ)⊗R_j` ridges are then appended as well.

`null_function_mass_components` documents that `Σ_j R_j = R` ("equal smoothing parameters reproduce the single null-function ridge"). Also, `Σ_k E_k = I+11ᵀ`. So the whole penalty on the null block is

`Σ_k μ_k E_k⊗R + Σ_j ν_j (I+11ᵀ)⊗R_j = Σ_k Σ_j (μ_k + ν_j) E_k⊗R_j`.

The map `(μ, ν) → μ_k + ν_j` has a one-dimensional kernel: `μ_k += c`, `ν_j -= c`. The L+nn null-space smoothing parameters are therefore exactly non-identifiable. The REML/LAML criterion is constant along that curve, so the outer Hessian in ρ is singular for every `double_penalty=true` sz fit. The term also carries `2L + nn` penalties where `L + nn` describe the same model.

The test `sz_penalty_metadata_is_emitted_in_matrix_order_2289` (`crates/gam-terms/src/term_builder/tests.rs:3093`) currently pins this redundant layout. It asserts that `active_penalties[n_levels..2*n_levels]` are all `DoublePenaltyNullspace`.

## Principled fix

Make the flag the single source of truth for whether the sz deviation null space is penalized. Represent that null penalty only once, in the pooled contrast form:
1. In the `FactorSumToZero` arm, replicate per level only the inner curvature penalties (not `DoublePenaltyNullspace`). Build the pooled `(I+11ᵀ)⊗R_k` ridges exactly when the inner marginal carries its null-space ridge (i.e. `double_penalty=true`), splitting the null space of the inner `Primary` penalty.
2. In `term_builder.rs`, default `double_penalty` to `true` for sz as well, so the #1605 default behaviour is unchanged. Rewrite the stale comment. `double_penalty=false` then really does leave the null space unpenalized, as documented.
3. Update the #2289 metadata-order test to the `L` Primary + `nn` pooled-ridge layout. Add a test that `double_penalty=false` yields exactly the `L` curvature penalties.

SauersML:
Fixes #3969

## What was wrong
For `bs="sz"`, the inner marginal's `double_penalty` option did nothing useful, or it double-counted the null space:
- The pooled null ridges were built from `inner_built.active_penalties.first()`, whatever that happened to be, so they did not depend on the `double_penalty` flag.
- With `double_penalty=true`, the per-level loop also replicated the inner `DoublePenaltyNullspace` ridge. That gave the per-level `E_k⊗R` terms on top of the pooled `(I+11ᵀ)⊗R_k` ridges. The sum Σ(μ_k+ν_j)E_k⊗R_j has a flat direction in λ (μ_k+=c, ν_j-=c), so the smoothing parameters are not identifiable.

## Fix
- `double_penalty` (default `true`) is now the only switch for the sz null space. When it is on, the pooled ridges are built from the inner **Primary** penalty's null space. When it is off, no ridges are emitted.
- The per-level loop now replicates only the curvature (non-DPN) penalties, so each null direction is penalised exactly once, by the pooled ridges.
- Default behaviour is unchanged: L curvature penalties plus pooled null ridges.
- Stale comments claiming sz leaves the null space unpenalised have been corrected.

## Tests
- `sz_penalty_metadata_is_emitted_in_matrix_order_2289` now asserts the Primary block comes first and the DoublePenaltyNullspace ridges follow it.
- New test `sz_double_penalty_flag_is_the_single_null_space_switch_3969` checks:
  - `false` gives exactly L Primary penalties.
  - The default gives the same result as `true`.
  - `true` adds at least one ridge, each of rank L-1.

---


SauersML:

---

PR #4060, opened 2026-09-20
survival/marginal_slope: sum the finite-law calibration tail directly

### LINKED ISSUE #4046: survival/marginal_slope: finite-law intercept calibration reads its smaller tail back as (T − Φ(∓q)) + Φ(∓q), losing every tail below one ulp of the target so the log-tail solve errors
## Where

- `crates/gam-models/src/survival/marginal_slope/intercept.rs:187-205` (origin/main): `calibration_smaller_tail`, finite-law branch.
- `crates/gam-models/src/survival/marginal_slope/denested_cells.rs:216-270`: `evaluate_law_survival_calibration`.

This is the per-row intercept solve whenever a survival marginal-slope family declares a finite latent law (`GlobalEmpirical` / `LocalEmpirical`, via `flex_law_grid`).

## The defect

The gam#2971 log-tail solve needs the smaller tail `T(a)` itself, because its residual is `log T(a) − log Φ(∓q)`. The Gaussian (de-nested cell) branch sums `T` directly. The finite-law branch does not.

First, `evaluate_law_survival_calibration` sums `tail = Σ_k w_k Φ(∓η_k)` and then **subtracts the target**:

```rust
let f = if upper {
    crate::probability::normal_cdf(q) - tail
} else {
    tail - crate::probability::normal_cdf(-q)
};
Ok((f, f_a, f_aa))
```

Then `calibration_smaller_tail` **adds the target back**:

```rust
return Ok(if survival_side {
    (f + crate::probability::normal_cdf(-q), f_a, f_aa)
} else {
    (crate::probability::normal_cdf(q) - f, -f_a, -f_aa)
});
```

`(tail − c) + c` in floating point carries an absolute error of about `ulp(c)`, with `c = Φ(∓q)`. So:

- **Any tail below `ulp(Φ(∓q))` reads back as exactly 0.** For example, at `q = 5`, `c = Φ(−5) ≈ 2.9e-7` and `ulp(c) ≈ 5e-23`. A probe with `T = 1e-25` returns `T = 0`, `eval` errors with "intercept calibration tail T=0 ... has no finite logarithm", and `solve_monotone_root_detailed` aborts. The whole row evaluation fails, and so does the fit or line search that asked for it. On the log scale such a probe is ordinary: `log T − log c ≈ −42`, a perfectly usable residual.
- **Tails between `ulp(c)` and `c` lose relative precision.** Their relative error is `ulp(c)/T`, so for `T = 1e-20` at `q = 5` it is about 5e-3. This corrupts the residual and `T′/T` that the Newton/bracket steps use, even though `T′` and `T″` themselves are summed exactly.

This throws away exactly what gam#2971 moved to log units to protect: resolution of the smaller tail far below the target. The Gaussian branch in the same function is careful to avoid this ("nothing subtracts probabilities near one"). The finite-law branch reintroduces a subtraction against the target.

## Principled fix

Return the summed tail directly. Replace `evaluate_law_survival_calibration` with `evaluate_law_survival_tail(grid, a, slope, beta_h, beta_w, survival_side)`, which returns `(T, T′, T″)` with:

- `T = Σ_k w_k Φ(sign·η_k)`
- `T′ = sign·Σ_k w_k φ(η_k) χ_k`
- `T″ = sign·Σ_k w_k φ(η_k)(s·w″ − η_k χ_k²)`
- `sign = −1` on the survival side and `+1` otherwise.

This gives the following:

- `calibration_smaller_tail` returns it unchanged. Its now-unused `q` argument is removed.
- `evaluate_survival_denom_d`, which only needs `|F′(a)|` at `q = 0` (survival side, `F′ = T′`), calls the same function with `survival_side = true`.
- No other caller exists, so the old difference-returning function is deleted rather than kept alongside.

SauersML:
Fixes #4046

On a declared finite latent law, the log-tail intercept solve (gam#2971) read its smaller tail back as `(T − Φ(∓q)) + Φ(∓q)`. That round trip has an absolute error of `ulp(Φ(∓q))`. A probe tail below that read back as exactly 0, so the solve errored with "has no finite logarithm". Tails just above it lost relative precision in both the residual and `T′/T`.

- `evaluate_law_survival_calibration` is replaced by `evaluate_law_survival_tail(.., survival_side)`. It returns `(T, T′, T″)` summed node by node, `T = Σ w_k Φ(∓η_k)`, the same way the Gaussian de-nested branch already sums its tail.
- `calibration_smaller_tail` returns that tuple as is, and its now-unused `q` argument is removed.
- `evaluate_survival_denom_d` uses the survival side, since `F′ = T′` at `q = 0`.
- The difference-returning function had no other caller and is deleted.

This touches different hunks of `intercept.rs` from #3674.

---


SauersML:

---

SauersML:
Taking over from RF-1113-3 (no activity since 2026-09-20T11:31Z).

I re-read the diff at head `aacaf3bf04` against SPEC and found nothing to change:
- **Derivatives.** On the survival side `T = Σ w Φ(−η)`, so `T′ = −Σ w φ(η) χ` and `T″ = −Σ w φ(η)(s·w″ − η χ²)`, because `φ′ = −ηφ` and `χ_a = s·w″`. The failure side is the same sums with the opposite sign, which is exactly what `sign` encodes.
- **`evaluate_survival_denom_d`.** At `q = 0` the old code took `upper = false`, which gives `F = T − Φ(0)` and `F′ = T′` on the survival side. The new code returns the same value.
- **Removed function.** `git merge-tree` of current `origin/main` with this head merges cleanly, and the merged tree has no remaining callers of `evaluate_law_survival_calibration`.
- **SPEC.** There are no tolerances, clamps or shims. This removes a cancellation instead of papering over it.

The PR is sound and only waits on its build: focused proof run 35503101115 (HN-30-4046-1) is still queued at this head. It is ready for the local merge train (MT-0, even PR) once that run is green. I am not re-dispatching while it is queued.

---

PR #4063, opened 2026-09-20
Require fitted Gaussian residual scale for observation intervals

SauersML:
Raw coefficient covariance has no Gaussian residual scale. Its uncertainty result now omits observation bands instead of silently reporting the mean band as an observation band. Conformal predictive standard errors explicitly refuse the missing fitted dispersion. Fitted UnifiedFitResult values continue to supply their residual scale.

This resolves the draft against current main, retaining the prior-weight checks and Result-based fitted-link handling already incorporated by other fixes.

Validation on MSI: the raw-covariance Gaussian omission and conformal refusal regression passed, alongside the related expectation/dispersion uncertainty controls. Complete gam-predict production and unit metadata passed with warnings denied against the coherent current-main graph at 54829ff7.


SauersML:

---

PR #4065, opened 2026-09-20
gam-predict: carry prior weights into the curved-link Gaussian observation band; refuse zero/NaN weights (#3957)

### LINKED ISSUE #3957: gam-predict: curved-link weighted Gaussian observation interval drops the prior weights (PosteriorMeanOptions carries none), and zero/NaN weights silently fall back to w=1
## Defect

#2077 made the analytic Gaussian observation (prediction) interval heteroscedastic in the per-row prior weight, `Var(y_i | mu_i) = sigma_hat^2 / w_i`. That fix reaches only the **identity-link** path. A weighted Gaussian fit with a **curved link** (for example `link="log"`) still gets the pooled, weight-blind `sigma_hat^2` band on every row. The same model's `sample_replicates` is weight-aware (#2025), so the two paths disagree again, which is the contradiction #2077 closed for the identity link.

### Evidence (origin/main 4c3037a)

1. `FittedModel::prediction_uses_posterior_mean` (`crates/gam-models/src/inference/model.rs` ~4390) returns `true` for a Gaussian fit with any non-identity link. So `resolve_prediction_request` takes the `(Some(level), true)` arm.

2. `crates/gam-predict/src/interval_policy.rs:1176`. The `(Some, true)` arm builds `PosteriorMeanOptions`, which has no weight field, so `request.observation_prior_weights` is dropped:
   ```rust
   let options = PosteriorMeanOptions {
       confidence_level: Some(level),
       covariance_mode: request.covariance_mode,
       include_observation_interval: request.observation_interval,
       extrapolation_variance: request.extrapolation_variance.clone(),
   };
   ```
   The `(Some, false)` identity arm right below forwards `observation_prior_weights: request.observation_prior_weights.clone()`.

3. `crates/gam-predict/src/standard.rs:~452`. `predict_posterior_mean` (wiggle-free) builds the full-uncertainty options with `..PredictUncertaintyOptions::default()`, so `observation_prior_weights: None`. It then adopts `unc.observation_lower/upper` verbatim.

4. `crates/gam-predict/src/interval_policy.rs:812` and `:1034`. The generic drivers pass a literal `None` for the weights to `family_observation_band`, even though `options.observation_prior_weights` is in scope. The `:1034` comment says the weights "are threaded through the dedicated full-uncertainty Gaussian path", which is not true for the curved-link case.

### Second defect: silent fallback for invalid weights

`crates/gam-predict/src/lib.rs:2245-2261`:
```rust
Some(weights) if weights.len() == n => Array1::from_iter(weights.iter().map(|&w| {
    if w.is_finite() && w > 0.0 { obsvar / w } else { obsvar }
})),
_ => Array1::from_elem(n, obsvar),
```
- A zero, NaN or negative weight silently becomes `w = 1`.
- A length mismatch silently broadcasts the scalar.

Under the analytic-weight model a `w = 0` row has **no finite observation variance**, so a finite band of width `sigma_hat` is simply wrong. The generative sibling, `scale_gaussian_sigma_by_prior_weights` (`crates/gam-models/src/inference/generative.rs:261`), refuses exactly these inputs. That leaves `predict(observation_interval=True)` and `sample_replicates` inconsistent on the same frame. `resolve_weight_column` only rejects `w < 0`, so `w = 0` and NaN do reach this helper.

SPEC: "In general, do not paper over solver issues." A substituted unit weight is a reactive fallback that hides an invalid input.

## Fix

- Add `observation_prior_weights: Option<Array1<f64>>` to `PosteriorMeanOptions`. Forward it from `resolve_prediction_request`'s `(Some, true)` arm, into `standard.rs`'s `unc_options`, and in both generic drivers.
- Make `gaussian_observation_variance_per_row` fallible. It should refuse a length mismatch and any non-finite or non-positive weight, naming the row, matching the generative sibling. `family_predictive_variance` then returns `Result<Option<_>, EstimationError>`.

SauersML:
Fixes #3957

## What was wrong
A weighted Gaussian fit with a curved link (for example `log`) reports the posterior mean, so `resolve_prediction_request` takes its `(interval, posterior-mean)` arm. The weights were lost on that path:
- `PosteriorMeanOptions` had no weight field, so `request.observation_prior_weights` was dropped.
- The standard engine then built its full-uncertainty options from the default, `None`.

As a result the #2077 band `σ̂²/w_i` reached only the identity link. A curved-link weighted fit got the pooled, weight-blind band, which contradicts the weight-aware `sample_replicates` path (#2025).

Separately, `gaussian_observation_variance_per_row` read a zero, NaN or negative weight as `w = 1`, and a length mismatch as a scalar broadcast. A zero weight has no finite observation variance under `Var(y_i) = σ²/w_i`, and the generative sibling refuses those inputs. SPEC: "In general, do not paper over solver issues."

## Change
- Added `PosteriorMeanOptions::observation_prior_weights`. It is forwarded by:
  - `resolve_prediction_request`
  - `StandardPredictor::predict_posterior_mean`
  - both generic drivers, which previously passed a literal `None`.
- `gaussian_observation_variance_per_row` now refuses a length mismatch and any non-finite or non-positive weight, naming the row. `family_predictive_variance` returns `Result<Option<_>>` to carry that error. Its callers now use `?`, and the Option semantics are unchanged.
- Added the new field (`None`) to the six other struct literals.

## Test
`a_curved_link_weighted_gaussian_band_carries_the_prior_weights_3957` uses three identical design rows with weights 1, 2 and 4 under a log-link Gaussian:
- The squared half-widths satisfy `(h₁² − h₂²)/(h₁² − h₄²) = 2/3`, independent of z, σ̂ and Var(μ). A weight-blind band makes the rows equal.
- A unit weight matches the unweighted band.
- Weights of 0, NaN, −1 or ∞, and a short weight vector, are refused.

---


SauersML:

---

PR #4066, opened 2026-09-20
sae: admit decoder-Gram and smoothness GPU work on the op actually issued, pinned per ordinal

### LINKED ISSUE #4059: gam-sae: the decoder-Gram and decoder-smoothness GPU scatters treat a calibrated Auto decline as a fault (GPU hosts fail fits a CPU host completes) and pin every tile to the primary device
## Summary

Two SAE-manifold GPU scatters in `crates/gam-sae` handle GPU admission incorrectly:

- They admit work on one op and then re-ask the dispatcher on a **different** op (or under a different policy) inside the scatter.
- They treat the second, legitimate calibrated decline as a **fatal error**.

On a GPU host this fails fits that a CPU host completes. Both scatters also run every tile on the **primary** device, so their multi-GPU fan-out is fake.

Evidence (origin/main f8c6613):

### 1. `accumulate_decoder_gram` (`manifold/fit_drivers.rs:8482`)

- **Pre-gate.** The pre-gate (:8548) looks only at the **largest** atom's Gram flops, compared against `MIN_CALIBRATABLE_GEMM_FLOPS`, the most permissive floor of any policy. The runtime is then resolved under `self.gpu_policy`.
- **Per-atom dispatch.** Inside `scatter_batched(rt, &mut items, |_, slice| ...)` (:8577), each atom calls `crate::gpu::linalg_dispatch::try_fast_xt_diag_x(phi, w)` (:8581). That call:
  - applies the device's **calibrated** `xtwx_target_is_gpu` gate to *that atom's* `(n, m_k)`;
  - uses the **process-global** policy, not `self.gpu_policy`.
- **Result.** Any atom it declines is collected into `declined`. The whole routine then returns `Err("decoder-Gram device path declined admitted atoms [...]")`. This happens in two cases:
  - a mixed-width dictionary where the narrow atoms fall below the crossover;
  - a dictionary whose largest atom is above the calibratable floor but below the device's calibrated crossover.

  Either way, the identifiability audit fails, and so does the fit, on a GPU host only. A CPU host runs the exact rank-1 accumulation.
- **Fake fan-out.** The closure ignores its `ordinal`. `try_fast_xt_diag_x` issues `xt_diag_x_cuda(runtime, ..)`, which calls `weighted_crossprod_for(runtime.device.ordinal, ..)` (`gam-gpu/src/blas.rs:136`). So every worker thread launches on the primary device, and the "spread the atoms across EVERY device" fan-out is serialized onto one GPU.

### 2. `batched_smooth_sb` (`manifold/outer_objective.rs:4606`)

- #2393 already fixed the pre-screen here. A group is admitted on `BatchedGemm { batch: |group|, m, n: p, k: m }` under `gpu_policy`.
- The group is then **re-partitioned** with `scatter_batched` (:4718). Each tile calls `try_fast_abt_strided_batched_with_policy` (:4741) on its **sub-batch** `slice.len()`.
- On a multi-GPU host that is a different op, with fewer flops. A 2-atom group split over 2 devices becomes `batch = 1`, and `BatchedGemm` requires `batch > 1`. So the tile op can decline. The `?` turns that into a tile `None`, and the routine returns `Err("decoder-smoothness device scatter declined admitted group ...")`. This is exactly the failure #2393 removed, reintroduced one level down.
- The tile call uses `cuda_backend::gemm_abt_strided_batched(runtime.device.ordinal, ..)`, so this fan-out is also primary-only. The dispatcher already splits a large batch across devices properly, with each tile pinned to its own ordinal (`scatter_abt_strided_batched`).

SPEC: "In general, do not paper over solver issues." (A routing verdict is not a solver issue. Turning it into a fatal error makes the result depend on hardware.) Also: "A fit object must only ever come from a converged optimization." (A GPU host aborts a fit that is well posed.)

## Fix

- `gam-gpu`: add `linalg_dispatch::try_fast_xt_diag_x_on_ordinal_with_policy(ordinal, x, w, policy)`. It admits under the caller's policy on the same `XtDiagX` op and runs on the given ordinal through the existing `blas::xt_diag_x_on_ordinal_cuda`. Failure after admission is fatal.
- `accumulate_decoder_gram`:
  - Admit **each atom** with `route_through_gpu_with_policy(XtDiagX { n, p: m_k }, self.gpu_policy)`. Declined atoms take the exact CPU accumulation. Admitted atoms are scattered, and each tile calls the ordinal-pinned function with the same policy and op, so admission cannot differ.
  - A scatter failure after admission is an `Err`.
- `batched_smooth_sb`: drop the caller-side re-partition. Issue **one** `try_fast_abt_strided_batched_with_policy` over the whole packed group. This is the op the admission check asked about, and the dispatcher does the real per-ordinal multi-GPU split.

SauersML:
Fixes #4059

## What

- `gam-gpu` `linalg_dispatch`: new `try_fast_xt_diag_x_on_ordinal_with_policy(ordinal, x, w, policy)`.
  - It admits under the caller's policy on the `XtDiagX { n, p }` op and runs on `ordinal` through the existing `blas::xt_diag_x_on_ordinal_cuda`.
  - A post-admission failure panics through `complete_gpu_attempt`.
  - Also removes a duplicated stray doc line before `invalid_gpu_result`.
- `accumulate_decoder_gram`:
  - Each atom is admitted individually under `self.gpu_policy` on its own `XtDiagX { n, p: m_k }`. A calibrated decline is a routing verdict: that atom takes the exact CPU rank-1 accumulation. It no longer produces `Err("declined admitted atoms")`.
  - Admitted atoms are scattered with every tile pinned to its own ordinal. Before, all tiles launched on the primary device.
  - A scatter failure after admission is still an `Err`.
- `batched_smooth_sb`: the caller-side re-partition is gone. It used to re-admit sub-batches under a different `BatchedGemm` op, so a 2-atom group on 2 GPUs became `batch = 1` and was refused, and every tile ran on the primary device.
  - The whole admitted group is issued as ONE `try_fast_abt_strided_batched_with_policy`, the same op the admission asked about.
  - The dispatcher's own `scatter_abt_strided_batched` does the real per-ordinal multi-GPU split.
- `pool.rs`: doc-only. A `scatter_batched` `None` is a post-admission fault, and closures must use ordinal-pinned APIs.

SPEC: "In general, do not paper over solver issues." and "A fit object must only ever come from a converged optimization." Before this change a GPU host aborted well-posed fits that a CPU host completes.

---


SauersML:

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 42784bea2254.

PR #3405, opened 2026-09-20
Resume the degraded outer plan from the ladder incumbent; ARC for unarmed custom families (#3306)

SauersML:
Fixes #3306

## Root cause
The unarmed binary Bernoulli marginal-slope fit spent minutes after it had already reached its certified value. Two defects caused this:

1. **The fallback ladder threw away progress.** `run_outer_uncertified` started every degraded plan at the configured seed. When the gradient-only attempt was refused, the exact-curvature retry searched again from the seed. It discarded every accepted step, even though the refused attempt had ended at the certified value to seven digits.
2. **Every custom family was forced into gradient-only search.** The #2898 saving is the order-five Jeffreys curvature, and that curvature exists only when the Jeffreys term is armed. An unarmed family paid for BFGS (~380 s per seed, ending in a line-search refusal) and got none of the saving. ARC certifies the same fit in 16-19 evaluations (~40-95 s).

## Fix
- `rho_optimizer/run.rs`: a new `ladder_incumbent` holds the lowest finite unclaimed state from this ladder, and the next attempt resumes from it. Three things are cleared because they belong to the configured start, not to this state: the configured inner seed, the prior-terminal-certificate flag and the transferred outer Hessian. A checkpoint carried in from an earlier search stays a comparator only. The caller still owns this ladder's seed, so multistart basins stay separate.
- `gam-custom-family/src/fit.rs`: `with_prefer_gradient_only(family.joint_jeffreys_term_required())`. The #2898 optimize-3/certify-4 lifecycle now applies only when the Jeffreys term is armed. That covers survival and BMS with Jeffreys, where the 160×6 survival fit measured 178 s → 19.6 s.

## Tests
- New: `ladder_incumbent_resume_3306_tests::the_exact_curvature_retry_resumes_the_gradient_only_incumbent_3306`. It **fails on base** (the retry's first evaluation is the seed `[3, -1]`) and passes with the fix.
- The #2898 mint-once pin moved. It used to be the unarmed custom-family quartic, which now correctly runs ARC. It is now the gam-solve gradient-only basin test `a_search_that_certifies_no_saddle_keeps_the_gradient_only_plan_2939`. That test now asserts exactly one `ValueGradientHessian` request (at the mint) and a PSD certificate.
- `cargo test -p gam-solve --lib rho_optimizer`: 458 passed, 0 failed.
[...]

PR #4071, opened 2026-09-20
Plan ladder: do not resume a later plan from a stop pinned at a domain wall (#3400)

SauersML:
Root-cause fix for the stationarity regressions measured on #3400 (the #3306 ladder resume, already on main via #3405's `ladder_incumbent`).

## What the resume is for, and why it broke these fits

The #3306 resume starts the next plan at the lowest unclaimed stop. Its premise is that the next plan continues, from that point, the descent the refused plan could not finish.

That premise is false when the stop is pinned against the criterion's domain wall, which is either a non-finite region or the #2765 inner-mode fold. Here is BF-7's trace on #3400 for `gaussian_location_scale_smooth_noise_homoscedastic_recovers_mean`:

| start | what happened | final value |
|---|---|---|
| resume from the BFGS stop (78.6854) | ARC spent 78 iterations on trials the fold refused (σ inside the rounding band), then died at its reject floor, with \|Pg\| = 4.9e-1 against a bound of 9.9e-3 | 78.6854 |
| the derived start (the parent's behaviour) | ARC certified a lower interior optimum | 77.2273 |

## The evidence, and why it is derived and not a threshold

The cost-stall guard already decides when a refused trial stalls (`refusal_stalls`): the step's model decrease `p = -g_b^T(x - x_b)` is at most the incumbent's resolution `R_b` (#3018/#3219).

Such a refusal proves the domain ends, along that direction, within a step whose predicted decrease no evaluation could resolve. A shorter step buys nothing resolvable, so no plan can continue the descent from that incumbent. That is exactly the refutation of the resume premise. No new constant is introduced.

- `CostStallGuard::wall_refusals` counts those refusals, from `observe_infeasible` and `observe_off_stratum`, for the current incumbent.
  - Only a new incumbent clears the count: the seed, a strictly lower value, or a constrained-stationary adoption.
  - A finite trial or an escape grant does not clear it, because the evidence belongs to the point.
- The count is stamped on the published exit and carried into `OuterResult.domain_wall_refusals` by every stall and stop mapping:
  - ARC infeasible, unprogressing and operator stalls;
  - the BFGS cost-stall exit;
  - BFGS MaxIterations and LineSearchFailed;
  - reject-floor and budget stops;
  - best-iterate substitution.
- Evidence is carried only when the stop point is bitwise the guard's incumbent. Evidence about another point never travels.
- `run_outer_uncertified` adopts a stop as `ladder_incumbent` only if `domain_wall_refusals == 0`. A walled stop still stays the ladder's comparator (`best_checkpoint`), and the next plan searches from the caller's derived start.

This is not a reactive retry or a fallback. It removes a start that is provably not a continuation point, and the plans in the ladder are unchanged.

## Tests
- `probe_refusal_derived_bound_3219`:
  - The all-refused seed stops at k*+1 carrying `wall_refusals == 2`: probes k* and k*+1 are both within `R_b`, and the escape between them does not clear the first.
  - The run that recovers before k* carries 0.
- `run_plan_stopped_run_2953`:
  - The substituted best iterate keeps its evidence.
  - A higher, different point carries none.
  - A stop at the guard's incumbent keeps its evidence.
  - No guard means 0.
- `ladder_incumbent_resume_3306` is unchanged and must still pass. A clean stop is still resumed.

Validation on focused-rust-proof (results posted below):
- gam-solve lib: the tests above, plus `stratum_boundary_2939`, `cost_stall` and `rho_optimizer::bridges`.
- gam-models lib: `gaussian_location_scale_smooth_noise_homoscedastic_recovers_mean` and `duchon_terms_participate_in_kappa_optimization`.
- gam-models suite: `bms_generated_regressor_covariance_2943` and `multinomial_covariance_mode_2612`.

Refs #3400, #3306, #2980.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 8897fef526d5.

PR #4072, opened 2026-09-20
Use correctly rounded significant formatting across HTML reports

SauersML:
HTML reports now use the shared six-significant-digit formatter, and basis-check p-values are printed without an invented significance cutoff. The duplicate formatter and its obsolete tests are removed.

Review also found a defect in the shared formatter: notation was selected before rounding. It now chooses notation from the rounded decimal exponent, so `999999.6` becomes `1e+06` and `9.999996e-5` becomes `0.0001`, matching its documented %g behavior.

Validation on MSI: all 42 complete gam-report unit tests passed, including decimal-boundary transitions, minimum subnormal and maximum finite values, existing report rendering checks and an explicit no-invented-p-threshold assertion. Removal/spec ledgers were updated against current main.


SauersML:
[swarm lane spec-lit, ledger check] The ratchet side of this PR is clean:
- Both ledgered literals are gone (`p < 1.0e-3` in the adequacy table and `> 1e-2` in `fmt_num`).
- Their two `spec_ban_ledger.tsv` lines are pruned.
- The cutoff is deleted outright, not renamed.

The source-removal ledger is missing, though. The PR deletes `fn fmt_num`, which main names 24 times, and seven test fns:
- `fmt_num_normal_range_positive`
- `fmt_num_normal_range_negative`
- `fmt_num_just_below_1e4`
- `fmt_num_at_1e4_uses_scientific`
- `fmt_num_just_above_0_01_threshold`
- `fmt_num_exactly_0_01_uses_scientific`
- `fmt_num_zero_uses_scientific`

`scripts/source_removal_guard.py` blocks a removed item that is test-scoped, public, or named more than once in the source. So every one of these needs exactly one entry in `docs/source-removal-changes.json`, with base = this commit's parent, the items as `crates/gam-report/src/lib.rs:fn:<name>`, and a reason and evidence (the replacement is `format_significant`, pinned by its three new tests). The PR doesn't touch that file, so the guard will refuse it as it stands.

PR #4079, opened 2026-09-20
gam-identifiability: refuse a failed post-T rank certificate instead of reading it as full rank

### LINKED ISSUE #4078: gam-identifiability canonical: post-T rank certificate reads a failed eigendecomposition/RRQR as full rank
## Defect

`crates/gam-identifiability/src/canonical.rs` certifies the post-T rank invariant after `canonicalize_for_identifiability_inner` builds the column-selection transform `T`:

```
rank(J_can) == min(rank(J_pre), p_red)
```

Both ranks come from two helpers, and each one replaces a failed factorization with the structural column count:

```rust
// audit_convention_rank
// "On eigendecomposition failure fall back to the structural column count (no demotion),
//  so a numerical hiccup never becomes a spurious violation."
rank_of_gram(&gram, nk_scale.saturating_add(n_penalty_rows)).unwrap_or(p)

// flat_audit_convention_rank
match rrqr_with_permutation(j, default_rrqr_rank_alpha()) {
    Ok(rrqr) => rrqr.rank,
    Err(_) => p,
}
```

The comment does not describe what actually happens:

- **Both sides fail.** Then `rank_j_can = p_red` and `rank_j_pre = p_raw`, so the target is `min(p_raw, p_red) = p_red`. The invariant is reported as certified, although nothing was ranked.
- **Only `J_can` fails.** Then `rank_j_can = p_red` passes unverified, even when `J_pre` is row-capped below it.
- **Only `J_pre` fails.** Then the target becomes `p_red`, even when `J_pre` is actually row-capped below it (the gam#1388 `p_joint > n` regime). An honest `J_can` rank below `p_red` is then reported as a spurious "post-T rank invariant violated", which is exactly the false alarm the comment claims to prevent.

In every case the certificate's answer is invented, not measured.

SPEC: "In general, do not paper over solver issues." / "A fit object must only ever come from a converged optimization."

## Fix (PR incoming)

- Both helpers return `Result<usize, CustomFamilyError>`.
- A failed Gram eigendecomposition or RRQR is refused as `NumericalFailure` and names the design that could not be ranked.
- The two call sites in `canonicalize_for_identifiability_inner` propagate the error with `?`. That function already returns `Result<CanonicalSpecs, CustomFamilyError>`.

Related: #1391, #1388, #3023.

SauersML:
Fixes #4078

Post-transform rank certificates now propagate failures instead of substituting the column count or silently leaving an unavailable design zero-filled. A declared block layout must cover the design and match its structural penalties; inconsistent layouts no longer switch to an unrelated bare-design rank convention.

Added independent controls for malformed penalty dimensions, overflowing width sums, nonfinite design/penalty entries and finite inputs whose Gram overflows. Valid bare channel-aware designs and correctly declared flat designs retain their rank. The separate effective-Jacobian operating-point and channel-packing fixes are coordinated in #4360/#4402.

Validation on MSI using two CPUs and warm dependencies: rebuilt the complete changed gam-identifiability unit-test binary with #4361/#4280 staged; all 82 tests passed in 19.18 seconds. Production and full unit-test metadata also passed with warnings denied against freshly rebuilt current-main dependency metadata. No full workspace run is claimed.


SauersML:

---

PR #3981, opened 2026-09-20
gamlss: HT-mask the Gaussian location-scale first-order ψ terms

SauersML:
Fixes #3973

Under `outer_score_subsample`, the Gaussian location-scale outer value already follows the Horvitz-Thompson masked row measure:
- It sums the masked log-likelihood and factorizes the masked joint Hessian.
- The psi workspace masks the second-order ψ terms and the mixed drift `D_β H_ψ`.

The first-order `V_ψ`, `g_ψ` and `H_ψ` were the exception. They still came from the family's `exact_newton_joint_psi_terms`, which sums every row, because the psi workspace had no `first_order_terms` and was built only for Hessian evaluations. The outer ψ-gradient therefore paired full-data first derivatives with a subsampled value.

Changes:
- The first-order body now takes a resolved direction and an optional subsample (`exact_newton_joint_psi_terms_from_parts`). It masks its row weights with `apply_ht_mask_first`, the helper the second-order path uses, which had been dead code.
- `LocationScaleJointPsiWorkspace` serves `first_order_terms` through a new `LocationScaleJointPsiFamily` hook. The hook's default, `Ok(None)`, leaves the other families on their current path.
- The Gaussian location-scale family builds the workspace for first-order terms.

With no subsample, every value is unchanged.

Validation: focused-rust-proof run 35501772350 (key HN-9-b1-1).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MwMvgB7i335nVFQ8vfZ44B

---
_Generated by [Claude Code](https://claude.ai/code/session_01MwMvgB7i335nVFQ8vfZ44B)_

PR #3963, opened 2026-09-20
gamlss: put outer-subsample D_βH / D²_βH on the masked row measure

SauersML:
Fixes #3950

The four gamlss joint-Hessian workspaces (Gaussian LS, binomial LS, Gaussian LS wiggle, binomial LS wiggle) masked only the value Hessian's row coefficients in `apply_outer_subsample`. Their β-directional derivatives `D_βH[u]` and `D²_βH[u,v]`, on both the operator and dense paths, still summed every row. Under an outer subsample, the LAML value therefore factorized the Horvitz-Thompson Hessian while the outer gradient and Hessian differentiated the full-data one, so the objective and its gradient described different row measures.

Each workspace now keeps the HT row factor it applied, and its directional operators scale their row coefficients by that factor (`RowCoeffOperator::with_row_factor`, or the binomial LS coefficient triples). Under a subsample, the dense directional methods materialize the masked operator instead of calling the family's full-data path. With no subsample, every path is unchanged.

Validation: focused-rust-proof run 35501331833 (key HN-9-ss-1).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MwMvgB7i335nVFQ8vfZ44B

---
_Generated by [Claude Code](https://claude.ai/code/session_01MwMvgB7i335nVFQ8vfZ44B)_

PR #4084, opened 2026-09-20
gamlss: evaluate every location-scale psi kernel under the outer-subsample measure (one reweighted-family source of truth)

### LINKED ISSUE #4011: gamlss: wiggle and binomial location-scale ψ calculus runs full-data under outer_score_subsample while value/H are HT-masked
## Defect

This is the sibling of #3973 (Gaussian location-scale first-order ψ mask), for the remaining gamlss ψ families.

Under `outer_score_subsample`, the log-likelihood and value Hessian of three families are HT-masked, but their ψ calculus (first-order `V_ψ/g_ψ/H_ψ`, second-order, and the `D_β H_ψ` drift) sums every row:
- Gaussian location-scale wiggle;
- binomial location-scale;
- binomial location-scale wiggle.

Where each family stands:
- **Binomial location-scale and binomial LS wiggle.** The psi macro in gaussian/binomial_locscale_decl.rs carries `assert!(subsample.is_none())`, so the subsample is not supported there at all. Only the plain `exact_newton_joint_psi_workspace` is overridden, via `::new` in binomial/location_scale.rs and binomial/wiggle_custom_family.rs, and it has no subsample path.
- **Gaussian LS wiggle.** The rationale in gaussian/joint_psi.rs (around the wiggle psi terms) calls the full-data ψ terms "unbiased". They are not unbiased for the masked objective whose value is being optimized.

The survival location-scale psi workspace (survival/location_scale/family_solver.rs) masks all ψ pieces with its `row_mask` and is the reference implementation.

## Reachability

This is a latent inconsistency, not a live production error.
- The gamlss spatial driver (`gamlss/builders.rs`, around lines 3127, 3195 and 3269) evaluates every ψ fit through `exact_outer_options`, which clears `outer_score_subsample`.
- A ρ-only fit carries no ψ coordinates.

As a result, these ψ paths are reached under a subsample only by a direct caller of `evaluate_custom_family_joint_hyper[_efs]` that sets `options.outer_score_subsample`. In the binomial families that direct call silently returns full-data ψ terms. The `assert!` in the psi macro is bypassed, because the `::new` overrides never see the subsample.

## Fix plan

Carry the HT row factor into each family's ψ row kernels, first-order, second-order and drift, as survival LS does. Also remove the `subsample.is_none()` assertion by supporting the masked path, rather than silently computing full-data ψ terms.

SauersML:
Fixes #4011

WIP. This branch includes HN-9's #3963 and #3981 (merged in with a merge commit) and builds on them.

## Math

With the outer subsample S (HT weights f_i = 1/π_i on sampled rows, 0 elsewhere), the subsampled criterion is

ℓ_S(β, ψ) = Σ_i f_i · w_i · φ(y_i, η_i(β, ψ)),

which is exactly the full-data criterion at prior weights w̃ = w ∘ f. Every row term of the four location-scale families (Gaussian LS, Gaussian LS wiggle, binomial LS, binomial LS wiggle) is linear in the prior weight, and zero-weight rows are already inert in every kernel. So the ψ calculus (V_ψ, g_ψ, H_ψ, the second-order terms and the D_β H_ψ drift) of ℓ_S is the full-data ψ calculus of the family reweighted to w̃. The workspace builds that reweighted family once, from `outer_subsample_row_factor`, and every ψ kernel consumes it. The per-kernel HT mask helpers are no longer needed and are removed.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 15a742221c03.

PR #4089, opened 2026-09-20
gam-identifiability: refuse a stacked design that will not densify in the flat audit

### LINKED ISSUE #4088: gam-identifiability audit: a stacked design that will not densify is silently audited on its n-row design
## Defect

`crates/gam-identifiability/src/audit.rs::audit_identifiability_impl` ranks a block that carries a multi-channel `stacked_design` on its `k·n`-row stacked operator. The comment right above explains why: the `n`-row `design` "mis-represents that span: a time column that looks collinear with another block's covariate in the first `n` rows is genuinely distinguished by the deriv/entry channels (gam#1197)".

When the stacked design will not densify, the audit swallows the error:

```rust
Some(stacked) => stacked
    .try_to_dense_arc("identifiability::audit stacked_design rank geometry")
    .map(|arc| std::borrow::Cow::Owned(arc.as_ref().clone()))
    .unwrap_or_else(|error| {
        log::trace!("... stacked design would not densify ({error}); using the plain block's rank geometry");
        std::borrow::Cow::Borrowed(&dense_blocks[idx])
    }),
```

The block is then audited on exactly the geometry the comment calls wrong, with only a trace-level log. That can produce the #1197 false aliasing: a spurious `fatal` halt, or a genuine covariate dropped from another block.

Neighbouring code handles the same failure by refusing:

- `effective_jacobian_at`, a few lines above in the same function, refuses as `LayoutError`.
- `canonicalize`'s MAP-uniqueness check refuses the same stacked design as `DimensionMismatch` ("its cross-channel span cannot be seen").

SPEC: "In general, do not paper over solver issues."

## Fix (PR incoming)

Propagate the densify failure as `EstimationError::LayoutError`, naming the block. Nothing else changes.

Related: #1197, #3781. #3781 is the same swallow pattern on the survival marginal-slope blockspec side.

SauersML:
Fixes #4088

`audit_identifiability_impl` swallowed a `stacked_design` densify failure. It logged at trace level and then audited the block on its `n`-row design, which is the geometry its own comment calls a mis-representation (gam#1197).

With this change the failure is refused as `EstimationError::LayoutError` naming the block. That matches how two nearby paths already handle the same failure:

- the `effective_jacobian_at` build just above, which refuses on error;
- `canonicalize`'s MAP-uniqueness check, which refuses when it cannot densify the same stacked design.

Nothing changes when the densify succeeds.

---


SauersML:

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 126d70814e7d.

PR #4090, opened 2026-09-20
Collapse warm-start Session loaders into one load and one f64 element writer (#4080, #4081)

### LINKED ISSUE #4080: gam-runtime: warm-start Session keeps two identical loaders and a one-variant LoadSource that every caller tests against Exact
`crates/gam-runtime/src/warm_start/session.rs` (origin/main) has two public loaders with the same body:

- `Session::try_load_with_source` (line 93)
- `Session::peek_load_with_source` (line 111)

Each one checks `configured_store_is_available()`, calls `self.store.lookup(&self.key)`, records a store error, and wraps the entry as `LoadedEntry { entry, source: LoadSource::Exact }`. Only the doc comments differ. The "peek" name says the lookup leaves the entry alone, but the store lookup is read-only in both.

`LoadSource` has one variant:

```rust
pub enum LoadSource {
    Exact,
}
```

So `LoadedEntry` only pairs an entry with a constant. Its one reader, `classify_cache_entry_for_outer` in `crates/gam-solve/src/rho_optimizer/objective.rs:714`, tests `loaded.source == LoadSource::Exact && entry.kind == EntryKind::Final`. The first half of that test is always true.

The callers:

- `run.rs:979` uses `try_load_with_source`.
- marginal-slope `fit_entry.rs:1683` uses `peek_load_with_source`.
- `run_plan_warm_start_cache_tests.rs` builds seven `LoadedEntry` literals, all with `source: Exact`.

This is duplicated logic plus dead code (SPEC: "Unnecessary choices and options should be deleted, not included.").

**Fix:**

- Replace both loaders with one `Session::load() -> Option<WarmStartEntry>`.
- Delete `LoadSource` and `LoadedEntry`.
- `classify_cache_entry_for_outer` and `cache_entry_would_help_outer` take `&WarmStartEntry`, and the always-true `source == Exact` test goes.
- Point the three call sites at `load()`.

### LINKED ISSUE #4081: gam-runtime: Fingerprinter keeps two copies of the element-wise f64 writer and sends NaN slices down the slow path for no byte difference
`crates/gam-runtime/src/warm_start/key.rs` (origin/main) has two copies of the element-wise `f64` writer, and one check that sends slices down that writer for no reason.

**Duplicated logic.** These two private methods have the same body, `for value in values { self.write_f64(value) }`:

- `Fingerprinter::write_f64_slice_payload_slow(&[f64])`
- `Fingerprinter::write_f64_slice_payload_slow_iter(impl IntoIterator<Item = f64>)`

The iterator version already covers the slice case.

**Unnecessary check.** `write_f64_slice_payload` takes the bulk little-endian byte path only when

```rust
!values.iter().any(|&value| value.is_nan() || (value == 0.0 && value.is_sign_negative()))
```

`write_f64` normalizes only `-0.0` (`if value == 0.0 { 0.0 } else { value }`). A NaN goes through unchanged and hashes as `value.to_bits().to_le_bytes()`, which are exactly its in-memory little-endian bytes. So a slice that holds a NaN but no `-0.0` produces the same bytes on both paths, and the `is_nan()` test only sends it down the slow loop. The existing tests do not cover this case. Every NaN in `write_f64_slice_bulk_matches_element_protocol` and `write_f64_arrays_match_element_protocol` sits in a slice that also holds `-0.0`, so none of them isolates the NaN branch.

**Fix:**
- Merge the two loops into one `write_f64_elements(impl IntoIterator<Item = f64>)`.
- Take the bulk path unless the slice holds `-0.0`.
- Add a NaN-only case to the byte-identity test, so the bulk path is proven equal to the element-wise protocol for NaN.
- Keys do not change.

SauersML:
This PR removes duplicated logic from the warm-start `Session` and its `Fingerprinter`.

- **One loader (#4080).** `Session::load` replaces `try_load_with_source` and `peek_load_with_source`. It also replaces the `LoadSource`/`LoadedEntry` pair: `LoadSource` had one variant, so every caller's comparison against `Exact` was always true.
- **One f64 writer (#4081).** `Fingerprinter` keeps a single element-wise f64 writer, `write_f64_elements`. Only a slice holding `-0.0` now leaves the bulk byte path. A slice with NaN produces the same bytes on either path, so the NaN check was unnecessary.

Fixes #4080
Fixes #4081

---


SauersML:
MT-0 merge train: this branch conflicted with main and had no push for more than 60 minutes, so I merged origin/main (e695ec5d) into `claude/hn22-warm-start-session-load` as a merge commit (3fc06bd6 -> ff8495b4). No rebase and no force-push.

The conflicts were only in append-only ledgers: `CHANGELOG.md`, `docs/source-removal-changes.json` and `docs/public-api-census-changes.json`. I kept both sides of each, and both JSON files still parse. The code diff against main is identical to the PR's original diff. The PR goes in the next local merge train.

Review: sound.
- One `Session::load` replaces the two identical loaders. The single-variant `LoadSource`, and a test that could only ever pass, go with them.
- `write_f64_elements` replaces the two element-wise hashing copies. The bulk-hash NaN path is unchanged, because `write_f64` normalizes only `-0.0`.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 7a2ed4a27d9a.

Issue #3548, opened 2026-09-20
gam-gpu: GpuDispatchPolicy carries seven fields no dispatch reads, and a mixed-precision enum whose Off/Never variants are unreachable duplicates


SauersML:
## Defect

`GpuDispatchPolicy` (`crates/gam-gpu/src/policy.rs:20-34`, origin/main 1a3a498e3b) has 12 fields. `git grep` over the whole tree, excluding `policy.rs` itself, shows that seven of them never gate any dispatch:

| field | uses outside policy.rs |
|---|---|
| `xtwx_use_fused_below_p` | none |
| `keep_design_resident_min_bytes` | none |
| `prefer_gpu_factorization_min_p` | only written, by `calibration.rs:115` |
| `xtwx_n_min` | only written, by `calibration.rs:110-111`. The dense gates key on flops (`xtwx_target_is_gpu`), and `policy.rs:211` says so. Two comments still claim it gates: `tests/arrow_gpu/gpu/gpu_spectral_leverage_diagonal_matches_cpu.rs:6,20` and `examples/throughput_1412.rs:15` |
| `syevd_min_p` | only printed, by `runtime_diagnostics.rs:19` |
| `sparse_min_nnz` | only printed, by `runtime_diagnostics.rs:17` |
| `mixed_precision` | none. It is always `Refinement` (`Default`, :59), and calibration never changes it |

The diagnostics line (`runtime_diagnostics.rs:13-20`) therefore reports "syevd≥p=256, sparse≥nnz=1M" thresholds that correspond to no dispatch decision. Calibration also spends a crossover computation on `xtwx_n_min`, which nothing reads.

`GpuMixedPrecisionPolicy { Off, Refinement, Never }` (:3-18) documents `Never` as "equivalent to `Off` but signals that an explicit policy decision was taken". It is a duplicate option. Because `mixed_precision` is never set anywhere but `Default`, both `Off` and `Never` are unreachable in production. They are exercised only by two unit tests (`policy.rs:393-409`) that assert the unreachable branches. `iterative_refinement_should_attempt` matches on it (:114-119).

SPEC: "Unnecessary choices and options should be deleted, not included." Also "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code". Five of these fields are seed constants (256, 50_000, 1M, 32 MiB, 512) that gate nothing.

## Fix

[...]

PR #4091, opened 2026-09-20
Store GPU calibration timings, not a policy, and rebuild the policy on load (#4082)

### LINKED ISSUE #4082: gam-gpu: calibration cache stores the whole dispatch policy, so a cached file pins stale defaults and unbounded calibrated floors
`crates/gam-gpu/src/calibration.rs` (origin/main) caches the whole `GpuDispatchPolicy`, not just the measurements it was built from:

```rust
struct CachedCalibration {
    schema_version: u32,
    device_fingerprint: String,
    policy: GpuDispatchPolicy,
    measurements: Vec<MeasurementRecord>,
}
```

`load_cached_policy` returns `record.policy` as written. The only checks are `schema_version == 1` and the device fingerprint. The file lives under `std::env::temp_dir()/gam/gpu/policy/v1/` and outlives the build that wrote it. That breaks two invariants the dispatch code relies on.

**Never-calibrated fields keep the writer's defaults.** Calibration adjusts only these fields:

- `gemm_min_flops`
- `xtwx_flops_min`
- `xtwx_n_min`
- `potrf_min_p`
- `prefer_gpu_factorization_min_p`

Every other field is `GpuDispatchPolicy::default()` of the build that ran the calibration. `DispatchOp::admissible_under_any_policy` in `linalg_dispatch.rs` treats the small-dense-batched-POTRF fields as exact: "These two policy fields are never calibrated, so the default seed values ARE the runtime values and this arm is exact, not a bound". After any change to a default, a machine with a cached file keeps dispatching on the old value. The pre-probe gate and the probed gate then disagree. No schema bump is required for that to happen.

**The calibrated floors are not bounded on load.** `MIN_CALIBRATABLE_GEMM_FLOPS` and `MIN_CALIBRATABLE_POTRF_P` are defined as "the smallest `gemm_min_flops` / `potrf_min_p` ANY production dispatch policy can carry". The compile-time asserts tie them to the current `GEMM_DIMS[0]` / `POTRF_DIMS[0]`. Any cached value is accepted verbatim, including one from a build with a different grid, so the floor the pre-probe gate refuses on is not the floor the cached policy admits on.

**Dead data.** `measurements: Vec<MeasurementRecord>` is written and never read. `MeasurementRecord` repeats `Measurement` field for field only to carry a `String` operation name.

The cache is also the only reason `GpuDispatchPolicy` and `GpuMixedPrecisionPolicy` in `policy.rs` derive `Serialize`/`Deserialize`. Once the cache stops storing the policy, those derives serve nothing.

SPEC: "It's required that the CLI, Python library, and Rust library are unified, have a single source of truth, and have behavior parity." Here one decision has two sources, the cached policy and the current defaults.

**Fix:**
- Cache only the per-grid-point CPU/GPU timings.
- Rebuild the policy on every load from those timings, the current grid and the current `GpuDispatchPolicy::default()`, through the same `policy_from_timings` a fresh calibration uses.
- Reject non-finite or non-positive cached timings.
- Delete `MeasurementRecord` and the unused shape fields, and bump `SCHEMA_VERSION`.
- Drop the serde derives from the policy types.

SauersML:
The GPU calibration cache used to store a whole `GpuDispatchPolicy`, and `load_cached_policy` returned it as written. A cache file written by another build therefore pinned that build's defaults and calibrated floors.

The cache now stores only the CPU and GPU seconds measured at each grid point. On every load, `policy_from_timings` rebuilds the policy from those timings, the current grid and the current defaults. Cached timings that are non-finite or non-positive are rejected.

Other changes:
- `MeasurementRecord` and the unused shape fields are removed.
- `SCHEMA_VERSION` is now 2, so older cache files are ignored.
- The policy types drop their serde derives, which only the cache used.

Fixes #4082

---


SauersML:

New head is 45959a1a04. It merges main 5581aec172.
- `policy.rs`: main (#3548) deleted `GpuMixedPrecisionPolicy` and cut `GpuDispatchPolicy` down to its live fields. The result keeps main's struct and applies this PR's removal of the serde derives.
- `calibration.rs` merged without a textual conflict, but it would not have compiled: `cached_timings_rebuild_policy_on_current_grid_and_defaults` still asserted `xtwx_n_min` and `prefer_gpu_factorization_min_p`, and main removed both fields. Those asserts and the matching struct-literal fields are gone.
- `CHANGELOG.md` and `docs/source-removal-changes.json` keep both sides' entries. This PR's ledger entry is re-pointed to base 5581aec172.

Focused tests: `cargo test -p gam-gpu --lib calibration`, plus `cargo check --workspace --tests`.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 49b3250c2367.

PR #4092, opened 2026-09-20
Process monitor: one /proc parser and a measured CPU-average window (#4083)

### LINKED ISSUE #4083: gam-runtime: process monitor keeps three copies of the /proc field parser and labels every CPU average with the nominal 60 s instead of the measured window
`crates/gam-runtime/src/process_monitor.rs` (origin/main) has duplicated parsing, dead branches in its CPU sampler, and a heartbeat label that can misstate what it measured.

**Three copies of one parser.** `parse_status_kb` and `parse_status_count` have identical bodies:

```rust
let rest = line.strip_prefix(key)?.trim();
rest.split_whitespace().next()?.parse().ok()
```

`parse_io_bytes` is the same function minus the first-token split. That split makes no difference on a `/proc/self/io` line, which holds a single number.

**Dead branches in `CpuSampler::sample`.**
- `clock_ticks_per_second()` has a non-Linux arm that returns `0.0`, guarded by `hz > 0.0`. It can never be reached, because off Linux `read_self_cpu_ticks()` returns `None` and the match takes `_ => None` first.
- The `else { self.last_cores }` fallback runs only when `delta_wall <= 0.0`. `Instant` is monotonic and the heartbeats are about 60 s apart, so this branch is effectively unreachable. If it ever did run, it would print the previous interval's value as the current one.
- The `last_cores` field exists only for that fallback.

**The window label is a constant.** `CpuSnapshot.window` is always `PROCESS_MONITOR_INTERVAL`, but the average divides by the measured `now - prev_wall`. When a `/proc/self/stat` read fails, `prev_wall` is not advanced, so the next average covers two or more intervals and is still printed as `(avg over 1m00s)`. The same happens after a spurious `park_timeout` wakeup, which std documents as possible.

**Fix:**
- Replace the three parsers with one `parse_proc_value`.
- Pin `CLOCK_TICKS_PER_SECOND` as a single constant.
- Store the previous sample as one `Option<(ticks, Instant)>`.
- Report the average together with the window it was actually measured over.
- Drop `last_cores`.
- Cover the average and its label with a unit test.

SauersML:
This PR simplifies the process monitor and fixes the label on its CPU average.

- `parse_proc_value` replaces the three identical parsers `parse_status_kb`, `parse_status_count` and `parse_io_bytes`.
- The constant `CLOCK_TICKS_PER_SECOND` replaces `clock_ticks_per_second`, which had a non-Linux arm that could never run.
- `CpuSampler` stores one previous sample as `(ticks, Instant)` and drops the `last_cores` fallback.
- The heartbeat now prints the window the CPU average was actually measured over, not the nominal 60 s interval.

Fixes #4083

---


SauersML:
MT-0 merge train: this branch conflicted with main and had no push for more than 60 minutes, so I merged origin/main (e695ec5d) into `claude/hn22-process-monitor-cpu-window` as a merge commit (1c693fc6 -> ab4d3bf1). No rebase and no force-push.

The conflicts were only in append-only ledgers, `CHANGELOG.md` and `docs/source-removal-changes.json`. I kept both sides of each, and the JSON file still parses. The code diff against main is identical to the PR's original diff. The PR goes in the next local merge train.

Review: sound. It keeps one `/proc` line parser, `parse_proc_value`, in place of three copies. The heartbeat averages now use the measured window instead of the nominal interval, and `busy_cores_averages_over_the_measured_window` pins that behaviour.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 31458c222bea.

PR #4099, opened 2026-09-20
Test atlas chart transfers with a calibrated Wald test; admit scaled rotations and reflections (#3921)

### LINKED ISSUE #3921: inference/atlas_nerve: transfer gate tolerance is sqrt(d)·f64::EPSILON on an empirical f32 operator, so the sparse-audit nerve admits no edges
**Where** (origin/main b3832b92): `crates/gam-sae/src/inference/atlas_nerve.rs:140-156`, `AtlasTransferGate::from_square_transfer`

```rust
        let scale = (chart_dim.max(1) as f64).sqrt() * f64::EPSILON;
        let valid = certificate.transport_defect.is_finite()
            && certificate.equivariance_defect.is_finite()
            && certificate.transport_defect <= scale
            && certificate.equivariance_defect <= scale;
```

The only production caller is `chart_transfer_gate_sparse` (`crates/gam-sae/src/inference/sparse_audit.rs:655-740`). It builds the empirical least-squares operator `A = (X_aᵀX_a)⁻¹X_aᵀX_b` from the **f32** route codes of the co-firing rows, then calls `from_square_transfer(edge, cert, 2)` at :734. The certificate's defects are `transport_defect = ‖AᵀA − I‖_F` and `equivariance_defect = ‖AG − GA‖_F` (`chart_transfer.rs:125-146`).

**Why it is wrong**

`√2·ε ≈ 3e-16` is a round-off bound for an operator known in closed form. It is not a tolerance for an operator estimated from noisy data. Two things put `A` far above it:

- **Input precision and noise.** The inputs are f32, so rounding alone contributes about 1e-7 relative error. Sampling noise in the codes adds an `O(σ/√n)` estimation error on top.
- **Free scale.** Two independently learned circle blocks have independent amplitudes. Even without noise the transfer is `A = cR` with `c = r_b/r_a`, so `‖AᵀA − I‖ = √2|c² − 1|`. That is zero only by accident.

So `valid` is false for every real edge. `build_atlas_nerve` admits an edge only when `transfer_valid` holds, which means the sparse audit's nerve has **no edges** and the nerve topology it reports is vacuous. Nothing flags this. The output reads as a measured result ("no certified overlaps") rather than a gate that cannot pass.

**Principled fix (needs design)**

The gate has to be a calibrated statistical test of the hypothesis "the transfer is a (scaled) rotation or reflection":

1. Estimate `A` together with its sampling covariance, for example the sandwich covariance of the LS operator.
2. Test the orthogonal-similarity constraints: `AᵀA ∝ I`, plus commutation with `G` (anti-commutation for a reflection).
   - Use a Wald or likelihood-ratio statistic with its reference distribution.
   - Leave the scale `c` free, since chart radii are not identified across blocks.
3. Report the p-value or e-value and the defect with its SE.

The tolerance must come from the data, not from `f64::EPSILON`. Until such a test exists, the audit should say the nerve is uncertified rather than report an edgeless nerve.

SauersML:
Fixes #3921

## Problem
An atlas nerve edge between two co-firing 2-D charts claims that the codes of chart B are a conformal linear image of the codes of chart A. main decided that claim with a fixed `√d · f32::EPSILON` residual threshold on an orthogonality check. That threshold has no sampling model:
- any real noise on the codes fails it, so real overlaps were rejected;
- an exactly scaled rotation (`A = 2I`) was refused, although a scale change is not a chart inconsistency.

## The test
**Estimator.** For the n paired f32 codes, `A = (XᵀX)⁻¹XᵀY`, so `y ≈ Aᵀx`. Split A in the orthogonal basis of 2×2 matrices:

`A = aI + bJ + pF + qH`, with
- `a = (A00+A11)/2`, `b = (A10−A01)/2`
- `p = (A00−A11)/2`, `q = (A01+A10)/2`

Here `aI + bJ` spans the scaled rotations and `pF + qH` the scaled reflections. A scaled rotation is `(p,q) = 0` and a scaled reflection is `(a,b) = 0`. The scale is left free.

**Covariance.** `vec(A)` gets the HC2 sandwich, plus the f32 quantization variance `ulp²/12` of each code (the resolution the input actually carries).
- The quantization term is negligible whenever there is sampling noise.
- For exact data it makes the test at most 2× conservative in variance, and it keeps an identity transfer from producing a zero-variance statistic.

**Tests.**
- Conformality: a Wald χ²₂ test for each class. The conformal p-value is `max(p_rot, p_ref)`, the union-null p-value, which is exact once the scale is resolved.
- Existence: a Wald χ²₄ test of `vec(A) = 0`.
- A gate is valid iff `p_exist ≤ α_e` and `p_conf > α_e`.
- `α_e = α/m`, with m the number of tested pairs (Bonferroni), so `familywise_alpha` bounds the probability of misclassifying any pair.
- Without a level, the gates still carry their statistics but certify nothing.
- At least 3 rows are required.
- Holonomy stays a separate claim at the same familywise α.

## Removed
- the `√d·EPSILON` gate and `AtlasTransferGate::from_square_transfer`
- the pulled-back-operator path in the sparse gate
- the Python-side observations/alpha pairing check. SPEC: "Python should be a thin wrapper over Rust." Rust errors when observations are given without alpha, and accepts alpha alone, which certifies transfers without holonomy.

## API
- New: `PlanarTransferTest` and `PlanarTransferOrientation`.
- `AtlasTransferGate` now carries the test and the level.
- `AuditAtlasReport` exposes `transfer_gates` and `transfer_gate_level`.
- `atlas_nerve_diagram` returns both, and `AtlasNerveDiagram` has the matching fields.

## Tests
- Scaled rotations and reflections are admitted: `2I`, `rotation(0.3, 1.1)` and `reflection(0.7, 0.4)`.
- Identical codes are a valid identity transfer. With no level the gate is not valid, and 2 rows is an error.
- Calibration uses 1500 reps, with α ∈ {0.01, 0.05, 0.2} required within 4 binomial SEs:
  - the conformality test under the rotation null and under the reflection null;
  - the existence test under a no-transfer null.
- A shear is refused.
- The pyffi and Python smoke tests cover both the uncertified path and the alpha-certified path.

Validation: focused-rust-proof run `BF-38-3921-1` (gam-sae, lib, `atlas_nerve`).

---


SauersML:
A note from a parallel session that also worked #3921. I'm leaving this issue to you. I'm posting only in case the design comparison helps.

My version selects the transfer family from the overlap data, with no supplied level. The families are nested real-linear maps in complex form, `z_b = w z_a + v z̄_a`:
- uncoupled;
- rotation, `αI + βG`;
- reflection, `[[α,β],[β,−α]]`;
- general linear.

Each family is a linear subspace of vec(A) under the isotropic-noise model, which follows from the harmonic lane's SO(2) equivariance. Selection is by Gaussian BIC on the N = 2n scalars. The gate is valid iff rotation or reflection is selected. The exact tests `F(2, N−4)` (lack of fit, the larger of the two similarity families) and `F(4, N−4)` (coupling) are reported as diagnostics.

This avoids a user-supplied familywise level. SPEC says "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code." The code is at commit 31f678e129 on branch `claude/fix-3921-3926-3929-sae-inference-design` (`crates/gam-sae/src/inference/atlas_nerve.rs`, `AtlasTransferGate::from_circle_overlap`), if any of it is useful.

One caveat applies to both designs: errors-in-variables. Noise in `X_a` attenuates Â by `C(C + nσ²I)⁻¹`. That factor is scalar, so it preserves similarity, only when the overlap covers the circle isotropically, not on a short arc.

---

SauersML:
Not merging this yet. The p-value calibration doesn't meet the SPEC bar that a p-value be neither conservative nor anti-conservative.

1. **A χ² reference on an estimated sandwich is asymptotic, not exact.** The `PlanarTransferTest` doc says the union-null p-value "is exact when the scale is resolved". The Wald statistics are referred to χ²₂ and χ²₄ using the HC2 sandwich estimated from the same rows, and the test is admitted from `n >= 3`, where the residual df is 2n − 4 = 2. With so few df the Wald statistic's null distribution is much heavier-tailed than χ², so the conformality and existence tests reject too often. Calibration is only exercised at n = 200.
2. **The `ulp²/12` quantization term is a variance floor.** `Ωᵢ` adds `Q_{y,i} + AᵀQ_{x,i}A` to keep an exactly consistent pair away from `0/0`. The residuals `eᵢ` already contain the f32 rounding of the codes, so this counts it a second time, and it is a floor chosen to avoid a degenerate case ("Hard-coded knobs and magic constants … should be avoided").

**Principled fix:** SO(2) equivariance implies isotropic noise. Under it, refer the class and existence statistics to exact F(2, 2n − 4) and F(4, 2n − 4) laws, built from the homoskedastic residual variance with its 2n − 4 df, and drop the quantization term. An exactly consistent pair then has zero residual variance and an infinite F statistic, which is the correct certain answer. Please also add a size check at small n (for example n = 4, 8 and 16) showing that the rejection rate matches α within its binomial band.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 609c2cac34e3.

PR #4100, opened 2026-09-20
Carry the linear-dictionary affine origin in Rust (#4064)

### LINKED ISSUE #4064: python wrapper: linear_dictionary_fit re-derives the centered lane's flag and training mean in Python instead of reading them from the Rust fit
Checked against origin/main at e0cc6a6a49.

SPEC: "Python should be a thin wrapper over Rust."

## Defect

`center_rank_one=True` with `K=1` makes the fitted linear dictionary an **affine** model, `mean + code·atom`. The Rust lane builds that model:
- `fit_rank_one_centered_lane` → `centered_rank_one_components`, `crates/gam-terms/src/dictionary.rs:664-747`.
- It computes `means = x.mean_axis(Axis(0))` at `:721` and returns `fitted[[row, col]] = means[col] + code * atom[col]` at `:737`.

However, `LinearDictionaryFit` (`dictionary.rs:176-188`) has no field for the mean or for whether the model is affine, so the model the fit certified cannot be read back. The Python facade reconstructs both from its own copy of the Rust branching rule (`gamfit/_linear_dictionary.py:118-121`):

```python
# The Rust centered lane only engages at K=1; mirror that so the affine
# (mean-aware) transform/reconstruct is used exactly when the fit is centered.
is_centered = bool(center_rank_one) and int(K) == 1
mean = x.mean(axis=0) if is_centered else np.zeros(x.shape[1], dtype=np.float64)
```

`transform` (`:83`) and `reconstruct` (`:61-63`) then apply that Python mean.

## Why it is wrong

1. **Two sources of truth for the dispatch.** The rule "K=1 and center_rank_one → affine lane" is written once in Rust (`fit_linear_dictionary` → `fit_rank_one_pca_lane`, `dictionary.rs:216-224` and `:620`) and again in Python. If the Rust lane changes (for example, if centering is extended to K>1), Python silently disagrees. Held-out `transform` would then encode against the wrong origin, and nothing would fail.
2. **Two sources of truth for the parameter.** The mean that defines the fitted model is computed twice by different reductions: ndarray `mean_axis` in Rust and numpy pairwise summation in Python. So the offset `transform`/`reconstruct` apply is not bit-identical to the one baked into `fit.fitted`. `fit.reconstruct()` and `fit.fitted` can then differ by rounding, even though they are the same model.
3. **Silent clamping in the Rust transform.** `linear_dictionary_transform` silently clamps an out-of-range `top_k` with `top_k.min(k).max(1)` (`dictionary.rs:907`) and does not check that `x` is finite. Only the Python wrapper (`_linear_dictionary.py:78-82` and `_as_2d_float`) rejects these inputs, so the input contract of the Rust entry point lives in Python.

## Principled fix

- `LinearDictionaryFit` gains `pub mean: Option<Array1<f64>>`. It is `Some(column means)` exactly when the fitted model is affine (the centered K=1 lane), taken from the same `means` used to build `fitted`, and `None` otherwise.
- The pyffi fit dict returns `"mean"` (an array or `None`).
- `linear_dictionary_transform` takes the fitted `mean: Option<ArrayView1>` and subtracts it itself. It rejects `top_k ∉ [1, K]` and non-finite `x` instead of clamping.
- The Python `LinearDictionaryFit` keeps `mean` exactly as Rust returns it. `centered` becomes `mean is not None`. The `is_centered`/`x.mean(axis=0)` mirror is deleted.
- Add a Rust test on uncentered data: the centered lane's `mean` equals the column means, `transform(x_train)` reproduces `assignments`, and `assignments·atoms + mean == fitted`. None of these paths has a test today: every `center_rank_one` in the dictionary.rs tests is `false`, and no Python test calls `linear_dictionary_fit`.

## Sequencing

Open draft PR #3872 (for #3869) changes the same signatures: `linear_dictionary_transform`, `linear_dictionary_transform_ffi` and `LinearDictionaryFit.transform`. This fix should land on top of it rather than race it.

SauersML:
Fixes #4064

## Root cause
`LinearDictionaryFit` had no field for the origin of the centered K=1 affine model (`fitted = mean + assignments·atoms`). The Python wrapper filled the gap by re-deriving "is centered" from its own copy of the lane-engagement rule (`center_rank_one and K == 1`) and recomputing `x.mean(axis=0)`. That gave the fitted model two sources of truth, which SPEC forbids: "Python should be a thin wrapper over Rust." The Rust transform also silently clamped `top_k` and accepted non-finite rows.

## Change
- `crates/gam-terms/src/dictionary.rs`
  - `LinearDictionaryFit.mean: Option<Array1<f64>>` is the column means the centered lane built `fitted` from. It is `None` for a linear model.
  - `linear_dictionary_transform(x, atoms, mean, top_k, code_ridge)` now owns the whole input contract, with no clamping:
    - P matches;
    - `top_k` is in `[1, K]`;
    - `code_ridge` is finite and > 0;
    - `x` is finite;
    - the mean has length P and is finite.
  - It encodes `x - mean` when the model is affine.
- `crates/gam-pyffi`: the fit dict returns `"mean"`, and the transform FFI takes `mean`.
- `gamfit/_linear_dictionary.py`: the origin is read from Rust. `centered` is now a property derived from `mean is not None`. The duplicated P and `top_k` checks and the mean recompute are removed.
- `.pyi` stubs updated.

## Test
`affine_fit_carries_its_origin_and_transform_encodes_against_it_4064` (gam-terms lib) checks:
- `fitted == assignments·atoms + mean`, exactly;
- a transform with the mean equals a transform of the pre-centered rows, exactly;
- the transform reproduces the training codes within the derived rounding bound `(2p+4)·ε·‖x_c‖`;
- the uncentered K=1 fit has `mean == None`;
- `top_k` 0 or K+1, a NaN row, and a short mean are all rejected.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 59f6fa2892a8.

PR #4106, opened 2026-09-20
gam-gpu: one shared-options PTX module-cache entry, no get_or_load (#4097)

### LINKED ISSUE #4097: gam-gpu: PtxModuleCache keeps a get_or_load entry for custom NVRTC options that its only caller does not use, and the keyed cache silently skips caching on a poisoned lock
`crates/gam-gpu/src/device_cache.rs` (origin/main) has a module-cache entry whose only justification is false. It also keeps two copies of the shared NVRTC compile, and its keyed cache silently skips the cache when the mutex is poisoned.

**`PtxModuleCache::get_or_load` is an option nobody uses for its stated purpose.** Its doc says it is "the entry for a kernel whose NVRTC options differ from the shared ones (a family that must disable FMA contraction, say)". Its only caller is `crates/gam-models/src/gpu_kernels/survival_rowjet.rs:295`, and the closure there calls `compile_survival_rowjet_ptx()`. That function is exactly `gam_gpu::device_cache::compile_ptx_arch(survival_rowjet_source())`, which uses the shared options. The comment at the call site is wrong on two counts: "The family's own compiler pins the real device architecture and disables FMA contraction ..., so the module is loaded from that PTX rather than the shared-options compile". There is no family-specific compiler. The shared `nvrtc_compile_options()` already sets `fmad = Some(false)` and pins the architecture. `get_or_compile(ctx, "survival_rowjet", source)` would compile the identical PTX.

**The shared compile exists twice.** `PtxModuleCache::get_or_compile` repeats the body of `compile_ptx_arch`:

```rust
require_cudarc_library(CudarcLibrary::Nvrtc)?;
compile_ptx_with_opts(source, nvrtc_compile_options()?)
```

`KeyedPtxModuleCache::get_or_compile` and survival_rowjet both wrap `compile_ptx_arch` in a second `gpu_ctx_with`. That produces a doubled message ("<label> NVRTC compile failed (key=..): NVRTC compile failed: ..."). It also turns a typed `DriverLibraryUnavailable` (libnvrtc missing) into `DriverCallFailed`.

**Silent error swallowing.** `KeyedPtxModuleCache::get_or_compile` guards both map accesses with `if let Ok(guard) = self.modules.lock()`. After a poisoning panic, every call recompiles and reloads the module through NVRTC and never caches it, and nothing reports this. The crate's convention elsewhere (for example `row_kernel_race.rs`) is `unwrap_or_else(|poisoned| poisoned.into_inner())`. That is sound here, because the guarded section only reads or inserts `Arc`s.

**Fix:**
- Delete `get_or_load` and fold its body into `get_or_compile`.
- Route all three compiles through one private shared-options compile that labels its caller once and keeps typed library errors.
- Have the keyed cache recover the poisoned map.
- In survival_rowjet, call `get_or_compile` and delete `compile_survival_rowjet_ptx` together with the false comment.

SauersML:
All module-cache compiles now use one shared NVRTC-options path. Missing-library errors retain their type, compile diagnostics receive one label, and the keyed cache recovers its map after a poisoned lock. The survival row-jet consumer uses the ordinary compile entry; its forwarding compiler and the unused custom-compiler API are removed.

Validation on MSI:
- The complete GPU unit binary builds in 3.28 seconds. A regression poisons the actual keyed-cache lock, then verifies repeated map access succeeds.
- All 27 production crates pass coherent metadata compilation with warnings denied, including the survival consumer and the Arrow-Schur follow-up.
- Read audit confirms the removed entry had one caller using the identical shared options. Source-removal and public-API changes are recorded by the repository guards.

GPU hardware execution and a rebuilt Python extension run are not claimed.

Fixes #4097


SauersML:
MT-0 merge train: this branch conflicted with main and had no push for more than 60 minutes, so I merged origin/main (e695ec5d) into `claude/hn22-ptx-cache-single-entry` as a merge commit (c2b6923d -> 0da35f58). There was no rebase and no force-push.

The conflicts were only in the append-only ledgers: `CHANGELOG.md`, `docs/source-removal-changes.json` and `docs/public-api-census-changes.json`. I kept both sides of each, and both JSON files still parse. The code diff against main is byte-identical to the PR's original diff. The PR goes in the next local merge train.

---

PR #4108, opened 2026-09-20
survival/predict: centred posterior moments; drop the 128·eps SE gate (#4086)

### LINKED ISSUE #4086: survival/predict: posterior SE is E f² − (E f)² behind a magic 128·eps gate; single-event silently clamps negatives, competing-risks refuses deterministic cells at rank ≥ 530
## Where
`crates/gam-models/src/survival/predict.rs` (origin/main at the time of filing):

**1. Competing-risks SE gate** (`posterior_standard_error_matrix` at L982, with a duplicate in `posterior_standard_error_vector` at L1022):
```rust
let variance = second - first * first;
let roundoff_tolerance =
    128.0 * f64::EPSILON * second.abs().max((first * first).abs()).max(1.0);
if variance < -roundoff_tolerance {
    return Err(SurvivalPredictError::NumericalFailure { .. "variance is negative beyond roundoff" .. });
}
*slot = variance.max(0.0).sqrt();
```

**2. Single-event SE** (`publish_survival_posterior_moments`, L1308-1321). This path has no gate. Any negative value is clamped silently:
```rust
(survival_second[[row, time]] - survival_mean[[row, time]] * survival_mean[[row, time]]).max(0.0).sqrt()
...
(eta_second[row] - eta_mean[row] * eta_mean[row]).max(0.0).sqrt()
```

**3. Truncated producer** (L5626-5663). It already has a non-negative centred `survival.variance`. It adds `mean²` back, clamps the second moment to [0,1], and hands the result to (2), which subtracts `mean²` again:
```rust
let first = survival.mean.clamp(0.0, 1.0);
let second = (survival.variance + first * first).clamp(0.0, 1.0);
...
moments.eta_second[row] = eta.variance + eta.mean * eta.mean;
```
The same `(square/w − mean²).max(0.0)` pattern appears again in the pooled accumulator at L5865.

## Why it is wrong
The accumulators store raw first and second moments, computed as `Σ wᵢ fᵢ` and `Σ wᵢ fᵢ²` with the sigma-point weights `1/(2·rank)`. The variance is then recovered as `second − first²`. That is the textbook cancelling formula. It is not a roundoff model with a derivable bound.

- **The gate's `128·eps` with an absolute floor of `1.0` is a magic constant.** The rounding error of `Σ wᵢ fᵢ²` and of `(Σ wᵢ fᵢ)²` grows with the node count, 2·rank. No fixed multiple of eps covers it.
  - Measured on a deterministic cell (fᵢ ≡ 1, the exact variance is 0): the sigma-point sums give `s = 1.000000000000029` at rank 530, so `second − first² = −131 eps`, and the gate raises `NumericalFailure`.
  - Up to rank 4000, 1207 ranks exceed the gate. The worst is −988 eps at rank 3973.
  - So on the competing-risks path, a zero-variance prediction on a model with ≥ 530 posterior directions is refused as a numerical failure. That is a false refusal.
- **The single-event path has the opposite defect.** The same cancellation is clamped with `.max(0.0)`. A genuinely negative variance, which would signal a broken accumulator, is swallowed silently. The two copies of the same quantity follow two different policies.
- **The truncated round trip adds two errors.** It adds cancellation error to a variance that was already computed centred and non-negative. Its `clamp(0,1)` on the second moment can also change the variance, since `second − first²` after clamping is not `survival.variance`.

## Principled fix (no tolerance at all)
Accumulate centred moments rather than raw ones. Merge each weighted node with the Chan–Golub–LeVeque pairwise update:

```rust
struct PosteriorMoment { weight: f64, mean: f64, centered: f64 }
fn merge(&mut self, scale: f64, other: Self) {
    let added = scale * other.weight;
    if added == 0.0 { return; }
    if self.weight == 0.0 { *self = Self { weight: added, mean: other.mean, centered: scale * other.centered }; return; }
    let total = self.weight + added;
    let d = other.mean - self.mean;
    self.centered += scale * other.centered + d * d * (self.weight * added / total);
    self.mean = self.mean * (self.weight / total) + other.mean * (added / total);
    self.weight = total;
}
fn variance(self) -> f64 { self.centered / self.weight }   // >= 0 by construction
```

- The variance is a sum of non-negative terms, so it is ≥ 0 by construction.
- A deterministic cell gives exactly 0 at any rank.
- The gate, its constant and both `.max(0.0)` clamps can all be deleted. The SE helper refuses only a non-finite mean or variance.
- Apply it everywhere:
  - Sigma-point (single-event and competing-risks) accumulation.
  - The exact-anchor Gauss–Hermite path: `ExactAnchorCellMoments` becomes a struct of `PosteriorMoment`s implementing `gam_solve::quadrature::GhqValue`.
  - The truncated producer: `PosteriorMoment::from_mean_variance(first, survival.variance)`, with no second-moment round trip.
  - The pooled accumulator at L5865.
- `SurvivalPosteriorMoments` then carries `Array2<PosteriorMoment>` in place of `survival_mean` / `survival_second` (and likewise for density and η). The single-event and competing-risks SE go through one helper.

**Regression test:** use an identity posterior covariance of dimension 530 and a constant functional of 1.0. Assert that the centred SE is exactly 0. Also assert that the raw `second − first²` is below `−128·eps`, which documents the old false refusal.

Found by local audit agent HN-6.

SauersML:
Fixes #4086

## Root cause
The survival posterior SEs were recovered as `E f² − (E f)²` from raw weighted sums. The error of that cancellation grows with the node count (2·rank), so no fixed multiple of eps bounds it. Two paths handled it in opposite ways:
- **Competing-risks:** a magic `128·eps` gate refused deterministic cells as `NumericalFailure` once rank ≥ 530.
- **Single-event:** `.max(0.0)` silently swallowed the same negatives.

The truncated producer also round-tripped its already-centred variance through a clamped second moment.

## Fix
- Add `PosteriorMoment { weight, mean, centered }`, merged by the Chan–Golub–LeVeque pairwise update. The variance is ≥ 0 by construction and exactly 0 for a constant functional, at any rank.
- It is used everywhere the posterior moments are accumulated:
  - sigma-point single-event and competing-risks accumulation;
  - the exact-anchor GHQ path, where `ExactAnchorCellMoments` becomes a struct implementing `GhqValue`;
  - the truncated producer, via `from_mean_variance(first, survival.variance)` with no second-moment round trip;
  - the pooled replicate accumulator, where replicates are merged on the pooling scales.
- One `posterior_standard_errors` helper, generic over dimension, refuses only non-finite moments.
- Removed: the gate and its constant, the two duplicated SE helpers, and every `.max(0.0)` clamp on a variance.
- Density and hazard stay plain weighted means. The hazard can be infinite, and its mean only separates a zero hazard from an infinite one.

## Location-scale response moments (commit 459c0497ba)
The location-scale chain had the same cancellation: `predict_survival_location_scalewith_uncertainty` read the response SD as `sqrt(max(E[S²] − E[S]², 0))`. On a row whose survival is nearly certain, `E[S²] ≈ E[S]²`, so the SD it reported was their rounding, or the zero that `max(0)` made of it. This commit extends the branch to cover it:
- **`location_scale/moments.rs`:**
  - The projected 3-D Gauss–Hermite rule is now generic over `GhqValue` (`low_rank_normal_expectation_3d_result<F, R>`).
  - Each node is merged into a `PosteriorMoment`. The nested link-wiggle rule's inner moment merges into the outer one, which gives `Var S = E[Var(S|y)] + Var(E[S|y])` with no subtraction.
  - `exact_survival_response_moments` now returns `(E[S], Var S)`.
- **`location_scale/truncated_moments.rs`:**
  - `ResponseMomentAccumulator` keeps one centred `PosteriorMoment` of `d = S − S_ambient` per row, on the same log weight scale. It is rescaled with `PosteriorMoment::scaled`.
  - Replicates pool by the same merge, so the spread between replicates enters as a non-negative term.
  - The ambient shift is kept, so `E[S]` of a near-certain row stays resolved relative to `1 − S`.
  - `RowCertificate.second` becomes `variance`.
- **`location_scale/predict.rs`:** the SD is `sqrt(Var S)`, with nothing subtracted and nothing clipped. A non-finite variance is refused with an `Err`.
- **Tests:** the `(mean, second)` callers now destructure `(mean, variance)`. The Jensen `+1e-9` allowance became an exact `var >= 0` check.
- **New test `response_moment_rule_resolves_a_small_variance_under_a_large_mean_4086`:**
  - Setup: `f = 0.75 + 2⁻³⁰·x₀` under a full-rank 3-D normal, so `Var f = 2⁻⁶⁰`, which is below `ulp(E[f]²) = 2⁻⁵³`.
  - The centred rule must resolve it within the rounding bound derived from the rule, `√N·ulp(c)/(a·√Σ₀₀)`.
  - Negative control: the raw `E[f²] − E[f]²` must miss by a relative error of at least 1, which is forced because both terms lie in `[0.5, 1)`.
  - A constant functional must give exactly 0.
  - An offline replica of the 15³ rule gave a centred relative error of 5.4e-7, against a bound of 6.9e-6. The raw form's relative error was about 1.2e4.

## Landing filters (never compiled in CI yet: run locally)
```
cargo check --workspace --tests
cargo test -p gam-models --lib posterior_quadrature_constant_functional_has_exactly_zero_standard_error_at_high_rank
cargo test -p gam-models --lib posterior_moment_merge_matches_two_pass_moments
cargo test -p gam-models --lib response_moment_rule_resolves_a_small_variance_under_a_large_mean_4086
cargo test -p gam-models --lib truncated_response_moments_beat_the_moment_matched_normal_2679
cargo test -p gam-models --lib nested_response_moment_rule_reproduces_the_scalar_gaussian_law_2446
cargo test -p gam-models --lib near_wall_wiggle_coordinate_keeps_cross_covariance_in_moments_2390
cargo test -p gam-models --lib survival::location_scale
cargo test -p gam-models --lib survival::predict
```

## Tests
- `posterior_quadrature_constant_functional_has_exactly_zero_standard_error_at_high_rank`: rank-530 identity covariance with a constant functional. It asserts that the raw `second − first²` is below `−128·eps` (the old false refusal) and that the centred SE is exactly 0.
- `posterior_moment_merge_matches_two_pass_moments`: the pairwise merge and scaled pooling match two-pass moments.
- The existing sigma-point SE tests are ported to the new helper.
- `response_moment_rule_resolves_a_small_variance_under_a_large_mean_4086`: covered above.


SauersML:
Review of 8b5a0edc against SPEC and the math, plus an extension (459c0497ba).

**Math: sound.**
- The Chan–Golub–LeVeque merge only adds `scale·centered` and `d²·W·w/(W+w)`. Both are non-negative for `scale ≥ 0`, and GH and sigma-point weights are positive, so the variance can't go negative. A constant functional gives `d = 0` at every node, so its variance is exactly 0 at any rank.
- `EMPTY.variance()` is `0/0 = NaN`. That is reachable only for a cell nothing was merged into, and `posterior_standard_errors` refuses it loudly rather than papering over it (SPEC: "do not paper over solver issues").
- When the difference is infinite, the merge falls back to a convex combination, which avoids `∞ − ∞`.
- The `128·eps` gate is gone and no new knob replaces it, which satisfies SPEC's rule on magic constants.

**Sibling sites the PR missed, now fixed on this branch (459c0497ba):** the location-scale response-moment chain still computed `sqrt(max(E[S²] − E[S]², 0))`.
- `location_scale/moments.rs`: the projected 3-D GH rule and the nested link-wiggle rule. The inner moment now merges into the outer one, which is the law of total variance.
- `location_scale/truncated_moments.rs`: the replicate accumulators now hold a centred `PosteriorMoment` of `S − S_ambient` per row, on the existing log weight scale. The pooled certificate merges replicates the same way.
- `location_scale/predict.rs`: the SD is `sqrt(Var S)`. A non-finite variance is an `Err`.
- The only change to `survival/predict.rs` is that `PosteriorMoment::scaled` becomes `pub(crate)`.

The new test `response_moment_rule_resolves_a_small_variance_under_a_large_mean_4086` has a deterministic negative control. With `Var f = 2⁻⁶⁰` and `E[f] = 0.75`, both `E[f²]` and `E[f]²` lie in `[0.5, 1)`, so their difference is a multiple of `2⁻⁵³` and has a relative error of at least 1. The centred rule must come within its rounding bound, derived from the rule, `√N·ulp(c)/(a·√Σ₀₀)`.

**Left as is:**
- The MC SE near `predict.rs:8249` is test-only.
- The comment near `predict.rs:5624` mentions `ResponseMomentAccumulator`. That accumulator is now centred as well, so the comment is still accurate.

Other E f² − (E f)² sites outside survival (multinomial_posterior, multinomial_predictive, gam-sae compose) are tracked in #4124.

The landing filters are in the PR body. This is an even PR, so it goes to MT-0.

---

SauersML:
Status (BF-89): this is reviewed and only waits on its build. It is ready for the local merge train (MT-0, even PR) once focused proof run 35511490149 is green at head `459c0497ba`. That head merges cleanly with current `origin/main`. The superseded run 35505310543 (old head 8b5a0ed) was cancelled. Once it lands, close #4086 with the rank-530 zero-SE and the 2⁻⁶⁰-variance test data.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, eeb8dbc39e0a.

PR #4117, opened 2026-09-20
gam-linalg decision: never certify a NaN rank or drop a non-finite excursion

### LINKED ISSUE #4067: gam-linalg decision: certified_rank certifies a NaN singular value as dropped, and spectral_excursion_lower_bound discards NaN/inf
## Where (origin/main)

`crates/gam-linalg/src/decision.rs`

**`certified_rank` (l.155-163)**
```rust
let mut sv: Vec<f64> = singular_values.to_vec();
sv.sort_by(|a, b| b.partial_cmp(a).unwrap_or(std::cmp::Ordering::Equal));
...
if let Some(&sigma_in_band) = sv.iter().find(|&&s| s > low && s < high) { ... Ambiguous }
let rank = sv.iter().filter(|&&s| s >= high).count();
```

**`spectral_excursion_lower_bound` (l.411-424)**
```rust
a.sort_by(|x, y| y.partial_cmp(x).unwrap_or(std::cmp::Ordering::Equal));
...
worst = worst.max((sa - sb).abs());
```

### 1. `certified_rank` treats a NaN as a dropped value
Every comparison against a NaN is false, so a NaN singular value:
- fails the in-band test, and
- fails `s >= high`.

It is therefore counted as **dropped**, and the function returns `Certified { rank }`. That is a certified rank that silently lost a direction the spectrum never resolved.

A NaN band edge (a NaN `tol` or `gap`) is worse: every value fails every test, so it certifies rank 0.

The sort has two further problems when a NaN is present:
- The comparator `partial_cmp(..).unwrap_or(Equal)` is not a total order. The sort then leaves the NaN wherever it lands, so `sigma_r = sv[rank-1]` and `sigma_next = sv[rank]` read arbitrary entries, and `margin_low` / `margin_high` come out as garbage.
- Since Rust 1.81, `sort_by` may panic on such a comparator; the toolchain here is 1.97.

The function's own doc makes the promise this breaks: "An `Ambiguous` outcome is the honest report that no such margin exists."

### 2. `spectral_excursion_lower_bound` under-claims on NaN or inf
The doc says this is the monitor a **fail-closed** gate uses to refuse reuse: "if this bound already exceeds the reference's `rank_transport_radius`, the reference certificate is PROVABLY void".

But the loop drops exactly the entries that make the displacement unmeasured:
- `f64::max` drops a NaN.
- `∞ − ∞` is NaN, so an infinity in both spectra is dropped the same way.

The function then reports the displacement of the remaining finite values as the bound. A gate reading `bound > radius` then reuses a certificate it should have re-decided.

Production callers:
- `gam-identifiability/src/audit.rs:3141-3153` (pilot-certificate transport)
- `gam-identifiability/src/families/compiler.rs:948`
- `gam-solve/src/psi_gram_tensor.rs:1190`

## Fix (small, local)

`certified_rank`:
- A NaN `tol`/`gap` gives `Ambiguous { rank_floor: 0, rank_ceil: n, sigma_in_band: NaN }`.
- A NaN singular value gives `Ambiguous`, with bounds:
  - `rank_floor = #{σ ≥ high}`
  - `rank_ceil = #{σ > low} + #NaN`
- The sort uses `total_cmp`.

`spectral_excursion_lower_bound`:
- Any non-finite entry in either spectrum returns `+∞`. The displacement is unmeasured, and `transport_certified_rank` already reads a non-finite excursion as `NoCertificate`. This matches the convention `projector_error_bar` already uses for malformed input.
- The sort uses `total_cmp`.

Existing behaviour is unchanged for finite input. Regression tests pin the NaN placements (first, middle, last), a NaN band, and the NaN/±inf excursion cases.

SauersML:
Fixes #4067

- `certified_rank` refuses to certify a rank when a singular value or band is `NaN`, instead of counting it as dropped.
- `spectral_excursion_lower_bound` returns no bound for a non-finite spectrum, instead of discarding the `NaN`/`∞` entries.
- New tests:
  - `a_nan_singular_value_or_band_is_never_certified`
  - `a_non_finite_spectrum_voids_the_excursion_lower_bound`

---


SauersML:

---

PR #4118, opened 2026-09-20
Centred Welford spread for the smooth-LR replay SE and the ring permutation null (#4086 sibling)

### LINKED ISSUE #4114: One-pass Σd²/N − d̄² spread (the #4086 defect) in smooth_term_lr paired_mean_with_error and gam-sae ring_permutation_evidence
## Where (origin/main 6ff5fb0185)
- `crates/gam-models/src/fit_orchestration/drivers/smooth_term_lr.rs:2568`, `paired_mean_with_error` (the Monte-Carlo standard error of the smooth-LR selection-replay tail shift, which the report's accuracy bound is built from at L3038 `bound + 2.0 * standard_error`):
  ```rust
  let variance = (sum_squares / count - shift * shift).max(0.0);
  ```
- `crates/gam-sae/src/manifold/curl.rs:851`, `ring_permutation_evidence` null ρ sd:
  ```rust
  let sk = (sum_r2 / df - mk * mk).max(0.0).sqrt();
  ```

## Why it is wrong
Same defect as #4086: the spread is recovered by differencing two raw moments of size `d̄²`, so it is resolved only in steps of `ulp(d̄²)`, and the `.max(0.0)` hides a negative result instead of making it impossible.

Measured with the exact f64 expressions:
| sample | exact variance | one-pass | centred (Welford) |
|---|---|---|---|
| `0.1` × 10 (constant) | 0 | 5.204e-18 | 0 |
| `1/3` × 7 (constant) | 0 | 2.776e-17 | 0 |
| `0.75 ± 2⁻²⁴`, 128 each | 3.553e-15 | 7.772e-16 | 3.553e-15 |
| `0.75 ± 2⁻²⁷`, 128 each | 5.551e-17 | 0 | 5.551e-17 |

So a constant paired-difference sample (every draw's selection shifted by the same closed-form amount) reports a positive standard error made of cancellation, and a small spread around a large shift is lost.

## Fix
Accumulate centred (Welford): `m ← m + (d − m)/i`, `M₂ ← M₂ + (d − m_old)(d − m_new)`. The rounded new mean is the rounding of a point between `m_old` and `d`, so both factors have the same sign and every increment is ≥ 0 in floating point. A constant sample gives exactly 0; no clamp.

Sibling sweep of #4086; the survival sites stay in #4086.

SauersML:
Fixes #4114

These are the sibling sites of #4086 outside the survival tree. The survival sites stay with the #4086 PR (see https://github.com/SauersML/gam/issues/4086#issuecomment-5749255277).

## Defect
Both sites computed a Monte-Carlo spread as `V₁ = Σdᵢ²/N − d̄²` and clamped it with `.max(0.0)`. Each term carries an absolute rounding error of order `ε·d̄²`, so `V₁` is resolved only in steps of `ulp(d̄²)`. It can also come out negative, which is why the clamp was there.

## Fix: `smooth_term_lr.rs::paired_mean_with_error` (two-pass, pivoted)
```
d̄ = d₁ + Σ(dᵢ − d₁)/N,   M₂ = Σ(dᵢ − d̄)²,   SE = √(M₂/N²)
```
- `M₂` is a sum of squares, so it is `≥ 0` with no clamp.
- A constant sample has every pivoted difference exactly `0`, so it reports `d̄ = d₁` and `M₂ = 0` exactly.
- Rounding in `d̄` enters `M₂` only at second order: `Σ(dᵢ − m)² = Σ(dᵢ − d̄)² + N(d̄ − m)²`.
- An earlier head used a running (Welford) mean here. On an ordered sample that mean drifts by `ulp(d̄)` per step, which is not small beside a spread far below `|d̄|`, and the PR's own exactness test failed on it (5b2f5239).
- The paired differences are materialized once (`SMOOTH_LR_SELECTION_DRAWS` values), which the replay already holds.

## Fix: `gam-sae/src/manifold/curl.rs::ring_permutation_evidence` (streaming Welford)
- The null ρ moments are streamed with Welford: `δ = r − m`, `m ← m + δ/i`, `M₂ ← M₂ + δ·(r − m)`.
- The rounded running mean stays in the closed interval between the old mean and `r`, so each increment is a product of two same-signed factors. `M₂ ≥ 0` holds by construction, and the clamp is gone.
- The replicates are i.i.d. permutation draws, not an ordered sample, so the drift described above does not arise.

## Test
`paired_standard_error_is_centred_and_exact_under_a_large_mean_4086` (`selection_replay_tests`):
- **Fixture.** Half the draws are at `c + s` and half at `c − s`, with `c = 0.75`, `s = 2⁻ᵏ` for k = 20..31, and N = 256.
- **Why exact equality is the derived bound (no ε band).** Every floating-point step of the pivoted two-pass reading is exact on this fixture:
  - `c ± s` are exact.
  - The pivoted differences `0` and `−2s` are exact (Sterbenz).
  - The partial sums, which are multiples of `2s`, are exact.
  - Dividing by `N = 2⁸` is exact.
  - The centred values `±s` and their squares `2⁻²ᵏ` are exact.
  - `M₂/N² = 2⁻²ᵏ⁻⁸` and its root `2⁻ᵏ⁻⁴ = √(s²/N)` are exact.

  The test therefore asserts `error == √(s²/N)` and `shift == c`. The same reasoning applies to a constant sample, whose `shift == value` and `error == 0` are also asserted exactly. The inputs were replayed in f64 in Python and match bit-for-bit.
- **Cancelling regime.** The test also asserts that the retired one-pass form visibly loses the variance on this fixture.

## Proof filters
- `cargo test -p gam-models --lib selection_replay_tests`. This covers the new test, the replay SE honesty test `the_published_standard_error_covers_a_four_fold_draw_increase` and the bit-identity test.
- `cargo test -p gam-sae --lib manifold::curl`.


SauersML:

---

SauersML:
Note for RF-1113-6 (leaving this PR to you): at head the PR's own test `paired_standard_error_is_centred_and_exact_under_a_large_mean_4086` fails. Welford's update is not exact on an ordered sample under a large common mean, and its rounding does not cancel. Replaying the test's inputs in Python reproduces the failure.

Reading the spread in two passes about a mean pivoted at the first draw passes the test: the mean is exact, the SE is within 4ε and a constant sample gives exactly 0. The pivoted mean is `shift = d₀ + Σ(dᵢ−d₀)/n`, then `M₂ = Σ(dᵢ−shift)²` and `SE = √(M₂/n²)`.

The fix is commit 5b2f5239 on branch `claude/fix-4118-two-pass-paired-se`, which sits on this PR's head merged with main (ce58f233). It touches only `smooth_term_lr.rs`; the ring-permutation Welford in `curl.rs` is unchanged. You can merge that branch into this PR.

---

SauersML:
BF-103 review (partition 2 sweep, #4114): this PR is ready for the local merge train (MT-0, even PR number). It merges cleanly with origin/main (mergeable_state clean).

- It has **no check run yet**, so the train should run the proof filters listed in the body:
  - `gam-models --lib selection_replay_tests`
  - `gam-sae --lib manifold::curl`
- **Math.** The Welford increment `δᵢ·(dᵢ − mᵢ)` is ≥ 0 in floating point, because the rounded `mᵢ` stays in the closed interval `[m_{i−1}, dᵢ]`. So the `.max(0.0)` clamp is correctly removed rather than replaced, and a constant sample gives exactly 0.
- **Test tolerance.** The new test's `4ε` relative bound on the SE is tight, and the body gives no derivation for it. If it fails on the train, the fix is a derived bound, not a loosened constant.

---

SauersML:
Taking over as BF-117 (handed off to the merge lane). The plan:
- Bring in the two-pass fix 5b2f5239. It is a descendant of this head, so the PR branch fast-forwards to it.
- Replace the test's underived `4ε` bound with the bound the fixture actually implies, which is exact equality (derivation in the commit).
- Merge main, then dispatch ONE focused run with the proof filters from the body.

---

PR #4125, opened 2026-09-20
manifold/intrinsic_seed: Borůvka bridging and k-selection in deterministic_knn_graph

### LINKED ISSUE #4122: manifold/intrinsic_seed: kNN graph bridging rescans all pairs once per component (O(C·n²)); kNN fully sorts every row
`crates/gam-sae/src/manifold/intrinsic_seed.rs`, `deterministic_knn_graph`. Production path: `intrinsic_knn_graph` → `LocalAtlas::build` (`local_charts.rs`, over every row the atlas is built on) and `intrinsic_geodesic_embedding` → `structure_harvest` (`race_intrinsic_coords` and the local-sheet challenger).

**Defect 1: the bridging loop costs one all-pairs scan per component.** After the kNN union, the graph is bridged to one component. Each pass of the loop:
- rescans every `(i, j)` pair,
- calls `find` for every pair (O(n²) union-find walks),
- adds a single edge, the global-minimum cross-component edge, and
- repeats.

`C` components therefore cost `(C − 1) · n²/2` distance evaluations (each is `O(p)`). Every component of a kNN graph needs at least `k + 1` nodes, so `C` can reach `n/(k+1)` with `k = max(2d+1, ⌈log₂ n⌉)`. Clustered activations are exactly the SAE regime, and there the bridging is `O(n³/log n)`, well above the `O(n²)` the kNN scan itself costs.

**Defect 2: the kNN pass fully sorts every row's `n − 1` distances to keep `k` of them.** That is `O(n² log n)` where a selection plus a sorted `k`-prefix gives `O(n² + n k log k)`.

**Fix (output bit-identical).**
- Bridging becomes Borůvka passes. Root labels are computed once per pass. One all-pairs scan records each component's nearest outgoing pair under the strict order `(d², i, j)`, and all those picks are added.
  - Under a strict edge order the minimum spanning tree of the component graph is unique.
  - The old loop is Kruskal on that graph, so it builds the same tree.
  - Borůvka picks are all edges of that tree, so the edge set, and therefore the adjacency lists, are bit-identical.
  - The component count at least halves per pass, so the cost is `⌈log₂ C⌉` scans.
- kNN uses `select_nth_unstable_by` on the same strict `(distance, index)` order, then sorts the `k`-prefix. The result is the same set in the same order.
- A regression test compares the new graph to an in-test copy of the one-edge-per-scan reference on nine separated clusters (k = 1..4), plus the single-row case. A numpy simulation over 200 random clustered layouts, including rounded coordinates to force distance ties, also matched exactly.

SauersML:
Fixes #4122

Two speedups in `deterministic_knn_graph`. Both leave the output bit-identical.

- **Bridging.** Each pass now computes root labels once. It records each component's nearest outgoing pair under the strict order `(d², i, j)`, then adds all of those pairs. That is Borůvka on the component graph. Under a strict order the minimum spanning tree is unique, so this builds the same tree as the old one-global-minimum-edge-per-scan loop (Kruskal). The cost falls from `C − 1` all-pairs scans to `⌈log₂ C⌉`.
- **kNN.** `select_nth_unstable_by` pulls out the `k` nearest under the same strict `(distance, index)` order, and only that prefix is sorted. It no longer fully sorts `n − 1` distances per row.

The new test, `knn_graph_bridging_matches_one_edge_per_scan_reference`, compares against an in-test copy of the old algorithm on nine separated clusters, k = 1..4, plus the single-row case.

---


SauersML:

---

SauersML:
BF-103 review (partition 2 sweep, #4122): this PR is ready for the local merge train (MT-1, odd PR number). It is only waiting on the queued focused check HN-1-4122-1 (job 106064465992).

- Head 33f89f898f merges cleanly with origin/main.
- **Bridging is still exact.** Each component's nearest outgoing pair is the lexicographic minimum under `(d², i, j)`: pairs are scanned in ascending `(i, j)` and replaced only on strict `total_cmp` less. Under a strict edge order the MST of the component graph is unique, so Borůvka (all per-component minima per pass) and the old global-minimum loop (Kruskal) add the same edge set. The dedup and the `find` check make a pair picked by both components a no-op. `total_cmp` and the old `<` agree on non-NaN, non-negative d².
- **The kNN change is safe.** `select_nth_unstable_by(keep-1)` followed by a prefix sort gives exactly the first `keep` entries of a full sort under the strict `(dist, idx)` order. `keep == 0` and `keep == len` are handled.
- `Ordering` is already imported. The test compares against an in-test copy of the old algorithm for k = 1..4 on nine components.

---

SauersML:
Taken over by BF-104 (HN-1 handed off). The diff looks sound: Borůvka bridging under a strict edge order yields the unique MST, so the graph is bit-identical, and the kNN step uses a partial selection. I'm waiting on proof run 35505488429 for head 33f89f89 and will merge on green.

---

PR #4127, opened 2026-09-20
gam-predict conformal: exact split-conformal rank

### LINKED ISSUE #4123: gam-predict conformal: split-conformal rank ⌈(n+1)(1−α)⌉ is taken in floating point and over-covers by one rank
## Defect

`crates/gam-predict/src/conformal.rs`, `conformal_multiplier`, which backs `predict_full_uncertainty_conformal` (the split-conformal route in `conformal_routes.rs`, reached from Python `predict(interval="conformal", calibration=...)` and CLI `gam predict --conformal --calibration`):

```rust
let rank = ((n as f64 + 1.0) * (1.0 - alpha)).ceil() as usize;
```

Every caller passes `alpha = 1.0 - level` (`lib.rs::predict_full_uncertainty_conformal`), and neither the decimal level nor the subtraction is exact in binary. The product sometimes lands just *above* an integer, so the `ceil` takes one rank too many. The multiplier `q̂` then becomes the next order statistic and the interval over-covers by `1/(n+1)`. The module docs promise the "EXACT" order statistic.

Checked against exact rational arithmetic:
- `level = 0.68, n = 74`: `75·(1 − (1 − 0.68)) = 51.00000000000001`, so the rank is 52, but the exact rank is 51. This happens for 69 of `n ≤ 5000` at level 0.68.
- A direct `alpha = 0.95, n = 19`: `20·(1 − 0.95) = 1.0000000000000009`, so the rank is 2, but the exact rank is 1.

The full-conformal path already fixed exactly this in `gam_models::inference::full_conformal::conformal_rank_threshold`: it snaps `α(n+1)` to the integer it stands for, with a comment describing the same over-coverage. The split path duplicates the rank computation without that snap.

## Fix

- Compute the rank as `(n+1) − ⌊τ⌋`, with `τ = conformal_rank_threshold(alpha, n+1)`. This equals `⌈(n+1)(1−α)⌉` for integer `n+1`.
- This leaves a single source of truth for the conformal rank.
- It matches exact rational arithmetic for all 500,000 (level, n) and (α, n) cases I checked.
- Add a unit test covering the 0.68/74 and 0.95/19 cases, plus the nominal 0.9 and 0.95 ranks.

SauersML:
Fixes #4123

`conformal_multiplier` computed `⌈(n+1)(1−α)⌉` in floating point. With `α = 1 − level`, the product can land just above an integer; level 0.68 at n = 74 gives rank 52 instead of 51. That takes one order statistic too many, so the interval over-covers by `1/(n+1)`.

## Changes
- **Rank formula.** The rank is now `(n+1) − ⌊τ⌋`, with `τ = conformal_rank_threshold(α, n+1)`. That is the snapped threshold full conformal already uses, so both paths share one source of truth for the conformal rank.
- **New test** `multiplier_rank_is_exact_for_decimal_levels`. It covers:
  - 0.68 at n = 74;
  - α = 0.95 at n = 19;
  - the nominal 0.9 and 0.95 ranks.

---


SauersML:

---

PR #4128, opened 2026-09-20
skip transcoder: Rust-owned profiled rank selection (#3920)

### LINKED ISSUE #3920: python wrapper: select_skip_transcoder runs its whole outer optimizer (LBFGS profiling + rank birth/death search) in Python and pays two redundant full profile fits per LBFGS step
## Evidence (origin/main 6ea77be)

`gamfit/torch/skip_transcoder.py:513-587`, `_continuously_profile_rank`:
```python
    current = evaluate(transition, warm_start)
    optimizer = torch.optim.LBFGS(
        [logs],
        max_iter=1,
        tolerance_grad=0.0,
        tolerance_change=0.0,
        line_search_fn="strong_wolfe",
    )
    while current.stationarity_defect > current.stationarity_tolerance:
        ...
        def closure() -> torch.Tensor:
            optimizer.zero_grad()
            profiled = evaluate("continuous", closure_seed)
            ...
        optimizer.step(closure)
        current = evaluate("continuous", current.smooth)
```
`gamfit/torch/skip_transcoder.py:618-736`, `select_skip_transcoder`, adds the greedy rank birth/death loop in Python, with its own eps-scaled evidence tolerance and acceptance rule and the certificate assembly. The stationarity tolerance `sqrt(eps)·gradient_scale` (line 370) and the step-acceptance roundoff rule (lines 579-585) are also formed in Python.

## Why it is wrong

1. **The outer optimizer lives in Python.** SPEC says: "Python should be a thin wrapper over Rust, and should avoid containing math or much logic", and "General outer optimizer work should be in opt crate, not here." Everything here is outer-optimizer logic that gam-opt already owns for every formula fit:
   - the quasi-Newton iteration, delegated to torch's LBFGS rather than the repo's own certified driver;
   - the stationarity certificate;
   - the non-descent refusal;
   - the discrete rank search and its acceptance tolerance.

   The torch-specific part is only `profile_trial`, the user's value-and-gradient oracle.
2. **It pays two redundant profile fits per iteration.** Each `profile_trial` call "must train/profile the same evidence to convergence" (docstring, line 628), so one call is a full inner fit.
   - With `max_iter=1`, `LBFGS.step` first calls `closure()` at the current iterate, which `current` already holds.
   - The strong-Wolfe search then calls it again at every trial point, including the accepted one.
   - After `step`, line 574 calls `evaluate("continuous", current.smooth)` at the accepted point once more.

   Every outer iteration therefore repeats two converged inner fits whose results are already known. The closure's result is discarded after it has set the gradient, so nothing reuses it.

## Principled fix

- Expose from gam-opt, through pyffi, a callback-driven minimiser over a small log-hyperparameter vector. It takes a Python `(value, gradient)` oracle and returns a stationarity-certified optimum, or a typed non-convergence carrying the checkpoint. The certificate and tolerance should be the same ones gam-opt uses for REML outer fits. Also move the integer rank birth/death acceptance rule into Rust, as `select_topology_candidate_lifecycle` already does for topologies.
- Python then keeps only the torch oracle adapter and the result dataclasses. `torch.optim.LBFGS`, the Python tolerances and the Python rank loop are deleted.
- The certified driver evaluates each iterate exactly once, so the two redundant profile fits per step go away. Until the move lands, even a memo keyed on `(rank, logs)` would remove them.

SauersML:
Fixes #3920

## What changed

`select_skip_transcoder` ran its whole outer optimizer in Python. That included torch LBFGS over the two log-hyperparameters, its own `sqrt(eps)*scale` stationarity certificate, the non-descent refusal, and the greedy rank birth/death walk. It also paid for two redundant converged inner fits on every LBFGS step. All of that now lives in Rust:

- **`gam_solve::profiled_rank_selection`** (new). One oracle `(rank, theta) -> (V, grad V, scale)` drives opt's BFGS for each rank.
  - A profile is accepted only at an iterate that its own evaluation certifies: `||grad V||_2 <= sqrt(eps) * scale`. Anything else is a typed `NonConvergence` that carries the checkpoint.
  - A bit-exact per-rank evaluation cache, with the seed passed in as the initial sample, means every distinct iterate is evaluated exactly once.
  - The rank walk (birth/death) moves only on strict improvement beyond `eps * (1 + |V_cur| + |V_best|)`. It profiles each rank once and returns the selected profile, both neighbour profiles and the move record.
- **`gam-pyffi`**: the `select_rank_with_profiled_hyperparameters` binding returns JSON and is registered in the geometry module. The stubs are updated.
- **`skip_transcoder.py`** is now a thin adapter. The PyTorch oracle must echo the requested point exactly (the `sqrt(eps)` echo tolerance is removed). `SkipTranscoderProfile` no longer computes its own certificate. `SkipTranscoderContinuousNonConvergence` reports the defect and tolerance that Rust measured.

## Checklist
- [ ] focused-rust-proof: gam-solve lib, filter `profiled_rank_selection` (key `BF-23-3920-1`)
- [ ] python contracts / torch skip-transcoder tests

---


SauersML:
BF-23 status: head `3fc29066e6` is unchanged. Two focused proofs are still queued on it:
- 35505506836: gam-solve lib, filter `profiled_rank_selection`
- 35505609313: gam-pyffi lib compile check

I'm still actively driving this and will merge once both are green. The Python skip-transcoder test has no CI path while Python Contracts is disabled, so this PR proves only the Rust oracle, the rank walk and the pyffi compile.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 93da1400c385.

PR #4139, opened 2026-09-20
manifold: a non-finite or unevaluable inner objective refuses the decrement certificate and restores

### LINKED ISSUE #4132: manifold/construction_undamped_inner + fit_drivers: an unevaluable or infinite penalized objective certifies the inner decrement and skips the incumbent/exit-warranty restores
Five production sites price a failed or non-finite `penalized_objective_total` as an infinite scale or objective. Every one of them resolves the wrong way: it certifies or keeps a state it should refuse or restore.

**Certificates (`½λ² / (|F| + 1)`, read by `inner_decrement_certifies`)**
1. `construction_undamped_inner.rs` ~L846, the limit-boundary decrement certificate in `converge_inner_for_undamped_logdet_gate_frozen`:
   ```rust
   let limit_scale = self.penalized_objective_total(..).map(|obj| obj.abs() + 1.0).unwrap_or(f64::INFINITY);
   let predicted_relative_decrease = 0.5 * decrement_sq / limit_scale;
   ```
   If the objective fails to evaluate, the scale is `∞` and the ratio is `0`. `inner_decrement_certifies(0)` is `true`, so the iterate is **accepted as converged precisely because the objective could not be computed**. This is silent error swallowing that turns into a false convergence certificate.
2. The same function, ~L1099: the final-gate excursion certificate `excursion_cert` has the identical pattern.
3. `construction_evidence_root.rs` ~L772 propagates `Err`, but an `Ok(±∞)` objective gives `scale = ∞`, `relative = 0`, and the root is `Certified`. A NaN objective already refuses, via `is_finite` inside the certify predicate.

**Restores (`fit_drivers.rs`)**
4. ~L8064, the incumbent restore. `final_obj = penalized_objective_total(..).unwrap_or(INFINITY)`, then `obj_scale = TOL·(1 + max(|final|, |best|)) = ∞`, then `!(∞ <= best + ∞)` is `!(true)`. The incumbent is **not** restored when the final state's objective is infinite or unevaluable, which is the worst possible degradation.
5. ~L8188, the #2228 exit-warranty restore. Same pattern with `warranty_tol`.

SPEC: "A fit object must only ever come from a converged optimization." Sites 1–3 certify convergence on an iterate whose objective is unevaluable or infinite. Sites 4 and 5 hand back such a state instead of the converged incumbent.

**Fix**
- Add `SaeManifoldTerm::inner_relative_decrement(decrement_sq, objective)`. It returns `½λ²/(|F|+1)` for a finite `F` and NaN otherwise, so the certificate refuses. Sites 1–3 use it.
- Sites 1 and 2 propagate the objective error, with context, instead of pricing it as `∞`.
- Sites 4 and 5 check `final_obj.is_finite() &&` before the tolerance comparison, so an infinite or unevaluable final state restores.
- Regression test: `inner_relative_decrement_refuses_a_non_finite_objective` (`tests_outer_quasi_laplace_probe_budget_2080.rs`).

Related, but a different site: #3992.

SauersML:
Fixes #4132

- New helper `SaeManifoldTerm::inner_relative_decrement(decrement_sq, objective)`. It returns `½λ²/(|F|+1)` for a finite `F` and NaN otherwise, and `inner_decrement_certifies` refuses NaN. The undamped-inner limit-boundary certificate, the final-gate certificate and the evidence-root certificate all use it. Before this, a failed or infinite objective made the scale `∞` and the ratio `0`, so the iterate was certified.
- The two `converge_inner_for_undamped_logdet_gate_frozen` sites now propagate the objective error, with context, instead of `.unwrap_or(f64::INFINITY)`.
- The `fit_drivers` incumbent restore and the #2228 exit-warranty restore require `final_obj.is_finite()` before comparing against the tolerance. Before this, an infinite final objective made the tolerance infinite too, and `inf <= best + inf` kept the blown state.
- Regression test: `inner_relative_decrement_refuses_a_non_finite_objective`.

---


SauersML:

---

SauersML:

---

SauersML:
Taken over by BF-104 (HN-1 handed off). The diff looks sound: a non-finite objective now gives a NaN decrement, which the certificate refuses, and the restores require a finite final objective. I'm waiting on proof run 35505559226 for head 9f9d732e and will merge on green.

---

Issue #954, opened 2026-06-09
[math-audit] Optimizer convergence test is objective-shift-non-invariant (can stop at a non-stationary point)


SauersML:
Severity: **MOST SEVERE** — false convergence (the optimizer reports success at a non-stationary point).

## Where
- `relative_stationarity` at `src/geometry/optimizer.rs:64-87`
- Used by trust-region at `src/geometry/optimizer.rs:187` and L-BFGS at `src/geometry/optimizer.rs:450`
- Same idea duplicated in the FFI diagnostic at `crates/gam-pyffi/src/lib.rs:13185-13225`

## Why it's wrong / counterexample
The test computes a stationarity measure of the form

```
S(x) = ‖grad f(x)‖_g · ‖x‖_typ / max(|f(x)|, 1) ≤ τ
```

This is invalid because minimization is invariant under a constant shift of the objective, `f ↦ f + C`: the minimizer, gradient, Hessian, trust-region model reduction, and Armijo slope are all unchanged. Yet the test value `S(x)` changes with `|f(x)|`.

**Counterexample.** Let `f(x) = x²`, `x0 = 1`, so `∇f(1) = 2 ≠ 0` (not stationary). With the shifted objective `f̃ = C + x²`:

```
S(1; f̃) = ‖∇f̃(1)‖ · ‖1‖_typ / max(|C+1|, 1) = 2 / (C + 1)
```

Choosing `C > 2/τ − 1` makes `S(1; f̃) < τ`, so the optimizer **stops at the non-stationary point `x = 1`**.

Additionally, `‖x‖_typ` is not intrinsic on a manifold — it changes under a chart/translation/periodic-angle representative — so even the geometry of the criterion is ill-defined.

## Correct version
Test only the Riemannian gradient norm, with an absolute floor and a scale-invariant relative part anchored to the initial gradient:

```
‖grad f(x_k)‖_g ≤ ε_abs + ε_rel · ‖grad f(x_0)‖_g
```

This is shift-invariant (the gradient does not depend on `C`) and the relative part is scale-invariant (both norms scale together under `f ↦ α f`). The exact stationarity condition this targets is

```
‖grad f(x)‖_g = 0  ⇔  Df_x[η] = 0  ∀η
```

by positive-definiteness of the metric `g`.

SauersML:
[...]

PR #4143, opened 2026-09-20
identifiability: certify identifiable_factor_fit at a stationary point of a well-posed objective

### LINKED ISSUE #3997: identifiability: identifiable_factor_fit returns a fixed-400-step Adam iterate of an objective with no minimizer (no convergence certificate)
**Area:** `gamfit/identifiability.py` — `identifiable_factor_fit` / `_one_fit` (origin/main `f8c6613`; the file is unchanged since `988f269`)

SPEC: "A fit object must only ever come from a converged optimization." / "Python should be a thin wrapper over Rust." / "In general, do not paper over solver issues."

### What the code does

`_one_fit` (identifiability.py:484) runs a fixed number of Adam steps and returns the last iterate. There is no stationarity test, no convergence flag and no error path:

```python
# identifiability.py:531
optim = torch.optim.Adam(params, lr=float(learning_rate))
...
# identifiability.py:565
for _step in range(int(max_iter)):          # max_iter defaults to 400 (line 650)
    ...
    loss = recon + aux_surrogate + mech_surrogate
    loss.backward()
    optim.step()
...
# identifiability.py:611-614
# ... they describe this fixed-weight converged fit only.
...
return encoder, decoder, rss, total_pen     # line 638
```

`identifiable_factor_fit` (line 641) wraps this iterate as an `IdentifiableFactorFitResult` (line 304), including `T_free`, the decoder, and a `profile_log_likelihood` computed from the final `(rss, total_pen)`. The comment at 614 calls this iterate "converged", but nothing measures that. Changing `max_iter` or `learning_rate` changes the returned latents.

### Why this is more than a missing stop test: the objective has no minimizer

The loss is `||X - W T||^2 + aux_prior(T_sup) + mech_w · smoothL1(W_free)`, with the mechanism-sparsity penalty on the free-latent decoder rows (line 540). `T_free` has no prior and no normalization. The encoder's last layer is linear, so for any `c > 1` the reparameterization `T_free -> c·T_free`, `W_free -> W_free / c` leaves the reconstruction term exactly unchanged. It lowers the sparsity penalty strictly, roughly as `1/c`. The infimum is therefore approached only as `||T_free|| -> ∞`, `W_free -> 0` and is never attained. Every finite Adam iterate is an arbitrary truncation of this drift. No iteration count could produce a converged fit, and `profile_log_likelihood` reports a penalty that depends on the budget.

### Silent error swallowing in the same function

```python
# identifiability.py:556-561
try:
    rust.conditional_prior_ivae(float(aux_w), zero_sup, aux_np, aux_scale)
    aux_prior_active = True
except ValueError:
    aux_prior_active = False
```

Any `ValueError` from the Rust prior constructor, not just the documented degenerate-auxiliary case, silently turns the supervised prior off. The fit then proceeds as an unsupervised autoencoder. The module doc (line 28) calls this "the fit always completes", which is exactly what SPEC forbids.

### Principled fix
1. Remove the scale gauge. Fix the scale of `T_free`, for example with a unit second-moment/whitening constraint applied as an exact reparameterization of the encoder output, or by putting a proper prior on `T_free`, so that the penalized objective has an attained minimum.
2. Replace the fixed-step Adam loop with an optimizer that certifies termination against a derived stationarity tolerance, preferably in Rust (the penalties and prior already are). Raise a typed non-convergence error carrying the checkpoint instead of returning an iterate.
3. Make the degenerate-auxiliary case an explicit precondition check that raises, or that returns a typed "prior not identifiable" outcome. Do not catch every `ValueError`.

Related closed issues: #576, #790. Neither covers the fixed-step exit or the scale gauge.

SauersML:
Fixes #3997.

## Root cause
`identifiable_factor_fit` had three problems:

- **No certificate.** It ran 400 fixed Adam steps and returned the last iterate. SPEC: "A fit object must only ever come from a converged optimization."
- **No minimizer.** Replacing `T_free → c·T_free` and `W_free → W_free/c` leaves the reconstruction unchanged and strictly lowers the mechanism-sparsity penalty for every c > 1. The step budget was the only thing that stopped the fit.
- **Silent prior drop.** An `except ValueError` around the Rust `conditional_prior_ivae` constructor discarded the iVAE prior when the auxiliary was degenerate, and the fit returned a model the recipe does not describe.

## Fix
- **Gauge.** `T_free` is now the raw free encoder output divided by its per-column root second moment, stored as `free_scale`. The objective is exactly invariant to the raw free scale, so the sparsity penalty prices decoder sparsity at a fixed latent scale. That is the setting of Lachapelle 2401.04890.
- **Optimizer.** L-BFGS with a strong-Wolfe line search on the exact objective value. The Rust penalty gradients enter as `s − s.detach()`, which is zero in value, so the line search compares the true f.
- **Certificate.** The fit returns only when `‖∇f(θ̂)‖∞ ≤ grad_tol·‖∇f(θ₀)‖∞` over all parameters, recomputed at the returned point. This is the relative stationarity measure of `gaussian_reml_optimize_latent` (`grad_tol = 1e-8`, #954). Anchoring to the initial gradient divides out the O(N·P) scale of the objective and is invariant to additive constants. A tolerance derived from floating-point precision is not available for this non-convex MLP objective without absolute error bounds on each term's gradient, so this uses the in-package precedent rather than inventing a new constant.
- **Failure is typed.** If the certificate does not hold within `max_evals`, the fit raises `FitConvergenceError` with `grad_inf`, `grad_inf_init`, `grad_tol`, `max_evals`, `n_iter`, `objective_value`, `checkpoint_encoder_state` and `checkpoint_decoder`. `max_evals` is a budget only.
- **Degenerate aux raises.** The iVAE prior is built once, before optimizing, and its error is not caught. An auxiliary that cannot identify the prior raises the Rust Khemakhem precondition error.
- **Breaking API change.** `learning_rate` and `max_iter` are removed. `max_evals` and `grad_tol` are new. The result gains `free_scale`, `stationarity` and `n_iter`.

## Tests
The tests move to `tests/torch/`. Torch is absent in python-contracts, where the whole module used to be importorskip-skipped. They now cover:
- the certificate;
- the unit second moment of `T_free`;
- an identical point when the budget is doubled;
- a typed failure that carries the checkpoint;
- refusal of a constant aux;
- `check()` flagging a constant aux against a certified fit;
- the existing recovery and RNG-isolation tests.

Validation: validate-one with `torch_filter`, issue BF-40-3997-1.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 3e886e5bba09.

PR #4144, opened 2026-09-20
tests/regressions: derived φ̂ and SE gates for the #771 Tweedie dispersion locks

### LINKED ISSUE #4131: tests/regressions: #771 Tweedie locks accept a φ̂ off by 2x (one-sided SE ratio >= 2 vs 5.16, φ̂ within 20%)
Two #771 regression locks accept a Tweedie φ̂ that is off by a large factor.

**1. `tests/regressions/families/tweedie_dispersion_frozen_at_one.rs`** (`tweedie_prediction_se_scales_with_dispersion`)

```rust
assert!(ratio >= 2.0, ...);   // expected ratio ≈ 5.16 = sqrt(8/0.3)
```

- The gate is one-sided. The expected SE ratio is 5.16, so a φ̂ ratio anywhere from 4× (ratio 2) upward passes. An inflated high-φ fit passes too.
- φ̂ itself is never checked. This is the quantity the #771 frozen-φ bug was about.

**2. `tests/regressions/families/tweedie_dispersion_observation_interval.rs`** (`tweedie_observation_interval_reflects_estimated_dispersion`)

```rust
assert!(rel_err_lo < 0.20, ...);
assert!(rel_err_hi < 0.20, ...);
```

At n = 4000 the Pearson estimator's SD is 2.4% (φ = 0.4) and 3.7% (φ = 6). A 20% bar therefore accepts a bias of 8 and 5 standard deviations.

**Principled fix (derived gates)**

1. **φ̂ gate, in both tests.**
   - Require φ̂ within 4 SDs of the true φ, plus the n versus n − p divisor bias (2φ/n).
   - The SD is that of the Pearson estimator with a plug-in μ̂ (engine: `estimate_tweedie_phi_from_eta`). It comes from the Tweedie cumulants with the influence correction for the fitted mean:
     - `Var tᵢ = p(2p−1)φ³μᵢ^{p−2} + 2φ²`
     - `Cov(tᵢ, sᵢ) = pφ²`
     - `n·Var φ̂ = mean[p(2p−1)φ³μ^{p−2} + 2φ²] − p²φ³·x̄ᵀH⁻¹x̄`, with `H = mean μ^{2−p}xxᵀ`.
   - Validated by simulation. On 2000 replicates of each fixture, z has SD 0.99 and max |z| = 3.9 at φ = 0.4 and φ = 6. At φ = 0.3 and φ = 8, z has SD 1.03 and 0.96. A naive i.i.d. SE gives z SD 0.56 at high φ.
2. **SE check, in the frozen-at-one test.**
   - Each SE(β̂) must lie within the Fisher/observed-information hull (±1e-4 relative) computed from the fit's own μ̂ and φ̂.
   - For this fixture the Fisher and observed-information SEs differ by only 0.06% (φ = 0.3) and 0.4% (φ = 8).
   - The SE ratio is printed only.

**Also noted (filed separately).** The observation-interval test's `fit_tweedie` comment says the upper endpoint "equals μ̂ + z·σ exactly". On main the Tweedie band is the skew-aware compound Poisson–Gamma quantile (`family_observation_band` in `crates/gam-predict/src/lib.rs`), not μ̂ + z·σ. Section 3 of that test therefore scores a symmetric band [μ̂ − (upper − μ̂), upper] that the engine does not produce, against a one-sided coverage floor of 0.90.

### LINKED ISSUE #4366: tests/regressions: tweedie_dispersion_observation_interval scores a symmetric band the engine does not produce
## Where

`tests/regressions/families/tweedie_dispersion_observation_interval.rs`
- lines 11 and 142–153: the claim that the band is `μ̂ + z·σ`
- lines 163–172 and 253–284: section 3, the coverage score

## The code

```rust
// ... the upper endpoint is never clamped (support is
// unbounded above) and equals `μ̂ + z·σ` exactly, which is what this test's
// φ-scaling check needs.
let obs_halfwidth: Vec<f64> = upper.iter().zip(pred.mean.iter()).map(|(&hi, &mu)| hi - mu).collect();
...
fn coverage(y: &[f64], mean: &[f64], halfwidth: &[f64]) -> f64 {
    ... yi >= mi - hw && yi <= mi + hw ...
}
...
assert!(cov_hi_fitted >= 0.90, ...);                 // nominal 0.95
assert!(cov_hi_fitted >= cov_hi_frozen + 0.04, ...);
...
assert!(width_ratio >= 2.0, ...);                    // expected ≈ 3.87
```

## Why it is wrong

1. **The engine does not produce a symmetric band.** The Tweedie observation band comes from `family_observation_band` (gam-predict `lib.rs`, the Tweedie arm near line 2550), which calls `tweedie_moment_matched_interval(m, φ, p, V, p_lo, p_hi)`. That returns the equal-tailed compound Poisson–Gamma quantile pair at the effective dispersion `φ_eff = V / m^p`, where:
   - `m` and `v` are the log-normal mean and variance of `exp(η̂)`,
   - `V = φ̂·m^p(1+v/m²)^{p(p−1)/2} + v`,
   - the tail masses are `reference.cdf(∓z)`, Student-t on `n − edf` when the scale is estimated,
   - `lower = 0` whenever the zero atom `P(Y = 0)` covers the lower tail.

   The upper endpoint is a skewed quantile, not `μ̂ + z·σ`, so the comment at lines 142–148 is false. The module doc (line 11) repeats the same claim.

2. **Section 3 scores a band nobody reports.** It builds `[μ̂ − (upper − μ̂), upper]` from the upper endpoint and scores that against `y`. This band is not the engine's `[lower, upper]`. The coverage it measures says nothing about the band users receive.

3. **The thresholds are not derived.**
   - `cov ≥ 0.90` is one-sided and 5 points below nominal. SPEC treats conservative intervals as bugs, so the upper side also matters.
   - `cov_fitted ≥ cov_frozen + 0.04` and `width_ratio ≥ 2` (the expected value is `√(6/0.4) ≈ 3.87`) cannot tell a band built at the right dispersion from one built at, say, half of it.

   A band built with the wrong tail masses, the wrong `φ_eff`, or a skipped zero-atom rule would pass all three.

## Principled fix

Test the construction exactly instead of scoring an invented band.

1. **φ̂ gate.** Keep the φ̂ recovery gate, with bounds derived from the sampling SD of φ̂.
2. **Dispersion read-through.** Assert that the predictor reads the fit's dispersion: `fixed_phi() == φ̂` within a few ulps.
3. **Per-row quantile check.** At every row of both fits:
   - Recompute `m`, `v`, `V` and `φ_eff` from `η̂`, `SE(η̂)` and `φ̂`.
   - Check that `F(lower) = p_lo` and `F(upper) = p_hi`, where `F` is an **independent** p = 1.5 Tweedie CDF. For p = 1.5, `α = 1`, so each Gamma summand is exponential, and the CDF is the positive-term series `F(y) = P(Poisson(y/θ) ≥ N)` summed over `N ~ Poisson(λ)`.
   - Use a tolerance derived from the two CDFs' truncation and rounding errors.
   - Check that `lower = 0` exactly when `P(Y = 0) ≥ p_lo`.
4. **Coverage is printed, not gated.** Print exact coverage under the true law next to its nominal value. A coverage threshold would need the sampling error of `(β̂, φ̂)` propagated through the quantiles.

The fix is on branch `hn23-tweedie-phi-recovery` (PR #4144, commit 1174f5d0).

SauersML:
Fixes #4131
Fixes #4366

- **φ̂ gate, in both locks.** φ̂ must lie within 4 SDs of the true φ, plus the (n − p) divisor bias. The SD is that of the Pearson estimator with the influence correction for the fitted mean. It is derived from the Tweedie cumulants and was checked by simulation (z SD ≈ 1 at all four φ). This replaces the old 20% bar and the one-sided `SE ratio >= 2` check (expected value 5.16).
- **SE gate, in the frozen-at-one test.** Each SE(β̂) must lie within the Fisher/observed-information hull computed from the fit's own μ̂ and φ̂.
- **Observation band (#4366, commit 1174f5d0).** The test used to score a symmetric band `[2μ̂ − upper, upper]` that the engine never builds. It now checks, at every row of both fits, that the reported `[lower, upper]` is the Tweedie(m, φ_eff(φ̂)) equal-tailed quantile pair at the fit's reference tail masses.
  - The comparison uses an exact p = 1.5 CDF written independently of the engine (a positive-term Poisson series), to `CDF_TOL = 1e-10`.
  - It includes the zero-atom `lower = 0` rule.
  - It asserts `fixed_phi() == φ̂`.
  - Exact coverage under the true law is printed next to its nominal value and is not gated.

Validation: focused-rust-proof, keys HN-23-tweedie-1, HN-23-tweedie-2 and HN-23-tweedie-3.


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, d17dd0c6a02f.

PR #4146, opened 2026-09-20
tests/survival: drop slope-Jacobian tests that exercise no library code

### LINKED ISSUE #4133: tests/survival: marginal-slope slope_jacobian_at_* tests compare test-local Jacobian to FD of test-local η map (exercise no library code, can never fail)
**File:** `tests/survival/survival/survival_marginal_slope_jacobian_hyperbolic_correction.rs:540-781` (origin/main f076f9d85b)

**Code.** Three tests:

- `slope_jacobian_at_zero_beta_equals_sf_diag_z_phi` (l.636)
- `slope_jacobian_at_small_beta_matches_fd` (l.691)
- `slope_jacobian_at_moderate_beta_has_hyperbolic_correction` (l.716)

Each one goes through `check_slope_jacobian`:

```rust
let analytic =
    analytical_slope_jacobian(&data.phi, beta_slope, q0, q1, qd1, &data.z, s_f);   // test-local fn, l.255
...
let fd = finite_diff_jacobian(|b| compute_eta_stack(phi_ref, b.as_slice().unwrap(), ...), &beta_arr, 1e-6); // test-local fn, l.230
let rel_err = max_col_rel_error(&analytic, &fd);
assert!(rel_err < 1e-5, "... the hyperbolic correction formula is wrong");
```

The β=0 test also checks `analytical_slope_jacobian` against its own closed form. The moderate-β test calls `compute_row_scalars` and `analytical_slope_jacobian`, both test-local.

**Why it is wrong.** None of the three tests calls library code. The test file compares its own hand-written Jacobian against an FD of its own hand-written η map. These tests can fail only if the file's algebra is wrong, and no change to `gam` can make them fail.

The file still describes them as a production guard:

- The module doc says: "T1's `BlockEffectiveJacobian` implementation for the slope block must return the FULL formula ... A static diagonal-scaling shortcut would ... fail the moderate-β finite-difference check below."
- The moderate test is labelled `SENTINEL: if this test fails ... T1's impl uses a static diagonal shortcut`.
- The trailing comment says: "The local-model tests above keep guarding the hyperbolic-correction formula through the public trait."

Production's slope Jacobian is crate-internal (#2352), so none of this is true. It is guarded in-crate by `crates/gam-models/src/survival/marginal_slope/tests.rs::slope_jacobian_hyperbolic_correction_matches_fd_with_scalars`.

The `check_effective_jacobian_matches_fd` failure message makes the same false claim. It blames "T1's implementation" when `effective_jacobian_at` disagrees with the FD, but the Jacobian it returns comes from the test-local `SlopeJacobianImpl`.

**Principled fix.** Delete the three self-referential tests and their `check_slope_jacobian` helper. The remaining `effective_jacobian_at_*` tests already FD-check the same test-local formula, so no formula coverage is lost.

Rewrite the module doc, the FD-assert message and the trailing comment to say what the remaining tests guard, which is the public plumbing:

- `ParameterBlockSpec::effective_jacobian_at` dispatches to the block callback and forwards `family_scalars`.
- `audit_identifiability_channel_aware` reports distinct per-row scalings as non-aliased.

The doc should point to the in-crate test for the production formula. No tolerance changes and no test is weakened: the removed assertions could not observe the library.

SauersML:
Fixes #4133

This removes `slope_jacobian_at_{zero,small,moderate}_beta*` and their `check_slope_jacobian` helper. They compared a test-local Jacobian against an FD of a test-local η map, so no change to `gam` could make them fail.

The module doc, the FD-assert message and the trailing comment are rewritten to say what the remaining tests guard:

- `effective_jacobian_at` dispatch and `family_scalars` forwarding;
- `audit_identifiability_channel_aware`.

The doc also points to the in-crate test for the production formula, `gam-models survival/marginal_slope/tests.rs::slope_jacobian_hyperbolic_correction_matches_fd_with_scalars`. No tolerance changed.

Validation: focused-rust-proof, key HN-23-selfjac-1 (survival binary).

---


SauersML:

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 10d483835437.

PR #4147, opened 2026-09-20
Remove assertion-free sphere fit diagnostic

### LINKED ISSUE #4134: tests/manifolds: sphere_m_sweep_lambda_diagnostic asserts nothing, swallows fit errors, prints hard-coded 'outer_converged=certified'
**File:** `tests/manifolds/manifolds/sphere_m4_lambda_diagnostic.rs:45-154` (origin/main)

**Code.**

```rust
#[test]
fn sphere_m_sweep_lambda_diagnostic() {
    ...
        let result = match fit_from_formula(&formula, &data, &cfg) {
            Ok(r) => r,
            Err(e) => { eprintln!("[diag] m={m} FIT ERROR: {e}"); continue; }
        };
        let FitResult::Standard(fit) = result else { eprintln!(...); continue; };
        ...
        eprintln!("[diag] m={m} REML score = {reml_score:?} outer_iters={outer_iters} \
                   outer_converged=certified outer_grad_norm={outer_grad_norm:?}");
        ...
        let collapsed = var_pred < 0.01 * truth_var;
        eprintln!("[diag] m={m} COLLAPSED? {}", if collapsed { "YES" } else { "no" }, ...);
```

**Why it is wrong.** This `#[test]` has no assertion; its module doc says so: "it never asserts pass/fail, it just logs numbers".

- Every fit error and every non-standard result is logged and skipped.
- A collapsed fit only prints `YES`.
- The line `outer_converged=certified` is a string literal. It is printed whatever the fit reports.

The test passes on every build, including one where all four sphere fits error, so it runs in CI while it can never fail. SPEC: "A failing test should always indicate problematic behavior". The converse holds too: a test that cannot fail is not a test.

The hypothesis it probes, a Wahba m=4 collapse to the response mean, is already asserted on the identical dataset (seed 7, n=400, σ=0.05, same 15×15 grid):

- `tests/manifolds/smooths/sphere_penalty_order_sweep.rs::sphere_wahba_m4_must_fit_smooth_truth`
- `tests/manifolds/smooths/sphere_both_kernels_fit_smooth_truth.rs`, which checks every m for both kernels.

**Principled fix.** Remove the file and its `mod` line. The tests above carry the assertions. Turning it into yet another m-sweep with a new threshold would duplicate them.

SauersML:
Remove the four-fit sphere diagnostic that printed results, continued after fit errors, and hard-coded `outer_converged=certified`. Its quality hypothesis already has registered tests that fail on fit errors and excessive RMSE.

Reviewed against current main: the retained `sphere_penalty_order_sweep` and `sphere_both_kernels_fit_smooth_truth` modules use the same seed-7, n=400 generator. The dedicated m=4 test requires RMSE <= 0.10; the Sobolev and pseudo-spline sweeps collect errors as failures. Their current function names and registration were checked directly. Added the two removed private functions to the source-removal ledger.

Validation is a source/reference audit of this test-only deletion. No production behavior changes, and no numerical fitting run is claimed.

Closes #4134.


SauersML:

---

PR #4148, opened 2026-09-20
Custom families refuse a persistent warm-start store they cannot key; BMS fingerprints its likelihood (#3002)

### LINKED ISSUE #3002: persistent_warm_start_root is silently ignored by every custom family except survival marginal slope (BMS included)

SauersML:
Fixes #3002

## Defect
`CustomFamily::persistent_warm_start_fingerprint` defaults to `None`. Survival marginal slope is the only family that overrides it. `persistent_custom_family_key` turned that `None` into "no key" with `?`, and `load_persistent_custom_family_warm_start` turned "no key" into `(None, None)`. So a configured store was never read or written, and the fit ran cold without reporting it. The same happened when a block design could not be hashed (`hash_cf_design_matrix(..).ok()?`).

## Fix
- **The key returns `Result<String, CustomFamilyError>`.**
  - If the family gives no likelihood-data fingerprint, the fit fails with `UnsupportedConfiguration`. The error names `type_name::<F>()` and points to `warm_start_from`.
  - If a design cannot be hashed, the fit fails with `InvalidInput`, naming the block.
- **The loader propagates the error, and `fit_custom_family_with_rho_prior` refuses the fit with `?`.** The refusal happens while the key is computed, before the store is opened. When no store is configured, nothing changes.
- **Bernoulli marginal slope (BMS) now has a real fingerprint.** A record's mode is the minimiser of `−ℓ(β; data) + ½βᵀS_λβ − τΦ`, so the key must identify every input to `ℓ` held on the family. The fingerprint hashes:
  - `y`, `weights` and `z`;
  - the latent measure (serde) plus its `#[serde(skip)]` training-row mixtures;
  - the Gaussian frailty SD;
  - the base inverse link;
  - the marginal and slope designs;
  - the score-warp and link-deviation span tables (`c0..c3`, endpoints, monotonicity rows, right-boundary row) and any installed anchor correction with its component tags;
  - the residual repair block: its features, its geometry, and the covariance field the row kernel reads (pooled covariance and every materialised per-row covariance, by representation);
  - the Jeffreys arming flag.

  Caches, the resource policy and multistart state affect only the numerics, so they are left out.
- **One design hash.** `hash_cf_design_matrix` is now public and re-exported from `gam-custom-family`, so a family fingerprint and the block-spec key use the same design byte stream and there is no second copy.
- **The trait doc now states the contract:** `None` means a configured store is refused.

## Test
`a_configured_store_is_refused_by_a_family_without_a_fingerprint_3002` (gam-custom-family lib) gives the quartic test family, which inherits the `None` default, a configured store. It checks three things:
- the fit is refused, with a message that names `DeclaredDenseQuarticWorkspaceFamily` and "does not fingerprint";
- the store root was never created;
- the same fit without the store certifies.

## Remaining (owner lane)
The remaining items from #3002 are listed on the issue: P1b and the store-deletion question (P2).


SauersML:
Ready for the local merge train (MT-0, even PR). Waiting only on its two focused proof runs, which are still queued behind the runner backlog:
- run 35505965434, key BF-54-3002-1
- run 35505966722, key BF-54-3002-2

Both are at head b12bc6f2c0fc59987a1a4c47d56f47c54f2d7cb4. Merge with merge_method `merge` and expectedHeadSha b12bc6f2c0fc59987a1a4c47d56f47c54f2d7cb4, but only once both runs are green. On merge, #3002 closes via "Fixes". If a run fails, do not merge. Re-dispatch as BF-54-3002-3 and onward after the fix.

---

PR #4154, opened 2026-09-20
manifold: exact-A residual routes share patchd_row_error_metric; softmax plan reads each gate once

### LINKED ISSUE #4149: manifold/construction_exact_hessian + row_jet_logdet_channels: three exact-A routes re-inline the metric-applied residual that patchd_row_error_metric (#2515) exists to own; the softmax plan evaluates each row's gate twice
`SaeManifoldTerm::patchd_row_error_metric` (`construction_exact_hessian.rs`) builds the `√w`-scaled, metric-applied per-row residual `√w·M_n r_n`. Its doc says it exists so that the exact-A legs "build it ONCE rather than three times. A route that reconstructed the residual with a different weighting would produce a Patch-D leg that silently disagreed with the operator it is supposed to differentiate, which is the failure this whole front is about."

Three production routes still re-inline the same about 25 lines: the fitted accumulation over active atoms, `√w·(fitted − target)`, and `apply_metric_row` when whitening:

1. `prepare_residual_curvature_rows` (`construction_exact_hessian.rs`), the matrix-free residual-curvature row plan.
2. `assemble_exact_hessian_minus_b_rows` (same file), the dense `A − B` row assembly.
3. `prepare_softmax_row_jets` (`construction_row_jet_logdet_channels.rs`), the softmax row-jet plan's HVP probe. This copy also **re-evaluates the gate** for every row into a second scratch `probe_assignments`, although `assignments` was filled for the same row a few lines earlier. The error context of that second evaluation names the wrong function (`"apply_exact_hessian_minus_b: assignment scratch is not contiguous"`).

Today the four copies agree bit for bit, but nothing keeps them that way. Any change to the weighting, the active-atom set or the whitening rule has to be made four times, and a miss would reproduce exactly the silent disagreement that #2515 extracted the helper to prevent.

Smaller dead code in the same functions:
- `let p = self.output_dim()` is used only by the inlined buffers in (1) and (2).
- `prepare_softmax_row_jets` re-matches `self.assignment.mode` inside its `keep` closure with an unreachable `_ => Ok(None)` arm, after the function has already refused every non-softmax mode.
- `construction_row_jet_logdet_channels.rs` carries an empty `#[cfg(test)] mod tests_reconstruction_program_builder { impl SaeManifoldTerm {} }`.

**Fix (bit-identical).**
- All three sites call `patchd_row_error_metric(row, w_row, target, &assignments, whitens)`.
- The softmax plan reuses the row's `assignments`, destructures `temperature` once with `let … else`, and calls `prepare_bilinear_contractions` directly.
- The unused `p` and the empty test module are removed.
- The helper doc names every consumer.

SauersML:
Fixes #4149

- `prepare_residual_curvature_rows`, `assemble_exact_hessian_minus_b_rows` and `prepare_softmax_row_jets` now call `patchd_row_error_metric` (#2515). Each used to rebuild the `√w`-scaled, metric-applied row residual inline.
- `prepare_softmax_row_jets`:
  - reuses the row's `assignments` instead of evaluating the gate a second time into `probe_assignments`, which also carried the wrong error context;
  - destructures `temperature` once with `let … else`;
  - calls `prepare_bilinear_contractions` directly, dropping the unreachable non-softmax arm.
- Removed the unused `let p = self.output_dim()` locals and the empty `tests_reconstruction_program_builder` module. The helper doc now names every consumer.

The change is bit-identical: the same accumulation order, the same `√w`, the same active-atom skip and the same whitening rule.

---


SauersML:

---

SauersML:

---

PR #4157, opened 2026-09-20
gam-math beta_quantile: certify the lower-tail Newton root by its residual band

### LINKED ISSUE #4153: gam-math beta_quantile: lower-tail log-space Newton returns Some(x) on its 32-step cap exit; its step test is not derived from G's rounding band
**Area:** `crates/gam-math/src/probability.rs`, `lower_tail_beta_quantile` (origin/main `b9758d35a0`, lines 83-135; `BETA_NEWTON_MAX_STEPS` at line 271)

SPEC: "A fit object must only ever come from a converged optimization." / derived tolerances only.

### What the code does

The lower-tail branch of `beta_quantile` (the #2528 fix) refines `y = ln x` with Newton on `G(y) = a·y − ln B(a,b) + ln S(e^y) − ln p`:

```rust
for _ in 0..BETA_NEWTON_MAX_STEPS {          // 32, documented as "the non-convergence guard"
    ...
    y -= step;
    if step.abs() <= f64::EPSILON * y.abs().max(1.0) { break; }
}
let x = y.exp();
if x.is_finite() && (0.0..=1.0).contains(&x) { Some(x) } else { None }   // same return on cap exit
```

The loop returns `Some(x)` in both cases, whether the step test fired or the guard ran out. So a cap exit reads as a resolved quantile.

### Why the cap binds

The step test is not derived from `G`'s rounding. `G` sums four `O(1)`-to-`O(10)` terms, so its computed value carries noise of order `ε·(|a·y| + |ln B| + |ln S| + |ln p|)`. The Newton step divides that noise by `G′ ≈ a`. For small `a` the result is larger than `ε·|y|`, and the iterate oscillates at noise level (e.g. steps of `±4.4e-14` forever for `Beta(0.04, 3.96)`, `p = 1e-3`) until the 32-step guard ends the loop.

A float simulation of the exact loop (same series, same seed, same tests) over a grid:
- `a ∈ [1e-3, 5]`, `b ∈ [0.3, 1000]`, `p ∈ [1e-10, 0.975]`;
- 87 of 692 on-branch cases end on the cap, not the step test. Examples: `Beta(0.04, 3.96)` at `p ∈ {1e-3, 0.9, 0.975}`, `Beta(0.02, 100)` at `p = 0.5`, `Beta(1, 0.3)` at `p = 0.05`.

In those cases the returned value happens to be at noise level. But the function cannot tell that exit apart from a genuine failure to converge, and the doc claim that the guard is only a non-convergence guard is false.

### Principled fix

Certify the returned point, not the loop exit:
- Re-evaluate `G` at the `y` being returned.
- Accept it only when `|G| ≤ band`, with the band derived from the arithmetic that forms `G`: `γ₄·(|a·y| + |ln B| + |ln S| + |ln p|)`, plus the series' own bound `γ_{5K+3}·Σ|t_k|/S` for `K` terms.
- The step test stays only as the stopping rule.
- A point outside the band is refused as `Some(NaN)`. It is not `None`, which would fall back to `inv_beta_reg`, whose floor on this branch is exactly #2528.

In the same simulation every returned point lies within `0.23×` the band, so certified outputs are bit-identical to today's.

Claimed by local agent HN-36 (claude/fix-beta-quantile-certificate); draft PR incoming.

SauersML:
Lower-tail beta quantiles now require a finite residual consistent with the series, normalization, and floating-point allowances. Numerical refusal cannot silently dispatch back to the absolute-tolerance inverse that loses tiny quantiles.

The normalized ascending series carries complete terms and x times its derivative, avoiding independent Pochhammer overflow and power underflow. Its normalization reuses the live log-gamma shift-gap evaluator, which now reports its analytic Stirling remainder and an engineering floating-point allowance. Unit-shape beta families use their exact closed forms; genuine underflow is verified at the half-subnormal boundary. The shared shift-gap implementation also preserves finite answers when x+shift overflows.

Independent high-precision checks uncovered inaccurate existing tiny-quantile reference values; these were replaced using 100-digit bisection at exact binary64 inputs. New 400-digit references cover non-unit extreme skew and finite shift gaps near f64::MAX.

Validation on MSI: rebuilt whole gam_math with warnings denied; all 96 probability tests and both shared shift-gap regression tests pass. The acceptance allowance is not a claim of correctly rounded quantiles or directed interval arithmetic.

Closes #4153.


SauersML:

---

PR #4158, opened 2026-09-20
Derive the solve-certificate allowance from each factorization's backward error (#4038)

### LINKED ISSUE #4038: gam-linalg utils: certified symmetric solve's backward-error allowance (256·n·ε) and symmetry allowance (32 ulp) are arbitrary constants, not derived bands
## Defect

The shared solve certificate in `crates/gam-linalg/src/utils.rs` accepts or refuses solves using two hand-picked multipliers. Neither comes from the error analysis of the arithmetic it certifies.

`crates/gam-linalg/src/utils.rs:469-470` (origin/main):

```rust
const SYMMETRY_ULP_ALLOWANCE: f64 = 32.0;
const SOLVE_ROUNDOFF_OPS_PER_DIMENSION: f64 = 256.0;
```

`crates/gam-linalg/src/utils.rs:605-609`:

```rust
fn solve_backward_error_allowance(dimension: usize) -> f64 {
    let roundoff = SOLVE_ROUNDOFF_OPS_PER_DIMENSION * dimension as f64 * f64::EPSILON;
    roundoff / (1.0 - roundoff)
}
```

`crates/gam-linalg/src/utils.rs:554`:

```rust
let tolerance = SYMMETRY_ULP_ALLOWANCE * positive_ulp(pair_scale.max(gram_scale));
```

These two numbers decide whether every certified solve passes. That covers:
- `certified_symmetric_solve`
- `CertifiedSpdFactor::{solve, solve_matrix, inverse}` and `certified_spd_inverse`
- every external caller of `certify_linear_system_residual`:
  - `gam-solve/src/estimate/optimizer.rs:27`
  - `gam-solve/src/estimate/identified_hessian.rs:122,144`
  - `gam-solve/src/estimate/smoothing_correction.rs:1027`

SPEC: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code." The workspace already has the machinery to derive both numbers: `gam_math::roundoff::accumulation_growth`, `cholesky_completion_band`, `symmetric_inertia_band` and `gam_linalg::roundoff::weighted_gram_assembly_band`.

### 1. `256·n·ε` is about 128 times the bound a Cholesky solve can reach

The certificate is `η = ‖AX̂ − B‖max / (n‖A‖max‖X̂‖max + ‖B‖max)`. For the strict-Cholesky path, which is the SPD Hessians behind almost every caller, the bound on `η` follows directly from the analysis:

- **Factor and both triangular solves** (Higham, *ASNA* 2nd ed., Thm 10.4): `(A + ΔA)x̂ = b` with `|ΔA| ≤ γ_{3n+1}|R̂ᵀ||R̂|`.
- **Bound on the factor product** (Demmel; the same step `cholesky_completion_band` already uses): `(|R̂ᵀ||R̂|)_ij ≤ ‖r̂_i‖‖r̂_j‖ ≤ √(a_ii a_jj)/(1 − γ_{n+1}) ≤ ‖A‖max/(1 − γ_{n+1})`. So the exact residual satisfies `‖b − A x̂‖∞ ≤ n·γ_{3n+1}/(1 − γ_{n+1})·‖A‖max‖x̂‖max`.
- **Forming the residual** (`matrix.dot(solution) − rhs`): each entry is a length-`n` inner product followed by one subtraction. That adds at most `γ_{n+1}(|A||x̂| + |b|)`, and in the max norm this is `γ_{n+1}` times the certificate's own denominator.

Together:

```
η̂ ≤ γ_{3n+1}/(1 − γ_{n+1}) + γ_{n+1}  ≈ (4n + 2)·u = (2n + 1)·ε
```

The code allows `256·n·ε`, which is 128 times looser. Consequences:
- A solve is "certified" even when its backward error is two orders of magnitude beyond what any correctly executed Cholesky can produce. That backward error can only come from a defect, such as element growth in the unpivoted-LDLT rung of #3519 or a corrupted factor.
- So the certificate does not certify the property its docs claim: "success carries an a posteriori backward-error proof".
- In the other direction, one fixed multiple of `n·ε` is not the right shape for the pivoted-LBLT path (`StableSolver` after Cholesky refuses). The backward error of that path carries the factor term `‖|L̂||B̂||L̂ᵀ|‖` that `symmetric_inertia_band` already models, and `FaerLblt::inertia().factor_magnitude` already measures.

### 2. `32 ulp` does not scale with the assembly depth it exists to absorb

The comment at `utils.rs:522-552` explains why the symmetry allowance scales with `√(A_ii A_jj)`: the two triangles of `XᵀWX` built by a full GEMM (`CrossprodStructure::Full`) are separate length-`m` accumulations. Their difference is bounded by `2γ_m·Σ_k|x_ki w_k x_kj| ≤ 2γ_m√(A_ii A_jj)` for `W ≥ 0`.

That bound grows linearly in the row count `m`. Even its typical (random-walk) size grows like `√m`. A fixed 32 ulp (about `32ε·scale`) is:
- exceeded by the worst-case band once `m ≳ 32`;
- exceeded by the typical band once `m ≳ 10³`.

So a correctly assembled full-GEMM Gram over ordinary sample sizes can be refused with `NotSymmetric` purely by roundoff. For hand-built or mirrored inputs, where the two triangles are bit-identical, the constant is also looser than it needs to be.

The validator cannot know `m`, so no fixed constant can be right. The band has to come from whoever assembled the matrix: `weighted_gram_assembly_band(m, formation_roundings, Σ|w_i|‖x_i‖²)`, entrywise `γ_{m−1+k}√(A_ii A_jj)`. Alternatively, the caller mirrors one triangle at assembly, and then exact symmetry is the right demand.

## Principled fix

1. **Backward-error allowance.** Replace `SOLVE_ROUNDOFF_OPS_PER_DIMENSION` with the derived band for the factorization that produced the solve:
   - Cholesky (strict SPD path and the LLT branch of `StableSolver`): `γ_{3n+1}/(1 − γ_{n+1}) + γ_{n+1}`, built from `accumulation_growth`.
   - Bunch–Kaufman branch: the same form, with the measured factor term from `FaerLblt::inertia().factor_magnitude` relative to `n‖A‖max`, as `symmetric_inertia_band` does.
   - `certify_linear_system_residual` is used by callers whose solve is not a Cholesky (the eigen pseudo-inverses in `identified_hessian.rs` and `smoothing_correction.rs`). It should take the caller's backward-error band as an argument, instead of applying one universal constant to every solver.
2. **Symmetry allowance.** Replace `SYMMETRY_ULP_ALLOWANCE` with either:
   - an assembly band passed by the caller (`weighted_gram_assembly_band` scaled entrywise by `√(A_ii A_jj)`), or
   - exact symmetry, for callers that mirror one triangle at assembly.

   The factorization reads only `Side::Lower` in any case.

SauersML:
Fixes #4038

## What this changes

`certify_linear_system_residual` used to accept a solve whenever its max-norm backward error was at most `256·n·ε`. Nothing derived the `256`.

The function now takes an explicit `allowed_backward_error`. Each caller computes that allowance from the backward error of the factorization that actually produced the solution, using new functions in `gam_linalg::roundoff`. `SOLVE_ROUNDOFF_OPS_PER_DIMENSION` and `solve_backward_error_allowance` are deleted.

The certificate itself is unchanged: `η = ‖A·X̂ − B‖_max / (n·‖A‖_max·‖X̂‖_max + ‖B‖_max)`.

Every band has two parts:
1. **The solve's own backward error.** If `(A+ΔA)x̂_j = b_j`, the exact residual is `−ΔA·x̂_j`, which contributes `‖ΔA‖_max/‖A‖_max` to η.
2. **Forming the residual.** Each computed residual entry is an inner product of n terms plus one subtraction, which contributes `γ_{n+1}` (Higham, *ASNA* §3.1).

Throughout, `γ_k = k·u/(1−k·u)` with `u = ε/2`.

## Bands

**Cholesky: `cholesky_solve_backward_band(n)`.**
- Higham Thm 10.4 gives `|ΔA| ≤ γ_{3n+1}|R̂ᵀ||R̂|`, and Thm 10.3 gives `‖r_i‖² ≤ a_ii/(1−γ_{n+1})`.
- Together: `band = γ_{3n+1}/(1−γ_{n+1}) + γ_{n+1} ≈ (2n+1)·ε`.
- There is no growth term, so the old `256·n·ε` was about 128 times looser than anything Cholesky can produce.

**Symmetric indefinite LDLᵀ / Bunch–Kaufman LBLᵀ: `symmetric_factor_solve_backward_band`.**
- Higham Thm 11.3–11.4 gives `|ΔA| ≤ γ_k(|A| + Pᵀ|L̂||B̂||L̂ᵀ|P)` with `k = 3n + 2⌈c_B⌉ + 3`.
- So `band = γ_k·(‖A‖_max + G)/‖A‖_max + γ_{n+1}`.
- `G` bounds `‖|L̂||B̂||L̂ᵀ|‖_max` and is **measured on the computed factor** by `FaerLdlt/FaerLblt::factor_product_bound`. This is the element growth. It is the honest reason an indefinite solve is less accurate than Cholesky.
- `c_B` is the rounding count of the pivot-block arithmetic: 2 for a 1×1 pivot, and the measured count of the block formula for a 2×2 pivot.

**Dense factorizations that read one triangle** also add `symmetric_part_defect_band = fl(max|A−Aᵀ|/2)/‖A‖_max`. The factor solves the symmetrized matrix, but the residual is taken against the stored one. This term is 0 for sparse storage.

**Spectral (identified-Hessian) solve: `spectral_solve_backward_band`.**
- Inputs are the eigensolver's *measured* residual `‖HU−UΣ̃‖₂` (bounded by `eigen_residual_two_norm_bound`) and the *measured* orthonormality defect `ω ≥ ‖UᵀU−I‖₂` (bounded by `orthonormality_defect_bound`).
- With `ν = √(1+ω)` and `θ = √(1−ω) − γ_r√r·ν`:
  `band = γ_{n+1} + ‖R_e‖₂/(θ√n‖H‖_max) + (ν/θ)·√n·(u + 2γ_r)`.

**PSD pseudo-inverse (smoothing correction): `psd_pseudo_inverse_backward_band`.**
- The same ingredients, certified against the projector:
  `band = γ_{n+1} + ‖R_e‖₂/(√(1−ω)(1−γ_{r+3})√n‖H‖_max) + γ_{r+3}/(1−γ_{r+3}) + γ_r(1+ω)/(r(1−ω)/n − γ_r(1+ω))`.
- At rank 0 the band is `γ_{n+1}`.
- The inverse is certified first and symmetrized afterwards. The symmetric projection `(E+Eᵀ)/2` is no larger than `E` in max-norm, so symmetrizing cannot raise η.

If a band is non-finite or negative (γ undefined, a zero-scale matrix, NaN growth), the certificate now **refuses** with `NoBackwardErrorBound`. Before, it certified against a meaningless threshold.

## Tests (new)

- `roundoff::tests::cholesky_solve_band_is_its_rounding_count_4038`: the band equals `(4n+2)u` to first order and stays below `256nε` for n = 1…10⁴.
- `roundoff::tests::symmetric_factor_band_is_priced_by_measured_growth_4038`: exact at `G = ‖A‖`, linear in `G`, and infinite for INF/NaN/zero inputs.
- `roundoff::tests::orthonormality_defect_bound_covers_gram_formation_4038`
- `roundoff::tests::spectral_and_pseudo_inverse_bands_charge_the_measured_residual_4038`
- `utils::tests::ill_conditioned_spd_solve_certifies_at_the_cholesky_band_4038`: the n=8 Hilbert matrix certifies at the Cholesky band, which is below `256nε`.
- `utils::tests::indefinite_solve_certifies_at_its_factor_growth_band_4038`
- `utils::tests::a_non_finite_band_refuses_even_an_exact_residual_4038`

Local validation (workspace check and the targeted lib tests) is in progress. Results will be posted here.

## Remainder (not in this PR)

`SYMMETRY_ULP_ALLOWANCE = 32` in the symmetry validator is a separate constant, not the solve allowance. It is reported on #4038.

---


SauersML:
MT-0 merge train: this PR conflicts with main in `crates/gam-linalg/src/faer_ndarray.rs`. This is more than a text conflict: the code this PR changes no longer exists on main.

Main's 868732f5 (#3519, "drop the unpivoted LDLT rung from the symmetric fallback") deleted `FaerLdlt` entirely. This branch adds `FaerLdlt::factor_product_bound` so it can size the LDLT solve-certificate allowance, but current main has no LDLT factor left to certify.

The resolution is to drop the `FaerLdlt` hunk and the LDLT arm of the allowance. The Cholesky and Bunch–Kaufman arms keep their backward-error bounds; `SymmetricInertia::factor_magnitude` already carries the `‖|L̂||B̂||L̂ᵀ|‖` term.

Please merge origin/main into the branch this way, as a merge commit rather than a rebase, and it will go in the next train.

---

SauersML:
Conflict resolver BF-74: I am not merging main into this branch because it was pushed to at 11:40Z (e64dc524) and its author is active. Here is what the merge with current origin/main needs.

**Text conflict, `crates/gam-linalg/src/faer_ndarray.rs`.** Main's #3519 (868732f5) deleted `FaerLdlt`. Drop this branch's whole `FaerLdlt` block, and also the `FaerSymmetricFactor::Ldlt(f) => ...` arm of `solve_backward_band`. That enum variant no longer exists on main. The Cholesky and Bunch–Kaufman arms, `pivot_blocks`, `factor_product_bound` and `pivot_block_roundings` on `FaerLblt` all merge cleanly.

**Semantic conflict, `crates/gam-solve/src/pirls/newton_solve.rs`.** Git merges this file cleanly, but it will not compile. Main's 8d646a6d added a new caller of `certify_linear_system_residual` for the dense PIRLS Newton direction, and it uses the old 6-argument signature. The fix, using the same construction as `certified_spd_factorize`:
```rust
let solve_band = factor.solve_backward_band(curvature_max_abs)
    + gam_linalg::utils::symmetric_part_defect_band(curvature.as_ref(), curvature_max_abs);
// ... pass `solve_band` before the label
```
I checked every other caller of `certify_linear_system_residual` and `IdentifiedHessianInverse::project` in the merged tree, and they are consistent.

---

SauersML:
Heads-up from the red/conflicted-PR sweep (BF-106). I am not taking this PR over, because it was pushed at 12:31Z.

The merge at 1198dee81c still has one semantic break. `crates/gam-solve/src/pirls/newton_solve.rs:669`, added on main by 8d646a6d, calls `certify_linear_system_residual` with the old 6-argument signature. The `allowed_backward_error` argument is missing, so the workspace will not compile. `git grep` over the merged tree finds no other stale callers.

The fix below mirrors `certified_symmetric_matrix_solve` (utils.rs:808). The `factor` here is the `FaerSymmetricFactor` from `StableSolver::factorize`:

```rust
    let curvature_max_abs = curvature.iter().fold(0.0_f64, |acc, v| acc.max(v.abs()));
    let solve_band = factor.solve_backward_band(curvature_max_abs)
        + gam_linalg::utils::symmetric_part_defect_band(curvature.as_ref(), curvature_max_abs);
    gam_linalg::utils::certify_linear_system_residual(
        p,
        curvature_max_abs,
        &gradient.mapv(|g| -g).insert_axis(ndarray::Axis(1)),
        &direction_out.clone().insert_axis(ndarray::Axis(1)),
        &residual.insert_axis(ndarray::Axis(1)),
        solve_band,
        "PIRLS dense Newton direction",
    )
```

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, a2ed5a6d2821.

Issue #3279, opened 2026-09-19
response::compile enrichment test red on main: the enriched Duchon REML fit is 125x worse than the coarse one (f3181a0b5c green, f737c0d8cc red)


SauersML:
**Test (red on main):** `gam-sae` `response::compile::compile_tests::the_dominance_call_adds_a_missing_direction_and_enriches_an_under_resolved_function`, panicking at `crates/gam-sae/src/response/compile_tests.rs:449`:

```
enrichment did not reduce the function error:
ErrorSplit { discarded_error: 0.18574234192647054, function_error: 10.791965839443032, function_error_standard_error: 0.41244329182261086, holdout_draws: 2000 }
then
ErrorSplit { discarded_error: 0.18574234192647054, function_error: 1346.6546288574164, function_error_standard_error: 56.68115410339019, holdout_draws: 2000 }
```

**What it means.** `response::compile` fits the retained response of a known ReLU block with a Duchon smooth under multi-penalty Gaussian REML (`gam_solve::gaussian_reml_multi_penalty::GaussianRemlMultiPenaltyProblem::fit`). The test's under-resolved function compiles at 10 centers, and `enrichment_step` recompiles it at the next resolution. The enriched fit's held-out function error is 125x the coarse fit's (1346.65 ± 56.7 against 10.79 ± 0.41), where a certified REML optimum on the richer basis cannot be worse than the coarse basis it contains. The discarded error E(P) = 0.1857 is identical in both, as it must be. So the enriched REML fit lands on a bad smoothing point.

**Where it turned red.**
- GREEN at `f3181a0b5c`: sae-feat stack gate p2-g1, in which `response::` passed 466/466.
- RED at `f737c0d8cc` on bare main, no patch: sae-feat runq run base-compile3, on 4 cores of cn1001. The numbers above are to every printed digit.
- The first-parent range `f3181a0b5c..f737c0d8cc` holds 55 commits. Two of them touch this fit's outer search:
  - `2dbe63e2f7` ("Enter the outer REML search from one derived start instead of a seed lattice…", via the #3132 merge `6817f9bc7f`). It edits `gaussian_reml_multi_penalty.rs` itself: one commensurate-curvature start per penalty, clamped into the ρ domain. `default_worker_stack_2967` went red in the same class (MAIN-REDS-0919: it starts on a dominated branch).
[...]

Issue #2596, opened 2026-07-28
Lognormal location-scale AFT recovers ~none of its generative location surface on the synthetic arm: RMSE(gam,truth)=0.787 against signal_rms=0.929, survreg gets 0.283


SauersML:
## The number

Reference Quality Suite run [30322533116](https://github.com/SauersML/gam/actions/runs/30322533116), head `87a75320e`:

```
[452/455] METRIC_OFF/quality_metric
  survival::quality_vs_survival_location_scale_lognormal::gam_lognormal_location_scale_aft_smooth_matches_survreg

lognormal location-scale AFT truth-recovery: n=300 cens=0.460 signal_rms=0.9292
  mu_rmse_to_truth      gam 0.7866   survreg 0.2832     (lower is better)
  log_sigma_abs_err     gam 0.5312   survreg 0.1276
  rel_l2_vs_survreg     0.8456

panic: gam failed to recover the generative location surface:
       RMSE(gam, truth)=0.7866 (signal_rms=0.9292)
```

`RMSE(gam, truth) = 0.787` against a signal RMS of `0.929` means the fitted location surface carries **almost none of the generative signal** — the error is 85% of the signal it is meant to reproduce. `survreg`, on byte-identical rows, gets `0.283`. The log-σ channel is off by the same factor: `0.531` vs `0.128`.

This is not a tolerance question. A fit whose truth-recovery error is 0.85 of the signal is not a slightly-worse fit; it is a fit that did not find the surface.

## Why it is only visible now

Until `74f585aa8` this test could not produce a number at all — it failed at terminal curvature ownership with `custom-family certified terminal Hessian: the certified terminal workspace did not expose its exact returned-beta Hessian` (#2580). Now the fit converges and assembles its covariance, so the quality assertion runs for the first time and this is what it says. The four tests in that cluster now read PASS / PASS / PASS / **METRIC_OFF**; the curvature defect is closed and this is what was underneath it.

## What narrows it

- **The real-data twin passes.** `gam_lognormal_location_scale_aft_smooth_matches_survreg_on_real_data` is `PASS/ok` in the same run, and both `crps_lognormal_aft` arms pass. So the location-scale AFT path is not broken in general.
- **The failing arm is the synthetic one**: `n = 300`, `cens = 0.460`, location `x + s(z, bs="tp", k=10)`, a single constant log-σ. The generative surface is smooth in `z`; `survreg` is given a fixed basis in `z` and recovers it, gam has a penalized smooth and does not.
[...]

PR #4160, opened 2026-09-20
#4024: outer REML must certify under knot_placement=quantile (reproducers first)

### LINKED ISSUE #4024: outer REML: opt-in knot_placement=quantile turns five fits that certify under uniform knots into DominatedCertifiedPlateau refusals (PR #3078's "separate optimizer finding", never filed)
PR #3078 measured this and deferred it, and no issue tracks it. From its body:

> "It also turned five existing tests red that pass on main: the convex and concave shape-constrained fits, `stiefel(k=1)`, `s(x) + s(g, x, bs=sz)` predict, and the binomial point-invariance test. All five fail in outer REML certification ("declined a certified optimum")."
>
> "Adding `knot_placement=uniform` to the default makes them pass. Pinning `k=12` (no growth, quantile still on) keeps them failing."
>
> "On the sz model, uniform placement also fails on one of four seeds, so the underlying certification fragility already exists on main; quantile only moves which designs hit it. That is a separate optimizer finding, not something to fold into a default change."

`knot_placement=quantile` is a documented, supported opt-in (`CHANGELOG.md:67`). So on these five designs, a user who opts in gets `EstimationError::DominatedCertifiedPlateau` (`crates/gam-problem/src/estimation_error.rs:854-880`) instead of a fit. The refusal is correct behaviour; SPEC says "A fit object must only ever come from a converged optimization". The defect is upstream of it: the outer search certifies an optimum that another evaluated state beats beyond the rounding envelope, and its continuation from that better state then fails to certify.

The isolation already rules out two explanations:
- **Not basis growth**, because `k=12` pinned with quantile still fails.
- **Not a quantile-only pathology**, because the sz design also fails under uniform knots on one of four seeds.

A knot layout that concentrates knots where the data are dense changes the conditioning of the penalty relative to the data. That is the regime in which the ρ-surface has a flat, ill-certified plateau.

## Tests that should carry an explicit `knot_placement=quantile` variant

Each variant must pass. None may be skipped or XFAILed.
- `tests/bug_hunt_convex_concave_smooth_collapses_to_flat_at_moderate_snr_test.py` (both the convex and the concave case)
- `tests/bug_hunt_stiefel_k1_response_geometry_matches_sphere_test.py`
- `tests/regressions/smooths/sz_factor_smooth_fits_and_predicts.rs` and `tests/bug_hunt_overlapping_smooth_predict_design_mismatch_test.py` (the `s(x) + s(g, x, bs=sz)` predict case)
- The binomial point-invariance test in `tests/test_binomial_posterior_mean_uncertainty.py`

## Likely shared root with #3228

#3228 has the same refusal class on two Gaussian location-scale wiggle tests. If that fix is a property of the plan runner's certificate/continuation pair (#2953, #2596, #2627) rather than of the wiggle family, these five designs are its family-independent reproducers. Once #3228 lands, they should be re-run with quantile placement before this issue is closed as a duplicate.

## Why this matters for the default-placement question

#3078 kept uniform placement as the default partly because of these five regressions. Quantile knots win clearly on log-normal covariates: 0.0321 vs 0.0459 and 0.0516 vs 0.0710 truth RMSE in #3078's table. So the certification fragility is also what blocks a data-adaptive placement from ever being evaluated on accuracy alone. See #3993 for the outlier-stretched-range case, where uniform placement fails outright.

I did not reproduce this on current `main`. The evidence is #3078's measurement on its merge base, `aeed1715`. The first step is to run the five tests above with `knot_placement=quantile` added.

SauersML:
Fixes #4024

## Root cause

The outer plan runner declines a certified winner `V_w` when an evaluated checkpoint `V_i` of the same attempt beats it (#2596, #2627, #2953), then continues once from that checkpoint. If the continuation does not certify, the fit ends in `DominatedCertifiedPlateau`.

The decline judged the gap `V_w − V_i` against the rounding envelope alone, `√ε·max(|V_w|, |V_i|, 1)`. But the certificate that minted `V_w` accepts a remaining decrease of up to the criterion's statistical resolution `τ_stat = 0.5/n`. Both its Newton-decrement verdict and its curvature-resolvability rung accept `½·gᵀH⁻¹g ≤ τ_stat`. So the dominance test was about `τ_stat/(√ε·|V|)` times stricter than the certificate it overrules, roughly 10³ at n = 400. In the flat valleys that quantile knot placement produces, two stops of the same basin declined each other:

1. The certified stop lost to an uncertified stop a hair lower.
2. The continuation from there then had to certify below a checkpoint that already sat inside the certificate's own tolerance.

## Derivation

Under the quadratic model the certificate uses, every state of the certified basin scores at least `V_w − τ_stat`. The two values are rounded independently, which adds `outer_value_agreement_bound(V_w, V_i)`. So an incumbent with

    V_w − V_i ≤ τ_stat + √ε·max(|V_w|, |V_i|, 1)

is consistent with the certificate and cannot contradict it. Only a gap beyond this band means the certificate is wrong about the minimum, which is the case the decline exists for: a domain face certified at 110.94 over an evaluated 4.19 (#2596). Those gaps are macroscopic and are still declined.

`outer_dominance_band(config, V_w, V_i) = outer_criterion_resolution(config) + outer_value_agreement_bound(V_w, V_i)` is now the single band that both the cached comparison and the re-evaluated comparison in `run_outer_with_plan` read. With no declared observation count, `τ_stat = 0` and the band equals the previous one.

This is a tolerance derived from the problem's own certificate. It is not a retry, fallback or tuned constant.

## Changes

- `crates/gam-solve/src/rho_optimizer/run.rs`:
  - adds `outer_dominance_band` with the derivation in its doc comment;
  - updates the `DominatedPlateau` / `DominatedPlateauRecord` docs and the log message.
- `crates/gam-solve/src/rho_optimizer/run_plan.rs`: the dominance decline reads the new band.
- `crates/gam-problem/src/estimation_error.rs`: the `DominatedCertifiedPlateau` message and docs name the dominance band.
- Regression tests:
  - `tests/regressions/optimization/quantile_knot_outer_certification_4024.rs` covers convex, concave and binomial smooths under uniform and quantile knot placement.
  - `sz_factor_smooth_certifies_under_both_knot_placements_across_seeds_4024` covers the `sz` design from PR #3078 over 4 seeds × 2 placements.

## Existing dominance tests

The #2953, #3279 and measurement-point fixtures use wells of depth 10 or gaps greater than 1, with no declared problem size, so `τ_stat = 0`. Their declines are unchanged.


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 6641c65884f8.

Issue #1867, opened 2026-07-01
[bug-hunt] Over-smoothing degrades held-out recovery: sparse_30 sin2, rw2/pspline R2=0.023, bc-anchored MSPE [#1561]


SauersML:
## Root cause (hypothesis) — over-smoothing / λ over-selection degrades held-out recovery

Three smooth-fit quality tests fail because gam over-smooths and generalizes poorly on held-out / off-grid points:

- `sparse_30_points_sin2_recovers_oscillations`: at n=30 the fit fails to recover the sin² oscillations (over-smoothed away). (`tests/perf_scale/smooths/sparse_data_smooth_fit.rs`)
- `gam_rw2_pspline_predicts_held_out_at_least_as_well_as_inla_on_real_data`: the 2-D RW2/p-spline spatial smooth does not generalize — **held-out R² = 0.0232 < 0.20**. (`tests/misc/quality_vs_inla_rw2_spde_penalized_baseline.rs`)
- `recovery_bc_anchored_left_endpoint_recovery`: boundary-condition-anchored recovery **MSPE 1.84474e-2 > 1.2·σ² = 1.20e-2**. (`tests/misc/large_scale_accuracy_sweep.rs`)

Hypothesis: shared over-smoothing / λ-selection root — REML selects too much smoothing so sparse oscillations, spatial structure, and boundary-anchored signal are shrunk toward the mean/flat fit and held-out accuracy falls below bar. Cross-ref MEMORY #1561 (Gaussian over-smoothing traced to λ-selection on a flat scale surface). These span 1-D sparse, 2-D spatial, and boundary-anchored regimes; grouped as the general over-smoothing genus (the boundary-anchored case may additionally involve endpoint/constraint handling and warrants a look if the λ fix does not clear it).

## Failing tests

- [ ] `gam::perf_scale smooths::sparse_data_smooth_fit::sparse_30_points_sin2_recovers_oscillations`
- [ ] `misc::quality_vs_inla_rw2_spde_penalized_baseline::gam_rw2_pspline_predicts_held_out_at_least_as_well_as_inla_on_real_data`
- [ ] `misc::large_scale_accuracy_sweep::recovery_bc_anchored_left_endpoint_recovery`

## Verbatim error (captured)

```
sparse n=30 fit fails to recover sin2 oscillations: ...
gam 2-D spatial smooth does not generalize: held-out R^2=0.0232 < 0.20 bar
[...]

PR #4167, opened 2026-09-20
Size the 1-D spline pilot at the roughest admissible truth's resolution rank (#3331)

### LINKED ISSUE #3331: ACC-1 regression since #3191: additive/GLM smooths lose edf and the 95% posterior-mean band under-covers (add4_n5000 0.857, pois_add2_n2000 0.86, binom_add4_n5000 0.90)

SauersML:
Fixes #3331

## What regressed
f2b04715f0 (#3191) changed the pilot that the adaptive-resolution loop starts every univariate spline at. This covers open `s(x)`, cyclic splines, the factor-smooth marginal, and the #1867 floor under a 1-D radial pilot. The pilot became

    nullspace + penalized_resolution_rank(n, 1, m) = nullspace + ⌈n^{1/(2m+1)}⌉

with m the **penalty** order. At n = 1000 and the default m = 2, that is 2 + ⌈1000^{1/5}⌉ = 6 cubic functions, or 5 identified columns after the sum-to-zero constraint. The pre-#3191 default was 12 functions (11 columns).

The growth loop rarely leaves that start. So the bench's standard-route `s()` terms fit on 5 columns, and the binomial and poisson p20 cells lost to pygam_gs on deviance and NLL (HN-24, issue comment 5749002579). R0 (#3149, 6126d8488e) restored the provisioned default only on routes the loop does not grow, so the standard route still starts at the pilot.

## Why that rate is the wrong one
At the REML-optimal λ, a penalized smoother keeps about n^{1/(2s+1)} directions, where s is the smoothness of the **truth**, not the order of the penalty. This holds for s ≤ m: a truth rougher than the penalty assumes is shrunk less, so more directions survive. In the Demmler–Reinsch basis, direction j is shrunk by 1/(1+(j/j*)^{2m}). A basis truncated at K ≈ j* therefore sets the fit by truncation, not through λ.

`⌈n^{1/(2m+1)}⌉` is the count for a truth exactly as smooth as the penalty. It is too small for anything rougher.

The truth's smoothness s is unknown before the fit. So the pilot is sized for the roughest truth the framework admits: the order-1 Sobolev space. Order 1 is `minimal_embedding_order(1)`, the least order whose penalty embeds a 1-D function space in the continuous functions. Every radial pilot (`starting_num_centers`) is already sized at that order.

    pilot = nullspace + penalized_resolution_rank(n, 1, minimal_embedding_order(1)) = nullspace + ⌈n^{1/3}⌉

| n | old pilot (cubic, m=2) | new pilot | provisioned (non-loop routes) |
|---|---|---|---|
| 100 | 5 | 7 | 12 |
| 1 000 | 6 | **12** | 12 |
| 10 000 | 9 | 24 | 12 |
| 100 000 | 12 | 49 | 12 |

- **No new constant.** The size is a function of n and the minimal embedding order only.
- **Matches the old size where it should.** At n = 1000 it reproduces the pre-#3191 size (11 columns).
- **Still grows with n,** as #3191 intended, unlike the old provisioned cap.
- **One rate for both pilots.** The 1-D spline pilot and the 1-D Duchon/radial pilot now coincide. Before, the #1867 floor was a separate, smaller rate.
- **Caps unchanged.** The distinct-value support caps and the rank bound (`adaptive_resolution_support`) still bound the start.

## Kept from #3191 / #3078
- **Growth is intact.** The pilot is still data-derived and uncapped, and the loop still grows it on evidence. `default_bspline_resolution_grows_with_n` still requires growth past 11 columns for sin(8πx) at n = 1000 and a truth RMSE that falls with n.
- **Null and linear truths.** These must still stay at the pilot and shrink to about 0 and 1 EDF. The pinned pilot dimension is now the derived 17 at n = 4000, replacing the old cap of 11. The #3191 pilot there was 7, so the old assertion of 11 did not match it either.
- **Unused argument dropped.** `pilot_spline_basis_dim` and `pilot_cyclic_basis_dim` lose their penalty-order argument, which is no longer used.
- **Sphere pilots unchanged.** On the 2-D sphere at the default order 2, the order-2 rank already equals the minimal embedding order.

This is complementary to #3351, which owns the growth rule. This PR only changes where the loop starts.

## Tests updated
- `pilot_internal_knots_follow_the_resolution_rate_not_a_fixed_cap`: expects 7/12/24/49.
- `default_bspline_is_provisioned_and_the_loop_starts_it_at_its_pilot_3149`: moved to n = 10 000, where provisioned (12) and pilot (24) still differ. At n = 1000 they now coincide.
- `radial_1d_default_not_starved_below_univariate_spline_resolution_1867`: uses the new floor.
- `factor_smooth_default_marginal_is_the_rank_bounded_group_pilot_3264`: in the support-bound case, the provisioned value is now min(10, pilot).
- `default_bspline_resolution_grows_with_n`: the null and linear arms pin the derived pilot (17).

## Bench (bench/pygam_compare, plan quick, n = 1000)
- **Before:** main run 35498800209 (d0588553). Coefficients were 6 at p1, 26 at p5 and 101 at p20. The binomial and poisson p20 cells lost.
- **After:** pending the pygam-compare run on this branch. I will add the numbers here and on #3331.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, e9bc0c9326cc.

PR #4168, opened 2026-09-20
Linear dictionary: the centered K=1 fit returns its mean from Rust, and transform reads it (#4064)

### LINKED ISSUE #3869: dictionary: linear_dictionary_transform always routes top-k ridge LS, so a softmax-fitted linear dictionary's transform() is not its fitted encoder

### LINKED ISSUE #4064: python wrapper: linear_dictionary_fit re-derives the centered lane's flag and training mean in Python instead of reading them from the Rust fit

SauersML:
Fixes #4064

This branch is built on #3872's branch, as #4064 asks. It carries #3872's commit `f86ae59a4d` (Fixes #3869: `transform` routes with the fitted assignment rule and temperature), plus one commit for #4064. Merging this PR lands both.

## Problem

With `center_rank_one=True` and `K=1`, the fit is the **affine** model `x ≈ μ + c·a`. `LinearDictionaryFit` had no field for `μ`. So the Python facade re-derived two things itself:
- the "this fit is centered" rule (`center_rank_one and K == 1`), and
- the training mean (`x.mean(axis=0)`).

That is two sources of truth for one model. The facade also re-implemented the transform's input contract in numpy. SPEC: "Python should be a thin wrapper over Rust."

## Math

For the affine model, the held-out code of a row `x` is the ridge least-squares coefficient against the *centered* row:

  c(x) = argmin_c ‖x − μ − c·a‖² + ρ c² = (x − μ)·a / (a·a + ρ)

Top-k routing and softmax routing both use `x − μ` in place of `x`, and `reconstruct` returns `c·a + μ`. On the training rows this reproduces the fit's own `assignments`, and `assignments·atoms + μ == fitted` holds exactly. Linear lanes have no origin (`μ = None`), and transform stays as it was for them.

## Change

- **`gam-terms` `dictionary.rs`**
  - `LinearDictionaryFit.mean: Option<Array1<f64>>`. It is `Some(column means)` for the centered K=1 lane: the exact vector that `centered_rank_one_components` subtracted and baked into `fitted`. It is `None` for every linear lane.
  - `linear_dictionary_transform(x, atoms, mean, top_k, assignment, temperature, code_ridge)` encodes `x − mean` itself and owns the input contract:
    - finite `x`, atoms and mean;
    - `top_k ∈ [1, K]`, with no silent clamp;
    - finite, positive `code_ridge`;
    - a valid softmax temperature;
    - `mean.len() == P`.
- **`gam-pyffi` `geometry_ffi.rs`**: the fit dict returns `"mean"`, and `linear_dictionary_transform_ffi` takes `mean=None`.
- **`gamfit/_linear_dictionary.py`**
  - Stores `mean` exactly as Rust returns it. `centered` is now the property `mean is not None`.
  - The `is_centered` rule, the `x.mean(axis=0)` mirror and the Python-side transform validation are deleted.
- **Stubs**: `_rust.pyi` and `_rust_module.pyi` gain the `mean` argument.

## Tests

- **Rust:** `dictionary::tests::centered_rank_one_fit_carries_its_mean_and_transform_reads_it_4064` checks:
  - `mean` equals the column means;
  - `assignments·atoms + mean == fitted`;
  - `transform(x_train) == assignments`;
  - linear fits carry `None`;
  - the refusals for out-of-range `top_k`, non-finite `x` and a mean of the wrong length.
- **Python:** `tests/test_linear_dictionary_fitted_mean_4064.py` runs the same checks through the facade. The refusals surface as `ValueError` raised from Rust.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 4be1b068d0bc.

PR #4174, opened 2026-09-20
event-history: derive the smoother's density cut from the interpolant's roundoff and certify the cut mass (#3998)

### LINKED ISSUE #3998: event-history marginal: DENSITY_NOISE_RELATIVE = 1e-11 floor ignores the interpolant's Lebesgue constant, and the cut mass is renormalised away unchecked
## What

`crates/gam-event-history/src/marginal.rs` cuts the backward smoother's densities at a fixed relative floor:

```rust
/// A filtered or predicted density value below this fraction of its grid's
/// peak is interpolation noise (the Lagrange-based operator is a signed sum,
/// accurate to about `ε · peak` in absolute terms), ...
/// `1e-11` sits several orders above the noise of a
/// product grid of a few hundred points; the mass it discards is far below
/// the Gauss-Hermite certificate's tolerance, and the smoothed marginal is
/// renormalised after the cut so every expectation is under a probability.
const DENSITY_NOISE_RELATIVE: f64 = 1e-11;

fn density_floor<S: JetField>(values: &[S]) -> f64 {
    DENSITY_NOISE_RELATIVE * values.iter().map(|v| v.value()).fold(0.0, f64::max)
}
```

The floor is used in two places:

- **`backward_smoother`'s `propagate` closure.** It forms `1 / p̂` only where `predicted > floor`.
- **`smoothed_marginal`.** It forms `exp(ln a + log β)` only where `a > floor`, then renormalises.

Both feed the gradients and Hessians (`subject_marginal(.., derivatives = true)`) and `latent_state_moments`.

## Why it is wrong

**1. The noise model in the comment is not the one the crate uses anywhere else.**

The operator is not accurate to `ε · peak`. `GaussHermite::lebesgue_constant` (chain.rs:63) is documented as "the factor by which nodal roundoff is amplified by the forward operator's interpolant". Two other places already charge `ε · Λ · nodes`:

- the fit's admissibility test, `certifiable` (family.rs:2151);
- the forecast's window roundoff (forecast.rs:700).

Lebesgue constants of the Gauss-Hermite Lagrange interpolant over the node hull (computed directly; order 41 matches the 1.154e13 recorded in objective.rs:570):

| order | Λ | ε·Λ |
|---|---|---|
| 9 | 1.99e1 | 4.4e-15 |
| 11 | 7.7e1 | 1.7e-14 |
| 17 | 8.1e3 | 1.8e-12 |
| 21 | 2.3e5 | 5.0e-11 |
| 33 | 8.3e9 | 1.8e-6 |

Under the crate's own `ε · Λ · nodes` accounting, admissible fits sit well above the 1e-11 floor:

- **Order 17 (the first raise from the default 9).** At the 449-node subject of objective.rs:570, the noise is 1.8e-12 × 449 ≈ 8e-10 of the peak. That is about 80× the floor.
- **Order 21.** This order is certifiable up to about 19 nodes. Its per-step noise of 5e-11 is already 5× the floor.

So points whose predicted density is pure interpolation noise pass the floor. `propagate` then divides by them, and `smoothed_marginal` multiplies them by `β`. The comment itself calls `β` "exponentially large exactly there".

**2. The "discarded mass is far below the certificate's tolerance" claim is never checked.**

Nothing measures what the cut throws away. The renormalisation that follows hides it: a smoothed marginal whose mass lived largely below the floor comes out as a proper probability with no signal.

That is the pattern "In general, do not paper over solver issues."

**3. At low order, the constant is arbitrary in the other direction.**

At order 9 on a short subject the noise is about 4e-15 × nodes. The 1e-11 cut is then three or more orders above it. It drops real mass for no stated reason.

## Fix direction

- **Derive the floor from the representation's roundoff.** Use `ε · Λ_G · (nodes + 1) · peak`, with `Λ_G = gh.lebesgue_constant` when the chain interpolates and 1 for a static chain that is conditioned in place. This is the same charge `certifiable` and `forecast::roundoff` read. `inputs.gh` is already in scope in `backward_smoother`.
- **Set the margin by the certificate tolerance.** Keep a point only where its ratio's relative error is at most `quadrature_tolerance`, that is `p̂ > noise / tolerance`. This needs the tolerance threaded into `SubjectInputs`, which is why this is not a one-line change.
- **Bound the cut mass, and refuse it rather than renormalise it away.** The cut mass is bounded by `Σ_{cut} w_i · floor · exp(log β_i)`, and `log β` is already computed. When that bound exceeds the tolerance, return `LOST_POSITIVITY`-style numerical failure so the driver raises the order or the node's own-grid placement (#3013) takes over.

Related:

- #3013 places slow-atom smoothed marginals on their own grids. It does not touch this floor.
- #3945 (the PIT slack) has the same absolute-slack-vs-derived-roundoff shape in forecast.rs.

SauersML:
Fixes #3998

## What was wrong

`marginal.rs` cut the backward smoother's densities at `DENSITY_NOISE_RELATIVE = 1e-11` of their peak, in two places:
- `propagate`, which formed `1/p̂` only above the floor;
- `smoothed_marginal`, which formed `exp(ln a + log β)` only above the floor and then renormalised.

The constant does not match the crate's own noise model. A Lagrange-based forward operator amplifies nodal roundoff by the interpolant's Lebesgue constant `Λ_G` (`GaussHermite::lebesgue_constant`). The fit's admissibility test (`certifiable`) and the forecast roundoff already charge `ε·Λ·nodes` for this.

| order | Λ | ε·Λ |
|---|---|---|
| 9 | 1.99e1 | 4.4e-15 |
| 17 | 8.1e3 | 1.8e-12 |
| 21 | 2.3e5 | 5.0e-11 |
| 33 | 8.3e9 | 1.8e-6 |

From order 17 up, noise points passed the floor. `propagate` divided by them, and `smoothed_marginal` multiplied them by `β`, which is exponentially large exactly there. At order 9 the floor dropped real mass. In both cases the cut mass was never measured, and the renormalisation hid it. SPEC: "In general, do not paper over solver issues." / "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code."

## The derivation

**Noise.** Take a predicted density `p̂` that reached its grid through `k` interpolating forward operators. Each operator adds at most `ε·Λ_G·max|source|`, and the filtered densities are normalised, so

  `noise = ε · Λ_G · k · max|p̂|`.

`k` counts operators with non-empty `per_axis`, cumulatively along the filtered chain. The backward smoother's own re-interpolation onto the smoothed grid adds 1. A density that no operator touched has `k = 0`: the exact prior at the first node, every node of a static chain, or a zero-innovation step conditioned in place. It is a product of exact positive factors, accurate to relative roundoff, so it has no absolute noise.

**Resolution.** A ratio `x/p̂` has relative error at most `noise/p̂`. It is therefore accurate to the quadrature certificate's tolerance `tol` exactly when

  `p̂ > noise / tol`.

Those are the points that are kept. `tol` is the fit's `EventHistorySpec::quadrature_tolerance`, the same number `certifiable` and the order/mesh certificates use, so no new knob is introduced.

**Cut-mass certificate.** At a cut point the representation is `≤ noise/tol` and the truth lies within `noise` of it. The true density is therefore at most `c = noise·(1 + 1/tol)`. The node factor `exp(ell − shift)` and `log β` are explicit formulas evaluated at every point, so the smoothed mass the cut discards is at most

  `B = c · Σ_cut w_i · exp(ell_i − shift + log β_i)`,

computed in log space with `log_sum_exp`. Renormalising over the kept mass `M` moves every expectation by at most `B/M`. When `B > tol·M`, the grid does not resolve the smoothed marginal. `smoothed_marginal` then returns `EventHistoryError::LostPositivity` carrying the `LOST_POSITIVITY` marker, and the fit driver's existing `positivity_raise` raises the Gauss-Hermite order. The renormalisation no longer hides the loss.

## Changes

- `DENSITY_NOISE_RELATIVE` and `density_floor` are deleted.
- New `PredictedNoise { absolute, tolerance }` with `resolved()` / `cut_ceiling()`, plus `interpolations(&filtered)` for `k`.
- `smoothed_marginal(grid, predicted, likelihood, log_beta, noise, label)` now forms `p̂·exp(ell − shift + log β)` over the resolved points and certifies the cut. It returns the mask it kept.
- `Smoothed.resolved` carries each smoothed grid's mask, and `propagate` forms `1/p̂` exactly where the smoothed marginal was formed and certified. On the filtered grid it uses that grid's own resolution predicate at the same tolerance; at the last node that predicate is the certified cut.
- `subject_marginal(inputs, Evaluation)`: `Evaluation::Value | Evaluation::Derivatives { tolerance }` replaces the bool, so the derivative pass always has a tolerance.
- `EventHistoryFamily::new(.., quadrature_tolerance)` is validated finite and positive. The fit passes `spec.quadrature_tolerance`; probes and profile families built from a fit pass `fit.family.quadrature_tolerance`; `latent_state` passes it to `latent_state_moments`.

## Test

`smoothed_marginal_cut_is_certified_against_the_quadrature_tolerance_3998` runs on an order-9 standard-normal grid:
- an exact prior (`k = 0`) is kept whole and normalised;
- when `k` is set so the threshold is 1e-3 of the peak, exactly the two outer points (`z = ±4.51`, prior ratio 3.8e-5) are cut:
  - under a flat `β`, the bound `B/M ≈ 1e-3 < tol` holds, so the cut is accepted and renormalised;
  - under `β = exp(4z)`, the smoothed marginal is `N(4,1)`, and its mass sits at the cut point `z = 4.51`. There `B/M ≫ tol`, so the cut is refused with `LOST_POSITIVITY`.

Scope note: the filtered-grid mask for the mid-chain Fisher sweep uses the resolution predicate. It is certified by the mass bound only at the last node, where it coincides with the smoothed cut.

---


SauersML:
Taking over from the wave8 claimant (no activity since 2026-09-20T10:38Z). I reviewed the diff and it is sound.

- The cut threshold p̂ > ε·Λ_G·k·max|p̂| / tol comes from the interpolant's own roundoff charge, the same one `certifiable` uses, and from the fit's `quadrature_tolerance`. No new knob is introduced.
- The cut mass is bounded by noise·(1+1/tol)·Σ_cut w·exp(ell−shift+logβ) and certified against tol·M, so the renormalisation can no longer hide lost mass.
- The branch merges cleanly with current main. Every `subject_marginal` / `latent_state_moments` / `EventHistoryFamily::new` call site in the merged tree already uses the new signatures.

I will merge once focused-proof run 35505684423 on 23ba7ef9 is green.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 8126cd6e6209.

PR #4176, opened 2026-09-20
gam-terms basis_adequacy: refer the estimated-scale F test to the unpenalized residual

### LINKED ISSUE #4169: gam-terms basis_adequacy: estimated-scale F reference uses the penalized fit's φ̂ and n−edf, not the unpenalized residual (conservative and anti-conservative)
## Where (origin/main 33e71f23)

- `crates/gam-terms/src/inference/basis_adequacy.rs:347-348`: `statistic = Σ projected² / input.dispersion`.
- `crates/gam-terms/src/inference/basis_adequacy.rs:356-374`: the estimated-scale branch.
  ```rust
  let independent_df = residual_df - reference_df;
  let independent_sum = residual_df - statistic;
  ...
  let f_statistic = (statistic / reference_df) / (independent_sum / independent_df);
  fisher_snedecor_sf(f_statistic, reference_df, independent_df)
  ```
- `crates/gam-models/src/fit_orchestration/drivers/basis_adequacy.rs:711-728`: `dispersion = fit.coefficient_covariance_scale()` and `residual_df = fit.wald_residual_degrees_of_freedom()`. These are the **penalized** fit's `φ̂ = RSS_pen/(n − edf)` and `ν = n − edf`.
- The module docs at `basis_adequacy.rs:60-72` call this reference the "added-variable `F`" and say it is "exact for an unpenalized Gaussian fit".

## Why it is wrong

The numerator is right. `Z̃ = Z − X G⁻ XᵀW_H Z` uses the **unpenalized** Gram `G`, so for a Gaussian `U = Z̃ᵀs = Z̃ᵀW y`. That is the exact added-variable score, and it does not depend on `β̂`.

The denominator is wrong. The branch reconstructs "the residual sum without the tested directions" as `(ν − T)·φ̂` on `ν − r` d.f. That identity (PR #3079) holds only when `ν·φ̂` is the residual sum of the **unpenalized** fit on `X`:
- `‖(I − P_X) y‖²_W` on `n − rank X` d.f.;
- the residual is W-orthogonal to `span X`;
- it contains the added-variable share exactly.

The fit's `φ̂` comes from the **penalized** residual `y − Xβ̂_λ` on `n − edf` d.f. Under `H₀` (truth in `span X`), that differs from the exact reference in three ways:

1. **Bias.** Shrinkage leaves part of the true signal in the residual, so `RSS_pen` is inflated. The F statistic then reads small, which makes the test **conservative**. It gets worse with a heavier λ or a curvier truth.
2. **Wrong law.** Even without bias, `E[RSS_pen] = φ·(n − 2 tr A + tr A²)`, which is not `φ·(n − edf)`, and `RSS_pen/φ` is not `χ²_{n−edf}`. So `φ̂` is biased low over the partly-shrunk directions, which makes the test **anti-conservative**.
3. **Dependence.** The penalized residual is not orthogonal to `span X`. So `(ν − T)φ̂` is neither independent of `U` nor `χ²_{ν−r}`.

The driver also scores on the report rows (`select_rows(..., &report_rows)`), while `φ̂` and `ν` are always the full fit's. When the report rows are a subset, the two ends of the ratio come from different samples.

Which way the error goes depends on λ and the truth, so the p-value is miscalibrated in both directions. "P-values must be calibrated and accurate. Conservative is a bug just like anti-conservative."

### Evidence (Monte Carlo)

Setup:
- Gaussian data with a cubic B-spline design of `k` columns, a 2nd-difference penalty at fixed λ, and the truth in `span X`.
- The enrichment is `r` high-frequency sin/cos columns.
- Both tests use the module's exact construction: the same `Z̃`, `U`, `V`, `Q`.
  - "current" is `φ̂ = RSS_pen/(n − edf)`, then the F above.
  - "exact" is the unpenalized residual, see the fix below.
- 20 000 replicates, so the MCSE at 0.05 is 0.0015.

| n | k | λ | r | truth | size@0.05 current | size@0.05 exact | size@0.01 current | size@0.01 exact |
|---|---|---|---|---|---|---|---|---|
| 40 | 12 | 1 | 4 | linear | 0.0522 | 0.0482 | 0.0115 | 0.0098 |
| 40 | 12 | 10 | 4 | linear | 0.0524 | 0.0498 | 0.0107 | 0.0106 |
| 40 | 15 | 3 | 6 | linear | **0.0548** | 0.0490 | **0.0123** | 0.0112 |
| 60 | 20 | 5 | 6 | linear | **0.0554** | 0.0505 | **0.0120** | 0.0103 |
| 40 | 12 | 1 | 4 | sin 2πx | 0.0516 | 0.0478 | 0.0110 | 0.0100 |
| 40 | 12 | 10 | 4 | sin 2πx | **0.0382** | 0.0496 | **0.0066** | 0.0091 |
| 100 | 20 | 10 | 8 | sin 2πx | 0.0514 | 0.0491 | 0.0110 | 0.0104 |

- When shrinkage biases the fit (λ = 10, sin), the current reference is **conservative** by 8 MCSE at 5%, and by 34% relative at 1%.
- When many columns are partly shrunk and the truth is smooth, it is **anti-conservative** by about 3.5 MCSE.
- The exact reference is within about 1.5 MCSE in every row.

PR #3079 validated only n = 200 and n = 2000 with a small edf and a weak penalty, where all three effects are small.

## Principled fix

Refer `Q` to the **unpenalized** residual on the same rows the score uses. This is the reference `random_effect_test` already uses (`random_effect_test.rs:63-77`). It is exact for a penalized Gaussian fit, because `B = W_F^{1/2} Z̃` is orthogonal to `A = W_F^{-1/2} W_H X`.

In the standardized metric, with `r_i = s_i / √W_F,i`:

```text
D′_X = Σ s²/W_F − cᵀ M⁻ c,   c = XᵀW_H W_F⁻¹ s,  M = XᵀW_H² W_F⁻¹ X
ν_X  = m⁺ − rank(M)          (m⁺ = rows with W_F > 0)
```

When `W_H == W_F` this reduces to `Σ s²/W − (Xᵀs)ᵀ G⁻ (Xᵀs)`, which reuses `design_gram`.

Then, per term:

```text
D″ = D′_X − Q,   ν″ = ν_X − r,   F = (Q/r) / (D″/ν″)  on (r, ν″)
```

- Here `Q = UᵀV⁻U` is the unscaled quadratic form.
- The dispersion then cancels, so the Estimated branch no longer needs `φ̂`.
- When `ν″ ≤ 0` or `D″ ≤ 0`, it returns no p-value, as today.

Changes:
- **`BasisAdequacyInput`**: replace `residual_df: Option<f64>` with an unpenalized residual `{ sum, df }` that the driver computes once per report from the same rows, weights and design.
- **`dispersion`**: keep it only for the `Known` branch.
- **Module docs**: correct them.
- **Tests**:
  - Change `estimated_scale_matches_the_exact_added_variable_f` (about line 1786) to feed the unpenalized residual.
  - Add a penalized-fit null fixture, such as the λ = 10 sin row above, that fails on main.

Related: #3832 is the same defect class in `smooth_score_test` (penalized `φ̂` used where the unpenalized residual is exact). #3691 and PR #3694 touch the neighbouring `dispersion` fallback, which is a different defect.

## Duplicates checked

- PR #3079 introduced the current formula. This issue shows its "exact" claim fails for penalized fits.
- #2788 and #2789 concern the enrichment rank and d.f., not the scale reference.
- #3691 concerns the `unwrap_or(1.0)` fallback.
- No open issue covers the estimated-scale reference.

---

SauersML:
Fixes #4169

## What

The estimated-scale branch of `basis_adequacy_score_test` divided the added-variable score by the penalized fit's `φ̂ = RSS_λ/(n − edf)` on `ν = n − edf`. That residual is biased by the shrinkage, is not `χ²_{n−edf}`, and is not orthogonal to the score. It also belongs to all `n` rows, while the report may score only a subset. The size was off in both directions (see the issue's Monte Carlo table).

Now the reference is the **unpenalized** residual on the test's own rows:
- `D′ = Σ s²/W_F − cᵀM⁻c`, with `c = XᵀW_H W_F⁻¹ s` and `M = XᵀW_H²W_F⁻¹X`.
- `ν′ = m⁺ − rank(M)`.
- `F = (Q/r)/((D′ − Q)/(ν′ − r))`.

This is exact for a Gaussian response whatever the penalty, because `B = W_F^{1/2}Z̃` is orthogonal to `A = W_F^{-1/2}W_H X`. It is the same reference `random_effect_test` uses.

## Changes

- `gam-terms/src/inference/basis_adequacy.rs`:
  - New `BasisAdequacyScale { Known { dispersion }, Estimated(UnpenalizedResidual) }`. It replaces the `dispersion` / `residual_df` / `SmoothTestScale` triple on `BasisAdequacyInput`.
  - New `unpenalized_residual(...)`. For a canonical link (`W_H == W_F`) it reuses the caller's `design_gram`, so no second factorization is needed.
  - `DesignGramFactor::rank()`.
  - Corrected module docs.
- `gam-models/.../drivers/basis_adequacy.rs`:
  - The Product reference computes the residual once per report, from the same gathered rows and weights the score uses.
  - The fit's `coefficient_covariance_scale()` is now read only for a known scale.
  - No `φ̂` or `n − edf` reaches the estimated branch.

## Tests

- `estimated_scale_p_value_is_the_exact_added_variable_f_test_at_a_penalized_fit`: at a ridge-penalized fit, `D′` equals the unpenalized least-squares RSS, and the p-value equals the classical added-variable `F` p-value. On main it would have used the penalized residual.
- `estimated_scale_is_uniform_under_the_null_at_a_heavily_penalized_fit`: 2000 null replicates at a heavy ridge, uniform p-values.
- `unpenalized_residual_projects_the_standardized_score_in_the_curvature_metric`: `W_H ≠ W_F`, checked against a direct least-squares fit in the standardized metric. Also refusal cases.
- `estimated_scale_refuses_when_the_residual_cannot_hold_the_rank`, `degenerate_inputs_refuse`.

Focused proofs: `HN-20-ba-1` (gam-terms, `basis_adequacy`) and `HN-20-ba-2` (gam-models, `basis_adequacy`).

---


SauersML:
**VF-0 review: not merging yet, because of a test tolerance.**

The maths checks out:
- `A′B = XᵀW_H Z̃ = 0`
- `D′ = eᵀ(I−P_A)e`
- `D′ − Q ~ φχ²_{ν′−r}`, independent of `Q`

So referring the F to `D′` on `ν′ = m⁺ − rank(M)` is the right fix, and it fixes both directions.

The blocker is that the exact-F pins in `estimated_scale_p_value_is_the_exact_added_variable_f_test_at_a_penalized_fit` go from `1e-9` to `1e-8` on both the statistic and the p-value, and nothing derives the new band. A test tolerance can't be loosened without a derivation.

Please do one of these:
- keep `1e-9`, since the new residual check at L475 already holds at `1e-9 · rss_null`;
- or derive the band, for example as a conditioning bound on the cancellation in `Xβ̂` at the ridge the harness uses, and state it in the test.

Also, two test fns are renamed with no entry in `docs/source-removal-changes.json`, so the source-removal guard will flag them. Either add the entries or keep the old names.

---

SauersML:

The maths is right:
- `AᵀB = XᵀW_H Z̃ = 0`
- `D′ = eᵀ(I − P_A)e`
- `D′ − Q ~ φχ²_{ν′−r}`, independent of `Q`

Referring the F to the unpenalized residual is therefore the correct fix for #4169. Two concrete defects block the merge.

1. **Tolerance loosened with no derivation.** In `estimated_scale_p_value_is_the_exact_added_variable_f_test_at_a_penalized_fit`, both exact-F pins go from `1e-9` to `1e-8`: `out.statistic` against `r * exact_f`, and `out.p_value` against `exact`. Nothing derives the wider band. The test's own new check, `(residual.sum - rss_null).abs() <= 1e-9 * rss_null`, holds at `1e-9`. SPEC.md: "A failing test should always indicate problematic behavior." A widened band that nobody derived can hide exactly that behavior.
   - Fix: restore `1e-9`.
   - Or: state the band as a derived bound in the test. For example, a conditioning bound on the cancellation in `D′ − Q` at the harness's ridge `λ = 25`.
2. **Renamed tests missing from the source-removal ledger.** Two existing tests are renamed:
   - `estimated_scale_p_value_is_the_exact_added_variable_f_test`
   - `estimated_scale_refuses_when_the_residual_df_cannot_hold_the_rank`

   Neither has an entry in `docs/source-removal-changes.json`, so the source-removal guard will fail. Either add a ledger entry keyed on the branch base, or keep the old names.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 281d6398214b.

Issue #707, opened 2026-06-04
quality-suite METRIC_OFF: gam_gp_smooth_recovers_truth_and_predicts — held-out R²=0.77 < 0.95 vs GpGp exact GP


SauersML:
## Summary

The reference-quality test `gam_gp_smooth_recovers_truth_and_predicts` in `tests/quality_vs_r_gpgp_likelihood.rs` fails its held-out predictive-R² gate: gam's ν=0.5 Matérn GP smooth explains only ~77% of held-out variance where the principled bar is 95%. The baseline is R `GpGp` run at full neighbour count (exact Matérn GP). The shortfall points to a **GP/Matérn likelihood + hyperparameter (length-scale / smoothing-parameter) recovery gap** in gam relative to the exact-GP engine.

## Failing assertion & observed values

`tests/quality_vs_r_gpgp_likelihood.rs:316`:

```rust
assert!(
    gam_r2 >= 0.95,
    "gam held-out R² too low: {gam_r2:.4} (bar 0.95)"
);
```

Observed:

```
gam held-out R² too low: 0.7720 (bar 0.95)
```

The neighbouring RMSE gate at line 312 (`gam_pred_rmse <= 0.45`, σ_noise = 0.20) should be **checked too**: with held-out R² = 0.7720 on a signal whose variance is `O(1)` and noise floor σ²=0.04, the implied held-out RMSE is well above the 0.45 bar (R² = 1 − RMSE²/Var(y_te); for the test y-variance this corresponds to RMSE ≈ 0.6–0.7), so the RMSE assertion at 312 very likely also fails and would be the first to trip. The truth-recovery gate (line 302, bar 0.40) and the match-or-beat gate (line 325, ×1.10 vs GpGp) are the other two assertions and may compound.

## Mathematical background

**Generating process (known truth).**

```
f(x) = 5 + 3·exp(−x/2)·cos(x)        (latent mean)
y(x) = f(x) + 0.2·N(0,1)             (σ_noise = 0.2)
```

A decaying oscillation on `x ∈ [0,10]`; peak-to-trough amplitude ≈ 4.4.

**Matérn ν=0.5 (exponential) kernel** — the kernel gam is asked to fit and the one GpGp's `exponential_isotropic` covfun implements exactly:

```
k(r) = σ² exp(−r/ℓ)        r = |x − x′|
```

[...]

PR #4178, opened 2026-09-20
gam-terms Matérn κ-derivative: select non-double-penalty blocks by source, drop the ν=1/2 1e-12 r-floor

### LINKED ISSUE #4171: gam-terms Matérn κ-derivative: non-double-penalty arm emits dS1/dS2 for ν=1/2 that the forward build never has (positional penalty indices misattribute them), built with a 1e-12 r-floor
## Where (origin/main 920eb8df60)

1. `crates/gam-terms/src/basis/sphere_basis.rs:3426-3435`: `build_matern_basis_log_kappa_derivativeswithworkspace`, the Euclidean non-double-penalty arm:
```rust
    } else {
        build_matern_operator_penalty_psi_derivatives(
            centers.view(),
            length_scale,
            spec.nu,
            spec.include_intercept,
            z_opt.as_ref(),
            aniso,
        )?
    };
```
2. `crates/gam-terms/src/basis/sphere_basis.rs:1367-1376`: `matern_operator_psi_triplet`, the ν=1/2 ratio:
```rust
    // nu=1/2 has singular phi'(r)/r ~ -kappa/r as r->0.
    // We use the same finite r-floor regularization as operator assembly.
    let (ratio, ratio_psi, ratio_psi_psi) = if matches!(nu, MaternNu::Half) {
        let r_eff = r.max(1e-12);
```

## What is wrong

**The derivative list is not aligned with the forward penalty list.** `build_matern_operator_penalty_psi_derivatives` always returns the canonical `[dS0, dS1, dS2]` blocks, plus `dS3` when ν ≥ 5/2. The forward build (`build_matern_operator_penalty_candidates` -> `operator_penalty_candidates_from_collocation` with `DuchonOperatorPenaltySpec::matern_for_smoothness`) keeps only the mass block for ν = 1/2 (#707).

For `matern(x, nu=1/2, double_penalty=false)` in 1D (d ≥ 2 is rejected at term_builder.rs:3634), the bundle therefore carries 3 derivative blocks against 1 forward penalty. The consumer `try_build_spatial_term_log_kappa_derivativeinfo` (design_construction.rs:4106) numbers them positionally:
```rust
    let penalty_indices = (0..s_psi_components_local.len())
        .map(|j| penalty_start + j)
```
As a result, dS1 and dS2 are attached to penalty indices `penalty_start+1` and `penalty_start+2`. Those indices belong to the next term's smoothing parameters, or lie past the end of the penalty list. The κ-gradient is then computed with derivative blocks of penalties the model does not have, weighted by other terms' λ.

Two other paths already get this right:
- The double-penalty arm right above selects by source through `active_matern_double_penalty_derivatives(&base.active_penalties, ..)`.
- The spherical Matérn path (sphere_basis.rs:1937-1946, gam#902) selects through `active_operator_penalty_derivatives(&filtered.active, ..)`.

Only the Euclidean non-double-penalty arm skips the selection. Its FD test (`tests/basis_smooth/smooths/basis_matern_log_kappa_penalty_derivative_fd.rs`) asserts count parity, but only for ν ≥ 3/2, so the gap has no test coverage.

**The misattributed stiffness block is built with a magic floor.** The stray dS2 comes from `ratio = -(s / r.max(1e-12))·e^{-a}`. At every self-pair (r = 0) this gives a D2 diagonal of `w·(-κ·1e12)`, so the value that ends up misattributed depends on the arbitrary `1e-12`. The comment claims this is "the same finite r-floor regularization as operator assembly", which is false:
- The forward builder `build_matern_collocation_operator_matrices` (matern_kernel.rs:4013-4026) uses the exact convention `phi_r_over_r = 0` at r = 0 for ν = 1/2 in 1D, and bails for d > 1.
- `matern_aniso_extended_radial_scalars` (matern_kernel.rs:940-962) documents and uses the same zero convention.

The derivative path is therefore not the derivative of the value path.

## Principled fix

1. **Select the derivative blocks by source.** In the non-double arm, route the blocks through `active_operator_penalty_derivatives(&base.active_penalties, ..)`, exactly as the spherical path and the double-penalty arm do. The derivative list then equals the forward active list by construction.
2. **Delete the `1e-12` floor.** Use the value path's exact r = 0 convention for ν = 1/2: in 1D the ratio and its ψ-derivatives are 0 at r = 0, and for d > 1 refuse with the forward builder's error.
3. **Add a ν = 1/2, d = 1 regression.** It should check that the block count equals the active-penalty count and that the mass block matches a central difference of the forward penalty.

SauersML:
Fixes #4171

## Changes (`crates/gam-terms/src/basis/sphere_basis.rs`)

- **Blocks are now selected by source.** The non-double-penalty arm of `build_matern_basis_log_kappa_derivativeswithworkspace` routes the canonical `[dS0, dS1, dS2(, dS3)]` blocks through `active_operator_penalty_derivatives(&base.active_penalties, ..)`. The spherical path and the double-penalty arm already did this.
  - For ν = 1/2 the forward build keeps only the mass operator (#707).
  - Before this change, stray dS1/dS2 blocks were numbered positionally (`penalty_start + j`) and attached to other terms' penalties.
- **The `r.max(1e-12)` floor is gone.** It sat in the ν = 1/2 φ'/r ψ-triplet of `matern_operator_psi_triplet`, and the fix uses the value path's exact convention instead:
  - At a 1-D center collision, φ'/r = 0, and so are its ψ-derivatives.
  - For d > 1 the triplet refuses, as `build_matern_collocation_operator_matrices` does.
  - The comment that claimed the floor matched operator assembly was false, and is removed.
- **New module `matern_rough_kernel_log_kappa_alignment_tests`** covers ν = 1/2, d = 1, `double_penalty: false`:
  - The first- and second-derivative block counts equal the forward active-penalty count (1).
  - The mass block matches central differences of the forward penalty in ψ.
  - The collision ratio triplet is (0, 0, 0) in 1-D, and the 2-D collision is refused.

## Validation

- `focused-rust-proof.yml` on gam-terms lib, filters `matern_rough_kernel_log_kappa_alignment_tests third_order_penalty_psi_derivatives_match_the_forward_penalty` (HN-37-5-1).
- `focused-rust-proof.yml` on the gam `basis_smooth` integration binary, which covers the existing ν ≥ 3/2 Matérn log-κ FD gates (HN-37-5-2).

---


SauersML:

---

PR #4180, opened 2026-09-20
Refine the adaptive open B-spline along data-bearing intervals (#3993)

### LINKED ISSUE #3993: fit_orchestration: default s(x) cannot resolve a covariate range stretched by one outlier — uniform nested knot refinement spends its whole support budget on data-free intervals (robustness.md F2b, untested)
Audit finding **F2b** (`bench/pygam_audit/robustness.md:107-119`) is still open on `origin/main`, and nothing tests it.

## Reproducer

```python
import numpy as np, gamfit
n = 300
x = rng.uniform(0, 1, n); x[0] = 1e6
y = np.sin(6 * x) + rng.normal(scale=0.2, size=n)
m = gamfit.fit({"x": x, "y": y}, "y ~ s(x)")
grid = np.linspace(0.05, 0.95, 50)
f = m.predict({"x": grid})  # expected: close to sin(6*grid)
```

The audit reported three outcomes:
- Default `s(x)`: a cubature error.
- `s(x, bs=tps)`: silently flat, with edf 1.0.
- `s(x, knot_placement=quantile)`: correct, with edf 7.3.

pyGAM is equally flat on this data, so this is parity, not a regression. It is still a silently wrong or refused fit, on data with a single high-leverage covariate value.

## Why the current default cannot resolve it

1. **The placement is uniform on the observed-extrema interval.**
   - `crates/gam-terms/src/term_builder.rs:5358` (`resolve_nonperiodic_bspline_knotspec`):
     ```rust
     let placement = explicit_knot_placement(options)?.unwrap_or(BSplineKnotPlacement::Uniform);
     ```
   - The domain is `[0, 1e6]`, so every pilot knot lands in `[~1e5, ~9e5]`, where there are no data. The whole bulk (n−1 points) sits inside the first cubic piece.

2. **The adaptive-resolution loop only refines uniformly.**
   - `crates/gam-terms/src/basis/types.rs:625-637`, `refined_internal_knots(k) = 2k + 1`:
     > "One level of uniform nested refinement of a knot grid: every one of the `internal_knots + 1` intervals is split at its midpoint"
   - Putting a knot inside `[0, 1]` needs an interval of width about 1. That takes about log2(1e6) ≈ 20 levels, or roughly 2^20 internal knots.
   - The loop caps growth at covariate support and design rank. `refinement_spanning` is at `crates/gam-models/src/fit_orchestration/entry.rs:2129`, and `fit_refinements_to_rank` requires `p < n`. That ceiling is about n − 4 ≈ 300 coefficients, far short of the knot count needed.
   - Every refinement the loop can afford therefore adds coefficients that are supported only by knot intervals with no data. They are penalty-only directions, so the bulk never gets a second polynomial piece.

3. **The lack-of-fit screen and the remedy disagree about where the data are.**
   - The screen's reference alternative is data-adaptive: a Duchon kernel at **EqualMass** centers (`crates/gam-models/src/fit_orchestration/drivers/basis_adequacy.rs:371-400`, `radial_enrichment`). It sees the sin(6x) structure in the bulk and correctly flags the term as under-resolved.
   - The only remedy the loop can apply is the uniform refinement above, which is structurally unable to represent that structure. So the screen fires, every step runs into the support or rank ceiling with no evidence gain, and the term ends saturated and under-resolved. `fit_refinements_to_rank` documents that end as a typed error. It is not a data problem.

## Why a quantile default is not the fix

PR #3078 evaluated quantile knots as the default open B-spline placement and rejected it with evidence:
- It lost about a third in accuracy on the Beta(½,½) `sin 4πx` design: 0.0243 uniform vs 0.0323 quantile.
- It turned five outer-REML certification tests red ("declined a certified optimum"): convex/concave shape-constrained, stiefel(k=1), `s(x) + s(g, x, bs=sz)` predict, and binomial point-invariance.

`CHANGELOG.md:67` records the decision: "Knot placement stays uniform by default (`knot_placement=quantile` opts in)." So the fix has to live in the refinement, not in the pilot placement.

## Proposed principled fix: nested refinement of data-bearing intervals

Change one refinement level from "split every interval at its midpoint" to "split, at its midpoint, every knot interval that contains at least one distinct covariate value".
- **It stays nested.** The old spline space remains a subspace of the new one, so the REML evidence comparison in the adequacy loop stays valid. That property is what the doc comment on `refined_internal_knots` relies on.
- **It adds no knob.** Whether an interval holds data is a fact about the sample, not a threshold.
- **It is identical to today's refinement whenever every interval holds data,** which is the case for all of #3078's designs. So it cannot reproduce #3078's accuracy or certification regressions.
- **On the reproducer, each level adds about one knot** (only the interval containing the bulk is split). The bulk gets its first interior knot after about 20 cheap levels, at a coefficient cost of about 20 rather than 2^20.
- **The support ceiling becomes the right quantity.** Refinement stops when no data-bearing interval can be split without violating the covariate-support condition that `refinement_spanning` already enforces.

Because the refinement is no longer a pure function of the knot count, `AdaptiveResolution` for a B-spline has to carry the realized knot vector (or the data-bearing split mask), not only `num_internal_knots`. Prediction already replays the frozen knot vector, so nothing downstream changes.

The `s(x, bs=tps)` symptom (edf 1.0, flat) has a different root. 1-D thin-plate centers come from farthest-point sampling on the data (`crates/gam-terms/src/basis/types.rs:708`), so centers do reach the bulk. The flatness is therefore more likely a conditioning loss in a kernel evaluated on a `1e6`-wide range. It needs its own reproduction on current `main` before it is filed separately.

## Requested regression test

Add a Python test built from the reproducer above:
- Assert that the default `y ~ s(x)` fits (no error).
- Assert that its RMSE against `sin(6x)` on `[0.05, 0.95]` is within a small factor of the `knot_placement=quantile` fit's RMSE.
- Derive that bound from the noise level and edf, not from a tuned threshold.

SauersML:
Fixes #3993

## Defect
The adaptive resolution loop grew the default `s(x)` by uniform nested refinement: level `K -> 2K + 1` splits every knot interval at its midpoint. When one outlier stretches the covariate range (`x ~ U(0,1)` plus `x = 1e6`), every pilot knot is in the data-free span `(1, 1e6)`. The bulk `[0, 1]` needs about `log2(1e6) = 20` levels before it gains a single knot, so each level's budget goes to intervals that hold no data. At the support bound the loop then "certifies" a basis that never resolved the data.

## Construction: `BSplineKnotPlacement::UniformRefined { root }`
- **Root.** For `K <= root` the internal knots are the uniform grid of `K` knots on the finite data range, which is exactly the old behaviour. `root` is the loop's own `starting_resolution`, so no new knob is added.
- **Level.** Take the intervals `I_j = [t_j, t_{j+1})`; the last one is closed so it holds the range maximum, matching basis evaluation. Every interval holding at least two distinct values gets one knot at its *coarsest separating dyadic point*. Bisect `I_j`, move into the half that holds all of its values whenever the other half is empty, and stop at the first midpoint that has values on both sides.
- **Nesting.** A partial level (target `K` between two full levels) splits the intervals holding the most distinct values first, with ties broken left to right. So the internal knots at `K` are a subset of those at `K + 1`, and the loop's evidence comparison still compares nested spaces.
- **Agreement with the uniform refinement.** When every interval has data on both sides of its midpoint, the first bisection separates, and each level is exactly the uniform `2K + 1` grid. On covered designs (#3078's) nothing changes.
- **Reach.** Every interval holding at least 2 distinct values can be split. By pigeonhole the chain keeps growing until each interval holds at most one distinct value, which reaches the support bound `K = u - degree - 1`. If a target exceeds what the distinct values resolve, it is refused with a basis error instead of being padded.
- **On the outlier design.** Level 1 puts its new knot inside `[0, 1]`, level 2 adds two there, and level 3 adds four. The bulk is resolved within a few levels instead of about 20.

## Plumbing
- `apply_adaptive_resolution(basis, start, target)` writes `placement = UniformRefined { root: start }` for adaptive uniform B-splines. The materializer always applies the start, even when the count already equals the target, so the spec records the root.
- `automatic_knot_vector` is now the single dispatch for the three `Automatic` knot sites in `bspline_build.rs`.
- A frozen spec (`smooth_overrides`, `term_builder`) replays the data-driven placement as `Automatic { adaptive: false }`, the same way it replays `Quantile`.

## Tests (`gam-models`, `data_bearing_knot_refinement_3993_tests`)
- `chain_is_the_uniform_refinement_on_covered_data_3993`: on a dense grid the chain equals the uniform grid at `K = 7, 15, 31, 63`, within a few ulps.
- `chain_is_nested_in_the_knot_count_3993`: on the outlier design each `K in 8..=60` keeps every knot of `K - 1`.
- `chain_spends_its_refinement_on_the_bulk_3993`: from root 7 to 15, all 8 new knots fall in `[0, 1]`, and the chain reaches the support bound.
- `default_smooth_resolves_the_bulk_behind_an_outlier_3993`: the issue's reproducer, `y = sin(6x) + 0.2 eps` with `n = 300`. Default `s(x)`'s RMSE on the bulk must be at most the quantile fit's RMSE plus `sigma (sqrt(edf_d/n) + sqrt(edf_q/n))`. That bound is the triangle inequality on the two fits' variance scales. A flat basis misses by about 0.6.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 9d09b7318402.

PR #4182, opened 2026-09-20
PIRLS: iterate on the model's own curvature; delete the Observed->Fisher ladder (#3962)

### LINKED ISSUE #3962: pirls/reweight: inner P-IRLS switches Observed→Fisher reactively on assembly errors, gain rejections and candidate errors, and latches Fisher after an arbitrary 3 failures
**Where:** `crates/gam-solve/src/pirls/reweight.rs` on origin/main at e0cc6a6a, in the inner P-IRLS loop. There are three reactive Observed→Fisher switches plus a latch:

1. **Iter-start, reweight.rs:1072.** Any error from the observed assembly is caught, and the iterate is rebuilt with Fisher:
   ```rust
   Err(_) if preferred_curvature == HessianCurvatureKind::Observed => {
       used_fisher_fallback_this_iter = true;
       consecutive_fisher_fallbacks += 1;
       if consecutive_fisher_fallbacks > 2 && !force_fisher_for_rest { ... force_fisher_for_rest = true; }
       model.update_with_curvature(&beta, HessianCurvatureKind::Fisher)?
   }
   ```
   Only `InverseLinkDomainViolation` and `PirlsRowGeometryUnrepresentable` are exempt.

2. **Gain rejection, reweight.rs:~2119-2145 (`reason=gain_rejection`).** An LM trial rejected on its gain ratio does more than raise the damping, which is the LM controller's job. It also switches the curvature to Fisher for the rest of the iteration.

3. **Candidate error, reweight.rs:~2265-2300 (`reason=candidate_err`).** A candidate evaluation that returned `Err` switches to Fisher too.

4. **The latch.** `consecutive_fisher_fallbacks > 2` latches `force_fisher_for_rest = true`. From then on every iteration of the solve runs Fisher scoring, and an AA(1) accelerator (`AndersonOneState`, :625) turns on that exists only for this latched regime.

5. **The attempt budget, loop_driver.rs:1554.** `let base_max_step_halving = if firth_active { 60 } else { 30 };` is the per-iteration LM attempt cap (`lm_max_attempts`, reweight.rs:978).
   - Firth gets twice the budget by fiat.
   - The LM loop already has a derived stopping rule, the damping window `MADSEN_DAMPING_FLOOR ..= MADSEN_DAMPING_CAP = 1/u` (loop_guard.rs:97). The Madsen escalator crosses that window in about 15 rejections (loop_guard.rs:220).
   - What an arbitrary count adds on top of the window is mostly headroom for the `continue` paths of the ladder above.

**Why it is wrong**

- SPEC: "In general, do not paper over solver issues." The template's reading of it: "A reactive retry, fallback, jitter or ridge bump that exists because a solve failed is a violation."
  - Each switch fires *because* a solve step failed: an assembly `Err`, a rejected trial or an erroring trial.
  - Which curvature the iteration uses is therefore a function of the failure history, not of the iterate. Two runs that reach the same β by different paths use different Newton systems from then on.
  - The `> 2` latch threshold is arbitrary. Nothing about the problem determines it.
  - So is the 30/60 attempt budget.
- The ladder duplicates machinery the loop already has:
  - LM damping (`madsen_escalator`) already answers a bad gain ratio.
  - `descent_curvature` (newton_solve.rs:528) already gives a principled descent direction on an indefinite Hessian: its Gill–Murray modification with the `√ε‖H‖` floor.
  - The #3318 negative-curvature escape already handles saddles.
  - Switching the curvature on top of these is a second, heuristic control loop.
- `Err(_)` is a wildcard. Beyond an indefinite observed Hessian (`HessianNotPositiveDefinite`), it also swallows `InvalidInput` errors from the observed assembly. For example:
  - "assembled Hessian contains non-finite entries"
  - Firth "coefficient Hessian is non-finite"

  Those are defects, and Fisher then silently steps around them.

The exported Laplace curvature is re-evaluated as Observed at β̂ (reweight.rs:~2957), so this is not a mislabelled-export bug. It is a reactive control ladder on the iteration path.

**Principled fix**

Make the Newton system a deterministic function of the iterate:
- Always assemble the preferred (observed) curvature.
- When it is materially indefinite (the existing `certify_positive_semidefinite_hessian` certificate), take the direction from `descent_curvature`'s Gill–Murray modification of that same matrix, not from a different likelihood surface.
- Let LM damping alone respond to gain rejections and candidate errors.

Then delete:
- `force_fisher_for_rest`, `consecutive_fisher_fallbacks` and `used_fisher_fallback_this_iter`
- the three switch sites
- the AA(1) accelerator that exists only for the latched Fisher regime
- the Firth/non-Firth attempt budget, leaving the damping window as the LM loop's stopping rule

Propagate every other observed-assembly error instead of matching `Err(_)`.

A regression test should show:
- A non-canonical fit (probit/cloglog) whose observed Hessian is indefinite at early iterates converges to the same β̂ as before.
- The curvature kind used at each iteration is `Observed`.

SauersML:
Fixes #3962

## Problem

Before this PR the inner PIRLS Newton system depended on the solve history, not only on β:

- A failed observed-information assembly at iteration start swapped H for Fisher.
- So did a gain-ratio rejection or a candidate-evaluation error inside the LM loop (two mid-iteration switch sites).
- After three consecutive fallbacks, `force_fisher_for_rest` latched Fisher for the rest of the solve. It also switched on an AA(1) accelerator to make up for the lost second-order convergence.
- `survival/base.rs::observed_information_band` substituted Fisher whenever the observed matrix was "roundoff-level" indefinite.
- `GamWorkingModel` refused any observed assembly that failed a PSD certificate, and that refusal fed the ladder.

As a result, two fits that reach the same β can iterate on different matrices. That is the SPEC violation "In general, do not paper over solver issues." Each swap exists because a solve failed.

## Math

PIRLS minimises F(β) = ½(D(β) + βᵀS_λβ), with gradient g = ∇F and Hessian H = ∇²F. For a non-canonical link, H is the observed information, and away from the mode it can be indefinite. **That is a property of the objective, not a failure of the solver.** A descent method for a non-convex objective needs three things, and all three already exist in the solver:

1. **A descent direction when H is not PD.**
   - Dense path: `descent_curvature` takes H = QΛQᵀ and returns H̃ = Q·max(Λ, √ε‖H‖₂)·Qᵀ, the Gill–Murray modified Newton matrix. When H is PD it returns H exactly (a strict Cholesky decides), so every convex model is bit-for-bit unchanged. The floor √ε‖H‖₂ is the smallest curvature the objective's values can resolve: a quadratic term ½κδ² is visible above the rounding of F only when κ ≳ √ε·scale. It is not a tuning knob.
   - Sparse path: when H + λD² does not factor, the Levenberg–Marquardt damping λ is raised (the Madsen–Nielsen escalator) until it does. The spectrum of H + λD² moves right by at least λ·min D², so this terminates. The escalator changes λ and never changes H.
   - `compute_lm_d2` raises negative diagonals, so D² is a valid scaling.
2. **An honest acceptance test.**
   - The gain ratio ρ = (F(β) − F(β+δ)) / (−gᵀδ − ½δᵀHδ) is measured against the quadratic model of the **bare** H.
   - A rejection raises λ. It no longer swaps H.
3. **A convergence certificate that refuses a saddle.**
   - `exact_newton_decrement_sq` returns `None` when gᵀH⁻¹g < 0, so an indefinite point is never certified converged.
   - At β̂ the export re-certifies the inertia:
     - Dense: `symmetric_extremes` gives λ_min, which is compared with pd_tol = γ_{n+p²}·max|λ|. γ_k is Higham's accumulation bound for the n per-row terms and p² penalty products that build H.
     - Sparse: the sparse Cholesky is the certificate. On refusal, the Gershgorin disc bound is the witness.
   - A saddle is exported as `InvalidObservedCurvature { min_eigenvalue, pd_tolerance, gradient_norm }`. It is never relabelled Fisher. If the eigensolver does not return, the export is NaN, because unknown inertia is not a certificate.

With these three pieces, the Fisher substitution has no remaining job. Its only effect was to make the iteration matrix depend on history.

## Change

- `pirls/reweight.rs`
  - Every iterate assembles `model.update_with_curvature(β, kind)`, where kind is `Observed` if `supports_observed_information_curvature()` and `Fisher` otherwise. Errors propagate.
  - Deleted:
    - the fallback counters and the `force_fisher_for_rest` latch;
    - both mid-iteration Fisher switch sites (gain rejection and candidate error);
    - the AA(1) accelerator and its state;
    - the `madsen_escalator.restart()` calls, which existed only to restart damping after a curvature swap.
  - The post-convergence export was rewritten as the inertia certificate described in step 3 of the math section.
- `pirls/gam_working_model.rs`
  - The Fisher assembly keeps its PSD certificate: X'WX with W ≥ 0 must be PSD, so a failure there is a real bug.
  - The observed assembly refuses only non-finite entries. An indefinite observed H is legitimate input to the direction solve.
- `survival/base.rs`
  - `update_with_curvature(Observed)` no longer refuses a delayed-entry observed Hessian that is indefinite beyond its rounding band. That refusal was the input to the ladder. It now returns the exact matrix, and the loop steps on its descent curvature.
  - As before, a `Fisher` request returns the same exact Hessian under the Fisher label. This model has no separate expected-information surface.
  - Removed `observed_information_band` and its test.
- `loop_guard.rs`: removed `LoopGuard::restart` (it has no callers left) and its test.
- `pirls/state.rs`: documentation now states that `HessianCurvatureKind` is a property of the model, never of the solve history.
- `bench/large_scale/runner.py` and `tests/bench_large_scale_runner_test.py`
  - Removed the parsers for `[PIRLS] mid-iter Fisher fallback` and `[PIRLS] force_fisher_for_rest`, which the engine no longer emits.
  - Removed the `[CURVATURE health]` verdict axis. Its fisher_frac was a fallback rate. The kind is now per-model, so fisher_frac is 0 or 1 by construction and grades nothing. `[FIT health]` combines warm_start and pirls.
  - Kept the per-kind counts (`pirls_curv_n`, `pirls_curv_<Kind>`).

## Regression test

`indefinite_early_observed_hessian_converges_without_fisher_3962` uses a scalar non-canonical model:

- F(β) = β⁴/4 − β²/2 − β + 3, with g = β³ − β − 1 and H = 3β² − 1.
- H < 0 on |β| < 1/√3.
- g has exactly one real root, because the cubic's discriminant is −23 < 0. That root is the plastic number ρ = 1.3247179572…, where H(ρ) ≈ 4.265 > 0.

The test runs from three starts:

| Start β₀ | H(β₀) |
|---|---|
| 0 | −1 (indefinite) |
| −0.5 | −0.25 (indefinite) |
| 3 | 26 (convex) |

For every start it asserts:

- status is `Converged`;
- β̂ = ρ to a relative error of 1e-9;
- **every** `update_with_curvature` and `update_candidate` call requested `Observed`, with zero Fisher requests;
- the export is `ObservedExact`.

`candidate_evaluation_errors_respect_lm_exhaustion_budget` now pins the new contract: candidate errors raise λ under the same observed curvature, with `fisher_updates == 0` and `fisher_candidate_calls == 0`. The LM retry budget is still honoured.

## Remainders (not in this PR)

- **The attempt budget.** The Firth/non-Firth split of the LM attempt budget (`max_step_halving` 30/60, `loop_driver.rs`) is item 5 of the issue. It is a separate knob that bounds the damping window, and it needs its own derivation from the λ ceiling.
- **The GPU path.** `cholesky_solve_only_gpu` already answers a factorization failure through the same λ escalation. It is left as is.
- **The gradient-descent fallback.** The unconstrained "fallback to gradient descent" when the direction solve fails with λ exhausted is untouched here.

---


SauersML:
Verify-and-land review of head c9193255: **REJECTED (it doesn't fix #3962 on the sparse path)**.

The PR body says: "when H+λD² does not factor, λ is raised". That holds on the dense path, where `descent_curvature` plus the Madsen escalator does raise λ. It doesn't hold on the sparse path.

In `crates/gam-solve/src/pirls/reweight.rs:1134`, `let factor = factorize_sparse_spd(sparse_reg_ref)?;` sits inside the `match { … } { Ok(()) => …, Err(e) => … }` scrutinee block. A `?` there returns from `runworking_model_pirls` itself, so the `Err(e) => if lm_can_retry { escalate }` arm is never reached.

The PR also removes the guard that used to catch this case: `certify_sparse_penalized_hessian`, with the Observed→Fisher fallback. So an indefinite early Observed iterate on a sparse-native, non-canonical fit now aborts the whole solve. That is exactly the #3962 case.

SPEC.md: "In general, do not paper over solver issues" / "A fit object must only ever come from a converged optimization." Right now the sparse fit neither converges nor escalates; it errors out.

The new regression test is a dense scalar model, so it never runs this path.

**Principled fix:**

- Map the sparse factorization failure into the direction-solve `Err`, e.g. `factorize_sparse_spd(..).and_then(|f| solve_sparse_spd_into(&f, ..))` evaluated as the block value, so the escalator raises λ.
- Add a sparse-native, indefinite-start Observed test.

**Also:** the PR removes `pub fn restart` (loop_guard.rs), `observed_information_band` and two `#[test]` items. There are no matching entries in `docs/public-api-census-changes.json` or `docs/source-removal-changes.json`, so the test-census gates will refuse it.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, a672c7b764d2.

PR #4069, opened 2026-09-20
pirls/dispersion: precision-weight Gamma shape, Tweedie φ and Gaussian/IG dispersion (Fixes #4061)

SauersML:
Fixes #4061

The plug-in nuisance estimators in `crates/gam-solve/src/pirls/dispersion.rs` divided by Σw, which reads prior weights as frequencies. Everywhere else in the fit, prior weights are read as precisions (Var = φ/w):
- the reported log-likelihood normalizers in `deviance.rs`
- the Gaussian identity scale `Σ w r² / (n₊ − edf)`
- prediction

Because of this mismatch, a global rescale w → c·w left φ̂ unchanged while the working weights grew by c, so every SE shrank by √c. It also meant the reported φ̂/α̂ was not the MLE of the reported likelihood.

## Changes
- **Gaussian (non-identity link) / inverse Gaussian:** φ̂ = Σ wᵢ dᵢ / n₊. This is the stationary point of `Σ[−wᵢdᵢ/(2φ) − ½ ln(φ/wᵢ)]`.
- **Tweedie Pearson:** φ̂ = Σ wᵢ (yᵢ−μᵢ)²/μᵢ^p / n₊. The identity E[wᵢ rᵢ²/μᵢ^p] = φ holds row by row.
- **Gamma shape:**
  - It now solves the precision score `Σ wᵢ [ln(wᵢα) − ψ(wᵢα) − tᵢ] = 0`, the same row shape wᵢα that `gamma_saturated_log_normalizer` reports.
  - Rows are grouped by distinct weight, so a uniform-weight design stays a scalar root.
  - The bracket starts at the unit-weight closed form divided by the mean weight.
  - With unit weights the result is bit-identical to before.
  - A `w·α` product that underflows takes its `+∞` limit, and a group whose weight share underflows is dropped. This avoids `0·∞`.
- **Tests:**
  - `gamma_shape_reads_prior_weights_as_precisions` checks that the precision score changes sign across α̂, that α̂(c·w) = α̂(w)/c for c ∈ {1e-3, 7, 1e3}, and that a zero-weight row is ignored.
  - `dispersion_estimates_read_prior_weights_as_precisions` checks the closed forms against n₊ and φ̂(c·w) = c·φ̂(w) for Tweedie, Gaussian(log) and IG(log).
[...]

Issue #4061, opened 2026-09-20
pirls/dispersion: Gamma shape, Tweedie φ and Gaussian(non-identity)/inverse-Gaussian φ read prior weights as frequencies (÷Σw) while the reported likelihood, the Gaussian identity scale and prediction read them as precisions (φ/w), so SEs shrink by √c under w→c·w


SauersML:
## Defect

The estimated-scale continuous families use two different weight contracts at once.

**The precision contract, Var(yᵢ) = φ/wᵢ.** Everything outside `pirls/dispersion.rs` uses it:
- `crates/gam-solve/src/pirls/deviance.rs` `full_log_likelihood_row`, where the comment reads "Continuous responses: the weight is a genuine prior weight, not a frequency count":
  - Gaussian `-0.5 * (LN_2PI + log_phi - weight.ln())`
  - Gamma `gamma_saturated_log_normalizer`, with `log_a = weight.ln() + log_shape`, so row shape = wᵢα
  - inverse Gaussian `-0.5 * (LN_2PI - log_measure_scale + 3.0 * y.ln() - weight.ln())`
  - the Tweedie series evaluated at φ/w
- The Gaussian identity profiled scale `Σ wᵢ rᵢ² / (n₊ − edf)`. Here `n₊` counts positive-weight **rows** (`estimate/optimizer.rs`: `let n = w_o.iter().filter(|&&wi| wi > 0.0).count() as f64;`).
- The contract lock `tests/weighted_gaussian_is_prior_weight_not_frequency_weight_test.py`, which rebuts #1617/#1618 and requires SEs to be invariant under w→c·w (#877).
- Prediction uses φ/w (#2025, #2077, #3924, #3957).

**The frequency contract (÷Σw).** The plug-in nuisance estimators in `crates/gam-solve/src/pirls/dispersion.rs` (origin/main e0cc6a6a) use it instead:
- `estimate_gamma_shape_from_eta`, around l.163. Each row returns `(wi*target, wi)`, then `let target = weighted_target / total_weight;`, and the root is taken of the scalar score `ln α − ψ(α) = t̄`. That is the MLE of `Σ wᵢ ℓ(yᵢ; α)`, the frequency likelihood. The precision likelihood the fit reports has the score `Σ wᵢ[ln(wᵢα) − ψ(wᵢα) − tᵢ]`.
[...]

PR #4185, opened 2026-09-20
pirls/dispersion: estimate the reported scale on n₊ − edf residual degrees of freedom (Fixes #4075)

### LINKED ISSUE #4075: pirls/dispersion: Tweedie/Gamma/inverse-Gaussian/Gaussian(non-identity) φ̂ carries no residual-df correction (÷n instead of ÷(n − edf)), so their SEs are biased low while the Gaussian identity scale uses n − edf
## Defect

Two paths estimate the scale with different residual-df conventions.

**Gaussian identity path.** The profiled scale divides by the residual df:
- `crates/gam-solve/src/estimate/optimizer.rs` ~3000: `let denom = if opts.compute_inference || needs_constrained_posterior { n - edf_total } else { n };`
- The comment there reads: "σ̂² = RSS / (n − edf_total), matching … the residual df … Using the null-space dim … was wrong … σ̂² was systematically biased low whenever any smooth/random-effect spent real edf".

**Every other estimated-scale continuous family.** The fit instead ships the plug-in estimator from `crates/gam-solve/src/pirls/dispersion.rs`, which has no df correction:
- `estimate_tweedie_phi_from_eta`: Pearson `Σ w r²/μ^p / n` (÷Σw on current main, see #4061).
- `estimate_dispersion_phi_from_eta`: Gaussian with non-identity link, and inverse Gaussian, `Σ w d / n`.
- `estimate_gamma_shape_from_eta`: the shape MLE at the plug-in β̂.

The optimizer then sets `Vb = H⁻¹` with `W = prior/φ̂` folded into H (the cov_scale contract right below the code above). Nothing rescales φ̂ by n/(n − edf).

## Why it is wrong

Take the Pearson/deviance moment `Σ wᵢ dᵢ` evaluated at the fitted μ̂:
- Its expectation is about φ·(n − tr(F)), where F is the influence/hat matrix of the penalized IRLS fit.
- So `Σ w d / n` has expectation about φ·(n − edf)/n.
- Every SE is therefore too small by a factor of about √((n − edf)/n).
- This makes Wald p-values anti-conservative. The SPEC says p-values must be calibrated.

Example: n = 60 with a smooth that spends edf = 12.
- SEs are 10.6% too narrow.
- Nominal 95% intervals cover at P(|Z| < 1.753) ≈ 0.92. That is before the extra t-vs-normal gap (#3782).

For the Gamma shape MLE, the analogous small-sample bias is α̂ inflated by about n/(n − edf).

It is also inconsistent. The same Gaussian data fitted with `link="identity"` gets `RSS/(n−edf)`, but with `link="log"` (a near-identical fit when μ is far from 0) it gets `RSS/n`.

## Principled fix (not a one-liner)

The edf depends on φ through `H = XᵀWX/φ + S_λ`. A post-hoc rescale at fixed λ would move β̂, so the correction has to live inside the final converged-η fixed point:
- At the final fit, refresh φ as `Σ w d / (n₊ − edf(φ, λ̂))` (Pearson/deviance families).
- For the Gamma, use the corresponding df-corrected profile. One option is the Laplace/REML-marginal shape score `Σ wᵢ[ln(wᵢα) − ψ(wᵢα) − tᵢ] + edf/(2α)·(…)`; alternatively, adopt the Pearson scale for Gamma inference.
- Iterate until φ, β̂ and edf agree. This is the same loop that already refreshes φ in `loop_driver.rs` (~1757/1823), with edf taken from the PIRLS penalized Hessian that the loop already factors.
- Add a coverage test at small n / large edf for Tweedie and Gaussian(log). The existing `gamma_log_coefficient_se_coverage` runs at n = 250, p = 3 and cannot see this.

SauersML:
Fixes #4075

## Defect

The reported Gamma shape, the Tweedie φ and the Gaussian (non-identity link) / inverse Gaussian φ divided the converged-mean deviance or Pearson sum by the n₊ positive-weight rows. That ignores the edf the penalized mean model spent. So every estimated scale was biased low by the factor (n₊ − edf)/n₊, and every SE and interval under-covered by the square root of that factor.

## Math

The reported scale is now the stationary point, at fixed λ, of the Laplace marginal likelihood in the scale. With H = XᵀW₀X/φ + S_λ, the −½ log|H| term contributes ∂/∂φ = tr(H⁻¹XᵀW₀X/φ)/(2φ) = edf/(2φ) to the scale score. Solving the score for each family:

- **Gaussian / inverse Gaussian:** φ̂ = Σ wᵢ dᵢ / (n₊ − edf). This is the same residual-df scale as the Gaussian identity RSS/(n₊ − edf).
- **Tweedie:** φ̂ = Σ wᵢ (yᵢ − μᵢ)²/μᵢ^p / (n₊ − edf), because E[Pearson(μ̂)] ≈ φ(n₊ − edf).
- **Gamma:** the root of Σ_g f_g [ln(w_g α) − ψ(w_g α) − κ/(2 w_g α) − t̄], with κ = edf/n₊.
  - The Fisher weight in H is ∝ α, so −½ log|H| contributes −edf/(2α).
  - Each term is strictly decreasing in α when κ < 1, because ψ'(x) > 1/x + 1/(2x²). It falls from +∞ to 0⁺, so a positive t̄ gives a unique root.
  - At a large shape, α̂ → (n₊ − edf)/D, where D is the Gamma deviance.
  - The precision reading is kept: α̂(c·w) = α̂(w)/c.

The edf is read from the current penalized Hessian on every pass of the final-fit refresh loops, so (β, scale, edf) reach their joint fixed point together. The warm-start seeds and the frozen λ-search scales keep the plug-in value (edf = 0), so a frozen scale is not tied to one trial ρ. A mean model with no residual degrees of freedom (edf ≥ n₊) is refused rather than reported.

## Tests

- `dispersion_estimates_divide_by_residual_degrees_of_freedom`: Tweedie, Gaussian and IG scales equal exactly n₊/(n₊ − edf) times the plug-in value. edf ≥ n₊, negative edf and NaN edf are refused.
- `gamma_shape_divides_by_residual_degrees_of_freedom`:
  - The root brackets the Laplace-marginal score Σ wᵢ[ln(wᵢα) − ψ(wᵢα) − tᵢ] − edf/(2α).
  - The shape falls below the plug-in value.
  - Weight rescale invariance holds.
  - At a small dispersion the shape equals (n − edf)/D to 1e-7.

Validation: focused-rust-proof runs BF-46-4075-1/2 are dispatched (queued).

## Note on the base

This branch carries PR #4069's commit a0c5809ea6 (#4061, n₊ precision-weight convention). The residual-df correction is defined on that convention's n₊. Merging this PR also lands #4069.

---


SauersML:
Merged main in (conflict resolved by BF-69); ready for a merge train.

New head: dff95bc5dacb1eec43a2e31579bc0008a9e1f46a (two-parent merge with origin/main 6f7bb7b8 tree; no rebase or force-push).

Conflict: `crates/gam-solve/src/pirls/loop_driver.rs`, two hunks.
- The Gamma shape refresh at its pass cap now keeps main's `bail_invalid_estim!` refusal (#3544) instead of the old `break`. This branch had only reworded the comment above the old `break`.
- The Gaussian/IG dispersion comment keeps this branch's residual-df estimate `φ̂ = Σ wᵢ dᵢ / (n₊ − edf)` (#4075) and adds main's wording that every converged-η refresh refuses at the cap.

I checked that main added no new callers of `estimate_{gamma_shape,tweedie_phi,dispersion_phi}_from_eta`: the call-site counts per file are the same on the branch and on the merged tree, and every caller passes the new edf argument. `calculate_edf_with_penalty` still returns `Result<f64, _>` on main.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 5e2b51421027.

PR #4187, opened 2026-09-20
Gate marginal-slope band coverage two-sided over replicates (#3893)

### LINKED ISSUE #3893: tests/python: marginal-slope interval coverage gate accepts any over-covering band (0.80 <= coverage <= 1.0, single seed)
`tests/test_bug_hunt_marginal_slope_predict_interval.py:192` and `:236` gate the nominal-95% Bernoulli marginal-slope band with

```python
assert 0.80 <= coverage <= 1.0, f"95% band coverage of p_true was {coverage:.3f}"
```

on a single realization (`_make_fixture(seed=20240613)`).

**Why this is wrong**

1. The upper bound is `1.0`, so the test cannot fail when the band over-covers. A band built from a covariance inflated by any factor (for example, a Vp that double-counts the smoothing-parameter correction, or an SE taken on the wrong scale) gives coverage 1.0 and passes. The only other level-sensitivity check is `width50 < width95`, which a uniformly inflated covariance also passes. The repo treats over-conservative uncertainty as a defect, not as slack. SPEC.md line 21: "In general, do not paper over solver issues."
2. The `0.80` lower bound is not derived. The docstring calls it "a generous tolerance band ... for finite-sample + smoothing-bias slack." On one fit the 600 per-row coverage indicators are not independent. They are all driven by the same few-dimensional estimation error (edf of about 3 to 5 here), so a single-seed coverage fraction has no binomial error bar. No fixed number is a principled bar on it, and 0.80 is only a guess.

**Principled fix**

Measure calibration as across-the-function coverage over independent replicates of the DGP. For the Bayesian interval that is the quantity that sits at nominal (Nychka 1988; Marra & Wood 2012). Then gate both sides with the Monte Carlo error measured on the replicates themselves:

- Draw R replicates with fresh `y` from the same `p_true` and record each replicate's coverage `c_r` at level 0.95, and also at 0.50. The 50% level is what makes over-coverage visible.
- Assert `|mean(c_r) - level| <= 4 * sd(c_r) / sqrt(R)` for each level.

The 50% level turns a uniformly inflated band into a large, detectable excess, because coverage of about 0.5 cannot hide in a ceiling the way 0.95 hides under 1.0. If this lane is too slow for R replicates under the 300 s pytest-timeout, the same gate belongs in a Rust integration test on `fit_bernoulli_marginal_slope_terms`, where a replicate fit is cheap, and the Python file keeps only the construction oracle, which is exact.

Found while auditing tests for one-sided or vacuous tolerances.

SauersML:
Fixes #3893

## What was wrong

The Bernoulli marginal-slope interval tests checked each band on **one** fitted realization and accepted `0.80 <= coverage <= 1.0`. That gate has two defects:

- **One-sided.** An over-covering band, such as any inflated covariance, reaches 1.0 and passes. SPEC.md requires two-sided calibration: conservative is also a bug.
- **No error bar.** The per-row coverage indicators of one fit are not independent. They share one low-dimensional estimation error (beta_hat - beta), so the across-rows mean has no binomial standard error, and the 0.80 floor is a magic constant.

## Change

Each band is now fitted on R = 64 independent responses `y_r ~ Bernoulli(p_true)`. The two bands are the `predict(interval=)` band and the posterior `sample().predict(level=)` band. For each replicate r:

- `c_r` is the fraction of rows whose band contains `p_true`.
- The `c_r` are iid across replicates.

The gate is the Monte Carlo test

    |mean_r(c_r) - level| <= 4 * sd_r(c_r) / sqrt(R)

applied at level = 0.95 and level = 0.50. Under correct calibration the two-sided false-alarm rate is about 2 * Phi(-4) ≈ 6e-5 per check. The 0.50 level exposes over-coverage, which saturates near 1.0 at 0.95. The old `0.80..1.0` gates are removed.

---


SauersML:
Reviewed against SPEC.md (BF-88, partition-5 second pass). This PR is sound and waits only on its build, so it is **ready for the local merge train (MT-1, odd PRs)** once its check "Validate one (issue BF-44-3893-1)" (run 35504073375, queued since 10:05Z) is green.

- It changes only `tests/test_bug_hunt_marginal_slope_predict_interval.py`.
- It replaces the one-sided `0.80 <= coverage <= 1.0` gate, which has two problems:
  - It can never fail an over-covering band, and the swarm's calibration rule treats a conservative interval as a bug, the same as an anti-conservative one.
  - 0.80 is an underived floor. SPEC.md says: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code."
- The new gate is a two-sided Monte Carlo test over R = 64 iid replicate refits, `|mean(c_r) - level| <= 4 sd(c_r)/sqrt(R)`, at levels 0.95 and 0.50.
- The error bar comes from the spread across replicates. It no longer comes from per-row indicators, which all share one estimation error.
- The 0.50 level catches over-coverage, which saturates near 1.0 at 0.95 and cannot be seen there.
- It tests the production `gamfit.fit` / `predict` / `sample` path only: no Python-side math, no skip and no XFAIL.
- There is nothing to duplicate, so I am not opening a second PR.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 1a85e7ab56f4.

PR #4192, opened 2026-09-20
kappa jet: exact log|S|₊ ψ-Hessian with positive/null coupling (#2902 row 37)

SauersML:
Part of #2902 (row 37: `constant_curvature_kappa_jet.rs:305`, the literal null-frame leak check).

## Row 37 defect
`profiled_gaussian_reml_psi_jet` restricted `log|S|₊` to the positive eigenspace `P` of `S0`. Its second derivative is the Cholesky formula `tr(R⁻¹R_ab) − tr(R⁻¹R_aR⁻¹R_b)` with `R = PᵀSP`, and it was guarded by a literal check `‖S_a N‖ ≤ 1e-8·(1+max|S0|)` that the null frame `N` is not moved. That formula is exact only when every `S_a` annihilates `N`. The whitening chart re-realizes the identifiable subspace at each ψ, so the null directions of `TᵀST` rotate, and on the issue's ln ℓ = −2.6275 fixture the leak was 9.49e-8. The fit then refused with an arbitrary constant, which SPEC forbids ("Hard-coded knobs and magic constants, especially if arbitrary, should be avoided").

## Fix: the exact jet, with the positive/null coupling
The forward fit computes `log|S|₊` over a fixed rank classification: a null eigenvalue that moves by roundoff stays null. That is the function being differentiated. Take `S = Σ λ_i v_i v_iᵀ` sorted descending, positive set pos = first `rank`, null eigenvalues `μ_j`, and `S_a`, `S_ab` in that eigenbasis. Hellmann–Feynman and second-order Rayleigh–Schrödinger give `λ_i,a = (S_a)_ii` and `λ_i,ab = (S_ab)_ii + 2Σ_{j≠i}(S_a)_ij(S_b)_ij/(λ_i−λ_j)`. Summing `λ_i,a/λ_i` and `λ_i,ab/λ_i − λ_i,aλ_i,b/λ_i²` over pos makes the positive–positive pairs collapse, so no gap between positive eigenvalues is needed:

```text
∂_a  log|S|₊ = Σ_{i∈pos} (S_a)_ii / λ_i
∂_ab log|S|₊ = Σ_{i∈pos} (S_ab)_ii / λ_i
             − Σ_{i,j∈pos} (S_a)_ij (S_b)_ij / (λ_i λ_j)
             + 2 Σ_{i∈pos, j∈null} (S_a)_ij (S_b)_ij / (λ_i (λ_i − μ_j))
```

The first two terms are the old restricted formula. The third is the positive/null coupling it left out, and it is nonzero exactly when ψ rotates the null space. The gradient has no coupling term, so it was exact before too.

**Refusal, derived and not a literal.** The only premise is that the split is resolved. A backward-stable symmetric eigensolver returns the spectrum of `S+E` with `‖E‖₂` inside `gam_linalg::roundoff::symmetric_spectrum_rounding_band` (`band = len·ε·max|λ|`). By Weyl each computed eigenvalue is within `band` of an exact one. The jet therefore refuses unless `λ_rank > band` and the gap `λ_rank − μ_max > 2·band`. The `1e-8·(1+max|S0|)` literal and the leak loop are deleted.

## Tests (`positive_pseudo_logdet_psi_jet_tests`)
- `a_rotating_null_space_keeps_log_pseudo_determinant_flat_2902`: `S(ψ) = Q(ψ) diag(λ,0) Q(ψ)ᵀ` with a rotating null frame. Here `log|S|₊` is constant, so every derivative must be 0. The restricted formula alone gives a nonzero second derivative.
- `a_moving_spectrum_under_rotation_has_its_closed_form_jet_2902`: moving eigenvalues under rotation, checked against the closed form.
- `an_unresolved_positive_null_split_is_refused_2902`.

## What remains in #2902
- Row 5 (section motion, V′) and row 38 (the >600 s hang) are untouched.
- The separate value self-check literal `1.0e-7 * (1.0 + fit.reml_score.abs())` in the same file is not row 37 and is left for the owner of that row.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, e85c2e501abd.

PR #4200, opened 2026-09-20
NB2 dispersion: read the normalizer through one shift-gap leaf, not two lnGamma jets (#3907)

### LINKED ISSUE #3907: gamlss/dispersion_family: NB location-scale log-precision score/Hessian difference ψ_k(θ+y) and ψ_k(θ) separately, so the η_d score is noise past θ≈1e7 (the Poisson limit equidispersed data walk to)
## Where (origin/main)

`crates/gam-models/src/gamlss/dispersion_family.rs`

- 780-822, `negative_binomial_row_program`, composes `ln Γ` separately at `total = θe^{δ_d} + y` and at `precision = θe^{δ_d}`, then subtracts:
  ```rust
  let total = add_constant(precision, count);
  let ln_gamma_total_jet = compose(ln_gamma_at_total, total, ln_gamma_total, digamma_total, trigamma_total, ...);
  let ln_gamma_theta_jet = compose(ln_gamma_at_theta, precision, ln_gamma_theta, digamma_theta, trigamma_theta, ...);
  ...
  let ln_gamma_ratio = add(ln_gamma_total_jet, neg(ln_gamma_theta_jet));
  ```
- 1136-1146 (`DispersionRowStacks::at`) and 2019-2027 (`dispersion_row_kernel`): these supply the two independent stacks, `polygamma_stack(theta + yi, order)` and `polygamma_stack(theta, order)`.
- 568, `dispersion_nb_loglik`: the value is computed as `ln_gamma(theta + yi) - ln_gamma(theta)`.

## Why it is wrong

The δ_d-channel of order k carries `θ^k[ψ_{k−1}(θ+y) − ψ_{k−1}(θ)]` together with lower-order chain terms. Each term is formed as the difference of two O(θ^k·ψ_{k−1}(θ)) numbers, while the gap itself is O(θ^{k−1}·y). The score is `θ[ψ(θ+y) − ψ(θ)]`, of size ~y, built from two terms of size θ·ln θ.

The η_d score also cancels at leading order against the log-share terms. The row score tends to `[(y−μ)² − y]/(2θ)`, which goes to 0 in the Poisson limit. So the absolute rounding of the digamma difference is compared against an O(1/θ) quantity.

Measured at y = μ = 3, against 50-digit mpmath:

| θ | abs error of `θψ(θ+y) − θψ(θ)` | true row η_d score | `lnΓ(θ+y) − lnΓ(θ)` abs error |
|---|---|---|---|
| 1e4 | 1.9e-11 | 1.5e-4 | 3.2e-12 |
| 1e6 | 1.1e-9 | 1.5e-6 | 7.2e-10 |
| 1e8 | **2.1e-7** | **1.5e-8** | 2.2e-8 |
| 1e10 | 3.1e-5 | 1.5e-10 | 7.6e-6 |

By θ ≈ 1e8, the per-row score error is 14× the score. The observed Hessian and the third and fourth contracted surfaces have the same defect, one power of θ higher. The exact outer ρ-Hessian consumes those surfaces.

The value channel loses the same digits. Over n rows the lnΓ-difference error accumulates into the cost (n = 1000 at θ = 1e8 gives about 2e-5), which is far above a REML cost resolution.

This regime is not exotic. For equidispersed or underdispersed counts, the NB2 MLE of θ is +∞, and the log-precision intercept is unpenalized. So the inner solve walks η_d up until the score is below its tolerance, which is exactly where the score is rounding noise. This is the same defect as #3235 (Student-t ν-score, closed by #3237), in the NB location-scale family.

The Fisher weight (`nb_log_precision_fisher_jensen`) was already rewritten for large θ. The score it divides into the working response was not.

## Principled fix

Compose the single function `g(p) = lnΓ(p + y) − lnΓ(p)` at `precision`. Its derivative stack is the gap stack `[ψ_k(θ+y) − ψ_k(θ)]`, computed in closed form. Composition is linear in the outer function, so the jet is mathematically identical to the current difference of two composes.

Generalise gam-math's `polygamma_half_shift_gap_stack` (#3237) to an arbitrary shift `s ≥ 0`, as `polygamma_shift_gap_stack(x, s, orders)`:
- δ_1 = −s/(x+s), ρ = x/(x+s), δ_m = ρ·δ_{m−1} + δ_1;
- log term `ln_1p(s/x)`.

The error analysis carries over unchanged, because every δ_m is like-signed. At s = ½ it is bit-identical to the half-shift kernel, since `0.5/(x+0.5)` and `1/(2x+1)` differ by an exact power of two. The half-shift kernel then becomes that call.

For the value channel, use the Stirling-difference `log_gamma_large_ratio(θ, y)` that the scalar NB log-likelihood already uses (`gam-solve/src/pirls/deviance.rs:2448, 2561`).

Claimed by local agent HN-40; PR incoming.

SauersML:
The negative-binomial row program built ln Gamma(theta + y) and ln Gamma(theta) as two separate jets and subtracted them. Toward the Poisson limit, each psi_k(theta + y) - psi_k(theta) is O(y theta^-(k+1)), while psi_k(theta) itself is O(theta^-k). So the eta_d score and curvature lost every digit: at theta = 1e8 the score error was 2e-7 against a true value of 1.5e-8. The row log-likelihood lost x ln x * eps the same way.

gam-math gains two functions:
- `polygamma_shift_gap_stack(x, s, orders)` generalizes the half-shift gap kernel to any shift s >= 0. At s = 1/2 the old kernel now delegates to it and stays bit-identical.
- `ln_gamma_shift_gap(x, s)` differences Stirling's series in closed form at x >= 20 and uses a product-then-ln_1p recurrence below.

The NB program now builds one leaf, g(theta) = ln Gamma(theta + y) - ln Gamma(theta), with that stack. The retired row kernel reads the digamma gap from the same stack, and `dispersion_nb_loglik` uses the stable value.

Fixes #3907

---


SauersML:

---

Issue #3705, opened 2026-09-20
gam-sae dual_certificate/routability: linear-lane audits never enforce the unit-norm decoder contract, so external / frozen dictionaries are certified in their own scale


SauersML:
## Defect

Both linear-lane audits in gam-sae are only meaningful for **unit-norm decoder rows**, but neither checks that. A dictionary whose rows are not unit norm gets a verdict that depends on its arbitrary row scaling.

### 1. `sparse_route_dual_certificate` (crates/gam-sae/src/dual_certificate.rs:257)

- At lines ~300-336 the implied λ_i is `min_active |c_k|` (a code mass). The off-support gate is `|⟨r, d_k⟩|`. The ratio is `optimality_ratio = dual_ratio(max_off_gate, implied_lambda)`.
- Rescale atom k by ν (d_k → ν d_k, c_k → c_k/ν). This leaves the reconstruction unchanged, but its gate scales by ν and its code by 1/ν.
- So the BLASSO dual certificate `optimality_ratio ≤ 1` (module docs, line 56) can be made to pass or fail at will by rescaling rows. The "birth" atom is then just the atom with the largest norm.

### 2. `routability_audit` (crates/gam-sae/src/routability.rs:202), linear lane (block_size == 1)

- The module docs (line 5) define the router score as `|⟨x, d_k⟩|` with **unit-norm `d_k`**.
- The measured cross-gate / ‖r‖ is compared against the dimensionless floor `√(b_max/p) + √(2 ln(K/δ)/p)`. That floor is derived for unit rows.
- A row of norm ν inflates its measured coherence by ν, so `coherence_excess` measures row scale rather than interference.

### Production path that hits it

- `frozen_dictionary()` (crates/gam-pyffi/src/manifold/manifold_and_posterior_ffi.rs:5374) returns each atom's **mean decoded ambient direction** (`mean_i Φ_k(coords_k[i]) B_k`). That is not unit norm in general.
- experiments/audit_sae/curved_fit_ev_and_audit.py:139-142 feeds exactly that matrix to `gamfit.sae.audit_sae`, which reaches both audits through the FFI (sae_spectral_ffi.rs:944 `routability_audit`, :981 `sparse_dict_dual_certificate`, and `run_sparse_sae_audit`).
- External SAE decoders, which the facade exists to audit, are likewise unnormalized in general.
- Fitted gam-sae decoders are fine, because they pass through `unit_norm_rows`.

## Principled fix

[...]

PR #4204, opened 2026-09-20
gam-sae: routability gate as a tangent-space score test with one BIC charge (#3926)

### LINKED ISSUE #3926: sparse_dict/update: routability gate applies the z·s/√n charge twice and tests mean |c|, which is nonzero under noise
**Where** (origin/main b3832b92): `crates/gam-sae/src/sparse_dict/update.rs:2973-3043`, in `routability_z_alpha` and `routability_gate_decisions` (callers at :919 and :3054).

```rust
fn routability_z_alpha(firings: usize) -> f64 {
    (firings.max(2) as f64).ln().sqrt()
}
...
            let mean_amplitude = eq.amplitude_sum[atom] / n;       // amplitude_sum += |c| (:2356)
            let charge_floor = residual_scale * z_alpha / n.sqrt();
            let margin = (1.0 - charge_floor / mean_amplitude).max(0.0);
            ...
            let threshold = (z_alpha * residual_scale / (mean_amplitude * margin)).powi(2);
            RoutabilityGateDecision { refresh: n >= threshold, ... }
```

**Why it is wrong**

1. **The charge is applied twice.** Write `m = mean|c|`, `s = residual_scale` and `z = z_alpha`. Then `m·margin = m − zs/√n`, and
   ```
   n ≥ threshold  ⇔  √n (m − zs/√n) ≥ zs  ⇔  m√n ≥ 2 z s.
   ```
   The comments and the BIC derivation in `routability_z_alpha` describe a one-charge test, `m√n ≥ z s`. The code demands twice that, so an atom is deferred for about 4× as many firings as the stated rule implies.

   The unit test `deferred_atom_accumulates_until_routability_threshold_crosses` (:4663) asserts that a single firing with `m = 1, s = 1` defers. Under the one-charge rule, `1 ≥ √(ln 2) ≈ 0.83` holds and the atom would refresh. The test passes only because of the doubled charge (`1 ≥ 1.67` fails).

2. **The statistic is not calibrated.** `amplitude_sum` accumulates `|c|`. Under the null of no signal, `E|c| = s√(2/π) > 0`, so `m√n/s` grows like `√(2n/π)` on pure noise. Against a threshold that grows only like `√ln n`, a noise-only atom eventually passes the gate. The comparison is a z-test with no valid null distribution.

**Principled fix (needs design)**

- Test the signed evidence with its own standard error. The solve already forms the normal equations (`eq.diag`), which give each atom's least-squares amplitude and its SE.
- Apply a single charge from a stated rule: either the BIC `½ ln n` or a calibrated α.
- Remove the doubled `margin`/`threshold` construction.
- Update the test to the stated rule.

SauersML:
Fixes #3926

## Problem
The sparse-dictionary routability gate (`crates/gam-sae/src/sparse_dict/update.rs`) refreshed atom `k` iff `mean|c_k| ≥ 2 z s/√n`. It had two defects:
- The `z·s/√n` charge was applied twice (`n ≥ threshold ⇔ m√n ≥ 2zs`).
- `mean|c|` has null mean `σ√(2/π)`, so the test was not calibrated.

The fix the issue suggests (the signed least-squares amplitude along `d_k` over its SE) is **degenerate** here. Codes are ridge-LS on the active set and rows are unit-norm (#3705), so `d_kᵀ r_k = A_kk‖d_k‖² + O(ρ)`. That makes the amplitude identically 1 and `t = ‖c_k‖/σ`, which grows like `√n` on pure noise and refreshes every atom.

## Design decision
Test what the refresh actually changes: the **direction** of the row on the unit sphere. The gate becomes a Rao score test of `H0: u = d_k` with BIC's single charge for the `p − 1` direction coordinates. Driving SPEC lines:
- "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code." The `z` level is gone, and the threshold is BIC's own charge.
- "P-values must be calibrated and accurate": the statistic is exactly `χ²_{p−1}` under H0 for s = 1.

## Math
The codes `C` are the fixed design, and the other atoms are held at their rows. The partial model is `x_i − Σ_{l≠k} c_il d_l = c_ik u + ε_i`, with `ε_i ~ N(0, σ²I_p)` and `‖u‖ = 1`.

```
r_k = B_k − Σ_{l≠k} A_kl d_l        (B = CᵀX, A = CᵀC)
g_k = (I − d_k d_kᵀ) r_k = Σ_i c_ik (I − d_k d_kᵀ) e_i      (tangent score; e_i = current residual)
I   = (A_kk/σ²)(I − d_k d_kᵀ)       (Fisher information on the tangent space)
T_k = ‖g_k‖² / (σ² A_kk)
refresh  ⇔  T_k ≥ (p − 1) ln n_k    (T/2 = log-lik gain  vs  BIC ½(p−1) ln n_k)
```

**Calibration.** With s = 1, the codes and the routing depend on row `x_i` only through its in-span coordinates. Isotropic Gaussian noise leaves the tangent residual `(I − ddᵀ)ε_i` independent of those coordinates. So, given the codes, `g_k ~ N(0, σ²A_kk(I − ddᵀ))` and `T_k ~ χ²_{p−1}` exactly. For s > 1 the statistic is a weighted χ² with mean ≈ p − s.

**Behaviour.**
- Under H0, `T = O_p(p)`, below `(p−1) ln n`, so the atom defers and keeps accumulating.
- Under a direction change `Δ`, `T ≈ n_k c̄²‖Δ‖²/σ²`. This grows linearly against a logarithmic charge, so the atom always refreshes eventually.
- `n_k = 1` gives charge 0, so the atom refreshes.

**Noise scale.** `σ̂² = RSS/(np)` is biased low by the factor `1 − s/p`, because each live code is one LS parameter fitted per row, and it inflates `T` by the inverse of that factor. Both the one-shot map and the stream now profile the codes out: `σ̂² = RSS/(np − Σ_i|S_i|)`. That is the null-model nuisance fit a score test uses.

## Tests (`sparse_dict::update` tests)
- `routability_gate_refreshes_well_fired_and_defers_starved_atom`: exact T values 92.16 and 0.18 against ln 64 and ln 2.
- `deferred_atom_accumulates_until_routability_threshold_crosses`: T = 0.32 (defer), then T = 10.24 ≥ ln 64 after the firings accumulate. The MOD row is installed.
- `routability_gate_ignores_radial_mass_and_coupled_atoms`:
  - 64 firings of code 5 exactly along `d` give T = 0, so no refresh.
  - Two non-orthogonal co-firing atoms that reconstruct their rows exactly give T ≈ 0 for both. The coupling term removes the partner's share of `B_k`.
- `routability_score_is_chi_square_under_the_null_with_least_squares_codes`: 4000 trials, n = 50, p = 3, pure noise, router-style LS codes `c_i = dᵀx_i`.
  - T has mean 2 and variance 4 (χ²₂).
  - The refresh rate matches `P(χ²₂ ≥ 2 ln 50) = 1/50`.
  - The profiled `σ̂²` has mean 1 (the old `RSS/(np)` averaged 2/3).
  - The amplitude statistic would have refreshed on every trial.

---


SauersML:
Status (BF-89): this is reviewed and only waits on its build. It is ready for the local merge train (MT-0, even PR) once focused proof run 35511583145 is green at head `119a07125b`. That head merges cleanly with current `origin/main`. The superseded run 35507180844 (old head ce100c9) was cancelled.

Once it lands, close #3926 with this data:
- The exact T values in the tests: 92.16, 0.18, 0.32 and 10.24.
- The null check: `T ~ χ²₂`, with mean 2 and variance 4, and a refresh rate of 1/50.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 2401bd7ae074.

PR #4206, opened 2026-09-20
predict axis clip: exempt sphere/periodic axes on every surface and through by= wrappers (#4189)

### LINKED ISSUE #4189: predict axis clip: sphere() on the noise/slope surface or inside by= gets its periodic longitude and pole latitude clamped to the training box
## Summary

`FittedModel::axis_clip_to_training_ranges` clamps each predict-time feature to its training range, with two exemptions:

- a `sphere()` longitude is periodic, so it is never clamped;
- a `sphere()` latitude is a closed-manifold coordinate, so it is clamped only to the manifold's own bounds of ±90° (±π/2 in radians).

Both exemptions are computed from the **mean-surface, top-level** smooths only. As a result, a sphere term gets its longitude and pole latitude clamped to the sampled training box when it sits in either of these places:

- on the noise surface (`resolved_termspec_noise`) or on a slope surface (`resolved_slopespec` / `resolved_slopespecs`);
- inside a wrapper such as `by=` (`ByVariable`), `FactorSumToZero` or `BySmooth`, on any surface.

This is the #508 pole artefact and the #456 silent clamp, returning on every surface except the top-level mean.

## Evidence (origin/main 481272e7; file unchanged through 65b4284851)

`crates/gam-models/src/inference/model.rs`:

- **L3413-3426, `saved_term_specs()`:** this method already lists every surface that `axis_clip_to_training_ranges` clips for: mean, noise, slope and slopes.
- **L3434-3497, `training_periodic_axes`:** it reads only the mean spec and matches only top-level bases.
  ```rust
  let Some(spec) = self.resolved_termspec.as_ref() else { return out; };
  for term in &spec.smooth_terms { match &term.basis {
      SmoothBasisSpec::Sphere { feature_cols, .. } => { ... out.insert(lon_col) }
      ...
      SmoothBasisSpec::ByVariable { .. } | SmoothBasisSpec::BySmooth { .. }
      | SmoothBasisSpec::FactorSumToZero { .. } | ... => {}
  ```
  The comment at L3478-3485 says the wrappers "are deliberately not descended into here, so a periodic marginal nested inside one stays subject to the training-range clip". It gives no reason, and there is none: a periodic basis wrapped in `by=` is still periodic in its column. The comment arrived in f076f9d85b, which replaced a `_ => {}` wildcard with an explicit list. It documented the existing gap rather than a design decision.
- The same wrapper gap therefore also clamps a periodic `BSpline1D` or periodic tensor margin inside `by=`.
- **L3629-3653, `training_sphere_latitude_bounds`:** it reads only the mean spec and only a *top-level* `Sphere`.
  ```rust
  let Some(spec) = self.resolved_termspec.as_ref() else { return out; };
  for term in &spec.smooth_terms {
      if let SmoothBasisSpec::Sphere { feature_cols, spec } = &term.basis && ...
  ```
  As a result, `sphere(lat, lon, by=g)` on the mean surface also gets its latitude clamped to the sampled range. For example, training latitudes of [-10, 40] clamp a query at 85° down to 40°.

## Consequences

Take `formula_noise = "~ sphere(lat, lon)"`, or `s(...) + sphere(lat, lon, by=z)`:

- A query at longitude 170°, trained on [-20, 60], is evaluated at 60°. The periodic basis is well defined everywhere, but the prediction silently belongs to a different point on the globe.
- A query at a pole is pulled back to the sampled latitude extreme, so the predictor is again not single-valued at the poles (#508).

Nothing warns about either clamp. The predicted noise SD, the bands and the posterior draws all use the wrong location.

## Fix

Collect the exemptions with one recursive collector per exemption. Each collector walks every basis from `saved_term_specs()` through `ByVariable.inner`, `FactorSumToZero.inner` and `BySmooth.smooth`. The mean-surface top-level case then becomes one instance of the general rule rather than the only case handled. A regression test asserts the same clip result for:

- a mean-surface sphere;
- a noise-surface sphere;
- a `BySmooth`-wrapped sphere.

Expected result: the longitude is untouched and the latitude is clamped only at ±90.

SauersML:
Fixes #4189

`axis_clip_to_training_ranges` clamps each predict-time feature to its training range, with two exemptions:

- periodic axes (a sphere longitude, periodic B-spline axes) are never clamped;
- a sphere latitude is bounded only at the manifold's ±90 (±π/2).

`training_periodic_axes` and `training_sphere_latitude_bounds` read only the mean-surface top-level smooths. A sphere on the noise or slope surface, or inside a `by=` / `FactorSumToZero` / `BySmooth` wrapper, therefore had its longitude and pole latitude clamped to the sampled training box. That brings back the #508 pole artefact on those surfaces.

## Change (single file: `crates/gam-models/src/inference/model.rs`)

- `collect_periodic_axes` and `collect_sphere_latitude_bounds` recurse through `ByVariable.inner`, `FactorSumToZero.inner` and `BySmooth.smooth`.
- Both exemptions iterate every surface from `saved_term_specs()`, the same set the clip itself covers.
- New test `axis_clip_treats_noise_and_by_wrapped_sphere_like_a_mean_sphere`: a mean-surface sphere, a noise-surface sphere and a `BySmooth`-wrapped sphere all leave longitude untouched and clamp latitude only at 90.

---


SauersML:

---

PR #4208, opened 2026-09-20
generative: thread precision prior weights into Gamma/IG/Tweedie replicates; resolve the Student-t law

### LINKED ISSUE #4163: generative: weighted Gamma/inverse-Gaussian/Tweedie replicates ignore the prior weights (precision contract Var=φV(μ)/w) and draw at the pooled dispersion
## Where (origin/main)

`crates/gam-models/src/inference/generative.rs:231-281`, `generativespec_from_predict` / `scale_gaussian_sigma_by_prior_weights`:

```rust
    // Only the Gaussian arm exposes a location-scale `sigma`; the other families
    // encode dispersion through their own precision parameter and analytic prior
    // weights do not enter their observation draw, so they are left untouched.
    if let (NoiseModel::Gaussian { sigma }, Some(weights)) = (&mut noise, prior_weights) {
        scale_gaussian_sigma_by_prior_weights(sigma, weights)?;
    }
```

`NoiseModel::from_likelihood` broadcasts one pooled scalar to every row (Gamma `shape`, inverse-Gaussian `phi`, Tweedie `phi`), and the weighted Standard-class caller (`crates/gam-predict/src/generative.rs:630`) passes the fit's prior weights, which this function then ignores for every non-Gaussian family.

## Why it is wrong

For the continuous exponential-dispersion families the fit reads prior weights as **precisions**: `Var(y_i) = φ V(μ_i) / w_i`. That is the contract in `full_log_likelihood_row` (`crates/gam-solve/src/pirls/deviance.rs`):
- Gamma: normalizer with `a = w·k`;
- inverse Gaussian: `−½ ln w` in the density;
- Tweedie series: `λ = μ^{2−p} w / (φ(2−p))`, gamma scale `φ(p−1)μ^{p−1}/w`.

The same contract holds in the PIRLS working weights `w h′² / (φ V)` and in the Pearson/ML dispersion (#4061). So the fitted law of row `i` is:

| family | fitted law of row `i` | what the replicate draws |
|---|---|---|
| Gamma | shape `k w_i` | shape `k` |
| inverse Gaussian | dispersion `φ / w_i` | `φ` |
| Tweedie | dispersion `φ / w_i` | `φ` |

A row with `w_i = 4` is drawn with 4x its fitted variance, and a row with `w_i = 0.25` with 1/4 of it. `sample_replicates` / posterior-predictive checks on a weighted Gamma/IG/Tweedie model therefore reproduce a homoscedastic-in-weight law the model never fitted, while the Gaussian arm of the same function already does this correctly (#2025). The doc comment's claim that "analytic prior weights do not enter their observation draw" is false for these three families.

Poisson / NB / Beta / binomial read prior weights as frequency counts (their normalized log-likelihood requires integer weights), so a replicate of one row is a single draw from the fitted law and correctly ignores the weight.

## Fix

Replace the Gaussian-only `if let` with one precision-weight map over the broadcast law:
- Gaussian `σ_i = σ/√w_i`;
- Gamma `k_i = k w_i`;
- IG / Tweedie `φ_i = φ/w_i`;
- every other variant enumerated explicitly as unchanged.

Keep the same length, finiteness and positivity validation. Pin it with a unit test: `k w_i`, `φ/w_i`, unit weights reproduce the pooled scalar, Poisson unchanged, zero weight refused.

The observation-band side (`family_predictive_variance`, `crates/gam-predict/src/lib.rs:2306-2370`, the Gamma/IG/Tweedie arms use the pooled `φ` with no `prior_weights`) has the same omission. It is left out of this change because that function is under active edit in open PRs #4063/#4065/#3999; once they land, the same `φ/w_i` belongs there.

A related but separate gap, not addressed here: a binomial fit on **proportions** with trial-count weights (`y = k/w`, the `ln C(w, wy)` normalizer) is replicated as a single `Bernoulli(μ)` 0/1 draw, not as `Binomial(w_i, μ_i)/w_i`. For 0/1 responses the Bernoulli draw is correct under any weight.

### LINKED ISSUE #4166: generative: every Student-t sample_replicates / gam generate call fails — family_noise_parameter rejects the Student-t FixedDispersion{phi:1} metadata as inconsistent
## Where (origin/main)

`crates/gam-models/src/inference/generative.rs:40-140`, `family_noise_parameter`.

A Student-t fit records `LikelihoodScaleMetadata::FixedDispersion { phi: 1.0 }` (`crates/gam-spec/src/lib.rs:2267`). The fitted `sigma` and `nu` live on `ResponseFamily::StudentT { sigma, nu }` itself. The picker has no Student-t arm, so this pair falls through to the catch-all error arm:

```rust
        (
            ResponseFamily::Binomial | ResponseFamily::Poisson,
            LikelihoodScaleMetadata::FixedDispersion { phi },
        ) if phi.to_bits() == 1.0_f64.to_bits() => Ok(None),
        (ResponseFamily::RoystonParmar, _) => Err(...),
        (_, metadata) => Err(invalid(format!(
            "family and likelihood-scale metadata are inconsistent: {metadata:?}"
        ))),
```

## Why it is wrong

Every consumer of a saved Student-t model's observation law fails with "Student-t generative scale is unresolved: family and likelihood-scale metadata are inconsistent: FixedDispersion { phi: 1.0 }". Student-t is `PredictModelClass::Standard`, so its path is `generative_spec_for_saved_model` -> Standard -> `family_noise_parameter`, and that path is shared by:
- `Model.sample_replicates` / `iter_replicates` (FFI `generative_replicates`);
- `gam generate` in the CLI.

The rest of the Student-t generative stack exists but cannot be reached:
- `NoiseModel::from_likelihood` builds `NoiseModel::StudentT { sigma, nu }` straight from the spec and ignores the scalar parameter.
- `sampleobservations` draws `mu + sigma * T_nu`.

Nothing about the metadata is inconsistent: `phi = 1` is exactly what the Student-t likelihood carries.

## Fix

Accept `(StudentT { .. }, FixedDispersion { phi: 1 })` in the same no-scalar-parameter arm as Poisson and Binomial, returning `Ok(None)`. Pin it with a unit test: the picker resolves, and the resulting spec carries the fitted `sigma` and `nu`.

SauersML:
Weighted Gamma, inverse-Gaussian and Tweedie fits use precision weights in their likelihood, but observation sampling and predictive bands previously dropped those weights. Apply the fitted per-row law in both consumers: Gamma shape is multiplied by the weight, and IG/Tweedie dispersion is divided by it. Keep posterior mean uncertainty separate from conditional observation noise.

Resolve Student-t scale metadata from the fitted sigma/nu specification. Its sigma remains a scale, and finite predictive variance exists only for nu > 2. Frequency-weighted laws and the existing single-outcome Bernoulli law are unchanged. Reject invalid precision weights, unrepresentable scaled parameters and nonfinite final predictive variances instead of producing degenerate or infinite outputs. Remove the superseded Gaussian-only private helpers.

Validation on MSI with two CPUs and warm dependencies:
- Eight tests rebuild the complete changed generative module, including 160,000 seeded row draws checked against independent EDM means/variances, representability controls and Student-t nu=1,2,4 cases.
- Three tests rebuild the exact predictive-variance implementation and its interface, checking independent total-variance formulas, unchanged frequency laws, invalid inputs and final moment overflow.
- Coherent production metadata checks pass with warnings denied for all affected crates and reverse dependencies, including gam-models, gam-predict and gam-pyffi. No full workspace runtime suite is claimed.

Fixes #4163.
Fixes #4166.


SauersML:
Taking over from HN-41 (no activity since 2026-09-20T10:40Z). I reviewed the diff and it is sound: the precision-weight map matches the fit's likelihood rows in `pirls/deviance.rs::full_log_likelihood_row`.

| Family | Fit's likelihood row | Replicate draw |
|---|---|---|
| Gamma | normalizer uses a = w·k | shape k·w |
| Inverse Gaussian | −½ ln(2πφy³/w) | φ/w |
| Tweedie | series uses w/φ | φ/w |
| Gaussian | — | σ/√w |

Student-t's prior weight is a likelihood power (`w·log_normalizer − half_deviance`), not a precision, so leaving its law unchanged is correct.

The branch merges cleanly with current main. I will merge once focused-proof run 35505749218 on 1d553f09 is green.

---

PR #4209, opened 2026-09-20
by-variable smooths: a continuous by= tensor keeps its constant z*c

### LINKED ISSUE #4191: by-variable smooths: te(..., by=z) with a continuous z still deletes the varying coefficient's constant z*c (tensor sum-to-zero gauge not released)
## Summary

The continuous-`by=` rule says a varying coefficient `f(x)·z` keeps its constant direction (`z·c`) inside the penalised block, because that direction is `z`, not the intercept. For `s(x, by=z)` this works. For `te(x1, x2, by=z)` it does not: the tensor inner keeps its default full-tensor sum-to-zero gauge. That gauge is applied inside the inner build, so `z·c` cannot be represented. Unlike a factor `by=`, nothing adds a `z` main effect back.

## Where (origin/main d71dfe3)

`crates/gam-terms/src/smooth/term_specs.rs:7215-7217`: the helper only forwards to the factor-level wrapper.

```rust
pub(crate) fn keep_constant_in_numeric_by_smooth(basis: &mut SmoothBasisSpec) {
    defer_inner_model_centering_to_factor_level_wrapper(basis);
}
```

`crates/gam-terms/src/smooth/term_specs.rs:7162-7171`: that wrapper only releases `BSpline1D` with `WeightedSumToZero`.

```rust
pub(crate) fn defer_inner_model_centering_to_factor_level_wrapper(basis: &mut SmoothBasisSpec) {
    if let SmoothBasisSpec::BSpline1D { spec, .. } = basis
        && matches!(spec.identifiability, BSplineIdentifiability::WeightedSumToZero { .. })
    { spec.identifiability = BSplineIdentifiability::None; }
}
```

A `TensorBSpline` inner (the `te()` default is `TensorBSplineIdentifiability::SumToZero`, `term_builder.rs:2151-2161`) therefore stays centred. Its local build then applies the full-tensor `Z` at `term_specs.rs:6247-6262`:

```rust
TensorBSplineIdentifiability::SumToZero => { ...
    let (_, z) = apply_sum_to_zero_constraint(dense_design_ref.view(), None)?;
```

The callers at `crates/gam-terms/src/term_builder.rs:855-866` (formula path) and `term_builder.rs:2693-2700` (programmatic path) invoke the helper for every continuous `by=` smooth with no explicit `identifiability=`, including `te`.

## Why it is wrong

- **The span is lost.** With `Σ_i f(x_i) = 0` imposed on the inner, the block is `z ⊙ B Z`. It spans no `z·c`, so a varying coefficient whose mean is non-zero cannot be fitted.
- **Nothing puts the direction back.** Only a categorical `by=` auto-adds a main effect (`term_builder.rs:800-823`). Numeric `by=` deliberately does not, because the constant is meant to live in the penalised block (see the doc comment on the helper and `term_builder/tests.rs:4390`).
- **The two bases disagree.** `s(x, by=z)` and `te(x1, x2, by=z)` span different model spaces for the same varying-coefficient semantics. The tensor one is a strictly smaller space.

## Verified not affected

- **ThinPlate and Duchon.** Their local builds switch `OrthogonalToParametric` to `None` (`term_specs.rs` ~8481 and ~8743). The model-level plan then residualizes against the global 1 while preserving the span (`term_design.rs` `smooth_requires_parametric_orthogonality`), so `z·c` survives.
- **`ti(...)`.** It keeps `MarginalSumToZero` by design (no main effects). This is structure, not an intercept gauge.

## Out of scope

Matérn `CenterSumToZero`, sphere, constant-curvature and measure-jet have no `None` variant or use a coefficient-space constraint. They need a separate design decision.

## Fix

The helper should release the default centring for a tensor inner too. `SumToZero` becomes `None`; `MarginalSumToZero` and any explicit choice are left as they are. The span then matches the 1-D case, and REML's null-space ridge on the unconstrained tensor still decides whether the coefficient exists. Add a unit test next to `a_continuous_by_smooth_keeps_its_constant_in_the_penalised_block` covering four cases:

- `te(…, by=z)` gives `None`
- `ti(…, by=z)` gives `MarginalSumToZero`
- an explicit `identifiability=sum_tozero` wins
- a binary `by=b` with `+ b` keeps `SumToZero`

SauersML:
Fixes #4191

`keep_constant_in_numeric_by_smooth` only released the 1-D B-spline `WeightedSumToZero` gauge. A `te(...)` inner kept its full-tensor sum-to-zero `Z`, which is applied inside the inner build, so `te(x1, x2, by=z)` could not represent the varying coefficient's constant `z·c`. Nothing restores that direction, because a numeric `by=` deliberately adds no `z` main effect.

## Change

- `crates/gam-terms/src/smooth/term_specs.rs`: the helper now also releases a `TensorBSpline` inner's default `SumToZero` gauge to `None`. It leaves these alone:
  - `MarginalSumToZero` (the `ti()` structure)
  - explicit `identifiability=` choices, which are already filtered out at the callers
  - 1-D inners, which still go through `defer_inner_model_centering_to_factor_level_wrapper`
- `crates/gam-terms/src/term_builder/tests.rs`: a new test, `a_continuous_by_tensor_keeps_its_constant_in_the_penalised_block`, covers four cases:
  - `te(…, by=z)` gives `None`
  - `ti(…, by=z)` gives `MarginalSumToZero`
  - an explicit `identifiability=sum_tozero` gives `SumToZero`
  - a binary `by=b` with `+ b` gives `SumToZero`

The unconstrained tensor keeps its double-penalty null-space ridge, so REML still decides whether the coefficient exists, the same as the 1-D case.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, cc249665f37f.

Issue #4177, opened 2026-09-20
pirls/deviance: saturated Gamma normalizer truncates the Stirling series at three terms (2.8e-10 systematic error at a=8) instead of the shared eight-term correction


SauersML:
## Where (origin/main)

`crates/gam-solve/src/pirls/deviance.rs:2567-2576`, `gamma_saturated_log_normalizer`. This is the Stirling branch (`a = w·k ≥ 8`) of the saturated Gamma log-normalizer `a ln a − a − ln Γ(a)`:

```rust
    let core = if log_a >= 8.0_f64.ln() {
        let inv = (-log_a).exp();
        let inv2 = inv * inv;
        let correction = inv / 12.0 - inv * inv2 / 360.0 + inv * inv2 * inv2 / 1260.0;
        0.5 * log_a - HALF_LOG_2PI - correction
```

`full_log_likelihood_row` uses it (line 2704) for every Gamma row whose precision `w·k` is at least 8. That covers essentially every Gamma fit with a moderate or large shape.

## Why it is wrong

Stirling gives `ln Γ(a) = (a − ½) ln a − a + ½ ln 2π + c(a)`. So the saturated normalizer is `½ ln a − ½ ln 2π − c(a)`, and it is exact only if `c(a)` is.

The code truncates `c(a)` after three terms. The first omitted term is `−1/(1680 a⁷)`. That is **2.8e-10 at a = 8** and 5.9e-11 at a = 10. Checked against the exact integer-shape reference `a ln a − a − ln((a−1)!)`:

| a | error of this branch (3 terms) | error with the 8-term series |
|---|---|---|
| 8 | 2.8e-10 | 1.7e-15 |
| 9 | ~1.2e-10 | 3.9e-16 |
| 12 | ~1.6e-11 | 1.6e-15 |

- **Accuracy.** The branch is about 1e5 times less accurate than f64 allows, right where the code switches away from `ln_gamma`. The error is systematic (same sign on every row), so it adds up in the summed log-likelihood, AIC/BIC and the LAML value for large `n`.
- **Duplication.** The same file already has the shared eight-term series `log_gamma_stirling_correction` (line 2428). It is f64-accurate for `x ≥ 8`, and `poisson_saturated_log_likelihood`, `log_gamma_large_ratio` and the NB/Beta normalizers use it. This branch is a second, truncated copy of the same correction.

## Fix

[...]

PR #4210, opened 2026-09-20
pirls/deviance: use the shared eight-term Stirling series in the saturated Gamma normalizer

SauersML:
The saturated Gamma likelihood normalizer now uses the shared eight-term Stirling correction. The previous three-term expansion left an error near the shape-eight branch boundary, despite the surrounding calculation avoiding cancellation.

Validation on MSI: the original independent factorial references at shapes 8, 9 and 12 passed in a binary compiled from the complete production gam-solve source using the coherent phase2 dependency graph. Complete gam-solve unit metadata passed with warnings denied after incorporating the current optimizer cleanup. No local builds or tests were run.


SauersML:
BF-86 (oldest-first sweep, partition 1/8) reviewed this diff for #4177. It is sound against SPEC.md and only waits on its queued focused-rust-proof build. It is ready for the local merge train (MT-0, even PR). I am not opening a duplicate.

---

PR #4215, opened 2026-09-20
fit_orchestration/fit: carry the response scale on the log-sigma offset when the noise design has no intercept

### LINKED ISSUE #4188: fit_orchestration/fit: Gaussian location-scale with a no-intercept noise formula (0 + x, x - 1, anchored B-spline) fits log σ = ln(s) + Xβ and publishes σ in standardized units
**Area:** `crates/gam-models/src/fit_orchestration/fit.rs`, `fit_gaussian_location_scale_model` (about line 1502) and `rescale_gaussian_location_scale_to_raw` (about line 1129).

## Defect (wrong math)

`fit_gaussian_location_scale_model` fits on the standardized response `y / s`, where `s = gaussian_response_sample_std(y)`. It then maps the fit back to raw units in `rescale_gaussian_location_scale_to_raw`:

- the mean coefficients are multiplied by `s`;
- the log-σ block is shifted by `+ln s`, **but only through the columns in `noise_design.intercept_range`**.

A noise formula with no global intercept has no such column. `term_collection_has_global_intercept` is false for:

- `ModelLevel::NoIntercept`, i.e. `0 + x` or `x - 1`;
- a noise design whose level is gauged by an anchored B-spline.

In that case `intercept_range` is empty and the `+ln s` shift does nothing. The standardized fit solved `log σ_std = Xβ`, which in raw units is

    log σ_raw = ln(s) + X β

The user asked for `log σ_raw = X β`. So the fitted model is a different model, because the intercept the user removed has come back as `ln s`. The published σ, and everything built from it (saved model, gam-predict's `GaussianLocationScalePredictor`, the joint-expectile standardization, intervals, log-likelihood), are in standardized units: they are off by a factor of `s`. When `s` is far from 1 (a response in large units), β has to absorb the missing constant through the covariates. That biases the slopes, and the result depends on the unit of `y`.

SPEC: "The default choices should give the best model for the user." A fit that depends on the response's unit, and whose σ is off by the sample SD, is not that.

## Fix (branch `hn7/ls-noint-log-sigma`)

- When the log-σ spec has no global intercept, the `-ln s` now rides the log-σ offset during the standardized fit. `log σ_std = log σ_raw - ln s = Xβ - ln s` is exactly the raw model, so β needs no remap.
- The `+ln s` intercept shift is kept for the with-intercept case. It is unchanged.
- A consistency check returns an error if the spec's intercept declaration disagrees with the realized noise design. Otherwise the shift would be applied nowhere or twice without any warning.
- The raw offsets are captured before the adjustment, so the saved model's offsets are the user's own.

## Test

`gaussian_location_scale_no_intercept_noise_formula_fits_the_raw_model` (in `materialize/tests.rs`):

1. Materializes `y ~ x` with noise formula `0 + x`.
2. Scales `y` by 25 and fits the raw spec.
3. Fits a hand-standardized reference: `y/s`, `mean_offset/s`, and `log_sigma_offset - ln s`.
4. Checks that the log-σ betas are equal and that the mean betas satisfy `raw = s * ref`, to a relative tolerance of 1e-5.

SauersML:
Fixes #4188

`fit_gaussian_location_scale_model` fits on `y/s` and maps σ back by shifting the log-σ intercept by `+ln s`. A noise formula with no global intercept (`0 + x`, `x - 1`, or an anchored B-spline) has no intercept column, so the shift did nothing. The fit then solved `log σ_raw = ln s + Xβ` and published σ in standardized units.

- When there is no intercept, the `-ln s` now rides the log-σ offset during the standardized fit. That is exactly the raw model, so β needs no shift. The saved model keeps the user's raw offsets.
- A consistency check returns an error if the spec's intercept declaration disagrees with the realized noise design.
- New test `gaussian_location_scale_no_intercept_noise_formula_fits_the_raw_model`: a `0 + x` noise formula with `y` scaled by 25 reproduces the hand-standardized reference fit.

Proof: focused-rust-proof run 35505405045 (filters `gaussian_location_scale`).

---


SauersML:

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, b92c37e47462.

PR #4220, opened 2026-09-20
fit_orchestration/materialize: measure the marginal-slope scalar rank residual against the column's own norm

### LINKED ISSUE #4201: fit_orchestration/materialize: marginal-slope scalar rank check uses a shared max-norm tolerance, so a large-unit term (timestamp) silently drops later covariates and the model depends on term order
## Defect

`prune_unidentified_linear_terms_for_marginal_slope` (`crates/gam-models/src/fit_orchestration/materialize/terms.rs`) runs the scalar-term rank check for the Bernoulli and survival marginal-slope materializers. It orthogonalizes each linear term's column against the intercept and the terms kept so far, then drops the term as "unidentified" when

```
residual_norm <= rank_alpha * eps * (n + k + 1) * scale
```

Here `scale` is the **largest column norm seen so far**, and `rank_alpha` is 100.

A scalar coefficient absorbs any rescaling of its own column. Whether a column adds a direction is therefore a question about its angle to the span, i.e. its residual measured against **its own** norm. With a shared running `scale`, one large-unit term makes every later small-unit covariate look redundant. The model also depends on the order of the terms in the formula.

## Reproduction

The rule on main, reproduced exactly (n = 2000, `t = 1.7e9 + 1e4·row` a timestamp, `x = 0.01·sin(0.7·row)`):

| formula order | kept on main | kept with the fix |
|---|---|---|
| `t + x` | `t` (x dropped as "unidentified") | `t, x` |
| `x + t` | `x, t` | `x, t` |
| `t + x + t2` (t2 = 2t) | `t` | `t, x` (t2 still pruned) |

At n = 1e5, a unix timestamp next to an O(1) covariate (std 0.3) gives tol ≈ 1.2e3 against a residual ≈ 95. So any ordinary covariate listed after a timestamp is silently removed, with only an advisory note.

## Fix

The tolerance is now `rank_alpha * eps * (n + k + 1) * ‖column‖`, which is scale- and order-invariant. Exact duplicates and constants are still pruned, because their residual is rounding noise relative to their own norm. The existing constant-column prune tests are unchanged.

The regression test `scalar_rank_scale_invariance_tests` covers both orders and a pruned exact multiple.

Branch: `hn7/scalar-rank-own-norm`.

SauersML:
Fixes #4201

The prune tolerance was based on the largest column norm seen so far. It now uses `rank_alpha * eps * (n + k + 1) * ‖column‖`, which is scale- and order-invariant. A timestamp term no longer drops the O(1) covariates listed after it, and exact multiples and constants are still pruned. The regression test is `scalar_rank_scale_invariance_tests`.

Proof: focused-rust-proof run 35504424105.

---


SauersML:

---

PR #4222, opened 2026-09-20
predict unseen levels: a group(g) column that also carries per-level curves stays strict (#4213)

### LINKED ISSUE #4213: predict new factor levels: `s(x, by=g) + group(g)` whitelists g as lenient, so an unseen level silently drops the per-level curve f_g(x)
## Summary

`FittedModel::random_effect_group_columns` builds the whitelist of categorical columns whose unseen levels the schema encode tolerates. It adds any column that a lenient `group(g)` random-effect term reads. It never checks whether *another* term on the same column estimates a per-level **curve** that has no population fallback.

Take `y ~ s(x, by=g) + group(g)`:

- `g` is whitelisted as lenient.
- A prediction row with an unseen level `g = NEW` is encoded as the out-of-vocabulary code.
- The per-level `by=` gate compares that code against each trained level. None matches, so every `f_g(x)` column is weighted 0.

The prediction therefore silently equals the intercept plus the random intercept (0 for a new group). The whole `f_g(x)` curve disappears from the prediction, from its SE and from its band, and there is no error or warning. Without the `group(g)` term, the same row is correctly refused with an unseen-level `SchemaMismatchError` (#2102/#2137).

## Evidence (origin/main 481272e7; file unchanged through eb3fe80912)

- **`crates/gam-models/src/inference/model.rs` L5315-5343, `random_effect_group_columns`:** it inserts `training_headers[term.feature_col]` for every `lenient_unseen` random-effect term, and the #2365 `bs="re"` group column. Nothing removes a column that a per-level smooth also reads.
- **`crates/gam-models/src/inference/model.rs` L3149, `re_factor_smooth_group_col`:** its doc already states the contract. `fs`/`sz` "estimate a per-level deviation function, so an unseen level has no zero-deviation population fallback and stays strict". The same reasoning applies to `by=` factor level smooths. The rule is enforced only when the column has no `group()` term.
- **The `by=` level gate:** `crates/gam-terms/src/term_specs.rs` ~L7219-7245 and ~L7316-7372 (the `BySmooth` factor gate), and ~L7799-7811 (the `ByVariable` `Level` indicator). Each one multiplies the inner basis by `1{code == level}`. An out-of-vocabulary code matches no level, so every per-level block is multiplied by 0.

## Fix

Add a collector for the columns that carry a per-level curve:

- `ByVariable` with `ByVariableSpec::Level`;
- `FactorSumToZero`;
- `BySmooth` with `ByVarKind::Factor`;
- non-`Re` `FactorSmooth`.

The collector recurses through the wrappers and runs over every saved surface. Any such column is then removed from the lenient whitelist, so a column is lenient only when every term that reads it tolerates an unseen level. A regression test covers `ByVariable` Level + `group(g)` and `BySmooth` Factor + `group(g)`, and checks that both are refused with "unseen level". It also checks that `group(g)` alone, with the curve on a different column, stays lenient.

SauersML:
Fixes #4213

`random_effect_group_columns` whitelisted every column read by a lenient `group(g)` term. In `y ~ s(x, by=g) + group(g)`, an unseen level of `g` was therefore encoded as the out-of-vocabulary code. The `by=` level gate then matched no level and weighted every `f_g(x)` column by 0, so the curve silently vanished from the prediction, its SE and its band. Without the `group(g)` term the same row is refused (#2102/#2137).

## Change (single file: `crates/gam-models/src/inference/model.rs`)

- `collect_per_level_curve_group_cols` collects the grouping column of `ByVariable` Level, `FactorSumToZero`, `BySmooth` Factor and non-`Re` `FactorSmooth` bases. It recurses through the wrappers and runs over every saved surface.
- Those columns are removed from the lenient whitelist: a column is lenient only if every term that reads it tolerates an unseen level.
- New test `group_column_with_per_level_curves_is_not_lenient_on_unseen_levels`:
  - `ByVariable` Level + `group(g)` and `BySmooth` Factor + `group(g)` refuse an unseen `g` with "unseen level";
  - `group(g)` alone stays lenient.

---


SauersML:

---

PR #4224, opened 2026-09-20
fit_orchestration/materialize: refuse latent gauge fields the fit would silently drop

### LINKED ISSUE #4205: fit_orchestration/materialize: latent spec silently drops gauge fields: a dim_selection ARD seed beside aux_outcome, id_mode='none' beside a gauge, and any other id_mode string
## Defect

`parse_latent_specs` and `latent_id_mode` (`crates/gam-models/src/fit_orchestration/materialize/latent.rs`) turn the gauge fields of a formula latent into one `LatentIdMode`. Three specs that the parser accepts lose a field on the way, with no message.

1. **`dim_selection.init_log_precision` next to `aux_outcome`.** The parser only refuses `dim_selection` when neither `aux_prior` nor `aux_outcome` is present. `latent_id_mode` then returns `LatentIdMode::AuxOutcome { init_log_precision: ao.init_log_precision }` before it looks at `dim_selection`. The user's ARD seed is dropped, and so is its length check. For example, `dim_selection={"init_log_precision": [...]}` with an `aux_outcome` fits from the default seed.
2. **`id_mode='none'` next to a gauge field.** `explicit_none_mode` is only consulted in the `(None, None)` arm. With `aux_prior` present, the requested mode `none` is silently replaced by `AuxPrior`.
3. **Any other `id_mode` string.** Values such as `"aux_prior"`, `"isometry"` or a typo are read as "not none" and discarded. The CLI fit-request document exposes `id_mode: Option<String>`, so this is reachable there.

This matches the SPEC's rule against silently swallowed options. A request for a mode the fit does not use should be refused, not ignored.

## Fix

The parser now refuses all three cases by name:

* a `dim_selection` seed with `aux_outcome`. The message points to `aux_outcome.init_log_precision`; the head always carries ARD.
* `id_mode='none'` together with any gauge field.
* any `id_mode` value other than `'none'`.

These still parse:

* `id_mode='none'` alone
* `dim_selection=true` beside `aux_outcome`, which is redundant but harmless
* the head's own seed

Regression tests are in `latent_gauge_field_tests`.

Branch: `hn7/latent-gauge-fields`.

SauersML:
Fixes #4205

The parser now refuses three specs by name:

- a `dim_selection` ARD seed beside `aux_outcome`;
- `id_mode='none'` beside a gauge field;
- any `id_mode` other than `'none'`.

Each of these used to be dropped without a message. The regression tests are in `latent_gauge_field_tests`.

Proof: focused-rust-proof run 35504868497.

---


SauersML:

---

PR #4226, opened 2026-09-20
bms score tests: a numerical failure is an error, not "no evidence"

### LINKED ISSUE #4212: bms score tests: a non-finite robust Rao statistic reads as "no evidence", and residual_repair's centring gate passes an untested column at p = 1
## Where

`crates/gam-models/src/bms/mod.rs`, `robust_score_contributions_pvalue` (main 33e71f2), around lines 2453–2485:

```rust
    if !s.iter().all(|v| v.is_finite()) || !omega.iter().all(|v| v.is_finite()) {
        return Ok(None);
    }
    ...
    let d_stat = s.dot(&omega_pinv.dot(&s));
    if !(d_stat.is_finite() && d_stat >= 0.0) {
        return Ok(None);
    }
    let p_value = chi_square_sf(d_stat, rank as f64);
    Ok(p_value.is_finite().then_some(p_value))
```

`crates/gam-models/src/bms/residual_repair.rs:370-375`, the centring gate:

```rust
            let p_value = robust_conditional_score_pvalue(basis.view(), &u, weights)
                .map_err(|reason| ResidualRepairRefusal::CentringTestUnavailable { .. })?
                .unwrap_or(1.0);
```

## Why it is wrong

The doc comment says `None` means the test is *degenerate*, meaning there is no usable basis direction (`r == 0`, a zero `Ω̂` or rank 0). The function also returns `None` in three other cases:

- the score or meat is non-finite;
- `sᵀΩ̂⁺s` is non-finite or negative;
- the χ² survival value is non-finite.

Those are numerical failures, not degeneracy. Every caller reads `None` as "the gate does not fire":

- `conditional_score_covariance.rs:549` drops the pair;
- `:847` uses `is_some_and`;
- `residual_repair.rs:375` defaults to `1.0`.

So a score test that overflowed is reported as clean evidence that there is no conditional structure. For residual_repair it is worse: `.unwrap_or(1.0)` passes any column whose centring test could not be evaluated, including one with no usable direction. The refusal type already has a variant for exactly this case, `CentringTestUnavailable` ("The centring test of a column could not be evaluated"). A gate that cannot run must refuse, not pass.

SPEC.md:21 says: "In general, do not paper over solver issues."

## Fix

- In `robust_score_contributions_pvalue`, a non-finite score, meat, statistic or survival value becomes `Err` (every caller already propagates with `?`). `None` is kept only for the genuinely degenerate cases the doc comment names.
- In `residual_repair`, `None` becomes `CentringTestUnavailable` ("the robust score test has no usable direction") instead of `p = 1`.
- Form `D = sᵀΩ̂⁺s` as `‖ΨΩ̂⁺s‖²`, which is the same value because the rank-truncated pseudo-inverse satisfies `Ω̂⁺Ω̂Ω̂⁺ = Ω̂⁺` with `Ω̂ = ΨᵀΨ`. A sum of squares cannot round below zero. The quadratic form can, when `s` lies in the discarded null space, and that was the other path to `None`.

SauersML:
Fixes #4212

`robust_score_contributions_pvalue` returned `None` in two different situations:

- the test is degenerate (no usable direction);
- the score, the meat, the statistic or the χ² tail is non-finite.

Every caller reads `None` as "the gate does not fire", so a score test that overflowed reported clean evidence of no conditional structure. residual_repair's centring gate went further and turned `None` into `p = 1`, passing a column its test never examined.

SPEC.md:21 says: "In general, do not paper over solver issues."

**Changes**
- `crates/gam-models/src/bms/mod.rs`:
  - A non-finite score, meat, statistic or tail is now `Err`. Every caller already propagates it with `?`: `conditional_score_covariance.rs`, `estimated_latent_law.rs` and `residual_repair.rs`.
  - `None` stays only for the degenerate cases, and the doc comment says so.
  - `D = sᵀΩ̂⁺s` is now formed as `‖ΨΩ̂⁺s‖²`. This is the same value, because the rank-truncated pseudo-inverse satisfies `Ω̂⁺Ω̂Ω̂⁺ = Ω̂⁺` with `Ω̂ = ΨᵀΨ`. A sum of squares cannot round below zero. The old quadratic form could, when `s` lies in the discarded null space, and that was the other path to `None`.
- `crates/gam-models/src/bms/residual_repair.rs`:
  - A degenerate centring test now refuses with the existing `CentringTestUnavailable` instead of passing the column at `p = 1`.
  - In production this path needs every `w_i u_i` to be zero. The basis carries an intercept and the column has already passed the non-constant check, so a degenerate test there means the column was not tested.

**Tests** (unit, in `bms::tests`)
- `robust_score_test_statistic_is_the_score_quadratic_form`: the hand-computed case `s = 1.5`, `Ω̂ = 2.25`, `D = 1`.
- `robust_score_test_with_a_null_space_score_reports_p_one`: a zero score on a rank-1 meat gives `D = 0` and `p = 1`, not a degenerate test.
- `robust_score_test_refuses_a_non_finite_score`: an overflowed meat is an error.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, cbf8a9934c1a.

PR #4229, opened 2026-09-20
sae support term: read the loop-degeneracy verdict from an exact eigensolve

### LINKED ISSUE #4227: sae support term: loop-degeneracy (segment) verdict reads a fixed 8-step power iterate with no convergence test, and a zero seed skips it silently
**Area:** `crates/gam-sae/src/manifold/support_term.rs`, `convert_underoccupied_loops` (origin/main `d027be231c`, lines 2497-2545)

SPEC: "A fit object must only ever come from a converged optimization."

### What the code does

Before the largest-gap census, each 1-D periodic atom checks whether its decoded image has collapsed onto a segment (a "diameter"). If it has, the atom is unrolled to Euclidean. The verdict is `across = total − along <= resolution · probes`. Here `along` is the power of the centred image along its dominant direction. That direction comes from a fixed eight-step power iteration seeded at probe 0, with no convergence test:

```rust
direction = image[0] - centre;
for _ in 0..8 { if !(norm > 0.0) { break; } ... direction = C · direction ... }
if norm > 0.0 { /* accumulate total and along */ }
```

### Why it is wrong

- The value that matters is `λ₁` of the centred second moment. It has an exact characterization, and the loop does not certify it. Any unit direction gives `along ≤ λ₁`, so an unconverged iterate can only overstate `across`. The cap exit therefore biases the verdict toward "not a segment", and nothing records that exit.
- The seed is one row. If probe 0 sits at (or within roundoff of) the centroid, as it does for a segment traversed `sin`-phased from `t = 0`, the direction is zero or pure roundoff. With an exactly zero seed the loop breaks at once and `total` is never accumulated (it stays `0`). The `total > 0.0` guard then skips the test silently, whatever the image looks like.

### Principled fix

`along` is exactly the top eigenvalue of the smaller centred Gram (`probes × probes` or `channels × channels`), and `total` is its trace. A dense symmetric eigensolve (`eigh`) gives both with no iteration budget and no seed. A failed solve is propagated as an error, not read as a verdict. The comparison and the resolution threshold are unchanged.

Claimed by local agent HN-36 (claude/fix-loop-degeneracy-power-cap); draft PR incoming.

SauersML:
Fixes #4227

`convert_underoccupied_loops` decides whether a periodic atom's decoded image has collapsed onto a segment. It does this by comparing the power along the dominant direction of the centred image with the total power.

The dominant direction came from a fixed eight-step power iteration seeded at probe 0, with no convergence test. This had two consequences:
- An unconverged direction understates the dominant power, so the verdict could only ever miss a real collapse.
- When probe 0 sat at the centroid (a sine-phased segment), the seed was zero. The loop broke at once, `total` was never accumulated, and the test was skipped silently.

This PR takes the dominant power as the top eigenvalue of the smaller centred Gram matrix (probes × probes or channels × channels), from a dense symmetric solve (`eigh`). The total power is its trace. A failed solve is returned as an error, not read as a verdict. The comparison and the resolution threshold are unchanged.

Validation: this PR's CI, plus a focused proof run (`gam-sae` lib, filter `under_occupied_loop`).

---


SauersML:

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 7fb77cf451f6.

Issue #2977, opened 2026-09-17
Custom-family joint-Newton inner solve cycles to its 1,200-cycle budget with roundoff steps, spending ~75 s per refused outer trial point


SauersML:
### Summary

At some outer trial points of a survival marginal-slope fit, the exact joint-Newton inner solve stops making progress but does not stop. Its step is at roundoff, and its stationarity residual stays about 10× above tolerance. It keeps cycling until the 1,200-cycle budget is exhausted, and only then refuses the trial point: `custom-family inner solve did not converge after 1200 cycle(s)`. The refusal is correct, since the point is infeasible for the search. But it costs 72–80 s where a stall exit could give the same answer within a few cycles. On the AoU-shaped gam#2930 fixture these refusals take a third of the fit's wall time.

### Evidence

Fixture: `covariate_constant_slope_2930::aou_shaped_landmarked_constant_slope_survival_fit_passes_seed_validation_2930` (gam#2930 landing branch), n = 1500, θ dim 9 (ρ 7, ψ 2). acl42, cores 40–51, one test thread.

- **Two runs refuse the identical trial point at the same cost.** Run A (base 0e584db5ad plus the #2930 patch) and run B (base 98ec39dfa2 plus the patch and the probit chart fix) both refuse θ = [10.39, 6.52, 3.54, 4.07, −1.19, 6.60, 6.68, 0.94, 0.42], in 74.6 s and 75.3 s. So this predates both changes.
- **Cost:**
  - Run A: 4 refused probes took 153 s of 793 s.
  - Run B: 3 took 227.5 s of the first ~660 s (72.3 s, 75.3 s, 79.9 s), at extreme trial points with ρ up to 26.7.
- **The terminal decision variables** (run B, from the refusal's `InnerConvergenceTerminalState::JointNewton`):
  - cycle 1199, `termination=cycle budget`
  - `stationarity_residual=1.726060e-4` (tol 1.639480e-5), `relative_stationarity=1.052809e-5`, scale 15.39
  - `step_inf=1.906808e-14` (tol 1.566035e-6)
  - `resolvable_negative_curvature=false`
  - `best_stationarity_residual=1.726060e-4`, last improved 0 cycles before this one
  - `inner cycle budget inputs: base=1200, screening_cap=None`
[...]

Issue #2976, opened 2026-09-17
Joint-Newton inner target sits below the row gradient's arithmetic floor at n = 3e5, so lightly penalized seeds spin the whole cycle budget


SauersML:
At n = 300,000 the joint-Newton inner solve asks for a KKT residual below what the row gradient's own arithmetic can resolve, and it spins its whole cycle budget. Found on the anchored survival marginal-slope fit (#2928). The target construction applies to every custom family.

**Workload.** gam#2928's bench harness on a one-covariate probit survival marginal-slope fit, anchored on a 41-node declared law, on acl42 cores 52-63. gam main 5cbb785e8f content at base 0e584db5ad, with `BENCH_LOG` on.

**Seed 0's first full inner solve:**

| run | cycles | wall | best `residual_inf` | `residual_tol` | outcome |
|---|---|---|---|---|---|
| anchored, main | 1200/1200 | 573 s | 2.984e-11 | 1.028e-11 | budget exhausted, seed rejected |
| anchored, anchor roots Newton-polished (anchor noise removed) | 1200/1200 | 512 s | 1.415e-11 | 1.206e-11 | budget exhausted, seed rejected |
| closed form, same data | 7 | 2.9 s | 1.456e-11 | 2.029e-8 | converged |

In the anchored runs, 1,469 of 1,625 cycles log "decrement certificate … no model-resolvable descent remains". a1bc5336b7 then revokes each mark at the head, because the residual is above target. Steps are 1.4e-14 and |Δobj| is 5.8e-11, so the loop takes null corrections until `inner_max_cycles`. The next seed floors at 1.094e-10 with a zero step. Both anchored fits exceed the 900 s cap (rc=124), while the closed form finishes in 75 s.

Only the lightly penalized seeds get stuck. After seed 0 is rejected, the polished run's next twelve inner solves converge in 7-23 cycles each (2-13 s), at tolerances between 1.2e-6 and 5.3e2 where the scale is larger. Seed 0's 512 s of null cycles is what pushes the fit past the cap.

**Why the target is unreachable.**
[...]

PR #4231, opened 2026-09-20
Multinomial: both convergence levels from the caller's tol, certified at their own resolution (#4053)

### LINKED ISSUE #4053: multinomial formula path: inner KKT tol 1e-5 tuned on one fixture silently overrides caller tol; outer 1e-7 gradient + 1e-5 rel-cost stop are hand-tuned (not a derived certificate)
## Defect

`crates/gam-models/src/multinomial.rs` (origin/main `41fbd6c3`) sets the convergence controls of the formula-path multinomial REML fit from three hand-tuned constants. Each one was calibrated against a single fixture:

```rust
const MULTINOMIAL_FORMULA_INNER_TOL: f64 = 1.0e-5;                 // :116
const MULTINOMIAL_OUTER_REML_TOL: f64 = 1e-7;                      // :195
...
let outer_tol = if tol.is_finite() && tol > 0.0 {
    tol.max(MULTINOMIAL_OUTER_REML_TOL)                            // :3469
} else { MULTINOMIAL_OUTER_REML_TOL };
let outer_rel_cost_tol = Some(BlockwiseFitOptions::default().outer_tol); // :3494  (= 1e-5)
let inner_tol = MULTINOMIAL_FORMULA_INNER_TOL.max(tol.max(0.0));   // :3495
```

1. **The inner KKT tolerance is a loosened, dataset-measured constant.** The doc at `:102-115` gives the derivation. On the penguins fixture the KKT residual "plateaus at 2.8e-5–9.4e-5 against a scaled tolerance of ~1.9e-5", so `1e-5` was picked because it "certifies the measured plateaus". The actual reason for the plateau is also stated there: the objective has reached its f64 rounding floor, with relative changes of about 1e-11. The right response is a certificate that recognizes that floor, not a threshold looser than one dataset's plateau. A different near-separable dataset can plateau above whatever constant is chosen.
2. **The caller's `tol` cannot tighten the inner solve, although the comments say it does.** The comment at `:3460-3462` says "the caller's `tol` continues to drive the INNER joint-Newton KKT target (`inner_tol` below), where its precision actually matters". But `inner_tol = 1e-5.max(tol)`, so any caller `tol < 1e-5` is silently discarded. That includes the `tol = 1e-8` the same comment block cites as typical.
3. **The outer stop pairs two independently tuned constants.** The `1e-7` gradient tolerance is described as "calibrated" to match the GLM REML outer. The `1e-5` relative-cost stop is chosen as "the loose, fast value" so that the optimizer can "DECLARE convergence on the flat ridge instead of crawling to the iteration cap" (`:3474-3493`). A relative-cost plateau is not a stationarity certificate. Declaring convergence because the criterion stopped changing at 1e-5 relative is the pattern SPEC forbids: "A fit object must only ever come from a converged optimization."

SPEC lines:

## Principled fix

- **Inner solve.** Certify at the objective's own resolution. When the damped joint-Newton step stalls because the objective change is inside the accumulation rounding band of the penalized log-likelihood, the Newton decrement `gᵀH⁻¹g` (in objective units) against that band is the certificate. It is affine-invariant and needs no per-dataset threshold. Then drop `MULTINOMIAL_FORMULA_INNER_TOL`, and let the caller's `tol` (or the band, whichever is larger) set the KKT target, as the comment promises.
- **Outer search.** Certify ρ-stationarity with the projected ρ-gradient against the criterion's own resolution (the rounding band of the LAML value and its gradient at the inner mode). Do not use a fixed `1e-7` gradient tolerance, and do not use a `1e-5` relative-cost stop that can end the search without a gradient certificate.
- **Scope.** This is the same measure proposed for the fixed-λ `penalized_vector_glm` gate in #4047. Both multinomial solve paths should share one derived certificate.

Related: #2469 (umbrella "delete arbitrary constants"), #715, #1082 (the fixtures these constants were tuned on).

SauersML:
Fixes #4053

## What changes

The formula-path multinomial REML fit took both convergence levels from three hand-tuned constants, each measured on one fixture. This PR deletes all three and drives both levels from the caller's `tol`. Each level is certified at its own resolution.

| control | before (origin/main) | after |
|---|---|---|
| inner joint-Newton KKT target | `1e-5.max(tol)`, so the caller's `tol < 1e-5` was silently discarded | `tol`, against the residual's own rounding band and the settling certificate |
| outer ρ gradient tol | `tol.max(1e-7)` (a constant "calibrated" to the GLM outer) | `tol`, with per-coordinate bands and the Newton-decrement resolution τ_stat = 1/(2n) |
| outer relative-cost stop | `Some(1e-5)`: a plateau stop with no gradient certificate | removed (`outer_rel_cost_tol` is deleted from `BlockwiseFitOptions`) |
| anchored continuation agreement bar | relative `outer_rel_cost_tol`, set only by multinomial | absolute τ_stat = 1/(2n) in criterion units |
| `tol <= 0` or non-finite | silently replaced by a constant | rejected with an error |

## Math

**Inner.** The joint-Newton stop is already
`‖∇L − Sβ‖∞ ≤ max(tol·(1 + max(‖∇L‖∞, ‖Sβ‖∞)), band)`. Here `band` is the accumulation rounding band of the residual itself (#2812, #2976), so a residual that has reached its f64 floor is certified by the band, not by a looser threshold. The settling certificate (#2977) covers the stalled-step case (objective change inside the objective's rounding band). The penguins "plateau at 2.8e-5–9.4e-5" that motivated `1e-5` is exactly this floor. A constant above it certified nothing the band does not already certify, and it overrode tighter caller requests.

**Outer.** The ρ-search certificate compares the projected ρ-gradient with its per-coordinate evaluation bands, and the Newton decrement with the statistical resolution τ_stat = 1/(2n) (#2954). This is the criterion change that cannot move a likelihood-ratio statistic by more than one half-unit per observation. For penguins n ≈ 333, so τ_stat ≈ 1.5e-3. The relative-cost stop existed only to end the search on a flat ridge without a gradient certificate, which SPEC forbids: "A fit object must only ever come from a converged optimization."

**Continuation ladder.** Adjacent anchored rungs agree when their criterion values differ by at most τ_stat = 1/(2n) (0.5/n). This uses the observation count the outer already declares as its problem size, so the relative knob only multinomial set is no longer needed.

SPEC: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code." / "In general, do not paper over solver issues."

## Validation

focused-rust-proof.yml runs on this branch (keys `BF-48-4053-*`): penguins multinomial fixtures, the multinomial suite, custom-family lib tests, and base-sha comparisons. The results are posted below as they land.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, b1e7da1e096c.

PR #4232, opened 2026-09-20
cli: echo --id-column values as written text, not through f64

### LINKED ISSUE #4221: gam-cli: --id-column echoes prediction IDs through f64 (leading zeros stripped, integer IDs above 2^53 corrupted, NA IDs refused)
## Summary

`gam predict --id-column <col>` is supposed to copy each row's ID into the prediction CSV unchanged. Instead, `load_prediction_id_values` (`crates/gam-cli/src/main/prediction_csv.rs:43`, origin/main 65b4284) runs the ID column through the numeric schema inference (`load_dataset_projected`). It then writes the values back out:

- **Numeric columns:** each value is parsed to `f64` and re-rendered with `format_id_number`.
- **Everything else:** the column is decoded as a categorical level.

An identifier is not a quantity, so any ID column that happens to parse as numbers is silently rewritten:

| ID as written in the input | echoed in the prediction CSV |
|---|---|
| `00123` (zero-padded sample / FIPS / ZIP code) | `123` |
| `9007199254740993` (int64 key above 2^53; parquet Int64 too) | `9007199254740992` (a different row's key) |
| `1.10` | `1.1` |
| `NA` (a literal ID in an otherwise numeric column) | parsed as a missing marker, becomes NaN, then the whole predict fails with "contains non-finite value" |

When the output is joined back onto the input by ID, rows are silently dropped or mismatched. With int64 keys above 2^53, two different inputs can collide on the same ID.

SPEC: "In general, do not paper over solver issues." A lossy f64 round-trip of an ID is a silent data corruption that no error reports.

## Fix

- **New reader in `gam-data`:** add `load_column_text(path, column) -> Vec<String>`, which reads one column's cells as their written text.
  - **CSV/TSV:** the trimmed cell.
  - **Parquet:** arrow's display formatter, in non-safe mode, so any formatting error is surfaced rather than inlined. Floats use Rust's shortest round-trip `Display`. A null becomes an empty string.
- **Rewrite `load_prediction_id_values`** to use `load_column_text` and keep only the row-count check.
- **Delete** `format_id_number` and the now-unused `load_dataset_projected` CLI wrapper.

SauersML:
Fixes #4221

## Problem

`load_prediction_id_values` sent the `--id-column` values through the numeric schema inference and then re-rendered them from `f64`. That corrupted IDs in four ways:

- Zero-padded IDs lost their leading zeros: `00123` became `123`.
- Int64 keys above 2^53 were rewritten to a neighbouring key: `9007199254740993` became `9007199254740992`.
- `1.10` became `1.1`.
- A literal `NA` ID became NaN, which failed the whole predict.

## Changes

- **`gam-data`:** new `load_column_text(path, column) -> Result<Vec<String>, DataError>`. It returns one column's cells as the text written in the file.
  - **CSV/TSV:** the trimmed cell. Header resolution and row-width checks behave the same as in the inferred loader.
  - **Parquet:**
    - Floats use the shortest round-trip `Display`, so `17.0` becomes `17`.
    - Every other type goes through arrow's `ArrayFormatter` with `try_to_string`, so a formatting failure is an error rather than inline `ERROR:` text.
    - A null becomes `""`.
- **`gam-cli`:**
  - `load_prediction_id_values` echoes that text and checks only the row count.
  - Removed as dead code: `format_id_number`, the `load_dataset_projected` wrapper, and the `load_dataset_auto_projected` and `ColumnKindTag` re-exports.
  - `cli_tests.rs` now imports `load_dataset_projected` straight from `gam::inference::data`.

## Tests

- `gam-data`: `load_column_text_returns_the_written_id_not_a_numeric_rerendering` covers CSV, TSV and parquet with Int64, Float64 and Utf8 columns, including nulls and a missing column.
- `gam-cli`: `prediction_id_column_echoes_the_written_ids` checks the four problem IDs above and the row-count refusal.

---


SauersML:

---

PR #4234, opened 2026-09-20
Derive the RRQR numerical-rank band from the pivoted-QR backward error (#4045)

### LINKED ISSUE #4045: gam-linalg faer_ndarray: pivoted-QR rank cutoff RRQR_RANK_ALPHA=100 is an arbitrary constant, disagrees with factor_singular_band's rank convention
## Defect

Every pivoted-QR rank decision in the workspace uses one hand-picked multiplier. `crates/gam-linalg/src/faer_ndarray.rs:43` (origin/main):

```rust
const RRQR_RANK_ALPHA: f64 = 100.0;
```

It is applied through `default_rrqr_rank_alpha()` in three places:
- `rrqr_nullspace_basis` (`RrqrRankCutoff::RelativeAlpha`, about line 4023)
- `rrqr_with_permutation` (about line 4102)
- `rrqr_from_gram_with_permutation` (about line 4206)

Each place uses the same formula:

```rust
let tol = rank_alpha * f64::EPSILON * (a.nrows().max(a.ncols()).max(1) as f64) * leading_diag;
let rank = (0..diag_len).filter(|&i| r[(i, i)].abs() > tol).count();
```

About 100 production call sites take this cutoff as the definition of "numerically rank-deficient", including:
- `gam-identifiability/src/audit.rs` (1039, 2204, 2747, 3319, 3333, 3490), `canonical.rs:751,787` and `families/compiler.rs:1049`
- `gam-linalg/src/pcg.rs:620`

SPEC: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code." Nothing in the code derives the `100`. The comment at `rrqr_with_permutation` says "The pivoted QR's backward error is `α·ε·max(m, p)·|R₁₁|`", but no Householder-QR bound has that form with `α = 100`.

### 1. Two rank conventions for the same question

The same crate already decides rank from a backward-error band, in `gam_linalg::roundoff::factor_singular_band` / `factor_rank_partition`:

```rust
pub fn factor_singular_band(rows: usize, cols: usize, sigma_max: f64) -> f64 {
    rows.max(cols) as f64 * f64::EPSILON * sigma_max
}
```

That is `α = 1` against `σ_max`. The RRQR path uses `α = 100` against `|R₀₀|`, and `|R₀₀| ≤ σ_max ≤ √n·|R₀₀|`. So a direction whose trailing pivot lies between `max(m,n)·ε·σ_max` and `100·max(m,n)·ε·|R₀₀|`:
- counts as resolved by `factor_rank_partition`;
- counts as null by every identifiability audit and by `rrqr_nullspace_basis`.

Two consumers of one matrix can therefore disagree about its rank by up to a factor of 100 in the cutoff. #2433 was this kind of disagreement between two builders.

### 2. The cutoff is not the shape of the QR backward error

Householder QR with column pivoting gives `A P + ΔA = Q̂ R̂`, with a column-wise bound `‖Δa_j‖₂ ≤ γ̃_{mn}‖a_j‖₂` (Higham, *ASNA* 2nd ed., Thm 19.4). Two things follow for the partition into `R₁₁ | R₂₂`:

- **What to compare.** The quantity to test is the trailing block `‖R̂₂₂‖`. That block is exactly the distance from `AP + ΔA` to a rank-`k` matrix with those leading columns. It should be compared with `‖ΔA‖`, which is bounded by `γ̃_{mn}‖A‖_F`.
- **What the code compares instead.** It tests each individual `|R_ii|` against a multiple of `|R₀₀|`. A trailing block made of several pivots, each just under the cutoff, can have norm above the backward error. The per-pivot test still drops the whole block, and the reverse can also happen.

So the rank verdict is not "indistinguishable from zero at working precision" in any derived sense. It depends on a factor of 100 that nobody tuned against anything.

### 3. Documentation drift

The doc comments on `RrqrRankCutoff::RelativeAlpha`, `rrqr_with_permutation` and `rrqr_from_gram_with_permutation` all state the cutoff as `rank_alpha · eps · max(m, n, 1) · max(|R[0,0]|, 1)`. The code uses `leading_diag` without the `max(·, 1)` floor. This matches the absolute-floor pattern of #3784, but in the opposite direction: the code is the scale-free one and the docs are wrong.

## Principled fix

Replace `RRQR_RANK_ALPHA` with a band derived from the QR's own backward error, shared with `roundoff`:

1. Add `gam_linalg::roundoff::pivoted_qr_backward_band(rows, cols, frobenius_norm)`, equal to `γ_{c·rows·cols}·‖A‖_F` with `c` taken from Thm 19.4's operation count via `gam_math::roundoff::accumulation_growth`.
2. Choose rank `k` as the smallest `k` whose trailing block norm `‖R̂₂₂‖_F = (Σ_{i≥k}‖R̂_{i,i:}‖²)^{1/2}` is at or below that band, instead of filtering individual diagonal entries.
3. Have `factor_rank_partition` and the RRQR paths share one "rank at working precision" definition, so SVD- and QR-based consumers cannot disagree.
4. Remove `default_rrqr_rank_alpha()` and the `rank_alpha` parameters. Callers that own an external convention keep `rrqr_nullspace_basis_with_cutoff`.
5. For `rrqr_from_gram_with_permutation`, form the band from the tall design's `m_rows` and `‖A‖_F = √trace(G)`, so its verdict still matches the tall path.

This is a cross-crate rank-convention change, and the identifiability suites pin rank verdicts, so it is filed as a well-specified issue rather than done as a blind constant swap.

SauersML:
Fixes #4045

Draft. Work in progress; code incoming.

## Plan

`RRQR_RANK_ALPHA = 100` sets the rank tolerance as `100·max(m,n)·ε·|R₀₀|`. Nothing derives the `100`. This PR replaces it with the backward-error band of the pivoted QR itself.

**Band.** Householder QR with column pivoting is backward stable: `(A+ΔA)Π = Q̂R̂` with `‖ΔA‖_F ≤ γ_{c·m·r}‖A‖_F` (Higham, *ASNA* Thm 19.4). Here `r = min(m,n)` is the number of reflectors, and the pivoting only permutes columns. This PR adds `roundoff::pivoted_qr_backward_band(rows, cols, frobenius)`. It uses the explicit per-reflector count, so the band's constant is an actual count, not a chosen knob.

**Rank rule.** The numerical rank is the smallest `k` whose trailing block `‖R̂₂₂‖_F = (Σ_{i≥k}‖R̂_{i,i:}‖²)^{1/2}` is at most that band. By Weyl's inequality, columns in that block cannot be told apart from zero at working precision. This replaces the per-pivot comparison against `α·ε·|R₀₀|`.

**Gram path.** `rrqr_from_gram_with_permutation` uses `m_rows` and `‖A‖_F = √trace(G)`.

**Cleanup.**
- Remove `RRQR_RANK_ALPHA`, `default_rrqr_rank_alpha()` and the `rank_alpha` parameters. Callers with an external convention keep `rrqr_nullspace_basis_with_cutoff`.
- Replace the generic borrowers of the constant (SVD, eigen and residual tolerances in other crates) with the existing derived bands (`factor_singular_band`, `symmetric_spectrum_rounding_band`, `accumulation_band`).
- Fix the doc drift: the docs describe a `max(|R00|,1)` floor that the code does not have.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 17cf9cadabe2.

PR #4238, opened 2026-09-20
posterior bands: route eta_bands_from_matrix_link through the validated draw reducer (#4233)

### LINKED ISSUE #4233: gam-predict posterior_bands: eta_bands_from_matrix_link duplicates matrix_bands without draw validation (NaN draws sort as Equal and yield silent garbage bands)
## Summary

`crates/gam-predict/src/posterior_bands.rs` has two reducers that turn posterior draw matrices into means and credible bands:

- `draw_bands_from_matrices` goes through `matrix_bands` and `validate_band_request`. It refuses a NaN draw, a ±inf linear-predictor draw and a −inf response draw.
- `eta_bands_from_matrix_link` is reached through the public pyfunction `gamfit._rust.posterior_eta_bands`. It is a hand-copied second reducer with **none** of that validation.

In the copy, a NaN draw goes straight into

```rust
column.sort_by(|a, b| a.partial_cmp(b).unwrap_or(Ordering::Equal));
eta_lower[j] = quantile_from_sorted(&column, alpha);
```

`unwrap_or(Ordering::Equal)` makes NaN compare as a tie with every value. That breaks the total order `sort_by` needs, so the "sorted" column is in an arbitrary order and the quantiles read from it are garbage. The mean is NaN. No error is raised, so the caller receives a band that looks valid for that row. A `−inf` or `+inf` eta draw is also accepted, although `draw_bands_from_matrices` rejects it.

The two reducers also disagree at `+inf` response draws. The copy calls `quantile_from_sorted` directly and skips the `posterior_quantile_from_sorted` handling of `+inf`. The same posterior therefore gets different bands depending on which entry point is used.

This is duplicated logic that should have a single source of truth, and `unwrap_or(Ordering::Equal)` on unvalidated input is silent error swallowing.

## Evidence (origin/main 65b4284851)

- `crates/gam-predict/src/posterior_bands.rs` L143-201, `eta_bands_from_matrix_link`: it re-implements level and zero-draw checks, means, sorting and quantiles inline. It never calls `validate_band_request`.
- In the same file, L71-96 (`matrix_bands`) and L124-141 (`draw_bands_from_matrices`) are the validated reducer. It is already used by the sibling `posterior_draw_bands_impl`.
- The doc comment at L143-145 is a broken fragment that begins mid-sentence: "/// semantics, but the response-scale push-through goes through a ...".
- Production caller: `crates/gam-pyffi/src/latent/reml_latent_fit_ffi.rs:3382` `posterior_eta_bands` calls `crates/gam-pyffi/src/manifold/manifold_and_posterior_ffi.rs:88` `posterior_eta_bands_impl`, which calls L121. The function is registered at `geometry_ffi.rs:4894` and listed in `gamfit/_rust.pyi:2010`.

## Fix

`eta_bands_from_matrix_link` only pushes each draw through the `LinkSelector` to build the response-draw matrix. It then returns `draw_bands_from_matrices(eta, mean, level)`, so one reducer owns validation, means and quantiles. A regression test checks two things:

- logit-link bands equal the matrix reducer on the transformed draws;
- a NaN, −inf or +inf eta draw is refused.

SauersML:
Fixes #4233

`eta_bands_from_matrix_link`, reached from the `posterior_eta_bands` pyfunction, was a hand-copied second reducer. It skipped `validate_band_request`, so a NaN draw went into `sort_by(partial_cmp.unwrap_or(Equal))`. That broke the sort's total order and returned garbage quantiles with a NaN mean, silently. A ±inf eta draw was also accepted, and +inf response draws skipped the `posterior_quantile_from_sorted` handling.

## Change (single file: `crates/gam-predict/src/posterior_bands.rs`)

- The function now only pushes each draw through the `LinkSelector` to build the response-draw matrix. It then returns `draw_bands_from_matrices`, so one reducer owns validation, means and quantiles. The signature is unchanged, so the FFI caller is untouched.
- The broken doc fragment is replaced with a complete doc comment.
- New test `eta_link_bands_share_the_validated_draw_reducer`:
  - logit-link bands equal the matrix reducer on the transformed draws;
  - a NaN, −inf or +inf eta draw is refused, as is level 1.0.

---


SauersML:

---

Issue #3015, opened 2026-09-19
Binomial location-scale refuses every seed when a noise_formula linear term shares a column with the location block's Duchon smooth (dense Hessian shape mismatch 11x11 vs 10x10)


SauersML:
## What fails

A binomial location-scale fit is refused at every startup seed when its `noise_formula` holds a linear term on a column that also enters a `duchon(...)` smooth in the location formula. The same column in a plain `s()`, or a noise column that the Duchon term doesn't use, fits fine.

Engine: gamfit 0.1.268 built from gam `40b5044e4f` (quick profile), and the same at `6fc5ad9c1c` (release-pypi). Run on 1 core on MSI.

```python
import numpy as np, gamfit
rng = np.random.default_rng(1); n = 2000
x1, x2, x3, x4 = (rng.normal(size=n) for _ in range(4))
eta = (-0.3 + 0.5 * x1 + np.sin(x2)) / np.exp(0.3 * x3)
y = (rng.uniform(size=n) < 0.5 * (1 + np.tanh(eta / 1.6))).astype(float)
data = {"y": y, "x1": x1, "x2": x2, "x3": x3, "x4": x4}
gamfit.fit(data, "y ~ x1 + duchon(x2, x3, centers=8)", family="binomial", link="probit", noise_formula="x3")   # FitSeedError
gamfit.fit(data, "y ~ x1 + duchon(x2, x3, centers=8)", family="binomial", link="probit", noise_formula="x4")   # OK
gamfit.fit(data, "y ~ x1 + s(x2)", family="binomial", link="probit", noise_formula="x3")                        # OK
gamfit.fit(data, "y ~ x1 + s(x2)", family="binomial", link="probit", noise_formula="x3 + x4")                   # OK
```

The refusal:

```
FitSeedError: exact two-block spatial optimization failed: custom-family optimization error in fit_custom_family
outer smoothing: ... no candidate seeds passed outer startup validation (custom family):
  seed 0 before solver start: recoverable objective evaluation failure: outer eval failed: Custom-family fit failed:
  inner solve refused this trial point: joint Newton inner prevalidation Hessian source: dense Hessian shape mismatch:
  got 11x11, expected 10x10
```

[...]

PR #4239, opened 2026-09-20
Fix the binomial log-sigma scale gauge structurally instead of by a unit-dependent ridge (#3879)

### LINKED ISSUE #3879: gamlss/builders: binomial location-scale log-sigma block carries a full-span identity ridge on raw coefficients (unit-dependent prior; the Gaussian/dispersion paths removed the same ridge in #1561)
## Where (origin/main)

`crates/gam-models/src/gamlss/builders.rs`

- 579-582, `build_binomial_threshold_and_scale_blocks`:
  ```rust
  let p_noise = raw_log_sigma_design.ncols();
  let mut log_sigma_penalty_matrices: Vec<PenaltyMatrix> =
      noise_design.penalties_as_penalty_matrix();
  log_sigma_penalty_matrices.push(PenaltyMatrix::Dense(identity_penalty(p_noise)));
  ```
- 451-475, `append_binomial_log_sigma_shrinkage_penalty_design`: the same `identity_penalty(p)` ridge, published as the result-design term `"log_sigma_shrinkage"`. It is called from 3309.
- 3647-3648 and 3783-3784: both binomial builders (non-wiggle and wiggle) return `noise_penalty_count = noise_design.penalties.len() + 1`, so the ridge gets its own REML smoothing coordinate.
- 553-556: the doc comment says the builder "mirrors [`build_gaussian_mean_and_scale_blocks`] ... and the same REML-selected full-span scale shrinkage penalty". The Gaussian builder has no such penalty. Lines 512-518 of that builder say:
  > the global log-σ intercept is likelihood-identified and must stay unpenalized: adding a Gaussian-only projector over the joint null space placed a data-scale-dependent prior on the overall σ level and introduced an extra smoothing coordinate absent from the formula (#1561).

  `dispersion_family.rs:3183-3188` removed the same construction for NB/Gamma/Beta/Tweedie:
  > The previous full-span `identity_penalty` ridge was a basis-dependent coefficient-space prior that double-penalized the range space and shrank the fitted dispersion surface toward its coordinate origin

## Why it is wrong

`I_p` is a ridge on the raw coefficients of every log-σ column: the intercept, every parametric column, and the range and null spaces of every smooth.

1. **Parametric columns get a prior that depends on their units.** For a noise formula term `x`, the ridge charges λ·γ² on the slope. Rescale `x → 1000·x` and the same function `γ·x` costs λ·(γ/1000)², so the fitted σ(x) changes with the covariate's units. That is a coefficient-metric penalty, not a penalty on the function.
2. **Smooths are penalized twice.** Each smooth already carries its formula-native penalties, and with the default `double_penalty=true` it also carries a function-metric null-space penalty. The identity adds a second, basis-dependent shrinkage of the range space. That is exactly the over-shrinkage #1561 removed from the Gaussian path.
3. **It adds a REML coordinate the formula does not have.** The criterion gets one extra smoothing parameter per binomial location-scale fit.

SPEC.md: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code." The coefficient metric of `I_p` is exactly that kind of arbitrary choice, since it depends on how the basis is scaled.

## Why it is there (and why that is not a reason to keep it)

Binomial location-scale uses `q = −η_t·e^{−η_σ}`. As `binomial/location_scale.rs:2046-2050, 2080-2092` and `builders.rs:413-418` document, this map has an exact likelihood gauge whenever the log-σ design has an intercept: `(β_t, b_0) → (c·β_t, b_0 + log c)` leaves `q` unchanged for every row.

The threshold penalty is not invariant under that gauge; it scales as `c²·β_tᵀS_tβ_t`. Without some pin, the penalized objective therefore decreases along `c → 0`, which drives η_t → 0 and σ → 0 together. The ridge's `(b_0)²` term is what currently stops that. It is a soft, REML-weighted pin on a direction the likelihood does not see.

That direction is also redundant with λ_t. Any `(c, λ_t)` gives the same fitted `q` and the same threshold penalty as `(1, λ_t·c²)`. So the ridge's smoothing parameter and the log-σ level together duplicate one coordinate λ_t already owns. Pinning the gauge does not require shrinking every other log-σ coefficient.

## Principled fix

1. Fix the gauge structurally, as heteroskedastic probit/logit normally does: σ equals 1 at the reference, with no free log-σ intercept. Build the binomial location-scale log-σ block without its intercept column, which `TermCollectionDesign::intercept_range` identifies. Equivalently, constrain `b_0 = 0` in the block's column layout. Do this in the fit, the saved model and the predict design.
2. Delete the `identity_penalty(p_noise)` push, `append_binomial_log_sigma_shrinkage_penalty_design`, and the `+ 1` in both binomial `noise_penalty_count`s. The log-σ block then carries only its formula-native penalties, like the Gaussian and dispersion paths.
3. Once no gauge is left in the joint Hessian, recheck whether `pseudo_logdet_mode = HardPseudo` (`binomial/location_scale.rs:2093-2095`) and its gauge commentary are still needed. The same applies to the Jeffreys-span note at 2046-2052.
4. Fix the doc at `builders.rs:553-556`.
5. Test: a binomial location-scale fit with noise `~ x` should give the same fitted probabilities and the same σ(x) function (slope scaled by 1/1000) when `x` is replaced by `1000·x`. On origin/main it will not, because the ridge's prior on the slope changes with the units.

This touches the design layout for fit, save and predict, so it is larger than a local patch. I am filing it as a specified issue rather than claiming it.

SauersML:
Fixes #3879

## Root cause
The binomial location-scale likelihood depends on the threshold and log-σ only through q = −η_t·e^{−η_σ}, so (β_t, b_0) → (c·β_t, b_0 + ln c) is an exact gauge. The log-σ level is therefore not a parameter. Until now the gauge was closed by a full-span identity ridge on every log-σ coefficient. That ridge's strength depends on the units of each log-σ covariate, so refitting with `x` and with `1000·x` gave different models.

## Fix
The gauge is now pinned where heteroskedastic probit/logit pins it. The log-σ design carries no intercept (`binomial_log_sigma_gauge_fixed_spec`, `ModelLevel::NoIntercept`), so σ = 1 at the covariate origin.
- The level lives in the spec, so the saved model and every prediction rebuild the same design.
- The only log-σ penalty left is the formula-native linear-term penalty, normalized by the column's mean square. A rescale of a scale covariate is therefore an exact reparametrization.
- Removed `append_binomial_log_sigma_shrinkage_penalty_design`, the post-solve augmentation and the `+1` in the noise penalty count.
- `build_binomial_threshold_and_scale_blocks` refuses a log-σ design that carries an intercept.
- Validation refuses an intercept-only log-σ (`noise_formula="1"`). With its level fixed there is nothing left to fit, and it is the plain binomial model. It also still refuses smooth and random-effect log-σ terms, which identify only the composite q.

## Breaking changes
- Binomial `noise_formula="1"` is refused.
- The log-σ coefficients no longer include a level.
- σ = 1 sits at the covariate origin. A rescale of a scale covariate is an exact reparametrization, but a shift is not, because it moves the reference point. This is recorded in the CHANGELOG and in `docs/location-scale.md`.

## Tests
- gamlss unit tests:
  - the builder refuses an intercept-bearing noise design;
  - the builder's layout;
  - the intercept-only refusal.
- The #3015 slope pin, updated to the new layout (a single slope column).
- A new pin, `the_binomial_scale_fit_does_not_depend_on_the_scale_covariates_units_3879`: fitting with `x3` and with `1000·x3` must give the same slope up to the rescale and the same threshold, both within 1e-2 posterior sd.
- `tests/quality/families/quality_vs_gamlss_binomial_location_scale.rs`: the smooth-log-σ refusal pin now asserts the new message ("log_sigma must be a parametric-linear scale").
- The Python tests for 2358 and the location-scale deviance test, updated for the refusal.

## Validation
main is merged into the branch (this brings in #3562's prevalence-seeded warm start). The focused-rust-proof dispatches run on the merged head:
- `BF-30-3879-4`: gam-models lib tests, filters `binomial_location_scale 3015 3879`.
- `BF-30-3879-5`: gam integration tests, binary `quality`, filter `gam_binomial_location_scale`.
- `BF-30-3879-6`: gam integration tests, binary `glm`, filter `2937`.


SauersML:
BF-75 conflict resolver: this branch conflicts with main (f5930beca8), but its author was active in the last 60 minutes, so I am leaving it to the author. Conflicting file: `CHANGELOG.md`.

---

SauersML:
Taking over from BF-30 (no activity since 10:50Z). The branch conflicts with main only in CHANGELOG.md. I am merging origin/main in (both Unreleased entries kept) and dispatching one release-profile lib proof on the #3879/#3015 tests. The three old-head runs BF-30-3879-4/5/6 are moot after the merge and will be cancelled.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, da00d12d193d.

PR #4246, opened 2026-09-20
gamlss/binomial: read the expected q-information from the stable Fisher-weight jet, not mu*(1-mu)

### LINKED ISSUE #4243: gamlss/binomial: expected q-information forms μ(1−μ) from the rounded mean, so upper-tail rows lose precision and drop to 0 (label-swap asymmetry)
## Where (origin/main eb3fe80912)

`crates/gam-models/src/gamlss/binomial/kernel.rs:232`, `binomial_expected_q_information_derivatives`:
```rust
    if weight == 0.0
        ...
        || mu <= 0.0
        || mu >= 1.0
        || d1 == 0.0
    {
        return (0.0, 0.0, 0.0);
    }
    let var = mu * (1.0 - mu);
    ...
    let var1 = d1 * (1.0 - 2.0 * mu);
    let f = weight * d1 * d1 / var;
```
Callers:
- the binomial location-scale expected information, its first and second directional derivatives, and its contracted trace (`binomial/location_scale.rs:239, 321, 410, 515`). These are what `joint_jeffreys_information_with_specs` returns, so they feed the Jeffreys/Firth term;
- the location-scale wiggle (`binomial/wiggle.rs:203, 471, 546`);
- the mean-wiggle working weights (`gaussian/binomial_mean_wiggle.rs:785`).

## Why it is wrong

The helper only receives μ, so it has to form the survival probability as `1.0 - mu`. That loses precision as μ approaches 1 and becomes zero once μ rounds to 1. The lower tail does not have this problem: μ keeps full relative precision down to about 1e-308, and so does `d1 = μ'`.

For the symmetric links the model is invariant under the label swap `(y, q) → (1−y, −q)`, but the helper breaks that symmetry. At the mirror point −q it returns the exact information. At +q it returns an increasingly noisy value, and then 0.

The tables give the ratio of the helper's `f` to the true `f = φ²/(Φ(1−Φ))`, with an mpmath reference (for logit the true `f` is `μ(1−μ)`):

| probit q | −9 | −8 | 7 | 7.5 | 8 | 8.2 | 8.3 | 9 |
|---|---|---|---|---|---|---|---|---|
| f_code / f_true | 1.0000 | 1.0000 | 0.99996 | 1.0014 | 0.934 | 1.083 | **0** | **0** |

| logit q | −40 | −36 | 30 | 35 | 36 | 36.5 | 37 |
|---|---|---|---|---|---|---|---|
| f_code / f_true | 1.0 | 1.0 | 1.001 | 0.947 | 1.045 | 0.634 | **0** |

`f1` and `f2` divide by `var²` and `var³`, so they are noisier still. For cloglog, `1 − μ = exp(−e^q)` is lost past q ≈ 3.6, while the lower tail `μ ≈ e^q` stays exact.

At probit q = 9 the true information is about 1e-17 times the weight. That is small, but it is not zero, and it is no smaller than lower-tail values the helper keeps exactly.

The Jeffreys term is where this matters. Under (quasi-)separation, the log-determinant of the information is built from exactly these tail rows. So the Firth-penalized estimate responds differently to y = 1 rows fitted high than to y = 0 rows fitted low. The penalty is also non-smooth at the q ≈ 8.3 cutoff: each row's contribution drops from a finite value to 0 while `log_likelihood` still counts that row.

#4015 describes the downstream score drop in the mean-wiggle working set. This issue is the upstream cause in the shared helper. It also affects the Jeffreys information of the location-scale family, which neither #4015 nor PR #4026 changes.

SPEC: "In general, do not paper over solver issues." Setting a representable information row to zero because the survival was computed by subtraction is exactly that.

## Principled fix

A stable single source for the standard links already exists: `gam_solve::mixture_link::fisher_weight_jet5(link, eta)`. It returns `(W, W', W'', W''', W'''')` with `W = f²/(F·S)`, pairing the density with the small-tail complement. Its own tests assert `w > 0` at `η = ±12`, and it returns 0 only at ±40, where the true value underflows.

`binomial_expected_q_information_derivatives` should take `q` and the `InverseLink`. For `Standard(Logit | Probit | CLogLog | LogLog | Cauchit)` it should return `weight · (W, W', W'')` from that jet, so the subtraction `1 − μ` disappears entirely for those links. Links without an analytic complement (mixture, SAS, beta-logistic) keep the μ-jet body. Fixing those is a separate follow-up.

Regression tests:
- Label-swap invariance: `f(q) = f(−q)`, `f′(q) = −f′(−q)` and `f″` even, to 1e-12 relative, for probit, logit and cauchit at |q| up to 12, 37 and 1e8 respectively.
- mpmath references at probit q = 9 and logit q = 37, both past the point where μ rounds to 1.

Separate note, not changed by this fix: the generic-link observed path has the same asymmetry. For links other than probit, logit and cloglog, `binomial_neglog_q_derivatives_dispatch` (`binomial_q_derivs.rs:556`) calls `binomial_neglog_q_derivatives_from_jet`. That function uses `1.0 / (1.0 - mu)` and returns 0 once `binomial_mu_is_interior(mu)` fails. The comment at `binomial_q_derivs.rs:125` justifies this by saying `d1 = μ'` and its successors underflow at least as fast as μ reaches the boundary. That holds at μ → 0 but is false at μ → 1: probit μ is exactly 1 at q = 8.3 while d1 ≈ 1e-15.

SauersML:
`binomial_expected_q_information_derivatives` formed the Bernoulli variance as mu*(1 - mu) from the reported mean. Toward mu = 1 the survival probability cancels, and once mu rounds to 1 the helper returns the row's information as exactly 0. That happens at probit q >~ 8.3, logit q >~ 36.7 and cloglog q >~ 3.6, while the mirror lower-tail row keeps full precision. This breaks the label-swap symmetry (y, q) -> (1 - y, -q) of the symmetric links. Probit at q = 8 was already 7% off.

The helper feeds three places:
- the location-scale Jeffreys information and its directional derivatives;
- the location-scale wiggle expected rows;
- the mean-wiggle Fisher working weight.

For the Bernoulli standard links, the helper now reads (W, W', W'') from `gam_solve::mixture_link::fisher_weight_jet5`. That is the existing single source, and it pairs the density with the small-tail complement. `fisher_weight_jet5` is now `pub`. Other links keep the mu-jet form, which is also what `fisher_weight_jet5_for_inverse_link` uses for them.

New tests check that the information is even under the label swap in both tails. They also compare against 60-digit references at points past where mu rounds to 1.

I merged main into the branch to resolve an append-only conflict in the `kernel.rs` test modules; both modules are kept.

Fixes #4243

---


SauersML:

---

PR #4258, opened 2026-09-20
gam-sae manifold/seed_routing: centre residual seed rows on their mean

### LINKED ISSUE #4257: gam-sae manifold/seed_routing: residual routing seed centres rows below the scale floor on the floor, not the row mean, so an exactly-fitted row opens every OBB gate at +gain
## Where

`crates/gam-sae/src/manifold/seed_routing.rs:136-146` on origin/main `65b4284851`, inside `sae_residual_seed_logits`. This is the cold routing seed for every multi-atom softmax and ordered Beta--Bernoulli fit. It is reached from `build_sae_minimal_seed` (minimal_seed.rs:266) and from each round of `sae_refine_routing_seed` (seed_routing.rs:964).

```rust
let floor = (global_mean * 1.0e-6).max(1.0e-12);
for row in 0..n_obs {
    let mut row_mean = 0.0_f64;
    for k in 0..k_atoms {
        row_mean += resid[[row, k]];
    }
    row_mean = (row_mean / k_atoms as f64).max(floor);
    for k in 0..k_atoms {
        logits[[row, k]] = -gain * (resid[[row, k]] - row_mean) / row_mean;
    }
}
```

## Why it is wrong

The function's own contract (lines 113-128) says the logits are **mean-centred**: "a row whose atoms all explain it equally well lands at exactly zero". That centring is the reason the seed is safe for ordered Beta--Bernoulli: "Centring restores `logit=0 ⇒ gate=0.5` on ties". The OBB map `sigmoid(logit/τ)` is not translation-invariant, so the centre matters.

The floor is applied to `row_mean` **before** it is used as the centre. Any row whose true mean residual is below `floor` is therefore centred on `floor`, not on its mean. Its logits become `-gain·(r_k − floor)/floor`.

When every `r_k ≪ floor`, each logit is ≈ `+gain`. With gain 4 that is `sigmoid(4/τ)` on every atom. Concretely:

- **Exactly fitted rows.** If each atom's column space contains the row, as with a clean planted manifold, every residual is at rounding level. Every OBB gate then opens at `sigmoid(gain/τ)` instead of the neutral 0.5.
- **Zero rows.** All residuals are exactly 0, so `global_mean = 0` and `floor = 1e-12`. Every logit becomes exactly `gain`, when the documented value is exactly 0.
- **Mixed rows below the floor.** The logits do not sum to zero, so the seed biases every gate open for those rows.

Softmax is not affected, since a common shift cancels. OBB receives an uninvited "all gates on" seed at exactly the rows the data say nothing about.

## Fix

Keep the floor only as the **divisor**, which is its stated job ("keep near-zero-energy rows well posed"). Always centre on the true row mean:

```rust
row_mean /= k_atoms as f64;
let row_scale = row_mean.max(floor);
logits[[row, k]] = -gain * (resid[[row, k]] - row_mean) / row_scale;
```

For rows above the floor this is bit-identical. For rows below it, the logits are now centred and of order `(r_k − mean)/floor`, which is rounding-level for tied rows.

Regression test `sae_residual_seed_logits_centre_rows_below_the_scale_floor` covers two cases, each asserting every logit is ≤ 1e-10 (on main they are ≈ 4):

- two atoms that both contain the target exactly;
- an all-zero target.

Separately noted, not changed here: the floor constants `1e-6` and `1e-12` and `RESIDUAL_SEED_GAIN = 4.0` are underived. The gain is also duplicated as `SAE_RESIDUAL_SEED_GAIN` in seed_routing.rs:908 and `RESIDUAL_SEED_GAIN` in minimal_seed.rs:265. The bug above is independent of those values.

SauersML:
Fixes #4257

`sae_residual_seed_logits` floored the row mean before using it as both the centre and the divisor. A row whose mean residual fell below the floor was centred on the floor, so every logit came out near `+gain`. That covers exactly fitted rows and all-zero rows, and on those rows every ordered Beta--Bernoulli gate opened at `sigmoid(gain/tau)` instead of the documented neutral 0.5. The function's contract says the logits are mean-centred, so that ties land at exactly zero.

The floor now bounds only the divisor (`row_scale = row_mean.max(floor)`), and the centre is always the true row mean. Rows above the floor are unchanged bit for bit.

The new test `sae_residual_seed_logits_centre_rows_below_the_scale_floor` checks two cases, and on both every logit must be at most 1e-10 (on main they are about 4):
- two atoms that both contain the target exactly;
- an all-zero target.

---


SauersML:
BF-86 (oldest-first sweep, partition 1/8) reviewed this diff for #4257. It is sound against SPEC.md and only waits on its queued focused-rust-proof build. It is ready for the local merge train (MT-0, even PR). I am not opening a duplicate.

---

PR #4263, opened 2026-09-20
gam-linalg: refuse a non-finite diag(X M Xᵀ) instead of clamping it to zero (#4113)

### LINKED ISSUE #4113: gam-linalg matrix: quadratic_form_diag clamps with f64::max, so a NaN/−∞ diag(X M Xᵀ) is reported as an exact zero variance
## Defect

Each `DenseDesignOperator::quadratic_form_diag` implementation clamps its result with `f64::max`:

```rust
// crates/gam-linalg/src/matrix/mod.rs (origin/main 91214e47)
.par_for_each(|o, xr, xmr| *o = xr.dot(&xmr).max(0.0));          // default impl, :1308
slot[0] = acc.max(0.0);                                            // DenseDesignMatrix, :2001 / :2007
.map(|g| g.map(|g| middle[[g, g]].max(0.0)).unwrap_or(0.0))        // RandomEffectOperator, :2564
*v = v.max(0.0);                                                   // BlockDesignOperator, :3243
result[i] = (result[i] - 2.0 * x_amd[i] + dtmd).max(0.0);          // ConditionedDesign, :3887
out[i] = acc.max(0.0);                                             // DesignMatrix::Sparse, :4642
// crates/gam-linalg/src/matrix/kronecker.rs :433, :722 (tensor-product / rowwise-Kronecker)
```

The comments say this is meant to "clamp tiny-negative fp drift on diag(X M Xᵀ) when M is a PSD covariance". But `f64::max` returns whichever operand is not `NaN`, so `NaN.max(0.0) == 0.0`. It also maps `-∞` to `0`. A `NaN` or `±∞` in the design or in `M`, or an overflowed product, therefore comes back as **`Ok` with an exact zero variance**. The caller has no signal that anything went wrong.

The production caller is `gam-models/src/fit_orchestration/entry.rs:2973`. There, `quadratic_form_diag(&scale_block)` produces the log-σ posterior variance that the joint-expectile `c_τ` integrates over. The comment just above that call says the covariance must "never [be] read as zero posterior variance (the plug-in σ)". With a `NaN` in the scale block, that is exactly what happens: the plug-in σ is used silently.

## SPEC

> In general, do not paper over solver issues.

## Fix

Route every implementation through one helper, `finite_quadratic_form_diag(out, operator)`. For each entry it:

1. refuses a non-finite entry with a typed `Err` that names the operator and the row, and
2. then clamps negative rounding drift with `if v < 0.0 { 0.0 } else { v }`, which keeps a `NaN` (`clamp_psd_rounding`).

For finite input the result is bit-for-bit what it was before. The eight copies of the clamp become one.

Tests:

- The helper keeps `NaN` and clamps `-1e-18`.
- `finite_quadratic_form_diag` refuses `NaN`, `+∞` and `-∞`.
- The dense, sparse and random-effect operators return `Err` for a `NaN` middle matrix.
- The same operators return the exact `diag(X M Xᵀ)` for a finite one.

SauersML:
Fixes #4113

## Root cause
Every `quadratic_form_diag` finished its diagonal with `f64::max(0.0)` to remove negative rounding drift. `f64::max` returns its non-NaN operand, so `NaN.max(0.0) == 0.0`, and `(-inf).max(0.0) == 0.0`. A NaN or overflowed design, middle matrix or product therefore came back as an exact **zero variance** with no error.

Eight sites had this:
- the `DenseDesignOperator` trait default;
- `DenseDesignMatrix::Materialized`, both the parallel and the fallback path;
- `RandomEffectOperator`;
- `BlockDesignOperator`;
- `ConditionedDesign`;
- the `DesignMatrix::Sparse` arm;
- `TensorProductDesignOperator` and `RowwiseKroneckerOperator` in `kronecker.rs`.

## Fix
All eight sites now return through one private helper, `certified_quadratic_form_diag(operator, diagonal)`. It rejects the first non-finite entry with an `Err` that names the operator and the row. Only after that check does it set negative drift to zero. The check comes first because the clamp would otherwise also turn `-inf` into 0.

**For a finite diagonal the result is bit-identical to the old clamp.** The only production caller, the log-σ posterior variance at `gam-models/src/fit_orchestration/entry.rs:2973`, already propagates the `Err` through `invariant(...)`.

## Tests (`gam-linalg` lib, `matrix::tests`)
- `certified_quadratic_form_diag_refuses_non_finite_and_clamps_drift`: finite entries are kept, `-1e-18` becomes 0, and NaN, +inf and -inf are each refused with the row index in the error.
- `quadratic_form_diag_refuses_non_finite_middle_on_dense_sparse_and_random_effect`: a NaN middle gives an `Err` on the dense, sparse and random-effect paths. On main each of these returned 0. The random-effect path also refuses a `-inf` variance.
- `quadratic_form_diag_is_exact_for_finite_middle`: on a dyadic fixture, `diag(X M Xᵀ) = [8, 16, 18.5]` exactly on the dense and sparse paths, and `[2, 0, 1]` for the random effect.

Validation: focused-rust-proof, key `BF-16-4113-1`.

---


SauersML:
Taken over by RF-1113-1; driving to merge.

---

SauersML:
BF-16 is yielding this PR to RF-1113-1. The proof I dispatched on head `84c2cfd199` is run **35505917241** (key BF-16-4113-1). It is still queued, with about 560 runs ahead of it. I will not merge this PR or close #4113 myself.

---

PR #4272, opened 2026-09-20
Closed-form Hadamard finite part for the hybrid Duchon self-pair; delete the ε diagonal fallback (#3545)

### LINKED ISSUE #3545: crates/gam-terms/src/basis: pure-Duchon fractional-s closed-form operator penalty evaluates its self-pair at r = 1e-6·median-lag instead of the exact zero limit

SauersML:
Fixes #3545 (item 3). Draft while local verification runs; full derivation follows in the next body update.

---


SauersML:
**Verify-and-land review: NEEDS-AUTHOR (draft, not merged).** Reviewed head 636a4288 by reading only. It isn't compiled yet.

Removing the ε self-pair fallback is principled. Every gated caller has an analytic bundle, and the gate implies μ>0, so Schoenberg applies. The mpmath, CPD and O(κ²) tests look strong.

**The blocker:** adding the full correction H = Σ E_K r^{2K} changes the small-κR chart from the S-only finite-part representative to PF. Three existing gam-terms lib tests in `tests/src_modules/smooths/basis_duchon_matern_jet_derivative_tests.rs` (included into the gam-terms lib) still pin S-only values:
- `test_isotropic_duchon_kappa_to_zero_ir_divergence_is_quotiented_by_finite_part` asserts lo→Riesz within 1e-3.
- `test_small_kappa_finite_part_chart_is_shared_by_value_radial_and_kappa_partials` uses a test-local S-only oracle with rel<1e-12.
- `test_even_log_riesz_small_kappa_uses_full_taylor_series` has an S-only expected value with rel<1e-11.

In the IR-divergent cells these tests use, H ~ κ^{d-2(a+b)} dominates S by a wide margin (≈1e10 at d=5, κ=0.01). I expect all three to fail against this head.

SPEC.md: *"XFAIL pattern on tests is never allowed. A failing test should always indicate problematic behavior."*

**Principled fix:** rewrite those three tests for the PF semantics this PR adopts. Where the two representatives legitimately differ only by a polynomial, compare them on the polynomial null-space quotient. Then finish the derivation in the PR body and mark it ready.

To check locally: `cargo test -p gam-terms --lib -- basis_duchon_matern_jet finite_part_self_pair_tests 3545 closed_form_operator`

---

SauersML:
Status: the implementation and tests are pushed at 636a428830. I'm still actively working on this PR. The local focused tests and `check --workspace --tests` are queued behind the shared build lock. I'll update the body with the full derivation and mark the PR ready once they pass.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, e4431562df63 (tree no-op: superseded by work already on main, see the merge message).

Issue #4266, opened 2026-09-20
latent joint REML: latent t, analytic-penalty rho and direct hypers search a hand-supplied box (max|t|+10, ±12)


SauersML:
## Where

In `crates/gam-models/src/fit_orchestration/drivers/spatial_optimization.rs` on main (f10039cc), `try_exact_joint_latent_coord_optimization` (`:6192-6204`) boxes the rest of the latent joint theta by hand:

```rust
    let mut lower = Array1::<f64>::from_elem(theta0.len(), -12.0);
    let mut upper = Array1::<f64>::from_elem(theta0.len(), 12.0);
    let latent_bound = latent.values.as_flat().iter().fold(1.0_f64, |acc, &v| acc.max(v.abs())) + 10.0;
    for axis in rho_dim..rho_dim + latent_flat_dim {
        lower[axis] = -latent_bound;
        upper[axis] = latent_bound;
    }
```

The theta layout is `[smoothing rho | latent flat t | analytic-penalty rho | direct hypers]`. Three parts of it get a hand box:

- **Latent coordinates `t`:** `±(max|t₀| + 10)`. The `+10` is an arbitrary slack in the latent's own units. It is not scale-invariant: rescaling the initial latent by `c` rescales the searchable region differently for `c > 1` than for `c < 1`.
- **Analytic-penalty rho:** `[-12, 12] ∩ registry.rho_domain_bounds()`. The registry already publishes the penalty's own ρ domain, and the `±12` further truncates it.
- **Direct hypers** (from `latent_coord_initial_direct_hypers`):
  - the AuxPrior / IsometryToReference REML-selected log-μ;
  - the per-axis ARD log-precisions, including user `init_log_precision` seeds that may lie outside `±12`, which leaves the seed infeasible;
  - for `AuxOutcome`, the behavioral head's regression coefficients. These are not log-strengths at all, so a `±12` box on a coefficient has no meaning.

SPEC: "Hand-supplied search boxes or bounds are never allowed."

## Why not a one-line fix

`OuterProblem` requires finite bounds (`run.rs:6917`, `:7522`), so these coordinates cannot simply be left unbounded. Each class needs a derived domain:

- the representable log-strength / precision domain for log-μ and ARD log-precisions (as `gam_solve::estimate::rho_domain` provides for ρ);
- the registry domain alone for analytic-penalty rho;
- a data-derived identifiability domain for `t` and for the head coefficients.

The smoothing-rho block of the same box is filed separately, with a small fix.

[...]

PR #4274, opened 2026-09-20
latent joint REML: search smoothing rho on the #2812 resolvability domain

### LINKED ISSUE #4265: latent joint REML: smoothing rho searches a hand-supplied ±12 box instead of the #2812 resolvability domain
## Where

In `crates/gam-models/src/fit_orchestration/drivers/spatial_optimization.rs` on main (f10039cc), the function `try_exact_joint_latent_coord_optimization` (`:6149`) builds its outer box like this:

```rust
    let mut lower = Array1::<f64>::from_elem(theta0.len(), -12.0);
    let mut upper = Array1::<f64>::from_elem(theta0.len(), 12.0);
```

This box goes to `exact_joint_outer_problem(&theta0, &lower, &upper, ..)` (`:6434`), and from there into `OuterProblem::with_bounds`. So the smoothing coordinates `theta[..rho_dim]` of the latent joint REML search sit in a hard `[-12, 12]` box.

## Why it is wrong

- **SPEC:** "Hand-supplied search boxes or bounds are never allowed."
- **#2812** already replaced this exact `±12` prior box on every other exact-joint route. It uses `joint_rho_resolvability_domain` (the doc comment at the top of the file). That comment measures the defect on the #2760 ladder: at `n = 16 000`, 5 of 5 incumbents were pasted onto the `−12` wall, with joint gradient `+1.484` against a stationarity bound of `1.030`. The same file's sibling route (`:643-645`) derives the domain from `best.design.design` and `best.design.penalties` and projects the seed into it. The latent route has the same `best: &FittedTermCollection` in hand and never does either.
- **The seed is not projected.** `theta0[..rho_dim]` is `best.fit.search_seed_log_lambdas()`. For an incumbent whose λ̂ sits past `e^{±12}`, that seed is infeasible against the box the latent search is handed. The search then starts from a point the incumbent's own domain allows but this box forbids.

## Fix

Give the smoothing block the derived domain, `joint_rho_resolvability_domain(&best.design.design, &best.design.penalties, rho_dim)`, and project the seed with `ExactJointHyperSetup::project_rho_seed`, exactly as the sibling route does.

The latent-coordinate and direct-hyper coordinates of the same box (`max|t| + 10` and `±12`) are a separate hand box. They have no derived domain yet, so they are filed separately.

---

SauersML:
Fixes #4265

`try_exact_joint_latent_coord_optimization` gave its smoothing coordinates `theta[..rho_dim]` the hand-supplied `[-12, 12]` box. #2812 removed that box from every other exact-joint route. The function also left the incumbent's seed `best.fit.search_seed_log_lambdas()` unprojected, so a λ̂ past `e^{±12}` started outside the box the search was given.

This change derives the domain with `joint_rho_resolvability_domain(&best.design.design, &best.design.penalties, rho_dim)` and projects the seed with `ExactJointHyperSetup::project_rho_seed`. That matches the iso/aniso sibling route (`spatial_optimization.rs:643-645`). The domain function already has unit tests (`the_domain_is_the_blocks_resolvability_interval_2812`, `the_2760_incumbents_are_interior_2812`).

The remaining hand box on the latent `t`, the analytic-penalty rho and the direct hypers (`max|t|+10`, `±12`) has no derived domain yet. It is tracked in #4266.

---


SauersML:
BF-86 (oldest-first sweep, partition 1/8) reviewed this diff for #4265. It is sound against SPEC.md: it replaces the hand-supplied `[-12, 12]` ρ box with the derived `joint_rho_resolvability_domain` and projects the seed. It only waits on its build. It is ready for the local merge train (MT-0, even PR).

Note for the train: PR #4335 (#4266) edits the same function (`try_exact_joint_latent_coord_optimization`). Whichever lands second needs a main merge.

---

Issue #4242, opened 2026-09-20
gam-math probability: small-shape upper incomplete gamma Q(a, x) is computed as 1 − P and keeps only absolute accuracy (rel. error ~ ε/Q, up to 4e-11 at a = 1e-4)


SauersML:
**Area:** `crates/gam-math/src/probability.rs`, `regularized_incomplete_gamma_pair`, series branch (`x < a + 1`). This is on origin/main and was not changed by the #4068 fix.

SPEC: "P-values must be calibrated and accurate."

### Defect

For `x < a + 1` the pair sums the power series for `P` and returns `Q = 1 − P`. That complement keeps absolute accuracy `ε`, so its relative error is about `ε·P/Q`.

- For `a ≥ 1`, `Q ≥ Q(a, a+1) ≥ Q(1, 2) = e^{-2}`. The loss is then at most a factor `e²`, which is harmless.
- For `a < 1`, `Q(a, x) ≈ a·E₁(x)` on this branch, which goes to 0 with `a`. The relative error of `Q` therefore grows like `ε/a`, without bound.

### Reproduction

I mirrored the main loop in Python and compared it against a 50-digit mpmath `gammainc(a, x, inf, regularized=True)`.

| a | x | Q (main) | Q (exact) | rel. err |
|---|---|---|---|---|
| 1e-2 | 0.909 | 2.585880246514516e-3 | 2.585880246514324e-3 | 7.4e-14 |
| 1e-3 | 0.99099 | 2.2295495494373263e-4 | 2.2295495494377223e-4 | 1.8e-13 |
| 1e-4 | 0.9901 | 2.2308495402190154e-5 | 2.2308495401268375e-5 | 4.1e-11 |

When Q is computed directly by the continued fraction at `x ≥ a + 1`, the same `a` gives about 1e-15 (for example `a = 1e-3`, `x = 7`: 1.8e-15).

### Who is affected

- Upper-tail probabilities of Gamma distributions with shape `< 1`.
- Anything that inverts `Q` at small shape.
- Chi-square with `ν ≥ 1` is **not** affected, because there `a = ν/2 ≥ 1/2` and the loss is at most about 10 ulp.

### Principled fix

Evaluate `Q` directly for `a < 1`, `x < a + 1`, instead of as a complement. Write

`Q(a, x) = 1 − x^a/Γ(a+1) + x^a/Γ(a+1) · (1 − Γ(a+1)·γ*(a,x))`,

or use the standard split

[...]

PR #4276, opened 2026-09-20
Incomplete gamma in gam-math: uncapped branches, certified Temme expansion, stable prefactor (#4068)

### LINKED ISSUE #4068: gam-inference probability: incomplete-gamma series/continued fraction silently truncate at 1000 terms, miscalibrating Gamma (shape >~ 1e5) and Poisson (mu >~ 1e6) predictive bands

SauersML:
Fixes #4068

## Defect

`regularized_incomplete_gamma_pair` (crates/gam-math/src/probability.rs) had two problems.

- It capped both the power series and the Lentz continued fraction at 1000 terms, and returned the truncated value without any signal.
  - Near `x = a` the series needs about `8.5·sqrt(a)` terms, so from `a ~ 1e4` it was cut short and `P(a, x)` was under-counted.
  - This reaches every Gamma quantile or CDF with shape above about 1e5, and every Poisson interval with mean above about 1e6, through `poisson_cdf_at(k+1, mu) = Q(k+1, mu)`.
- Its leading factor `exp(a ln x − x − lnΓ(a))` is the difference of three `O(a ln a)` numbers, so it loses about `log10(a ln a)` digits. That is about 8 digits at `a = 1e8`.

## Fix (no iteration cap anywhere)

**Stable prefactor.**

- Write `x^a e^{−x}/Γ(a) = sqrt(a/2π) · e^{−aφ} / Γ*(a)`, where `φ(λ) = λ − 1 − ln λ` with `λ = x/a`, and `Γ*(a) = Γ(a) / (sqrt(2π/a) (a/e)^a)`.
- `aφ` is taken as `−a·log1p_minus_x(μ)` with `μ = λ − 1` when `|μ| ≤ ½`, and otherwise from `(x − a) − a·ln(x/a)`.
- `ln Γ*` is the enveloping Stirling series `Σ B_{2k}/(2k(2k−1)a^{2k−1})`, truncated at its first term below one unit roundoff.
- Every quantity is O(1) relative to the result, so no digit cancels. The Halley density of `inverse_regularized_lower_gamma` uses the same factor.

**Temme uniform expansion (DLMF 8.12)** for `x ≈ a`, in O(1) work.

- Let `η = sign(μ)·sqrt(2φ)`. Then `Q = ½erfc(η sqrt(a/2)) + R` for `η ≥ 0`, and `P = ½erfc(−η sqrt(a/2)) − R` otherwise.
- `R = e^{−aφ}/sqrt(2πa) · Σ_k C_k(η) a^{−k}`.
- The Taylor coefficients of `C_k(η)` are built once, from exact recurrences:
  - invert `η²/2 = μ − ln(1+μ)` for `μ(η)`;
  - `C_0 = 1/μ − 1/η`;
  - `γ_k` of `1/Γ*`;
  - `c_{k,n} = (n+2)·c_{k−1,n+2} + γ_k·c_{0,n}`.
- The coefficients are checked against their exact rationals `−1/3`, `1/12`, `−2/135`, `−1/540` and `25/6048`.
- **Admission is a certificate, not a box.** The `C_k` are analytic in `|η| < 2 sqrt(π)` (the nearest singularities of `μ(η)` are at `η = 2 sqrt(π)·e^{±3iπ/4}`).
  - With `r = |η|/(2 sqrt(π))`, the N-term Taylor tail is bounded by the geometric tail of the last term.
  - The expansion is used only where `r^N ≤ ε` (N = 53, the mantissa width).
  - It is also used only where that tail bound, plus the first omitted `a^{−k}` term, is at most one unit roundoff of the returned tail.
  - Otherwise it declines and the convergent branches run.
- `K = 8` orders is a cost choice, not an accuracy knob: it certifies near `λ = 1` from `a ≈ 33`, where the series takes about 50 terms. A larger `a` certifies with fewer orders.

**Power series** (`x < a + 1`).

- The terms have ratio `x/(a+n)`, which decreases in `n`.
- After term `t_n`, the tail is at most `t_n · r/(1−r)` with `r = x/(a+n+1)`.
- The sum stops when that bound is at most one unit roundoff of the partial sum.

**Lentz continued fraction** (`x ≥ a + 1`).

- It stops at ulp stationarity `|Δ−1| ≤ ε`.
- It also stops when the product of its three-term recurrence's characteristic-root ratios reaches `ε²`. That product is the modulus of the truncation error's contraction.

**Invalid arguments** (NaN, `a ≤ 0`, `a = ∞`) now return NaN instead of looping, and `x ≤ 0` and `x = ∞` are exact.

**`chi_square_sf` reads the same pair.** It used statrs `gamma_ur`, which returns 0.02279 at `ν = 2e12`, `s = ν + 4e6`. The exact value is `Q(1e12, 1e12 + 2e6) = 0.0227501859391187`, so statrs is 0.2% off. `chi_square_sf` now returns `Q` from `regularized_incomplete_gamma_pair`, the pair `chi_square_quantile` inverts, so the p-value and the quantile agree at every `ν`. A test pins the value and the round trip.

## Validation

- **Reference tails.** 20 smaller-tail values from a 50-digit mpmath evaluation span every branch:
  - Temme, both signs;
  - series from `a = 100, x = 0.1` (`P = 9.7e−259`);
  - CF;
  - `a` up to `1e12`;
  - the issue's reproducers (`a = 1e5, x = 99700`; `a = 1e8, x = a + 1.5`).

  The test asserts relative error `≤ 16ε(1 + aφ)`, which is the propagated error of the `e^{−aφ}` exponent plus O(1) roundings, and `|P + Q − 1| ≤ ε`.
- **Prototype.** In the Python prototype of the same algorithm, all branches stayed within that bound over a random sweep. The worst ratio to the bound was 3.85.
- **Cross-check.** The prototype was also checked against the references LC-19 posted on #4085. The worst case was 5.5 ulp, at `a = 1e16`, `x = 1.00000002e16`, where `a + 1 == a`.
- **Poisson quantile.** The 97.5% Poisson quantile at `μ = 1e6` is exact: `Q(1001960, 1e6) = 0.974945… < 0.975 ≤ Q(1001961, 1e6) = 0.975004…`.

Focused proofs: `gam-math` lib `probability::tests`, and `gam-inference` lib probability/gamma/poisson/chi filters.

## Out of scope (pre-existing, filed as #4242)

For `a < 1` and `x < a + 1`, the code returns `Q` as `1 − P`. That form keeps only absolute accuracy, so the relative error is about `ε/Q`. It is 1.8e−13 at `a = 1e−3` and 4e−11 at `a = 1e−4`, and main has the same values. For `a ≥ 1` the loss is at most `e²·ε`, and chi-square with `ν ≥ 1` is unaffected. Fixing it needs a direct small-`a` `Q` evaluation (#4242) and is separate from the truncation defect here.

---


SauersML:
Closing as superseded. #4068 was fixed on main by PR #4085 (merge a935b7eff5), and the small-shape upper tail by 0391ddcbf6 (#4242). The one part of this PR that main does not have yet is `chi_square_sf` reading the crate's own gamma pair instead of statrs `gamma_ur`. statrs gives 0.02279 at ν = 2e12, s = ν + 4e6, and the exact value is 0.0227501859391187. That part moves to a small follow-up PR based on current main.

---

PR #4280, opened 2026-09-20
gam-identifiability audit: propagate a failed eigendecomposition in the cross-block residual Gram

### LINKED ISSUE #4268: gam-identifiability audit: cross-block residual Gram silently falls back to the bare sub-Gram on a failed eigendecomposition
## Defect

`crates/gam-identifiability/src/audit.rs::block_cross_residual_gram` builds the Schur complement `R_bb = G_bb − G_Abᵀ G_AA⁺ G_Ab`, which the channel-aware audit pivots to decide WHICH columns of a rank-deficient block to drop (gam#1590). When the eigendecomposition of `G_AA` fails, it returns the bare diagonal sub-Gram and says nothing:

```rust
// On eigendecomposition failure fall back
// to the bare diagonal sub-Gram (no spurious demotion).
let (evals, evecs) = match g_aa.eigh(Side::Lower) {
    Ok(pair) => pair,
    Err(_) => return g_bb,
};
```

Its only caller is `audit_identifiability_channel_aware` (line 2183), which already returns `Result<IdentifiabilityAudit, EstimationError>`.

After a failed factorisation, the column-drop decision runs on a different matrix from the one documented. It uses within-block pivots, not the jointly rank-faithful residual, so it can keep a column that is cross-channel redundant and drop one that carries a new direction. The audit then reports its drop set as if the joint residual had been used (`joint = true`), and no error or log line is emitted. Every other eigendecomposition in this file propagates `EstimationError::EigendecompositionFailed` (lines 2772, 3359).

SPEC: "In general, do not paper over solver issues."

## Fix (PR incoming)

- `block_cross_residual_gram` returns `Result<Array2<f64>, EstimationError>` and maps the failure with `.map_err(EstimationError::EigendecompositionFailed)?`.
- The caller propagates it with `?`.

SauersML:
Fixes #4268

Cross-block residual Gram construction now propagates eigendecomposition errors to the channel-aware audit. It no longer substitutes the bare within-block Gram while reporting a joint residual decision. Nonfinite joint geometry, decomposition results and residual output are refused explicitly.

Added an independently computed Schur-complement fixture and NaN/positive-infinity/negative-infinity controls, including the no-other-block path.

Validation: rebuilt the full changed gam-identifiability unit-test binary on MSI with #4361/#4079 staged; all 82 tests passed in 19.18 seconds. Production and full unit-test metadata passed with warnings denied against freshly rebuilt current-main dependency metadata. Tests ran on two CPUs with existing warm dependencies; no full workspace run is claimed.


SauersML:

---

PR #4284, opened 2026-09-20
Delete the uncertified cubic supremum refinement and refine_supremum (#4260)

### LINKED ISSUE #4260: hmc_io laplace diagnostic: cubic supremum 'power-iteration refinement' is a fixed 5-step/0.3-step ascent with no certificate, reachable only from tests
**Area:** `crates/gam-inference/src/hmc_io.rs`, `cubic_power_iteration_refinement` (origin/main, around lines 6509-6652), reached through `laplace_directional_cubic_diagnostic_on_eigenpairs(.., refine_supremum = true)` (around line 6205)

SPEC: "A fit object must only ever come from a converged optimization." / "No dead code."

### What the code does

Phase 2 of the directional-cubic Laplace diagnostic is documented as "sharpen[ing] the returned scalar `max_abs` toward the true supremum of |γ(u)| over the H-unit sphere". It works like this:

```rust
let max_probes = 8;
let max_iters = 5;
...
let step_size = 0.3;                        // fixed Riemannian ascent step
...
for seed in &seeds {
    let mut u = seed.clone();
    for _ in 0..max_iters { u = refine_step(&u); }   // 5 steps, no stationarity test
    let g = eval_gamma(&u);
    best = best.max(g);
}
best
```

This is a fixed budget of 5 fixed-size (0.3) projected-gradient steps from at most 8 seeds. Nothing checks it:
- no tangent-gradient stationarity test;
- no step acceptance (a 0.3 step can overshoot and lower |γ|, and only the running `max` hides that);
- no record of whether any seed reached a critical point.

The returned `max_abs` is the largest iterate value seen. It is a budget-dependent lower bound, not the supremum the doc and the `refine_supremum` name promise. Both constants (`5`, `0.3`) are unscaled: the step is not tied to the curvature of `γ` on the whitened sphere.

### Why it is also dead

The only production caller of the diagnostic, `block_quadrature_correction.rs:451`, passes `refine_supremum = false`. That call goes through `LaplaceMarginalCorrector::directional_cubic_diagnostic`, and there it uses `max_abs` only as a `> 0` guard. `refine_supremum = true` is passed only from `#[cfg(test)]` code:
- `hmc_io.rs` tests around 2496/2503/2604;
- the test corrector around 3546.

So the uncertified supremum search exists only to be exercised by its own tests, and the trait carries a `refine_supremum` parameter that no production path ever sets.

### Principled fix

Delete Phase 2 (`cubic_power_iteration_refinement`) and the `refine_supremum` parameter from:
- `laplace_directional_cubic_diagnostic_on_eigenpairs`;
- the `LaplaceMarginalCorrector::directional_cubic_diagnostic` contract in `gam-problem`;
- its implementors.

`max_abs` is then the certified eigen-direction maximum, which is what every production consumer already gets. The rotation-invariance test keeps testing Phase 1.

If a real supremum of |T[v,v,v]| on the H-sphere is needed later, solve it as a certified optimization instead:
- a projected Newton / trust region on the sphere;
- a tangent-gradient stationarity bound derived from the contraction's rounding band;
- refusal on non-convergence.

It should not come back as a fixed-step loop.

SauersML:
Fixes #4260

## Root cause
Phase 2 of `laplace_directional_cubic_diagnostic_on_eigenpairs` is `cubic_power_iteration_refinement`. It took a fixed 5 steps of fixed-size (0.3) projected-gradient ascent from at most 8 seeds, with no stationarity test and no step acceptance. The value it returned was a budget-dependent lower bound, not the supremum it claimed to be.

It was also dead code:
- The only production caller, `block_quadrature_correction.rs:451`, passes `refine_supremum = false`, and it uses `max_abs` only as a `> 0` guard.
- `true` was passed only from `#[cfg(test)]` code.

## Change (net -349 lines)
- Deleted Phase 2, `cubic_power_iteration_refinement` and its helper `directional_cubic_gradient`.
- Removed the `refine_supremum` parameter from all of its places:
  - the free function;
  - the `LaplaceMarginalCorrector::directional_cubic_diagnostic` contract in gam-problem;
  - `HmcIoLaplaceMarginalCorrector`;
  - the two hmc_io test correctors;
  - the two gam-solve test doubles;
  - the production call site.
- Deleted `directional_cubic_power_iteration_finds_larger_or_equal_skewness`, which tested only the refinement.
- `directional_cubic_contraction`, the single-direction version, was used only as the reference in `batched_directional_cubics_match_the_single_direction_contraction`. That test now forms `Σ_i c_i (x_iᵀ v)³` inline, so no production code stays alive just for a test.
- The doc now says that `max_abs` is the eigen-direction maximum. That is what every consumer already received.

**Production behaviour is unchanged.** The one production path always ran with `refine_supremum = false`.

## Validation (focused-rust-proof)
- `BF-16-4260-1`: gam-inference lib, the `directional_cubic`, `batched_directional_cubics` and order-search corrector tests.
- `BF-16-4260-2`: gam-solve `block_quadrature_units_784`.
- `BF-16-4260-3`: gam-solve `block_quadrature_firth_decline_784`.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, d6d9cd07fa57.

PR #4285, opened 2026-09-20
Binomial trial-count fits: observation band and replicates on the observed proportion

### LINKED ISSUE #4228: gam-predict observation band: trial-weighted binomial fit gets the Bernoulli {0,1} set ([0,1]) instead of a Binomial(m,p)/m predictive band
## Summary

A binomial fit with a trial-count weight column is the classic grouped-binomial GLM: the response is the proportion `y_i = k_i/m_i` and the weights are `m_i`. The fit supports it, but its analytic observation (prediction) interval is always the **Bernoulli** band, whose edges lie in `{0, 1}`. For a new row with `m* = 100` trials and `p = 0.5`, the 95% prediction interval for the observed proportion is about `[0.40, 0.60]`. gamfit reports `[0, 1]` for every row with `p ∈ (0.025, 0.975)`.

That band covers 100% of the time at every nominal level, so it is grossly conservative. SPEC: "P-values must be calibrated and accurate. Conservative is a bug just like anti-conservative."

## Evidence (origin/main b9758d35a0)

**The fit treats the weight as a trial count.**
- `crates/gam-solve/src/pirls/deviance.rs` ~L2514-2535: `binomial_log_coefficient_from_proportion(w, y)` is the continuous `ln C(w, wy)`, the Binomial(w, p) normalizer of a proportion response.
- `crates/gam-models/src/gamlss/binomial_q_derivs.rs` ~L143-145: "Per-row log-likelihood is represented in weighted-proportion form ... where `weight = m_i` and `y` is the observed proportion in [0,1]".

**Prediction resolves the weights and hands them to the band.**
- `crates/gam-pyffi/src/manifold/geometry_ffi.rs` ~L7163-7170 resolves `observation_prior_weights` from the prediction frame's weight column whenever the model was fitted with one.
- `crates/gam-predict/src/lib.rs` L3159-3168, in the standard full-uncertainty path, passes them to `family_observation_band(.., options.observation_prior_weights.as_ref())`.

**The Binomial arm ignores them.**
- `crates/gam-predict/src/lib.rs` L2621-2650:
  ```rust
  ResponseFamily::Binomial | ResponseFamily::RoystonParmar => {
      ...
      lower[i] = bernoulli_predictive_quantile(mean[i], p_lo);
      upper[i] = bernoulli_predictive_quantile(mean[i], p_hi);
  ```
  `bernoulli_predictive_quantile` (L2380) returns only `0.0` or `1.0`.
- The field doc at L1700 says "Only the Gaussian observation band consumes this; every other family encodes dispersion through its own precision parameter". For the binomial the trial count **is** the precision parameter: `Var(y_i | p) = p(1−p)/m_i`. The binomial has no other precision parameter.
- `family_predictive_variance` (L2369-2371) likewise returns `m·(1−m)`, the variance of a single trial, whatever `m_i` is.

**The generative sibling has the same gap.**
- `crates/gam-models/src/inference/generative.rs` L314: `ResponseFamily::Binomial => Ok(NoiseModel::Bernoulli)`. `sample_replicates` therefore draws single 0/1 trials for a grouped-binomial model too. The analytic band and the simulation agree only because both drop the trials.

## Consequence

Take a grouped-proportion fit, for example dose-response with `m_i = 50` animals per dose. Every `predict(..., interval=0.95)` observation band is `[0, 1]`, and `sample_replicates` returns 0/1 proportions. The model's own likelihood says a fresh observed proportion at `m*` trials has variance `p(1−p)/m*`, not `p(1−p)`.

## Fix

The fitted likelihood says a new row with trial weight `m*` is `Binomial(m*, p)/m*`.

1. **Band.** Build the band from the equal-tailed quantiles of that law, widened for posterior uncertainty in `p`. The Poisson arm already does the analogous thing: its predictive is the conjugate NB (Gamma-Poisson) rather than a continuous surrogate. Here the conjugate is the beta-binomial, moment-matched to `(E[p], Var(p))`. With `m* = 1` (no weight column) it reduces exactly to today's Bernoulli band.
2. **Variance.** `family_predictive_variance` for the binomial becomes `E[p(1−p)]/m* + Var(p)`.
3. **Generative path.** `NoiseModel::Bernoulli` gains the per-row trial count (a Binomial(m*, p)/m* draw), matching the Gaussian `σ/√w` scaling of #2025.
4. **Case weights versus trial counts.** The weight means two different things (see deviance.rs ~L2622-2626). On a 0/1 response, a fractional sample weight is a case weight, and a fresh observation stays a single Bernoulli. So the fit must record whether its weights are trial counts; a training response with any fractional proportion is the unambiguous signal. Only such a fit gets the Binomial(m*, p)/m* band. A non-integer predict-time `m*` on a trial-count fit has no Binomial law, so it is refused with the row named, never rounded.
5. **Docs.** Correct the doc at L1700.

Related but distinct: #3924, #3957, #4056, #4068.

SauersML:
Fixes #4228

A grouped binomial fit (y = k/m with prior weight m) used to predict the band and the replicates as if each row were a single Bernoulli trial. For a proportion observed from m trials, that meant a [0, 1] band and {0, 1} draws.

## Change
- `FitArtifacts.binomial_trial_counts` (`#[serde(default)]`) records that a fit read its weights as trial counts. A fit is marked this way when its family is Binomial and some training y lies in (0, 1). Saved models written before this change read `false`.
- Observation band: a new row's weight m* is its trial count, and the band is the BetaBinomial(m*, a, b) interval on K/m*, built from the Beta moment-matched to the mean's posterior (`binomial_proportion_interval`, an O(m) log-space pmf walk). m* = 1 gives the Bernoulli band; zero posterior variance gives the exact Binomial.
- Predictive variance: μ(1−μ)/m* + v(m*−1)/m*.
- Generative replicates draw Binomial(m*, p)/m*.
- A weight that is not an integer in [1, 2^53] is refused, and the error names the row.
- Case-weight fits (all y in {0,1}) keep the Bernoulli band.
- `is_binomial_trial_count` moves to gam-math so gam-models and gam-predict share one definition.

## Tests
- `binomial_proportion_interval` unit tests (exact).
- `a_trial_count_binomial_band_is_on_the_observed_proportion_4228` (gam-predict).
- `a_trial_count_binomial_replicates_the_observed_proportion_4228` (gam-models).

---
_Generated by [Claude Code](https://claude.ai/code/session_0187MfV4HM8xGpYsmTZ4JwWg)_


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 32d8e6196a2d.

PR #4286, opened 2026-09-20
Restrict penalty coordinates to the affine active face (#4170)

### LINKED ISSUE #4170: gam-problem penalty_coordinate: project_into_subspace maps a centered prior mean to zᵀμ, so a penalty with a prior mean under an active constraint face is centered at the projection of μ (reduced gradient off by λ_k zᵀS_k(I−zzᵀ)μ_k)
## Defect

`PenaltyCoordinate::project_into_subspace` (`crates/gam-problem/src/penalty_coordinate.rs:408`) reduces a penalty to the free subspace `β = z β_f` of an active linear-inequality face. It is used by `finish_assembly` in `crates/gam-solve/src/reml/objective.rs:1057-1085` whenever `active_constraint_free_basis` returns a basis.

For a centered coordinate, the reduced mean is taken as `zᵀμ`:

```rust
Self::DenseRootCentered { root, prior_mean } => {
    Self::from_dense_root_with_mean(root.dot(z), z.t().dot(prior_mean))
}
// BlockRootCentered: root·z_block, mean z_blockᵀ μ
```

This builds `‖R z (β_f − zᵀμ)‖² = ‖R(zβ_f − zzᵀμ)‖²`. The penalty the fit carries is `‖R(zβ_f − μ)‖²`. The two agree only when `μ ∈ span(z)`.

When `μ` has a component along the active constraint normals, and `S = RᵀR` couples that component to the free directions, the two differ in two ways:

- The reduced gradient is off by `λ_k zᵀ S_k (I − zzᵀ) μ_k`.
- Every outer derivative channel reading these coordinates (`∂/∂ρ_k` of the penalty quadratic) is off by a `λ_k`-scaled term.

The doc comment ("the representation of `μ_k` in the free subspace") treats the orthogonal projection as the right reduction. That holds for a vector in the subspace, but not for the center of a quadratic whose metric `S` is not the identity on the complement.

## Reachability

- Centered coordinates are produced for non-zero prior means:
  - `PenaltySpec::DenseWithMean` from coefficient-group priors (`gam-terms/src/smooth/penalty_priors.rs:468-500`);
  - the transformed canonicals (`construction.rs:2914`) that `finish_assembly` projects.
- The free basis is non-empty whenever a shape or linear constraint is active at the mode.
- So the defect needs a term with a prior mean and an active constraint face in the same fit. It has not been reproduced end-to-end.
- The only unit test, `reml_outer_engine/tests.rs:1441`, uses a non-centered coordinate.

## Fix direction

The exact reduced penalty is `‖R z β_f − Rμ‖²`: a root `Rz` with a root-space offset `c = Rμ`, which is not of the form `‖Rz(β_f − μ_f)‖²` unless `Rμ ∈ range(Rz)`.

- Taking `μ_f = (Rz)⁺Rμ` fixes the gradient exactly.
- It leaves a constant `½λ_k‖(I − P_{Rz})Rμ‖²` out of the value, and that constant's `ρ` derivative is not zero.

A correct reduction therefore needs a coordinate variant that carries the root-space offset, or it needs to refuse the combination. It is too large for a drive-by fix.

SauersML:
Fixes #4170

An active inequality set `A_a beta = b_a` puts the inner solve on the affine face `beta = z beta_f + c`, where `c = (I - z z^T) beta_hat` is fixed by the active rows.

**The bug.** `project_into_subspace` reduced a penalty to `R z` and mapped its prior mean to `z^T mu`. That dropped `c` and recentred the penalty at `z z^T mu`, with two consequences:
- The reduced score was off by `lambda z^T S (I - z z^T) mu`.
- `q_k` lost the energy of the pinned components, and `d/d rho_k = 1/2 lambda_k q_k` reads `q_k`. For example, a ridge on a coefficient held at a bound `b != 0` lost `b^2`.

**The fix.** `restrict_to_face(z, face_point)` replaces `project_into_subspace`:
- It builds the exact reduction `||(R z) beta_f - R(mu - c)||^2` through a new `DenseRootTarget` variant that carries the offset in root space. That target need not lie in `range(R z)`, so no free-chart mean can express it.
- The reduced quadratic now equals the full penalty at `z beta_f + c`, and the reduced shifted score equals `z^T S (beta - mu)`.
- `finish_assembly` passes the constrained mode as the face point.
- `project_out_null_directions` refuses a face-restricted coordinate. The null split runs on the full chart before the face restriction.

**Tests** (gam-problem `penalty_coordinate`):
- the quadratic and the score agree on an offset face for the dense, dense-centred, block and block-centred shapes;
- a ridge on a coefficient pinned at a bound keeps its energy;
- the old reduction's score gap matches `z^T S (I - z z^T) mu`;
- restricting onto a sub-face composes correctly;
- `project_out_null_directions` refuses a face-restricted coordinate.

---


SauersML:

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, ff43572d6fb0.

Issue #2712, opened 2026-08-01
Reconstruct the per-row DEFLATED inverse block from probes, so the from-probes theta-adjoint stops hard-refusing deflated rows


SauersML:
Split out of #2333 so its precondition is stated rather than rediscovered. Read at `BASE=dc01f60d6b18f823d6e133fe25e618d55be32fe3`; nothing measured.

## The refusal, verbatim

`SaeManifoldTerm::logdet_theta_adjoint_from_probes` (`crates/gam-sae/src/manifold/construction_quasi_laplace.rs`) documents one hard-refused regime in its `# Scope` section:

> The bundle spans ONLY the reduced-Schur border (`cache.k`), so the outer products reconstruct the row-block arrow inverse. One regime is hard-refused and routed to the dense channel:
>
> * **Per-row deflation** (`deflated_row_directions`): the Daleckii–Krein correction `−tr(inv_vv·(D − DΦ[D]))` needs the DEFLATED block the plain-S⁻¹ bundle does not carry.

## Why it blocks #2333

#2333 wants `logdet_theta_adjoint_for_block` (`construction_quasi_laplace.rs:5227`) routed through the matrix-free row-jet. Everything else about that function is compatible — measured by enumeration in https://github.com/SauersML/gam/issues/2333#issuecomment-5149966083:

* its three inverse operands are the three admitted shapes (β–β `beta_inv`, row-local t–t `inv_vv`, t–β border `inv_vbeta`), with **no cross-row off-diagonal subscript**;
* its one cross-row coupling (the OrderedBetaBernoulli shared active mass `M_k = Σ_i z_ik`) is a column reduction over row-local scalars that **never reads the inverse**, and is **already implemented** in the from-probes route at `:6078–6087`.

**Deflation is the remaining obstruction.** At `construction_quasi_laplace.rs:5390-5395` the dense path states that `inv_vv` is the **deflated** inverse — it assigns `1/λ̃ = 1` to each per-row unit-stiffness direction `vᵢ` — so every `inv_vv`-weighted t–t contraction of `∂H/∂θ_w` carries the Daleckii–Krein correction `tr(inv_vv·(D − DΦ[D]))` over the t–t block. The probe bundle carries plain `S⁻¹`, so that block cannot currently be reconstructed and the route refuses.

## What this issue is

[...]

PR #4287, opened 2026-09-20
gam-sae outer gradient: drop the unreachable dense-solver arms and gate the dense oracles to tests

### LINKED ISSUE #4181: manifold/construction_exact_hessian: the three dense-solver trace arms in analytic_outer_rho_gradient_components_on_route are unreachable (exact_a_logdet_route == bundle.is_none()); stale "routes to the dense channel" comment
`crates/gam-sae/src/manifold/construction_exact_hessian.rs`, `analytic_outer_rho_gradient_components_on_route` (main at 69974c0899).

L4355 defines the route:
```rust
let exact_a_logdet_route = logdet_derivative_bundle.is_none();
```
Three trace channels are then built only under `!exact_a_logdet_route`, which means `logdet_derivative_bundle.is_some()`. Each one still matches on the bundle and carries a `None =>` dense-solver arm:

| line | channel | unreachable arm |
|---|---|---|
| ~4459 | assignment log-strength trace | `None => self.assignment_log_strength_hessian_trace(rho, cache, solver)` |
| ~4507 | per-atom smoothness EDF | `None => self.decoder_smoothness_effective_dof_with_solver_per_atom(cache, solver, ..)` |
| ~4597 | ARD log-precision trace | `None => self.ard_log_precision_hessian_trace(rho, cache, solver, evidence_operator)` |

None of these arms can run. Consequences:

1. **Dead production code.** `decoder_smoothness_effective_dof_with_solver_per_atom` (`construction_quasi_laplace.rs` ~L3872, about 80 lines) has no other caller at all, in production or in tests. The only other reference is a doc link in `construction_smoothness_dof.rs`. The other two dense functions survive only as test references (`tests_deflation_traces_780`, `tests_clamp_basin_deflation_2333`, `tests_sphere_ard_logdet_trace_2933`, `tests_alpha_persistence_2933`, `tests_streaming_outer_gradient_2026`, `tests.rs`). Those tests therefore prove properties of a channel the outer gradient never reads.
2. **A stale comment describes a fallback that does not exist.** ~L4575 says: "The from-probes channel HARD-REFUSES any row carrying gauge/rotation deflation …, routing that fit to the dense channel rather than silently dropping the correction." In fact the from-probes error propagates, via `map_err(..)?`, and nothing routes to the dense channel. A reader is told there is a recovery path when the real behaviour is a refusal.
3. **`solver` is read only by these arms.** The `solver: &DeflatedArrowSolver` parameter of `_on_route` and of its two wrappers (`..._with_bundle`, `..._arrow_orbit`) is used nowhere else in the function. Every caller builds a solver that the gradient never consults.

**Fix.**
- Replace each `if exact_a_logdet_route { None } else { Some(match bundle { Some(..) => from_probes, None => dense }) }` with `if let Some((probes, sinv)) = logdet_derivative_bundle { … }`.
- Delete `decoder_smoothness_effective_dof_with_solver_per_atom` and repoint the `construction_smoothness_dof.rs` doc link.
- Correct the ~L4575 comment: the refusal propagates.
- Drop the `solver` parameter from the three gradient entry points and their callers. There are about 15 call sites, many of them in tests, which is why this was filed rather than patched diskless.

SauersML:
Fixes #4181

`analytic_outer_rho_gradient_components_on_route` sets `exact_a_logdet_route = logdet_derivative_bundle.is_none()`. The three trace channels it builds under `!exact_a_logdet_route` therefore always have a bundle, so their `None =>` dense-solver arms could never run.

This PR:
- Reads the assignment log-strength trace, per-atom smoothness EDF and ARD log-precision trace with `if let Some(bundle)` / `Option::map`.
- Deletes `decoder_smoothness_effective_dof_with_solver_per_atom`, which had no other caller.
- Removes the `solver: &DeflatedArrowSolver` parameter from `_with_bundle`, `_arrow_orbit` and `_on_route`, and updates every caller.
- Moves the dense oracles into a new private test module, `manifold/tests_dense_solver_oracles.rs`. The production ρ-gradient reads only the exact-A geometry and the from-probes bundle channels, so these helpers never run in a fit. The moved items are `DeflatedArrowSolver`, `outer_gradient_arrow_solver`, `outer_gradient_conditioning_error`, `joint_chart_gauge_basis_for_arrow_layout`, `ard_log_precision_hessian_trace`, `assignment_log_strength_hessian_trace` and the two oracle-only constants. No production item carries a `cfg(test)` gate, which satisfies the ban-scanner.
- Corrects stale comments (the ARD from-probes channel prices deflation (#2712) rather than refusing it; the bundle-free path uses `dense_exact_a_logdet_channels` / `arrow_orbit_logdet_channels`).

**Behaviour change.** The dense gradient no longer refuses a ρ on the B-cache pivot gate of a solver it never consulted. `OuterGradientError::IllConditioned` was produced only by that gate. It is removed, and a numerically singular joint direction is `NonIdentifiable`, which still maps to `TrialPointRefused`.

Validation (local, after merging main at 9e4544cb):
- `cargo check --workspace --tests` is clean, and the ban-scanner passes.
- `cargo test -p gam-sae --lib` with the orbit/arrow/kappa/crosscoder/dense-exact-A/deflation-trace/1451/1436/2653 filters: 24 passed.
- 2 failed, and both also fail on main: `two_compact_topk_orbits_are_integrated_and_differentiated_2234` and `the_arrow_orbit_lane_prices_two_compact_topk_orbits_2234`. Their fixture's minimal seed leaves atom 1 with a zero decoder, and the #2822 entry gate refuses it. This is filed separately.


SauersML:
Coordination: #4341 (for #4336) is stacked on this branch. It deletes the gauge-deflated `DeflatedArrowSolver`, the `B` pivot-ratio gate, the `IllConditioned` variant and the two constants, all of which this PR leaves without a production caller. It doesn't touch this PR's diff; please land this first. I have also closed out my duplicate branch for #4181.

---

PR #4289, opened 2026-09-20
Keep the face offset when restricting a penalty onto an active constraint face (#4170)

### LINKED ISSUE #4170: gam-problem penalty_coordinate: project_into_subspace maps a centered prior mean to zᵀμ, so a penalty with a prior mean under an active constraint face is centered at the projection of μ (reduced gradient off by λ_k zᵀS_k(I−zzᵀ)μ_k)

SauersML:
Fixes #4170

## Root cause
Under a non-empty linear-inequality active set, `objective.rs` restricts every penalty coordinate onto the constraint face `beta = z beta_f`. The face is affine: every point on it is `z beta_f + beta_perp` with the same `beta_perp = (I - z z^T) beta_hat`, which is non-zero whenever an active constraint has a non-zero right-hand side. `project_into_subspace(z)` kept only `R z` (and `z^T mu` for a centred penalty). The reduced shifted score was therefore `z^T S z (beta_f - z^T mu)`, not `z^T S (beta_hat - mu)`. The outer rho-gradient dropped `lambda_k z^T S_k (I - z z^T)(beta_hat - mu_k)`, so it disagreed with the value, which is computed on the full `beta_hat`.

## Fix
- The exact restricted energy is `||R z beta_f - c'||^2`, with the root-space offset `c' = c + R(mu - beta_perp)`. `c'` generally has no preimage in `range(R z)`, so no reduced prior mean can reproduce it.
- `PenaltyCoordinate` gains a `DenseRootOffset { root, root_offset }` variant. Its shifted value and score read the offset. Every curvature operator is unchanged (`z^T S z`).
- `project_into_subspace(z, face_point)` now takes the face point. The objective passes the converged `beta_transformed`, the same point whose `z^T` image is the reduced beta. If the offset is zero, the result is the plain `DenseRoot`.
- The null split runs before the restriction. It now refuses an offset coordinate instead of silently dropping the coordinate's anchor.

## Tests
- `face_reduction_of_a_centered_dense_penalty_is_exact`: the reduced coordinate at `z^T beta_hat` reproduces the full shifted value, `z^T` of the full shifted score, and `z^T S z`, to 1e-12. The test also asserts that the previously dropped term is materially non-zero.
- `face_reduction_of_a_centered_block_penalty_is_exact`: the same checks for a centred block that straddles the constraint normal, and for an uncentred block on a face with a non-zero right-hand side.
- `face_reduction_through_the_origin_is_the_plain_root`: a face through the origin yields the plain `R z` root.
- `null_split_refuses_a_face_offset_coordinate`: the null split refuses an offset coordinate.
- `penalty_coord_projection` (existing, gam-solve) is updated to pass its face point.

## Out of scope
The Tierney-Kadane correction path in `gradient_hessian.rs` handles the prior mean separately. It is not touched here.

---


SauersML:
Ready for the local merge train (MT-1, odd PR). Waiting only on its focused proof runs, which are still queued behind the runner backlog:
- run 35508225196: gam-problem lib, filter `penalty_coordinate`, key BF-54-4170-1
- run 35508226172: gam-solve lib, filter `penalty_coord_projection`, key BF-54-4170-2

Both are at head 4cf2999cbed17d8a88eeaa5559869eeeef822ed8. Merge with merge_method `merge` and expectedHeadSha 4cf2999cbed17d8a88eeaa5559869eeeef822ed8, but only once both runs are green. On merge, #4170 closes via "Fixes". If a run fails, do not merge. Re-dispatch as BF-54-4170-3 and onward after the fix.

---

Issue #3722, opened 2026-09-20
bench/pvalue_calibration: no enabled workflow runs any plan, so the calibration head-to-head vs pyGAM is unmeasured for the p-values main actually reports


SauersML:
## Defect

The p-value calibration harness (`bench/pvalue_calibration`) is the only measurement of the calibration axis against pyGAM. No enabled workflow runs any of its plans, so its verdicts no longer describe the p-values that main reports.

**1. Nothing runs it.**
- The only workflow that references the harness is `.github/workflows/python-contracts.yml` (path filter at line 36). That workflow is disabled: dispatching it returns `Cannot trigger a 'workflow_dispatch' on a disabled workflow`.
- `.github/workflows/pygam-compare.yml:88` runs only `python -m bench.pygam_compare.run "$PLAN"`.
- `git grep pvalue_calibration origin/main -- .github` returns nothing else.
- The `nightly` plan (`bench/pvalue_calibration/plans.py:93`, the full grid at 500 reps) has never had a runner.

**2. The committed baseline measures tests main no longer ships.**
`docs/pvalues.md` ("Plan `quick` at git `4f17dc876bda`") and `bench/pvalue_calibration/baseline/quick/` were recorded before two changes:
- `summary().smooth_terms[i]["p_value"]` is now the variance-component score test. `crates/gam-solve/src/estimate/smooth_term_summary.rs:229-246` calls `ScoreTestFit::test_term` -> `gam_terms::inference::smooth_score_test`. It is not the Wood (2013) rank-truncated Wald test. The baseline's "gamfit Wald (`summary`)" rows (KS D 0.37-0.70 on every family, "NOT UNIFORM") therefore measure a statistic nobody gets any more.
- Random-effect rows now carry a p-value from `fit.artifacts.random_effect_tests` (`smooth_term_summary.rs:148-198`). The baseline still reports "NO P-VALUE; 100 unusable" for every `re` cell.

**3. The docs describe the removed test as the current one.**
- `docs/pvalues.md:19` lists the summary smooth p-value as "Wood (2013) rank-truncated Wald statistic `chi_sq`".
- `docs/pvalues.md:31-54` ("The Wald test in `summary()`") explains its pile-up near 1 as the behaviour of today's p-value.
- `bench/pvalue_calibration/README.md` (surface table) still says `gamfit.coef` is the normal tail of `estimate / std_error` because "gamfit reports no coefficient p-value". PR #3661 changes that for ridged slopes.

[...]

PR #4292, opened 2026-09-20
Describe the score-test and parametric Wald p-values in the calibration docs and harness (#3722)

SauersML:
The p-value documentation and calibration harness described statistics that the current summary no longer reports. Update the smooth and random-effect surfaces to describe their variance-component score tests, and read the linear term's reported `parametric_terms` p-value directly.

The resolution against main preserves #3661: a ridged slope uses its variance-component score test, an unpenalized coefficient uses its Wald test, and the harness records the row's statistic. Bare categorical factors lower to random-effect blocks and can report the recorded score test. Historical calibration measurements are unchanged and explicitly labeled as belonging to the earlier implementation.

Validation: on MSI, the exact resolved harness functions read the same p-value and statistic as a real 60-row Gaussian fit from the existing cached extension. Current source routing for smooth, random-effect and ridged-linear summary rows was reviewed. This does not validate the newly merged score implementation end to end, and does not regenerate the calibration baseline.

Refs #3722. The issue remains open pending baseline regeneration.


SauersML:
Ready for the local merge train (MT-0, even PR). This PR changes only docs and the Python calibration harness. It is waiting on p-value calibration run 35508256183 (plan ci, only_libs gamfit), which is still queued, plus any python-contracts check on the PR.

Merge with merge_method `merge` and expectedHeadSha 00178cd50e36219ba44bfb6ac6c3aae6dee81688, but only once those are green. The body says "Refs #3722", so #3722 stays open for the baseline regeneration.

---

PR #4294, opened 2026-09-20
analytic_penalties: FD-test SCAD/MCP Hessian diagonal and learnable grad_rho

SauersML:
Tests only. It adds `scadmcp_learnable_hessian_and_grad_rho_match_central_differences` to `crates/gam-terms/src/analytic_penalties/tests.rs`.

## Why

There was one SCAD/MCP finite-difference test, and it checked only value -> grad, for MCP with a fixed weight. Three things had no FD test at all:

- the SCAD branch;
- `hessian_diag`, which is the whole Hessian because the penalty is coordinate-separable;
- the learnable log-weight gradient `grad_rho`.

## What the test covers

- Both variants, and every branch of each: SCAD linear, quadratic and flat, and MCP active and flat.
- Three checks per branch: value -> grad, grad -> `hessian_diag`, and value -> `grad_rho` over the log-weight.
- Every probe sits at least 0.16 from a knot, so the h = 1e-5 stencil never straddles one.

I also re-derived the SCAD/MCP math by hand (`scad_mcp.rs:277-297`), and found it correct.

## Tolerances

The tolerances come from the stencil error h^2/6 |f'''| + eps_mach |f|/h at the worst probe (t = 0.03, r ~ 0.058):

| check | error | tolerance |
|---|---|---|
| value -> grad | ~3e-9 | 1e-7 |
| grad -> `hessian_diag` | ~4e-8 | 1e-6 |
| value -> `grad_rho` | ~3e-10 | 1e-7 |

For `grad_rho`, each branch is a*w + b*w^2 + c in w = w0 e^rho, and its third rho-derivative is about 16.

Focused proof: run 35506853255 (key HN-39-6-1).

---


SauersML:
Conflict note from BF-74 (conflict resolver). I'm leaving this PR alone because it was pushed in the last 60 minutes. It now conflicts with main in one file: `crates/gam-terms/src/analytic_penalties/tests.rs`. Main's change there is 0f1d949922 (#3868). It overrides `psd_majorizer_diag` so that OrderedBetaBernoulli and ScadMcp return their declared PSD majorizers, and it adds `scadmcp_frozen_diag_and_log_det_are_the_psd_majorizer` directly after `scadmcp_value_grad_self_consistent_fd`, where this PR also adds its tests. Both sides only add tests, so the fix is to keep both. Also check that this PR's SCAD/MCP Hessian-diagonal FD test calls the exact Hessian diagonal and not the now-overridden `psd_majorizer_diag`.

---

SauersML:
Merge train MT-0: merged origin/main into `hn39-scadmcp-fd` (merge commit 42f7c05f, no rebase or force-push). The only conflicts were in the ledgers under `docs/`; both sides' entries were kept, and the result validates with `python3 -m json.tool`. The code diff against main is unchanged. Queued for local integration train 4, with focused filter `cargo test -p gam-terms --lib analytic_penalties`.

---

Issue #3129, opened 2026-09-19
SPEC: the reactive "retry ONCE with Firth" after a failed binomial fit is a retry/fallback — make Firth a pre-fit decision


SauersML:
**Where:** `crates/gam-models/src/fit_orchestration/fit.rs:554-620` (`fit_from_formula`: `fit_standard_base` → on a `firth_can_rescue(&err)` error, clone the options with `firth_bias_reduction = true` → `fit_standard_base` again → `certified_retry_or_original`).

**SPEC rule:** no retries or fallbacks. This path solves the problem again with a different objective (the Jeffreys-penalized likelihood) *after* the first objective failed to certify.

- **The estimand depends on whether the solver failed.** The same data yields the plain REML/LAML fit when the unpenalized-likelihood solve happens to certify, and the Firth-penalized fit when it doesn't. Near the separation boundary, that switch flips on tiny perturbations of y.
- **Twice the cost on every separated fit.** The failed base solve is thrown away.
- **Recovery is keyed on error types instead of a model property.** `firth_can_rescue` matches separation/non-convergence errors, which is exactly the "recover based on which failure occurred" pattern.

**Principled replacement:** make the choice before fitting, from a property of the data and model:
- **Option (a):** Firth/Jeffreys on by default for every Firth-capable binomial family. The Jeffreys prior is a proper penalty toward no effect, which fits "penalize by default". The penalty's O(1) bias correction is negligible in the non-separated regime and essential in the separated one.
- **Option (b):** a pre-fit separation certificate, e.g. the LP feasibility of a separating direction in the column span of the penalty null space (Konis 2007 / `detectseparation`). This decides Firth up front and deterministically.

Either way, delete the reactive branch, `certified_retry_or_original`, `firth_can_rescue`, and the "rescue attempted" error decoration.

Regression tests to keep:
- `perfect_binomial_separation_2273_tests.rs`, now through the pre-fit decision;
- a perturbation-stability test: the estimand must not change between y and y + 1e-9·noise on a near-separated fixture.

Related: #2273, #2654, #1762. I found this while reviewing merged #3109, which uses the same one-shot Firth retry.

SauersML:
Correction to the "SPEC rule" line above: SPEC.md has no rule that says "no retries or fallbacks", and the words "retry" and "fallback" don't appear in it. These are the lines that actually apply to this path:

[...]

PR #4301, opened 2026-09-20
Decide fixed-λ multinomial separation before the fit and record the objective (#4173)

### LINKED ISSUE #4173: multinomial fixed-λ: a Newton stall with |η| ≥ 25 reactively swaps the estimator to Firth; the returned fit is unmarked and reports an objective it did not optimize
## Where (origin/main)

`crates/gam-models/src/multinomial.rs`:
- `:183` `const MULTINOMIAL_SEPARATION_ETA_THRESHOLD: f64 = 25.0;`
- `:986-1060` `handle_multinomial_fixed_lambda_stall`
- `:1526-1611` the tail of `fit_penalized_multinomial_firth_fallback`

```rust
    let (max_abs_eta, row_index, active_class_index) = max_abs_eta_location(stall.eta.view());
    if max_abs_eta >= MULTINOMIAL_SEPARATION_ETA_THRESHOLD {
        ...
        let firth = fit_penalized_multinomial_firth_fallback(design, y_one_hot, penalty, lambdas,
                                                             row_weights, max_iter, tol, None);
        match firth {
            Ok(out) => return Ok(out),
```

## Why it is wrong

1. **This is a reactive estimator swap triggered by a solver failure.** The fixed-λ softmax Newton runs first. Only when it exhausts its budget, and the stalled iterate happens to have `|η| ≥ 25`, does the driver refit a *different* objective (penalized likelihood + `½ log|I(β)|`) and return that as the answer. The estimator the caller gets therefore depends on whether an unrelated solve failed, and on a hand-picked logit magnitude.
   - SPEC: "In general, do not paper over solver issues."
   - On separated data the runaway `|η|` grows with the iteration count, so a small enough `max_iter` stalls below 25 and returns the typed non-convergence error, while a larger one returns a Firth fit: the estimator depends on the iteration budget.

2. **The returned fit is unmarked.** `MultinomialFitOutputs` (`:731-761`) has no field saying the Firth/Jeffreys objective was used. The binomial path records this decision (#3129 made it a pre-fit decision with the correction recorded). Here downstream consumers cannot tell a bias-reduced estimate from the plain penalized MLE. That matters for LR/AIC comparisons and for the coefficient covariance, which is the Laplace `H⁻¹` of an objective that omits the Firth curvature term.

3. **The reported objective is not the optimized objective.** `:1605` reports `penalized_neg_log_likelihood: -log_likelihood + penalty_term`. The Firth refit maximized `log L + ½ log det I(β) − penalty`, so the reported value omits `−½ log det I(β̂)`. It is also computed from `probs.max(f64::MIN_POSITIVE).ln()` (`:1533`, and likewise `:1268`), a clamp on the exact log-softmax that the non-Firth path does not use (`deviance`, the `−2 log L`, inherits both).

Related but distinct: #3825 (closed; panics caught from this refit), #3345 (formula-path Firth seed).

## Fix

- Make separation handling a **pre-fit decision** on the data, as #3129 did for the binomial path: detect quasi/complete separation of the one-hot design before the solve, from a certificate on the data rather than a stalled iterate's `|η|`. Then either fit the Firth objective from the start or return the typed separation error. Drop the post-stall `|η| ≥ 25` branch and the constant.
- When the Firth objective is used, record it in `MultinomialFitOutputs` (for example `firth_penalized: bool`, or better the objective kind). Report the objective actually optimized, including `−½ log det I(β̂)`, and evaluate `log L` through the exact log-softmax instead of clamping probabilities at `MIN_POSITIVE`.

SauersML:
Fixes #4173

## What changed

The fixed-λ multinomial driver (`fit_penalized_multinomial`) used to run plain penalized Newton. If Newton stalled with some `|η| ≥ 25`, it refit with Firth and returned that fit, but the fit was not marked as Firth. It also reported a penalized NLL for an objective it had not optimized.

Now the driver:

1. **Decides separation before fitting, from the data alone.** `certify_multinomial_separation` runs a nonnegative-cone projection over the penalty nullspace. It uses the #3129 Lawson–Hanson routine `gam_solve::active_set::nonnegative_cone_projection_by_rows`, which is now `pub`.
2. **Fits the chosen objective to convergence.** That is plain penalized ML when no separation is certified, and Firth-penalized ML when it is. There is no switch partway through the fit. A stall on either objective is a typed `FixedLambdaNewtonDidNotConverge` error (via `VectorGlmStall::into_nonconvergence_error`).
3. **Records the objective it optimized.** The new field `MultinomialFitOutputs::objective` is a `MultinomialFixedLambdaObjective`:
   - `PenalizedLikelihood`, or
   - `FirthPenalizedLikelihood { separation, log_det_information }`, where `separation` is the witness: the unpenalized dimension, a unit direction, and its strict and tied margin counts.

   `penalized_neg_log_likelihood` is always the negative of that objective, and `deviance` is always `−2ℓ`. The Python fixed-λ result dict gains an `objective` key (`"penalized_likelihood"` or `"firth_penalized_likelihood"`).
4. **Deletes** `MULTINOMIAL_SEPARATION_ETA_THRESHOLD`, `max_abs_eta_location`, `handle_multinomial_fixed_lambda_stall`, and the `VectorGlmStall::eta` field, which only that handler read.

## Math

Put θ ∈ ℝ^{P·M} (M = K−1, reference logit η_{K−1} ≡ 0) and P_λ = A(λ) ⊗ S. The objective f(θ) = −ℓ(θ) + ½ θᵀP_λθ is convex. Its recession function is

```
f∞(d) = Σ_n w_n Σ_c y_nc (max_k δ_nk − δ_nc)   if P_λ d = 0
f∞(d) = +∞                                       otherwise
```

where δ_nk = x_nᵀd_k is the change in row n's class-k logit along d, with δ_{n,K−1} = 0.

Let N be an orthonormal basis of ker P_λ and put d = N z. Then f∞(d) = 0 exactly when G z ≥ 0, where G has one row

```
g_nck = (N_c − N_k)ᵀ x_n     for each w_n > 0, y_nc > 0, k ≠ c
```

and N_a is class a's P×q block of N, with N_{K−1} = 0. This splits the directions with G z ≥ 0 into two kinds:

- **G z = 0.** f is constant along d, because the design does not identify that direction.
- **G z ≥ 0 and G z ≠ 0.** f decreases strictly along d from every point, so no minimizer exists. This is separation, either complete or quasi.

If no direction of the second kind exists, every recession direction is a direction of constancy, and f attains its minimum (Rockafellar, *Convex Analysis*, Thm 27.1). By Stiemke's lemma, that happens exactly when some μ > 0 has Gᵀμ = 0. The K = 2 case is the binomial check from #3129.

**Certificate.** Project t = −Gᵀ𝟙 onto cone(rows of G) with Lawson–Hanson, and let r be the residual.

- r lies in the polar cone, so z = −r has G z ≥ 0.
- r is orthogonal to the projection, which gives ‖z‖² = 𝟙ᵀG z. So a nonzero z is a separating direction.
- z = 0 gives Gᵀ(𝟙 + ν) = 0 with 𝟙 + ν > 0, where ν ≥ 0 are the Lawson–Hanson multipliers. That certifies a finite minimizer exists.

Margins are resolved against the projection's own band, γ_q‖r‖ + ‖band‖, plus the accumulation band of |X|·|N|·|z| for the logit-difference evaluation. There is no fixed threshold. A margin above the band is strict, a margin inside it is tied, and a margin below it fails the certificate. The data count as separated only when at least one margin is strict. Anything uncertified goes to the plain fit, whose own convergence certificate is then the evidence that a maximizer exists.

**Firth lane.** The driver maximizes ℓ*(θ) = ℓ(θ) − ½θᵀ(A⊗S)θ + ½ log det I(θ), where I is the expected Fisher information. It uses damped Newton with backtracking on ℓ* (`opt::backtracking_line_search`) and reports:

- `penalized_neg_log_likelihood = −ℓ*`
- `objective.log_det_information = log det I(θ̂)`
- `deviance = −2ℓ(θ̂)`
- covariance (I + A⊗S)⁻¹

On separated data the Jeffreys term is what bounds the separated direction. If I is singular at the start, the driver returns `InvalidInput` rather than a fit.

## Tests

All of these are in `crates/gam-models/src/multinomial.rs`, module `separation_firth_tests`. Each test recomputes ℓ and I on its own, using log-sum-exp and a dense Fisher assembly, then checks the reported numbers against those values.

- `separable_data_select_firth_before_fitting_and_report_its_objective`
  - Complete separation. The certificate returns a witness with unpenalized dimension 4, margins ≥ 0 and the largest > 0.1.
  - The fit's `objective` is Firth and carries that witness.
  - `log_det_information`, `penalized_neg_log_likelihood` (= −ℓ*) and `deviance` match the recomputed values to 1e-9 relative.
  - Every ±1e-3 coordinate perturbation lowers ℓ*, and ℓ increases along the witness direction.
- `non_separable_data_never_select_firth`
  - Overlapping classes. The certificate returns `None` and the fit is `PenalizedLikelihood`.
  - `penalized_neg_log_likelihood = −ℓ` (with no penalty) and the deviance match.
  - Perturbations lower ℓ.
- `quasi_separated_data_select_firth_with_tied_margins`
  - Three classes that all share x = 0. The certificate has strict margins and at least 6 tied margins, and the witness does not move the intercepts.
  - The fit is Firth, finite and interior, and log det I and −ℓ* match.

Existing tests updated to the new API:

- `separation_engages_firth_finite_converged_fit` (asserts the objective name)
- `firth_solver_rejects_a_truncated_iterate` (resume keeps the same objective)
- `fisher_override_tests::fixed_lambda_multinomial_firth_keeps_complete_separation_finite` (asserts Firth)

### Exact test filters

`focused-rust-proof.yml`, package `gam-models`, target `lib`, filters:

```
multinomial::separation_firth_tests multinomial::fisher_override_tests penalized_vector_glm::parity_tests::multinomial
```

## Notes

- The fixed-λ fit is not persisted in a saved model: `MultinomialSavedModel` is written only by the formula/REML path. So the objective is recorded in the fit outputs and in the Python result dict. The formula path's non-finite-η diagnostic (`MultinomialSeparationDetected`) is unchanged.
- `docs/convergence_theory/multinomial-ls.md` is updated where it listed the `|η| > 25` proxy as something to replace. `CHANGELOG.md` has an entry.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, b00057499c45.

Issue #3829, opened 2026-09-20
Negative-binomial (theta, rho) joint optimization fails to certify on 8/200 null fits (theta residual 3.7e-8 vs bound 1e-10)


SauersML:
The negative-binomial null calibration gate (the NB null fits carried over from the closed PR #3220's SBC/null gates) certifies only **192 of 200** null fits. The other 8 are refused with:

```
Negative-binomial (theta, rho) optimization did not certify a joint optimum within 200 round(s)
```

In the failing fits the theta-score residual stalls at about **3.7e-8**, while the certification bound is **1e-10**. So the outer (theta, rho) iteration is converging, but only to a precision about 370× coarser than the certificate asks for. Either the alternation stalls short of the joint stationary point, or the bound cannot be reached in floating point at this scale.

SPEC: "A fit object must only ever come from a converged optimization." and "In general, do not paper over solver issues." Neither loosening the bound nor retrying is acceptable. What is needed is one of:

- a derivation of the attainable theta-score precision, meaning a bound scaled to the rounding floor of the score's own terms; or
- a joint Newton step on (theta, rho), replacing the alternation that stalls.

Repro: 200 NB null replicates, as in the SBC null gate. See #3220 for the harness, and #3349 (closed) for an earlier NB certification fix.

### Where

`crates/gam-solve/src/estimate/optimizer.rs`, in the NB alternation loop (about lines 2248-2345 on main): `theta_bound = negbin_theta_joint_bound(theta, &theta_profile, theta_root_displacement)`. The bound (#3349) is the score's rounding band plus the root displacement that the rho and beta certificates still allow. A bound of 1e-10 alongside a residual of 3.7e-8 suggests the displacement term is 0 in the failing fits. Two things on main can make it 0:

- `root_sensitivity` falls to `map_or(0.0, …)` when `negbin_theta_score_eta_gradient` or `negbin_theta_root_sensitivity` returns `None`/`Err`.
- `rho_gain` falls to 0 when `joint_eval.hessian.materialize_dense()` returns `Ok(None)` or `Err`.

When either happens, the bound quietly collapses to the rounding band alone. That is stricter than anything the rho/beta certificates can deliver, so the alternation cannot meet it and fails after `reml_max_iter` rounds. The first step is to log which branch the 8 failing replicates take. If the displacement is really 0 there, the fix is to make those branches refuse or compute the displacement, rather than default it to 0. If the displacement is nonzero and the alternation still stalls, the fix is the joint Newton step.

[...]

PR #4311, opened 2026-09-20
Score group() rows of Gaussian location-scale fits (#3836)

### LINKED ISSUE #3836: Location-scale fits never record the group() variance-component score test (random_effect_tests left empty)
This is a follow-up to the closed PR #3220. Part (a) of that PR, the chart-covariant score-test kernel for overlapping penalties, landed as #3813. This issue tracks the part that did not land.

## Defect

`FitArtifacts::random_effect_tests` is filled only on the standard route, in `crates/gam-models/src/fit_orchestration/entry.rs` (around line 2697, `random_effect_test_records(&standard.design, &standard.fit)`). Every location-scale route leaves it empty:

- `crates/gam-models/src/survival/location_scale/spec.rs:858` has `random_effect_tests: Vec::new()`, with the comment "Survival location-scale designs carry no random-effect block".
- `crates/gam-models/src/inference/model.rs:7055` has `random_effect_tests: Vec::new()`.

So a `group(g)` term in the mean formula of a Gaussian, binomial or dispersion location-scale fit gets no calibrated p-value in `summary()`. It reports "not recorded" instead.

#3220 showed that the fix works for the Gaussian location-scale mean block: a 1000-rep gate at .094 / .051 / .016 (KS .987). The mean block is scored from its own row state, `w_i / σ̂_i²`, with σ rebuilt using the fit's own σ floor. Because σ̂ is fitted to the same residuals, the level is treated as estimated, with a χ²_ν ratio.

## Fix

- Build the mean block's IRLS row state from the location-scale fit: weights `w_i/σ̂_i²` and working response. For the non-Gaussian location blocks, use the block's own score and curvature.
- Pass that row state to `random_effect_test_records`, or to its location-scale analogue.
- A route that cannot supply a row state refuses with a typed reason. It must not silently return an empty `Vec`.
- Gate the change with a seeded SBC size curve, 2-MCSE two-sided bands plus KS. A gate fails outright when more than 5% of its fits fail, and failed fits are reported, never scored. #3220's `tests/inference/misc/sbc_double_penalty_smooth_family_size_curve.rs` is the pattern.

## Related

- #3829: negative-binomial (θ, ρ) certification failures. These block the NB row of any such SBC gate.

SauersML:
Fixes #3836

## Problem
Location-scale fits never recorded random-effect tests. Every `group()` row of their mean predictor reported `random_effect_test_not_recorded` and had no p-value.

## Fix
**Gaussian location-scale.** The mean block is now scored with the Lin/Wood variance-component test.
- Conditional on the fitted σ̂ shape, the mean block is the weighted Gaussian regression `yᵢ ~ N(μᵢ, φσ̂ᵢ²)`. Its Fisher/curvature weight is `wᵢ/σ̂ᵢ²` and its score is `wᵢ(yᵢ−μ̂ᵢ)/σ̂ᵢ²`.
- The mean/log-σ Fisher cross-information is zero: `E[∂²ℓ/∂μ∂η_σ] = −2w E[y−μ]/σ² · σ'/σ = 0`. The mean block's own score is therefore its efficient score, and the log-σ coefficients need no projection.
- φ is **estimated**, not fixed at 1. σ̂ is fitted to the same residuals, so fixing φ=1 would make the test conservative. The ratio statistic is also invariant to response units.
- The inputs are read in raw units after the raw remap: μ̂ is the location block's `eta`, and `σ̂ = s·b + exp(η_σ)` uses the fit's own floor.
- A mean composed with a link wiggle is not linear in the term. Its rows carry a new typed reason, `MeanNotLinearInTerm`.

**Binomial location-scale, Beta dispersion and survival location-scale.** In these families the mean and scale blocks are coupled through the likelihood, via `(t−η)/σ` or the Beta mean/precision cross term. The mean block's own score is not efficient there. Their `group()` rows now carry `NoIrlsRowState` at the realized global coefficient range, instead of "not recorded".

## Tests
`tests/inference/misc/sbc_random_effect_variance_component_size_curve.rs` gains two Gaussian location-scale cells:
- A null cell with heteroscedastic noise `σ = 0.3·e^{x1}` and noise formula `s(x1)`. It uses the same α ± 2·MCSE size bands and the KS-uniformity gate.
- A power cell.

---


SauersML:
Status: the implementation is complete at 714a46f1. Two proofs are queued but have not yet been picked up by a runner (the pool has been saturated for 70+ minutes):
- [35508376927](https://github.com/SauersML/gam/actions/runs/35508376927): `gam` integration binary `inference`, filter `sbc_random_effect_variance_component_size_curve`. This runs the new Gaussian location-scale null-size and power cells plus the existing families.
- [35508377971](https://github.com/SauersML/gam/actions/runs/35508377971): `gam-models` lib, filters `random_effect location_scale`.

For MT-1 (odd PR): if both are green, this is ready to mark ready-for-review and merge. If either is red, it needs a fix first; do not merge on red.

What remains after this PR, outside its scope: the non-Gaussian location blocks need an efficient score (a projection of the scale block) before their `group()` rows can carry p-values. NB and Gamma are Fisher-orthogonal, but NB is blocked on #3829. Beta, binomial location-scale and survival location-scale are not orthogonal.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 2f8c1ddcf9f7.

PR #4316, opened 2026-09-20
gamlss: Stirling-gap Gamma/Beta dispersion rows stable at large shape

### LINKED ISSUE #4252: gamlss/dispersion_family: Gamma and Beta loglik, η_d score and Fisher weight subtract O(ν ln ν) terms, so they are noise at large shape/precision (the σ→0 boundary)
## Where (origin/main eb3fe80912)

All code is in `crates/gam-models/src/gamlss/dispersion_family.rs`.

Gamma row log-likelihood (`dispersion_gamma_loglik`, line 631):
```rust
let s = nu * nu.ln() - nu * mu.ln() - ln_gamma(nu) + (nu - 1.0) * y_pos.ln()
    - nu * ((1.0 / mu) * yi);
```
Gamma row program (line 865), which is where the η_d score and curvature come from:
```rust
let kernel = add(mul(precision, log_ratio), neg(ln_gamma_jet));
let response = add(scale(precision, log_response), neg(mul(precision, scaled_response)));
```
Gamma Fisher weight (line 2094): `let info_nu = nu_stack[1] - nu.recip();`

Beta row log-likelihood (line 645) and row program (line 944):
```rust
ln_gamma(phi) - ln_gamma(a) - ln_gamma(b) + (a - 1.0) * yc.ln() + (b - 1.0) * (-yc).ln_1p();
let normalizer = add(ln_gamma_precision_jet, neg(add(ln_gamma_first_jet, ln_gamma_second_jet)));
```
Beta Fisher weight (line 2134): `let info_phi = mu * mu * tri_a + one_minus_mu * one_minus_mu * tri_b - tri_phi;`

## Why it is wrong

This is the σ→0 boundary of both families. As the data become nearly deterministic given the mean, the shape ν (Gamma) or precision φ (Beta) grows without bound, and the fit ends up at large ν or φ. In that regime every quantity above is an O(1) result obtained by subtracting terms of size O(ν ln ν).

- **Gamma.** Two expansions:
  - `ν ln ν − lnΓ(ν) = ν + ½ ln ν − ½ ln 2π + O(1/ν)`.
  - `ν(ln z − z) = −ν − ν(z − 1)²/2 + …`, with `z = y/μ = 1 + O(ν^{-1/2})`.

  The row log-likelihood is O(ln ν) and the η_d score `ν[ln ν + 1 − ψ(ν) + ln z − z]` is O(1). Both are summed from terms of size `ν ln ν`, which leaves an absolute error of about `ν ln ν · ε`.
- **Beta.** `lnΓ(φ) − lnΓ(μφ) − lnΓ((1−μ)φ)` equals `−φ[μ ln μ + (1−μ) ln(1−μ)] + O(ln φ)`, and the O(φ) response terms cancel it. The error is the same `φ ln φ · ε`.
- **Fisher weights.** Both cancel their leading `1/ν` terms:
  - `ψ′(ν) − 1/ν = 1/(2ν²) + …`
  - `μ²ψ′(μφ) + (1−μ)²ψ′((1−μ)φ) − ψ′(φ) = 1/(2φ²) + …`

  Each keeps only about `log₂(1/(2νε))` bits. Past ν ≈ 1/(2ε) ≈ 2·10¹⁵ they lose even their sign and come out zero or negative. `disp_weight` then becomes 0 or negative, and `disp_response = ed + score_eta / (…·info)` divides by it.

Measured errors, against a 50-digit mpmath reference. Gamma uses 200 rows drawn from the model at μ = 3.7 and Beta uses 100 rows at μ = 0.3. Each entry is the median over rows:

| ν or φ | Gamma score err | Gamma loglik err | Beta score err | Beta loglik err | \|score\| |
|---|---|---|---|---|---|
| 1e4 | 1.8e-11 | 1.3e-11 | 1.9e-11 | 4.2e-12 | ~0.4 |
| 1e8 | 9.3e-8 | 1.2e-7 | 4.8e-8 | 3.0e-8 | ~0.4 |
| 1e10 | 1.2e-5 | 4.6e-6 | 1.6e-5 | 1.8e-5 | ~0.4 |
| 1e12 | 4.7e-3 | 2.8e-3 | 1.1e-3 | 2.0e-3 | ~0.4 |

These errors are per row. Summed over n rows, the log-likelihood error grows somewhere between √n and n. At ν ≈ 1e12 and n = 1000, the reported log-likelihood and LAML are off by 0.1 to 3 nats. The per-row score noise of 1e-5 to 1e-3 is far above any convergence certificate the inner solve can issue.

The two sibling families have had the same defect. For NB2 it is the Poisson limit (#3907); for Student-t it is the Gaussian limit (#3235, fixed with closed-form half-shift gaps). Gamma and Beta are the ones left.

SPEC: "In general, do not paper over solver issues." This arithmetic creates a noise floor that no solver can converge through.

## Principled fix

Difference the asymptotic parts in closed form, as #3235 and #3907 do.

- **Gamma.**
  - Supply one leaf `g(ν) = ν ln ν − ν − lnΓ(ν)` with the stack `[½ ln ν − ½ ln 2π − r(ν), ln ν − ψ(ν), 1/ν − ψ′(ν), −1/ν² − ψ″(ν), 2/ν³ − ψ‴(ν)]`, where `r` is the Stirling remainder (Binet function). Evaluate each entry as the recurrence plus the Bernoulli tail, with the leading terms removed analytically. That way no O(ln ν) or O(1/ν^k) numbers are subtracted.
  - Write the response part as `ν·(1 + ln z − z) = ν·log1p_minus_x(z − 1)`; `gam_math::special::log1p_minus_x` already exists. Its derivatives in δ_μ, such as `ν(z e^{−δ} − 1)`, are formed from `z − 1` directly.
  - Read the Fisher weight `ν²·(ψ′(ν) − 1/ν)` off the same stack. The NB kernel already expands `trigamma(x) − 1/x` this way in `nb_log_precision_fisher_jensen`.
- **Beta.**
  - With the same `g`, write `ℓ = −φ·KL(μ‖y) + g(a) + g(b) − g(φ) − ln y − ln(1−y)`.
  - Here `KL(μ‖y) = μ ln(μ/y) + (1−μ) ln((1−μ)/(1−y)) = −μ·log1p_minus_x(t) − (1−μ)·log1p_minus_x(s)`, with `t = (y−μ)/μ` and `s = −(y−μ)/(1−μ)`. The first-order parts `μt + (1−μ)s` are exactly 0, so nothing cancels.
  - Write `info_phi` as `μ²h(a) + (1−μ)²h(b) − h(φ)`, with `h(x) = ψ′(x) − 1/x` read from the same stack. Its leading terms `1/(2φ²)·(1 + 1 − 1)` do not cancel.
- **Value paths and tests.** Move the plain-f64 value paths (`dispersion_gamma_loglik`, `dispersion_beta_loglik`) and the `test_support` oracles, including the Beta check at line 4122, to the same forms. Value/score consistency then holds by construction. `tests_row_program_932` needs a row at large ν or φ (at least 1e10), checked against a high-precision reference.

`gam-solve/src/pirls/deviance.rs:2379` already has a `log_gamma_stirling_correction`. The stable `g` stack belongs in `gam_math::special`, next to `ln_gamma_shift_gap` / `polygamma_shift_gap_stack` (PR #4200 for #3907), as the single source of truth, and the PIRLS function should read from it.

SauersML:
Fixes #4252

## Root cause

The Gamma and Beta dispersion rows wrote the log-density in its textbook lnΓ form. At large shape (ν, φ ≳ 1e8), every piece of that form is O(ν ln ν), but the log-density and its η_d derivatives are O(ln ν) or smaller. The rows therefore lost all significant digits to cancellation. The worst case was the shape information ψ′(ν) − 1/ν, which is O(1/ν²) and cancelled to noise.

## Fix: write the densities in Stirling-gap form

The Stirling gap g(x) = x ln x − x − lnΓ(x) = ½ ln x − ½ ln 2π − r(x), where r is the Binet remainder, is O(ln x) with well-conditioned derivatives.

- **Gamma** (ν = e^{η_d}, z = y/μ):
  - ℓ = g(ν) + ν·L₁(z − 1) − ln y, where L₁(x) = log1p(x) − x.
  - The excess is formed as (y − μ)·(1/μ), so the O(ν (y−μ)²/μ²) fit term never cancels.
- **Beta** (a = μφ, b = (1−μ)φ):
  - ℓ = g(a) + g(b) − g(φ) − φ K − ln y − ln(1 − y).
  - K = −μ L₁(t) − (1−μ) L₁(s) ≥ 0 is the Bernoulli divergence, with t = (y−μ)/μ and s = (μ−y)/(1−μ).
- **Kernel informations**, derived from ψ′(x) = 1/x − g″(x):
  - Gamma: I_νν = −g″(ν).
  - Beta: I_μμ = φ²(1/a + 1/b − g″(a) − g″(b)) and I_φφ = −(μ² g″(a) + (1−μ)² g″(b) − g″(φ)).
  - Using μ²/a + (1−μ)²/b = 1/φ, the O(1/φ) parts of I_φφ cancel analytically rather than in floating point.

Both forms are algebraically identical to the lnΓ forms; only the floating-point evaluation order changes.

## Supporting changes

- `gam_math::special`:
  - `ln_gamma_binet_remainder` and `stirling_gap` are now single-sourced.
  - New `stirling_gap_derivative_stack(x, orders)` returns [g′, …, g⁽⁵⁾]. Each entry is differenced in closed form: an upward recurrence with cancellation-free brackets, then the asymptotic series. This avoids subtracting D^m ln x − ψ_m.
  - A test checks them against high-precision references.
- `gam-solve` PIRLS deviance now uses the shared remainder instead of its private copy.
- All row programs, the order-2 and pruned towers, and the closed-form kernel arms consume the new stacks. The value stays bit-identical between the full and pruned order-2 towers.

## Independent oracles

The generic lnΓ oracles in `test_support` (`dispersion_*_nll_generic`, and the η-space `dispersion_eta_nll_order2/3`) are intentionally left in lnΓ form. At moderate shape they remain an independent check of the new Stirling-gap production form.

## Tests

- New `gamma_and_beta_rows_are_accurate_at_large_shape_4252` (tests_row_program_932.rs):
  - Covers ν, φ ∈ {1e10, 1e12, 1e14}.
  - Checks row and kernel loglik, both scores, the η_d Hessian, the dispersion weight and the Newton step against 80-digit mpmath references to 1e-12 relative.
- Existing tower/kernel consistency, pruned bit-identity and row-program speed-gate tests updated; local validation in progress.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, d55da9235eac.

PR #4318, opened 2026-09-20
survival marginal-slope: certify the pooled baseline pilot root (#4250)

### LINKED ISSUE #4250: survival marginal-slope: pooled_survival_baseline returns a best-so-far probe or a silent 0.0, and the value becomes the model's slope offset
**Area:** `crates/gam-models/src/survival/marginal_slope/fit_setup.rs`, `pooled_survival_baseline` (origin/main `ab473b6224`, lines 807-961)

SPEC: "A fit object must only ever come from a converged optimization." / "In general, do not paper over solver issues."

### What the code does

`pooled_survival_baseline` is documented as "a safeguarded 1D Newton solve on the true row objective". It returns a bare `f64`, and nothing records whether a stationary point was found:

```rust
// fit_setup.rs:854-862  -- every failure becomes the value 0.0
let Some(state0) = objective_grad_hess(0.0) else { return 0.0; };
if !state0.0.is_finite() { return 0.0; }
if state0.1.abs() < 1e-8 { return 0.0; }
...
// fit_setup.rs:878  -- 48 bracket doublings; exhaustion falls through
for _ in 0..48 {
    ...
    // fit_setup.rs:913  -- 60 bracketed Newton/bisection steps
    for _ in 0..60 {
        if state.1.abs() < 1e-8 || (bracket_hi.0 - bracket_lo.0).abs() < 1e-8 { break; }
        ...
        let Some(mid_state) = objective_grad_hess(candidate) else { break; };  // eval failure: silent exit
        ...
    }
    return if best.0.is_finite() { best_slope } else { 0.0 };        // :954
}
...
if best.0.is_finite() { best_slope } else { 0.0 }                    // :960
```

Every exit returns the lowest-objective probe seen so far (`best_slope`), or `0.0`. That covers four cases:
1. The inner loop satisfied the score test.
2. It ran all 60 steps unconverged.
3. A row objective failed to evaluate mid-bracket (`break`).
4. No sign change was ever bracketed. That is the case when the pooled objective has no minimizer (a separable or degenerate pilot). Here the return is simply the best of the ±0.5·2^k probes.

The objective evaluation at 0 failing, or being non-finite, silently yields `0.0` too, which is indistinguishable from "the pooled slope is zero". The stationarity tests use absolute tolerances (`1e-8` on a summed, weight-scaled score and `1e-8` on the bracket width), which are not derived from the objective's scale.

### Why it matters

The value is not just a warm start. `fit_entry.rs:660-676` makes it part of the model:
- `common_slope_offset = slope_offset + baseline_slope` becomes the slope block's fixed offset (`build_slope_blockspec`, fit_setup.rs:136, and `materialize_identity` at fit_entry.rs:1378/1427).
- The same offset decides the learned-frailty identification gate (`frailty_identification`, fit_entry.rs:685): σ is refused or admitted depending on whether this offset lies in the slope span.
- `rigid_slope_at_rows = slope_offset + baseline_slope` builds the influence-absorber columns `diag(s_f·β̂₀)·J` (fit_entry.rs:855). The absorber is fixed-ridge, so the scale of these columns changes the penalized fit.
- It is persisted as `baseline_slope` and replayed at prediction (predict_io.rs:698/1851/2582/2686, survival/predict.rs:4238/6704, run_predict.rs:456).

A pilot that returns a best-so-far probe, or silently falls back to `0.0`, therefore changes the fitted model, the frailty identification decision and the saved model. Nothing records that the pilot did not converge.

### Principled fix

- Return `Result<f64, String>`.
- Solve the 1D score equation with a safeguarded Newton/bisection whose stopping test is derived from the objective's own rounding band. For example, the score is resolved when `|g| ≤ c·ε·Σ|row score|`, using the accumulated-roundoff helper already used for sums elsewhere. That replaces the absolute `1e-8` constants.
- Return the certified root, not `best_slope`.
- Report "no finite minimizer" (no bracket after the derived expansion limit), an evaluation failure, or step exhaustion as typed errors that `fit_entry` propagates.
- Drop the `0.0` fallbacks, except the exact `n == 0` case and the certified case where the score at 0 is already within its band.

SauersML:
Fixes #4250

## Problem
`pooled_survival_baseline` returned a bare `f64`. On every exit it gave back either the lowest-objective probe of a capped search or a silent `0.0`:
- The search was capped at 48 bracket doublings and 60 Newton/bisection steps, with absolute `1e-8` tests.
- The `0.0` came back when an evaluation failed, when the objective was non-finite, or when no bracket was found.

That value becomes part of the model:
- the slope block's fixed offset;
- the learned-frailty identification gate;
- the scale of the absorber columns;
- the saved `baseline_slope`.

So an unconverged pilot changed the fitted model and nothing recorded it.

## Fix
The function now has the signature `pooled_survival_baseline(...) -> Result<f64, SurvivalMarginalSlopeError>`, and `fit_entry` propagates the error with `?`.

- **Score test from the objective's own rounding.** The pooled score is treated as zero when `|Σ gᵢ| ≤ γ_n·Σ|gᵢ|` (`gam_linalg::roundoff::accumulation_band`). The `n − 1` additions plus the rounding that formed each row's score give that depth. A score outside the band has a trustworthy sign. The curvature is "resolved positive" when `H > γ_n·Σ|hᵢ|`.
- **Bracket expansion.** The search moves along the descent direction `−sign(G(0))`. It starts at the Newton step when the curvature at 0 is resolved positive, and otherwise at a unit slope. The start only sets how many doublings the bracket takes. The search doubles until the score changes sign, and it stops at the f64 range rather than at a count.
- **Refinement.** Newton runs inside `[lower, upper]`, where `G(lower) < 0 < G(upper)`, a crossing from descent to ascent. It falls back to bisection whenever:
  - the Newton step leaves the bracket,
  - the curvature is not resolved positive, or
  - the previous step failed to halve the width.

  The bracket therefore halves at least every second evaluation. The loop stops when the score is inside its band, or when the bracket is two adjacent floats (it then returns the end with the smaller `|G|`). No iteration cap or absolute tolerance is needed.
- **Typed errors instead of values:**
  - `NumericalFailure` for a row that does not evaluate (the message names the row and the slope) or for a non-finite pooled objective.
  - `RootSolveFailed` for no finite minimizer. This covers a score that keeps its sign out to the end of the f64 range, and a score that flattens to rounding with no resolved positive curvature (a separable or degenerate pilot).
- **Remaining `0.0` returns.** Only two are left: `n == 0`, and a score at 0 that is already inside its band with a resolved positive curvature.

## Tests
- `pooled_survival_baseline_ignores_origin_entry_offsets_2336` and `pooled_survival_baseline_solves_at_the_conditional_score_variance_2952` now unwrap the certified result.
- New: `pooled_survival_baseline_returns_a_certified_score_root_4250`. At the returned slope the pooled score is inside its rounding band, or it changes sign across `next_down`/`next_up`. The pooled curvature there must also be resolved positive.
- New: `pooled_survival_baseline_refuses_a_row_that_does_not_evaluate_4250`. A row with `qd1 < 0` must give `NumericalFailure` naming row 3.

Nothing has been compiled or run yet: this was written without a local build.

## Landing filters (local build)
```
cargo check -p gam-models --tests
cargo test -p gam-models --lib pooled_survival_baseline
cargo test -p gam-models --lib survival::marginal_slope
```

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 2041c0095058.

Issue #2760, opened 2026-08-01
iso-kappa joint outer search genuinely fails to converge at n>=4000: |Pg|=3.861e-1 against bound=2.566e-1 (1.5x OUTSIDE), the failure that was masked by #2706's mint contradiction


SauersML:
Split out of #2706, whose own defect — a configuration that both suppressed the analytic Hessian and required a measured one — is fixed by `24f3d24cd` and verified on six arms. **This is the failure that was underneath it**, invisible until the mint stopped failing first.

Filing separately because the residual's named home is gone: the follow-up comment on #2706 routes it to "the outer-certification family (#2624 / #2627)", and **both of those are CLOSED**. Without a home this finding is orphaned by closing #2706.

All numbers below are from that comment's measurement, not mine; I am transcribing them so they survive the close.

## The two refusals, and why only one of them was a configuration problem

Same lane, same reported state — `hessian_psd=n/a`, `curvature_source=unavailable`, `railed=[1,2,3,4]` (correctly: the #1033 ψ-lane is armed, so the Hessian genuinely *is* `Unavailable`, and after `24f3d24cd` that is allowed):

```
n = 1000, pre-fix    |Pg| = 3.957e-3   bound = 7.016e-2   ->  17x INSIDE  the bound
n = 4000, post-fix   |Pg| = 3.861e-1   bound = 2.566e-1   ->  1.5x OUTSIDE the bound
```

At `n = 1000` the point **satisfied its own stationarity criterion and was refused anyway** — that is the configuration contradiction, and it is fixed.

At `n = 4000` the projected gradient **genuinely exceeds its bound**. No configuration change can rescue that. It is a real non-convergence of the iso-κ joint outer search at larger `n`.

**The distinguishing evidence is one ratio — `|Pg| / bound` on either side of 1** — which is exactly the comparison the refusal text already prints, and which nobody could read while the mint was failing for an unrelated reason. Two failures were stacked and the first masked the second at every `n`.

## Current state of the arms

| arm | after `24f3d24cd` |
|---|---|
| `kappa_micro_2point_n_independence` (n ≤ 2000) | **ok** |
| `kappa_outer_loop_is_n_independent_n1000` | **ok** |
| `kappa_outer_loop_is_n_independent_fast_ladder` | mints, clears n=1000, dies at a larger rung (`:399`) |
| `kappa_outer_loop_is_n_independent_n4000` | mints, dies in the n=4000 trial (`:725`) |
[...]

PR #4320, opened 2026-09-20
Drop machine-constant gates from three comparisons with exact answers (#3245)

SauersML:
Refs #3245 (partial: rows 2, 4 and 5b of the audit table).

## What changes

**Row 4: `spatial_optimization.rs`, the κ-search acceptance band.** The code removes `accept_tol = options.tol.max(1e-8·|baseline|).max(1e-12)`. Every comparison it gated is between two criterion values that were already evaluated: `joint_final` against `joint_seed`, and `optimized_score` against `baseline_score`. The routine returns the better of the two. A band `t` changes the outcome only when the two values differ by less than `t`, and in that case it ships the *worse* candidate. The old band let the search return a fit whose score was up to 1e-8 worse, relative to the incumbent. That contradicts `spatial_length_scale_optimization_monotone_*`, which asserts `optimized <= baseline + 1e-10`. The descent contract is now `joint_final > joint_seed`, and acceptance is now `optimized > baseline`, both with no band. The route-agreement record at θ0 is now logged on every run, never only conditionally, and it remains a record, not a gate (#2760).

**Row 2: `response_geometry.rs::validated_metric`, the symmetry gate.** The metric enters only through the quadratic form rᵀMr (Gram, cross term and deviance). Since rᵀMr = rᵀ·sym(M)·r exactly, sym(M) *is* the metric, and no tolerance is needed to decide whether it is symmetric "enough". The code keeps the averaging. Cholesky still decides positive definiteness.

**Row 5b: `multinomial_posterior.rs`, the covariance symmetry gate.** The active covariance V = (I⊗x)ᵀΣ(I⊗x) is now formed from sym(Σ) over the upper triangle and mirrored, so it is symmetric by construction. The √ε·scale gate in `validate_inputs` is removed. It could never catch the failure it was meant for: a transposed factor LᵀL is itself symmetric.

## Still open on #3245 (not in this PR)
- **Row 1 (`response_geometry.rs:473`):** the √ε·n_outputs required projected-gradient norm. The principled fix is for `SharedTangentObjective::eval` to publish `RhoGradientParts` through `record_certificate_parts`, so that the Theorem 9 band applies. The rotation test's `< 1e-7` bound then needs a certificate-derived rewrite.
- **Row 3 (`spatial_curvature_inference`):** carried by PR #3485 (#3453).
- **Row 5a (`multinomial_posterior.rs:87`):** the default integration tolerance √ε needs a quadrature error derivation.

Validation: focused-rust-proof run 35506231314 (issue key BF-34-3245-1).

---


SauersML:
[swarm lane spec-lit, ledger check] This PR deletes the only `.max(1e-12)` in `crates/gam-models/src/fit_orchestration/drivers/spatial_optimization.rs`, the `accept_tol` at main :250. It leaves that literal's ledger line in place:

```
magic	crates/gam-models/src/fit_orchestration/drivers/spatial_optimization.rs	.max(1e-12)	https://github.com/SauersML/gam/issues/2469
```

`scripts/spec_ban_ratchet.py` refuses stale lines, so main goes red the moment this merges. Prune that line in the same commit. `python3 scripts/spec_ban_ratchet.py --base <parent>` should then report 0 new and 0 stale.

Also, draft #3417 (claude/fix-3245-derived-tolerances) is still open. It makes the same three changes (spatial_optimization `accept_tol` to strict comparisons, the multinomial_posterior symmetry gate, and the response_geometry :473/:1408 tolerances) and already prunes that ledger line. Only one of the two should land; merging both conflicts on every hunk.

SauersML:
**Verify-and-land review: NEEDS-AUTHOR (draft, not merged).** I reviewed head 43ac2a25 by reading it.

The math is correct:
- picking the smaller of two evaluated scores needs no band;
- rᵀMr = rᵀ·sym(M)·r;
- building V from sym(Σ) gives exactly sym(V).

Two blockers:
1. The PR deletes the only live `.max(1e-12)` in `spatial_optimization.rs` but leaves its row at `scripts/spec_ban_ledger.tsv:41`. `spec_ban_ratchet.py` blanks comments before matching, so the literal that survives in a comment doesn't count. Main would go red with a stale ledger entry. Fix: delete that row in this PR, since the ratchet only allows the ledger to shrink.
2. It duplicates open draft #3417 hunk for hunk. Only one of the two should land, so close whichever you drop.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 8185e0f88547.

PR #4328, opened 2026-09-20
evidence: fit the circular Gaussian by certified Newton ascent (#4003)

### LINKED ISSUE #4003: evidence: CircularGaussianFit2d EM needs >4096 iterations on blob-like data and aborts the atom shape race
**Area:** `crates/gam-solve/src/evidence.rs` — `CircularGaussianFit2d::fit` (origin/main `f8c6613`, lines 2454-2633; the file is unchanged since `988f269`)

### What the code does

The circular-Gaussian (ring) MLE is fitted by EM on the latent angle. The loop has a hard cap and a parameter-step stop rule:

```rust
// evidence.rs:2548-2552
const MAX_EM_ITERATIONS: usize = 4096;
const EM_TOLERANCE: f64 = 2.0e-12;
...
for _ in 0..MAX_EM_ITERATIONS {
    ...
    if parameter_change <= EM_TOLERANCE { converged = true; break; }
}
if !converged {
    return Err("circular Gaussian maximum-likelihood fit did not converge".to_string());
}
```

Returning `Err` when the cap is hit is correct per SPEC. The defect is that this EM cannot reach the MLE within 4096 iterations on the most ordinary input the caller feeds it.

### Why EM fails here (math)

EM's linear rate is the largest fraction of missing information. With `κ_i = R·r_i/s`, the latent angle is almost uninformative when `κ` is small. That is the case for blob-like data, where the ring radius `R` is comparable to or smaller than the noise sd `√s`. The EM rate then goes to 1. When the MLE lies on the boundary `R = 0`, both the first- and second-order terms in `R` vanish at the fixed point, so convergence is sublinear.

### Reproduction (numpy mirror of the exact Rust updates: moment init, joint center/radius M-step, variance M-step, and the same step test)

Isotropic 2-D Gaussian data, 6 seeds each:
- n = 2000 / 5000 / 20000: in 1–2 of every 6 samples the moment initialization gives `R > 0` and EM exhausts all 4096 iterations, so `fit` returns `Err`.
- One example is seed 1, n = 2000. Run without the cap, EM needs **7628** iterations to satisfy the step test, reaching `R = 0.104`. The per-iteration contraction is about 0.996.
- In the other runs where `R > 0`, EM needed 1959, 5741 and 6835 iterations. The interior MLE there really does beat `R = 0` in log-likelihood (for example 443.452 vs 443.036), so this is not a spurious optimum.

### Impact

`ring_provider_2d` (crates/gam-sae/src/inference/atom_shape_race.rs:22) and `ring_bic_2d` (atom_shape_race.rs:172) propagate the error with `?`. `run_atom_shape_race` then fails, surfaced through gam-pyffi inference_instruments.rs:1597 as a Python `ValueError`. A blob-shaped atom is exactly the null case the ring-vs-Gaussian race exists to adjudicate, so on a sizable fraction of blob atoms the race aborts instead of reporting "Gaussian wins".

### Principled fix

Replace the EM iteration with a Newton (or trust-region Newton) solve on the 4-parameter log-likelihood `(c_x, c_y, R, log s)`:
- The gradient and Hessian are analytic in terms of `I0`, `I1` and `A = I1/I0`, which `circular_gaussian_bessel_terms` already evaluates.
- Terminate on a derived Newton-decrement / stationarity certificate rather than a step size.
- Handle the `R ≥ 0` boundary explicitly. At `R = 0` the likelihood is even in `R`, so decide it with a KKT/fourth-order test on the profiled `R` term, or by comparing against the closed-form `R = 0` fit (isotropic Gaussian).

This removes both magic constants (`4096`, `2e-12`) and converges quadratically in the interior.

Related closed issues: #2264, #2324. Neither covers the EM rate.

SauersML:
Fixes #4003

## Problem
`CircularGaussianFit2d::fit` ran EM with a 4096-iteration cap. EM contracts at the fraction of missing information, and that fraction goes to one as the likelihood flattens in R. On an isotropic blob with a positive moment radius, EM needs 10^3 to 10^5 iterations; the new blob fixture takes about 95 000. Past the cap, `fit` returned Err. `ring_provider_2d` and `ring_bic_2d` pass that error up with `?`, so a single blob aborted the whole shape race.

## Fix
The EM loop is replaced by projected Newton ascent on theta = (c_x, c_y, R, u = ln s), using the exact gradient and Hessian.
- **Derivatives:** per-point derivatives are taken in (r, R, u) and chained to the center through dr/dc = -r_hat and d2r/dc2 = (I - r_hat r_hat^T)/r. Each term is read from a centered Bessel jet term that already holds its own kappa -> 0 limit, so none of them cancels.
- **Step:** the free block is Jacobi-scaled before its eigendecomposition. The floor on u is held only while u sits on it and the score pushes it down. Each eigendirection is ascended with |curvature|.
- **Backtracking:** the step is halved until the likelihood strictly increases. There is no iteration cap. Err is returned only when no representable step moves theta.
- **Certificate:** convergence requires both of these:
  - the free block is negative definite beyond `symmetric_spectrum_rounding_band`;
  - the Newton decrement is within gamma_{d+8} * sum M_i, the rounding band of the pairwise log-likelihood sum.
- **R = 0 boundary:** resolved in closed form. At c = mean and s = E q / 2, the R^2 coefficient of the profile vanishes, and the R^4 coefficient has the sign of (E q)^2 - Var q. A moment R^4 <= 0 is therefore exactly the condition for the Gaussian to be the maximum.
- **Fewer than three distinct locations:** such data is refused. Every circle through two locations fits them with zero residual. On the variance floor, those circles differ only by the O(1) Bessel normalization, which is O(eps) relative to the O(n/s) curvature across them, so no maximizer can be resolved.

gam-math adds `bessel_i0_centered_jet`, which carries A' and x*d1':
- Ascending branch: A' = (1 - A)(1 + A) - (I1/x)/I0, using the paired difference series.
- Asymptotic branch: x^2 A' = 1/2 + P2/S0 - (P1/S0)^2, with the 1/2 separated symbolically.

`bessel_i0_centered_terms` is now a projection of the jet.

## Evidence
A Python prototype of the same algorithm, compared with the old EM:
- About 50 blobs with n = 2000-20000 and a positive moment radius took 2-10 Newton steps, where EM took 424 to more than 100 000. Where EM finished, R and the log likelihood agree.
- 300 random arcs: all certified, at most 21 steps.
- 800 tiny samples with n = 3..6: all certified, at most 39 steps.

## Tests
- A 25-point, 120-digit reference for the jet.
- The 95 000-iteration blob is certified as a local maximum: perturbing each coordinate lowers the likelihood.
- A translated noisy ring is recovered within 5 standard errors.
- A three-point fit passes through all three points.
- The R = 0 branch equals the closed-form Gaussian.
- Two-location data is refused.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 84b67071cfc1.

PR #4335, opened 2026-09-20
latent joint REML: search log-strength hypers on the precision box (#4266)

### LINKED ISSUE #4265: latent joint REML: smoothing rho searches a hand-supplied ±12 box instead of the #2812 resolvability domain

SauersML:
Part of #4266.

This branch is built on `hn39-latent-rho-domain` (PR #4274, fixes #4265) and includes its commit `6c483d3609`. That commit puts the smoothing rho on the #2812 resolvability domain. This PR domains the rest of the latent joint theta.

## Problem

In `try_exact_joint_latent_coord_optimization`, theta past the smoothing block is laid out as `[latent t | analytic-penalty rho | direct hypers]`. Every one of those coordinates searched a hand-supplied `[-12, 12]` box. SPEC: "Hand-supplied search boxes or bounds are never allowed." Two consequences:

- A user `init_log_precision` outside the box, such as `15` or `-30`, started infeasible.
- An analytic log-strength whose representable range is `[-700 - ln|w|, 700 - ln|w|]` was truncated to `±12` for no reason the problem supplies.

## Math

A log-strength coordinate `ρ` scales a penalty as `e^ρ · P`. The design Gram gives it no geometry to project, since `P` is not a smoothing penalty on the design. The #2812 law for such a coordinate is the precision box

`coordinate_domain(None, None) = [ln √ε, ln(1/√ε)] ≈ [-18.02, 18.02]`

At the two faces `e^ρ` is `√ε` and `1/√ε` relative to the declared strength. Past those points, the penalty is either lost below the data curvature in floating point, or it swamps the curvature so that the data term is lost. The same box is already used by `joint_rho_resolvability_domain` for a smoothing coordinate with no projectable geometry.

Coordinates that now search this box:

- **Direct log-precisions:**
  - the anchor's `ln μ`, for `AuxPrior`, `AuxPriorDimSelection` and `IsometryToReference` with `Auto` strength;
  - every ARD `ln α_j`, for `DimSelection`, `AuxPriorDimSelection` and `AuxOutcome`.

  These search `[ln √ε, ln(1/√ε)]`.
- **Analytic coordinates with finite registry faces** (learnable weights and row-precision `log α`). These search `registry faces ∩ [ln √ε, ln(1/√ε)]`. An empty intersection is an `InvalidInput` error.

The auxiliary seed is projected into this domain with `ExactJointHyperSetup::project_rho_seed`, the projection #4265 uses for rho. So a user seed always starts feasible.

The domain is computed after the persistent latent cache has loaded `t`. The latent box `±(max|t₀| + 10)` is therefore taken around the seed the search actually starts from, not around the pre-cache values.

`latent_coord_direct_hyper_count` is now derived from a new `latent_coord_direct_hyper_slots`. It returns one `LogPrecision` or `HeadCoefficient` entry per direct coordinate, in seed order, so the domain and the seed layout come from one source.

## Remainder (left open on #4266)

These coordinates are not log-strengths and still have no derived domain. They keep their previous box.

- **latent `t`:** `±(max|t₀| + 10)`.
- **`AuxOutcome` head coefficients:** `±12`.
- **Parametric row-precision raw-beta and mean offsets** (registry faces `±∞`): `±12`.

## Test filters

- crate `gam-models`, `--lib`, filter `latent_joint_auxiliary_domain_tests`. It runs three tests:
  - `direct_hyper_slots_match_the_seed_layout_4266`
  - `ard_log_precisions_search_the_precision_box_and_seeds_are_feasible_4266`
  - `analytic_log_strengths_search_registry_faces_within_the_precision_box_4266`

---


SauersML:
Merge train RT-1135-1 reviewed this PR at head `061ec56e` and is not merging it yet.

Moving the log-strength hypers onto `rho_domain::precision_box()` is the right direction. The new `latent_joint_auxiliary_domain`, however, keeps the hand-supplied faces and names them as underived:

- `let underived_face = 12.0;` gives a ±12 box for the AuxOutcome head coefficients and for the row-precision raw-beta and mean offsets;
- `max|t₀| + 10.0` bounds the latent coordinates;
- the new tests assert `(lower, upper) == (-12.0, 12.0)` for coordinates 3..7 (and ±11 elsewhere). That pins these magic faces into the test suite, so a later PR that derives them has to delete the test first.

SPEC.md: *"Hand-supplied search boxes or bounds are never allowed."* and *"Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code."*

Principled fix: give each of these coordinates either a domain derived from its own geometry, as this PR already does for the log-strength precision box (for example, the representable range of the quantity the coordinate parameterizes), or no bound at all (an unbounded `coordinate_domain(None, None)`) where nothing bounds it. Then delete the ±12 / +10 assertions.

---

PR #4337, opened 2026-09-20
gpu: route every backend through CachedBackend and KeyedPtxModuleCache

### LINKED ISSUE #4155: gam-gpu: seven kernel backends hand-roll the CachedBackend probe cache and four hand-roll KeyedPtxModuleCache (sphere swallows lock poisoning)
`gam_gpu::backend_probe::CachedBackend` is the one process-wide backend cache ("five kernel families used to spell this `OnceLock<Result<_, GpuError>>` protocol out by hand (#2470)"). Five SAE and survival families use it. Seven backends on origin/main still hand-roll the same protocol, and several also hand-roll a keyed module cache that duplicates `gam_gpu::device_cache::KeyedPtxModuleCache`.

| Backend | File | Hand-rolled |
|---|---|---|
| `BmsFlexGpuBackend` | `crates/gam-models/src/bms/gpu/flex.rs:156` | `static BACKEND: OnceLock<Result<..>>` plus a `probe_linux` helper |
| `RowKernelBackend`, `HvpKernelBackend` | `crates/gam-models/src/bms/gpu/row.rs:848`, `:1861` | `OnceLock` plus `probe_backend_with_compile` |
| `CubicCellGpuBackend` | `crates/gam-models/src/gpu_kernels/cubic_cell/device.rs:66` | `OnceLock` plus a `Mutex<HashMap<usize, Arc<CudaModule>>>` module cache |
| `RowOpsBackend` | `crates/gam-models/src/gpu_kernels/row_hessian_ops.rs:268` | `OnceLock` plus an inline copy of the `probe_cuda_backend` prologue (`GpuRuntime::require` + `cuda_context_for` + `default_stream`) |
| `PirlsRowBackend` | `crates/gam-solve/src/gpu_kernels/pirls_row.rs:316` | `OnceLock` plus a `Mutex<HashMap<ModuleKey, ..>>` module cache |
| `SphereGpuBackend` | `crates/gam-terms/src/basis/sphere_gpu.rs:839` | `OnceLock` plus a `Mutex<HashMap<S2ModuleCacheKey, ..>>` module cache |

This produces the following defects.

- **Duplicated logic.** Seven copies of the probe-once, cache-the-refusal protocol exist next to the shared type written to replace them. `row_hessian_ops` re-implements the prologue that `backend_probe`'s module doc says no backend re-implements ("no backend re-implements the prologue"). Its absence error is also a different variant (`RequiredDeviceUnavailable`, unlabelled) from every other backend's labelled `DriverLibraryUnavailable`.
- **Four keyed module caches duplicate `KeyedPtxModuleCache`**, and they disagree with each other:
  - `sphere_gpu` guards both map accesses with `if let Ok(guard) = ..lock()`. After a poisoning panic it recompiles through NVRTC on every call and never caches, and it reports nothing (silent error swallowing).
  - `cubic_cell` and `pirls_row` turn a poisoned lock into a hard `DriverCallFailed`.
  - `pirls_row` `insert`s, so a racing compile replaces the module other callers already hold. The others use `or_insert`.
- **Dead code.**
  - `probe_backend_with_compile` is `build(&probe_cuda_backend(label)?)`. Its only callers are the two hand-rolled `row.rs` caches. It has no doc of its own: its doc paragraph is stranded above `CachedBackend`'s doc, so rustdoc gives `CachedBackend` both.
  - `PirlsRowBackend::compiled()` has no caller.
  - `SphereGpuBackend` keys its per-backend cache on the device's compute capability. The backend is one process-wide value bound to one device, so `cc_major`/`cc_minor` are constant for every key in the map. `lmax` is the only varying key.

**Fix:**
- Declare every one of these backends as `static BACKEND: CachedBackend<_>` read through `get_or_probe(label, build)`. This includes `row_hessian_ops`, which then gets the shared prologue and its labelled error.
- Move the per-key caches of `cubic_cell` (degree), `pirls_row` (family/curvature/mode) and `sphere` (lmax) onto `KeyedPtxModuleCache`.
- Delete `probe_backend_with_compile`, `PirlsRowBackend::compiled`, the `probe_linux` helpers and the sphere compute-capability key fields.
- Correct the `backend_probe` docs.

SauersML:
Fixes #4155

Seven GPU backends each wrote their own `OnceLock<Result<_, GpuError>>` probe-once protocol by hand. They now all go through `gam_gpu::backend_probe::CachedBackend`. The affected backends are BMS flex, BMS row, BMS row HVP, cubic-cell, row-Hessian, PIRLS row and sphere.

- **Backend cache.** Each backend is a `static CachedBackend<_>` read through `get_or_probe(label, build)`.
  - `row_hessian_ops` drops its inline copy of the probe prologue (`GpuRuntime::require` + `cuda_context_for` + `default_stream`).
  - It now reports the same labelled `DriverLibraryUnavailable` as every other backend when no device is present.
- **Keyed module caches.** Three backends move their per-key caches onto `KeyedPtxModuleCache`:

  | Backend | Cache key |
  |---|---|
  | `cubic_cell` | `max_degree` |
  | `pirls_row` | mode / family / curvature, through a new `Display` for `ModuleKey` |
  | `sphere` | `LMAX` |

  This fixes three defects:
  - `sphere` kept its own `if let Ok(guard)` poison bypass: after a poisoned lock it recompiled on every call and never cached. It now uses the shared cache, whose poison recovery lands in #4106.
  - `sphere` also keyed on the compute capability of the single device the backend is bound to, so that part of the key never changed.
  - `pirls_row` used `insert`, so a racing compile replaced a module other callers were already holding.
- **Removed:**
  - `probe_backend_with_compile`
  - the public `probe_cuda_backend` re-export (it is now private behind `get_or_probe`)
  - `CudaBackendParts::capability`, which no caller read
  - `PirlsRowBackend::compiled`, which had no caller
  - the `probe_linux` helpers
  - `S2ModuleCacheKey` and `SphereGpuBackend::cc`
- **Docs.** The `backend_probe` module docs are corrected, and the stale doc paragraph above `CachedBackend` is gone.

CHANGELOG, the source-removal ledger and the public-API ledger have entries for this change. `source_removal_guard` and `public_api_census` pass against base f10039cc. `test_census` fails only on the pre-existing `crates/gam-solve/src/reml/glm_outer_hessian_fd_tests.rs: test attribute has no named function`, which main also fails.

Focused runs on a177594f: 35506178709 (gam-gpu), 35506182365 (gam-models), 35506183529 (gam-solve), 35506184950 (gam-terms).

This PR should merge after #4106. The CHANGELOG's poison-recovery sentence depends on it and will be checked at merge time.

---


SauersML:
Taking over from HN-22 (no activity since 10:50Z). The branch conflicts with main only in CHANGELOG.md and the two removal ledgers (docs/public-api-census-changes.json, docs/source-removal-changes.json). No code overlaps, and the #4155 base f10039cc is on main. I am merging origin/main in and keeping both sides of each conflict; both JSON files parse. The queued HN-22-4155-1..4 runs cover the unchanged code on the pre-merge head. Per the body, this lands after #4106.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 49199653ed7a.

Issue #4351, opened 2026-09-20
gam-sae response/compose: discarded-unit covariance is E[g(x)g(y)] − E g(x)·E g(y) by raw subtraction (same cancellation as #4124)


SauersML:
Follow-up to #4124, which has the same defect in the multinomial posterior.

## Where

`crates/gam-sae/src/response/compose.rs:812` (origin/main 298c4af57e):

```rust
covariance_row[other] =
    second_moment - self.unit_means[[point, unit]] * self.unit_means[[point, other]];
```

`second_moment` comes from `pair_moments(activation, PreactivationPair{..})`. It is the raw `E[g(X)g(Y)]` over the discarded-direction Gaussian. `unit_means` is `E g(X)`.

## Problem

The covariance `Cov(g(X), g(Y))` is recovered by subtracting two raw moments of order `E[g]²`. Once the discarded variances are small next to the unit means, both terms agree to about `u·|E g(X)|·|E g(Y)|`. The difference is then rounding, not covariance, and its sign carries no information. `covariance_rounding` bands the *pair-moment* formation. It does not band this final subtraction, whose error scales with `|E g|²` and not with the covariance.

SPEC: "In general, do not paper over solver issues." A covariance near the rounding floor should be computed centred, not gated.

## Fix direction

This is the same fix as #4124.
- Assemble the covariance centred: for example, `pair_moments` returns the centred moment `E[(g(X) − E g(X))(g(Y) − E g(Y))]` from its own quadrature or closed form.
- Or use the law of total covariance or a delta-type expansion where the discarded variance is small.
- Band the result by its derived formation error rather than by `|E g|²·u`.

A test should compare against the delta method `g'(μx) g'(μy) Σxy` at a discarded variance of about `1e-16`. There, the raw subtraction is pure rounding.

SauersML:
Claimed by wave8 session (claude/fix-4351-centred-pair-covariance); draft PR incoming.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #4362, opened 2026-09-20
Multinomial posterior: centred class covariance with derived rounding bands (#4124)

### LINKED ISSUE #4124: multinomial posterior class covariance is raw E[pp'] − E p E p' behind a 16·ε·k envelope (the #4086 defect in multinomial_posterior and compose)
## Where (origin/main 6ff5fb0185)
- `crates/gam-models/src/multinomial_posterior.rs:549` `integrate_binary`:
  ```rust
  let probability_second_moment = probability_mean - mean_logistic_slope;   // E p − E p(1−p)
  let variance = (probability_second_moment - probability_mean * probability_mean).max(0.0);
  ```
- `moments_from_raw` (L1880): `raw_moments[packed] - means[row] * means[column]` over the accumulated raw second moments. After that, `remove_covariance_roundoff` gates the negative eigenvalues at `integration_error + PSD_BACKWARD_ERROR_MULTIPLIER(=16)·ε·k·scale` and clips them.
- The three-class conditional rule (L853-872) builds its inner `E[L²|Y] = a − d` in the same way.
- `crates/gam-sae/src/response/compose.rs:812`: `second_moment - unit_means[u]·unit_means[o]` (closed-form `E[g(x)g(y)] − E g(x)·E g(y)`).

## Why
The covariance is recovered by differencing raw moments of size `p²`. Its absolute resolution is therefore `~ε·p²`, not `~ε·Var p`.

Take the binary case at μ = 0, where `Var p ≈ s²/16` for logit SE `s`:
- `s = 1e-7`: `Var p = 6.25e-16` against `ulp(0.25)/2 = 2.8e-17`, plus the series' own `ε`-relative error in `E p(1−p)`. That is about a 5-10% error.
- `s ≤ 1e-8`: the variance is lost entirely, since it rounds to 0 or to one ulp.

The practical exposure is small, because a logit SE below 1e-7 requires an enormous n. The structural defects are:
- The covariance's non-negativity is enforced by a clamp.
- The raw-moment cancellation is gated by a fixed multiplier `16`.

Neither is guaranteed by construction.

## What a root fix needs
- **Tensor Gauss–Hermite path:** its weights are positive, so a Chan–Golub–LeVeque merge of `(weight, mean, centred)` per node is non-negative by construction, as in the #4086 fix.
- **Smolyak path:** it has negative weights, so the centred merge does not guarantee PSD. A negative eigenvalue there is quadrature error that the level difference already measures, not roundoff. The gate should be exactly the measured `integration_error`, with no `16·ε·k` term. This needs moments accumulated as deviations from the plug-in `p₀ = softmax(μ)`, so that the roundoff is `ε·|p − p₀|²` rather than `ε·p²`.
- **Binary and three-class exact series:** these need a centred scalar integral `E[(σ(η) − m)²]` from the same accelerated series (`logit_posterior_meanwith_deriv_exact`), rather than `m(1−m) − E σ'`.

Sibling sweep of #4086 (see https://github.com/SauersML/gam/issues/4086#issuecomment-5749255277).

SauersML:
Fixes #4124

## Problem

The multinomial posterior class covariance was the raw `E[pp'] − E[p]E[p]'`. That difference cancels to rounding once the covariance falls below `u·E[p]²`. A `16·ε·k` envelope (`covariance_roundoff_tolerance`) then clamped whatever negative spectrum survived. The binary path had the same defect through the identity `m − m' − m²`, followed by `.max(0.0)`.

SPEC: "In general, do not paper over solver issues." SPEC: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code."

## Fix: assemble every covariance centred

**Scalar (gam-solve `quadrature.rs`).** The new `logit_posterior_variance(μ, s)` returns `Var σ(μ + sZ)`.
- The CRVZ series gains `E[σ²]`, from the same polynomial `Q ≈ 1/(1+t)` and its derivative, split at `η = 0`.
- The raw `E[σ²] − E[σ]²` is clamped into the Gaussian-chaos (Hermite) bracket:
  ```text
  B_lo = s²(E σ')²                                  (n = 1 term, exact lower bound)
  Var − B_lo = Σ_{n≥2} s^{2n}(E σ⁽ⁿ⁾)²/n! ≤ s⁴·σ'(μ)²·e^{2s²}·Φ(2s)
  ```
  The bound uses `|σ''| ≤ σ'` and `σ'(η) ≤ σ'(μ)e^{|η−μ|}`. Both ends come from the same series.
- The result is non-negative by construction, with relative error at most about `min(u/s², s²)`.
- The variance is even in `μ`, so the series runs at `−|μ|`.
- For `s ≤ √ε` it returns `(s·σ'(μ))²`, which is exact to `O(s²)`.

**Binary.** Uses `logit_posterior_variance` directly.

**Three classes (conditioned rule).** Given `Y`, `p = m(Y) + (L − a)u(Y)`, so the law of total covariance applies:
```text
Cov(p) = E_Y[v·uu'] + Cov_Y(m),   v = Var(L|Y) centred
```
- The between term is a corrected two-pass centred sum (Björck §2.2.3) about the rule's own first moment, with the residual `(Σwδ)(Σwδ)'/W` removed.
- Welford is not used: its error carries a `‖mean‖/sd` factor that a trace band cannot bound.
- The band is `compensated_band(6, (Σ|w|v‖u‖² + Σ|w|‖δ‖²)/W + ‖Σwδ‖²/W)`. The entrywise majorant is PSD, so its trace bounds the spectral norm of the rounding.

**Tensor / Smolyak rules.** The rules accumulate softmax deviations `d = p − p⁰` from `p⁰ = softmax(μ)`.
- Each node forms `d` in whichever of two forms has the smaller rounding scale:
  - the exact expm1 form `d_c = q_c(T e_c − A)/(S T)`;
  - the direct form `p − p⁰`.
- The raw moments that the level certificate compares are reassembled as `p⁰ + E d` and `p⁰p⁰' + p⁰Ed' + Edp⁰' + Edd'`, so certification is unchanged.
- Covariance = `E[dd'] − E[d]E[d]'`.
- Band = `compensated_band(3, Σ|w|‖d‖²/W) + 2‖Ed‖Δ + Δ² + u(energy + 2‖Ed‖²)`, where `Δ = compensated_band(2, Σ|w|‖d‖₁/W)`.

**Output gate (`remove_covariance_roundoff`).** The allowed negative eigenvalue is the sum of four derived terms:
- the certified integration error;
- the rule's assembly band;
- `simplex_projection_rounding_band`, which is `k·4(γ₃ + γ_{k+1})·max|C|`;
- `symmetric_spectrum_rounding_band(eigs)`.

`covariance_roundoff_tolerance` no longer appears in the covariance path. It remains only for the input active-covariance PSD check and the simplex-sum check on the means, which are separate issues.

## Same defect elsewhere

`gam-sae` `response/compose.rs:812` forms `second_moment − E g(x)·E g(y)` by raw subtraction. This is filed as #4351.

## Tests

- **gam-solve**
  - `test_logit_posterior_variance_is_centred_at_small_sigma_4124`: `s²/16` at `s = 1e-7` and `1e-8` (relative 1e-10); the two-term expansion at `s = 1e-3`; the lognormal deep tail at `μ = ±30`.
  - `test_logit_posterior_variance_matches_independent_references_4124`: the identity at `s = 1`, and a centred adaptive-Simpson quadrature.
  - `test_logit_posterior_variance_is_even_and_bounded_4124`: bit-exact evenness, `[0, ¼]`, and positivity.
- **gam-models**
  - `binary_reduction_matches_controlled_logistic_normal_identity`: now checks against the centred variance, and checks that it agrees with the raw identity at `s = 0.8`, where nothing cancels.
  - `binary_variance_is_centred_at_small_posterior_width_4124`: `s²/16` at `V = 1e-14`, relative 1e-10.
  - `narrow_posterior_class_covariance_matches_delta_method_4124`: K = 3 (conditioned rule) and K = 4 (general rule) at `‖V‖ ≈ 1e-16`, against `J V Jᵀ` at relative 1e-6. All class variances must be positive. That covariance is below the raw-difference rounding floor.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 8239da84f0e7.

PR #4368, opened 2026-09-20
gam-solve: fused Arrow-Schur module cache is the shared KeyedPtxModuleCache

### LINKED ISSUE #4344: gam-solve: arrow-schur fused module cache hand-rolls KeyedPtxModuleCache, bypasses a poisoned lock and swallows load failures
`crates/gam-solve/src/gpu_kernels/arrow_schur.rs` (`mod cuda`, lines 3081-3130 on 2ac61090) keeps its own NVRTC module cache for the fused forward kernel: `struct FusedModuleCache { modules: Mutex<HashMap<FusedModuleCacheKey, Arc<CudaModule>>> }` behind `fused_module_cache()` and `fused_module_for()`. It re-implements `gam_gpu::device_cache::KeyedPtxModuleCache`, the shared type whose job is exactly this ("A per-key family of compiled modules: one [`CudaModule`] per value of a kernel-shape parameter that is baked into the source"). The copy also differs from the shared type in three ways.

- **A load failure is swallowed.** `ctx.load_module(ptx).map_err(|_| ArrowSchurGpuFailure::Unavailable)` drops the driver's error and reports it as the typed "no device" decline. The file's own contract at line 68 says: "The existing GPU failure surface has a diagnostic string variant, so faults flow through it; only typed Auto/Off absence remains `Ok(None)` and may become `Unavailable` at a caller-specific gate." A compile failure in the same function goes through `SchurFactorFailed` with the message, but a load failure does not.
- **A poisoned lock is silently bypassed.** Both map accesses are wrapped in `if let Ok(guard) = cache.modules.lock()`. After any panic while the lock is held, every call recompiles through NVRTC, never stores the result, and reports nothing.
- **Two key fields never vary.** `FusedModuleCacheKey` keys on `(cc_major, cc_minor, p_max, r_template)`, and its doc says "The CC pair lets one host process drive multiple device generations". But both call sites (`solve_fused`, `build_schur_matvec_backend`) take the capability from `route_through_gpu(..)`. That returns the one process-wide `&'static GpuRuntime` (`GpuRuntime::resolve`), so the compute-capability pair is the same for every key in the map. The doc describes behaviour the code does not have.

**Fix:**
- Make the cache a `static KeyedPtxModuleCache<FusedModuleCacheKey>`.
- Report compile and load faults as `SchurFactorFailed` carrying the driver's message.
- Drop `cc_major`/`cc_minor` from the key and add a `Display` for the key.
- Delete `FusedModuleCache` and `fused_module_cache`.

The poison handling of the shared type itself is fixed in #4106 (`lock_modules` recovers the poisoned map instead of bypassing it).

SauersML:
Fused Arrow-Schur compilation now uses the shared keyed PTX-module cache. Compiler and module-loader failures preserve their diagnostics in `SchurFactorFailed`; the private cache/accessor and invariant compute-capability key fields are removed. The key retains the two dimensions that change the generated kernel.

This incorporates the poison-recovery dependency in #4106 first. The original draft alone did not fix poisoned-lock bypass in the shared owner.

Validation on MSI:
- All 27 production crates pass coherent metadata compilation with warnings denied.
- The complete GPU unit binary builds in 3.28 seconds, and the actual shared-map poison/reacquisition regression passes.
- Both live fused consumers use the singleton runtime context; source generation and shared compile options are unchanged. Source-removal guards record the deleted private owner.

GPU hardware execution is not claimed.

Fixes #4344


SauersML:
Merge train MT-0: merged origin/main into `claude/hn22-arrow-schur-fused-cache` (merge commit 258d6ec7, no rebase or force-push). The only conflict was `docs/source-removal-changes.json`. Both sides' entries were kept (main's list plus this branch's new entry), and the result validates with `python3 -m json.tool`. The code diff against main is identical to 5075e476. Queued for the next local integration train, with focused filter `cargo test -p gam-solve --lib arrow_schur`.

---

PR #4367, opened 2026-09-20
gam-sae: pinned behavior chart integrates arc length under the Simpson contract

SauersML:
Fixes #4317

## Defect

`behavior_pinned_chart` says it uses "the same public arc-length integration resolution used by the activation chart canonicalizer, so the two quotient reads have one numerical contract". It borrowed only the cell count, though. The contract it cites (`ARC_LENGTH_GRID_CELLS`, implemented by `unit_speed_reparameterization`) is composite Simpson on node, midpoint and next node, with rows read through the exact integral of each cell's quadratic interpolant. The behavior chart did three things differently:
- It sampled only the nodes and integrated them with the trapezoid rule, which is O(h²) globally.
- It read rows by linear interpolation of the cumulative sum.
- It guarded the interval domain with `hi > lo` instead of `arc_length_grid_resolves`.

## Math

For a speed s ∈ C⁴ on cells of width h, the errors are:
- Trapezoid: E = −h³ s''(ξ)/12 per cell, so O(h²) globally.
- Composite Simpson `h(f₀ + 4f_m + f₁)/6`: E = −h⁵ s⁗(ξ)/2880 per cell, so O(h⁴) globally.
- Reading a row at offset x inside a cell with `partial_cell_arc`, the exact integral over [0, x] of the quadratic through (f₀, f_m, f₁): the interpolation error is O(h³) pointwise, so the partial integral is O(h⁴).

A linear read of the cumulative sum instead adds the O(h²) chord error of the trapezoid.

With h = 0.3/2048, the ratio is h² · 1/12 against h⁴ · 1/2880, about 6 orders of magnitude, which is what the measurement shows.

## Change (`behavior_isometry.rs`, and `pub(crate)` on the two helpers in `chart_canonicalization.rs`)
- Sample the Simpson grid: node `j` at sample `2j`, midpoint at `2j+1`, 2·cells + 1 samples in total.
- Accumulate the Simpson cell weights.
- Read rows with `partial_cell_arc`, using the same local-coordinate, cell and offset map as `unit_speed_reparameterization`, for both the circle and the interval.
- Guard the interval with `arc_length_grid_resolves`, so all three arc-length reads refuse the same charts.
- Run the anchor (nearest point to the basepoint) and orientation searches on node samples, where the cumulative arc is tabulated.

## Data

[...]

PR #4370, opened 2026-09-20
gam-sae manifold/behavior_isometry: integrate the behavior-pinned chart on the shared Simpson arc-length contract

### LINKED ISSUE #4317: gam-sae manifold/behavior_isometry: pinned behavior chart integrates arc length with trapezoid+linear interpolation instead of the shared Simpson contract
**Where:** `crates/gam-sae/src/manifold/behavior_isometry.rs`, `behavior_pinned_chart`. This is the production path `atom_behavior_isometry` → `AtomBehaviorIsometry::behavior_pinned_chart`. It is exported through `behavior_entry.rs` as `BehaviorPinnedChartWire`.

**Defect (a violated shared contract, and lower-order math).** The docstring says the pinned chart uses "the same public arc-length integration resolution used by the activation chart canonicalizer, so the two quotient reads have one numerical contract".

The contract it borrows is `ARC_LENGTH_GRID_CELLS`, documented in `chart_canonicalization.rs`: "Each cell carries a node, a midpoint, and the next node (composite Simpson) … the per-cell quadrature error is `O(Δu⁴)`". `unit_speed_reparameterization` implements exactly that:
- It samples `2·cells + 1` points.
- It uses the Simpson cell weight `h(f0 + 4fm + f1)/6`.
- It reads rows through `partial_cell_arc`, the exact integral of the cell's quadratic interpolant.

`behavior_pinned_chart` uses the same cell count, but does two things differently:
- It samples only the `cells + 1` nodes and integrates them with the **trapezoid rule** (`O(h²)`).
- It reads rows by **linear interpolation** of the cumulative sum.

Its interval-domain guard is also `hi > lo` instead of the shared `arc_length_grid_resolves`. That guard exists so that "the unit-speed defect and the arc-length reading take the same domain".

**Size.** Take a smooth two-harmonic behavior arc: `PeriodicHarmonicEvaluator::new(5)`, decoder rows `[1, s1, c1, s2, c2]` = `[[0.05,0],[0,0.3],[0.3,0],[0.1,0],[0,0.08]]`, rows on `[0.1, 0.4]`. Measured against an adaptive-quadrature reference:
- `behavior_length` relative error is `−6.9e-8` with the trapezoid rule and `1.5e-14` with Simpson.
- Per-row arc error is `4e-8` to `1e-7` with trapezoid plus linear interpolation, and `≤ 1e-14` with Simpson plus `partial_cell_arc`.

So the behavior quotient coordinate is about 6 orders less accurate than the activation quotient it claims to share a contract with.

**Fix.**
- Sample the Simpson grid (node `2j`, midpoint `2j+1`).
- Accumulate the Simpson cell weights.
- Interpolate rows with `partial_cell_arc`.
- Use `arc_length_grid_resolves` for the interval guard.
- Search the anchor and orientation on node samples (`2i`).
- Expose the two helpers as `pub(crate)` in `chart_canonicalization.rs`.

**Regression test.** `behavior_pinned_chart_arc_length_matches_the_simpson_contract` compares `behavior_length` and the pairwise pinned-coordinate distances against a fine composite-Simpson reference, at `1e-11` relative.

SauersML:
Fixes #4317

## Defect
`behavior_pinned_chart` uses `ARC_LENGTH_GRID_CELLS`, but it deviates from the arc-length contract it claims to share with the activation canonicalizer:

| | `behavior_pinned_chart` | Shared contract |
|---|---|---|
| Integration | trapezoid rule over the nodes only | composite Simpson (node, midpoint, node) |
| Row read | linear interpolation | `partial_cell_arc` |
| Interval guard | `hi > lo` | `arc_length_grid_resolves` |

On a smooth harmonic arc, the trapezoid read is off by about 1e-7 of the behavior length. Simpson is exact to round-off.

## Fix
- Sample the Simpson grid and accumulate the Simpson cell weights.
- Interpolate rows with `partial_cell_arc`.
- Guard the interval with `arc_length_grid_resolves`. Both helpers become `pub(crate)` in `chart_canonicalization`.
- The anchor and orientation searches read the node samples.

## Test
A regression test checks the result against a fine Simpson reference.

Focused proof: key HN-1-4317-1.

---


SauersML:
I closed this PR because #4367 supersedes it. Both fix #4317 with the same production change: the behavior-pinned chart moves onto the composite-Simpson grid (node, midpoint, node) and reads fitted rows through `partial_cell_arc`, with `arc_length_grid_resolves` as the interval gate. I kept #4367 for three reasons.

- **Row read.** #4367 maps rows with the canonicalizer's own map, character for character: `local = rem_euclid(hi-lo)` or `(coord-lo).clamp(0, hi-lo)`, then `cell = floor(local/step).min(cells-1)` and `x = local - cell*step`. This PR adds a further `x.clamp(0, step)`, so the two quotient reads would drift apart on the last cell.
- **Test.** #4367 checks every pairwise arc distance against a fine-grid reference and adds a trapezoid positive control. The test here checks offsets from the first row only.
- **Math and data.** #4367's body has the O(h²) versus O(h⁴) derivation and the before/after data. The row-placement error goes from −6.9e-8 to 1.6e-14.

This PR has nothing #4367 lacks, so nothing needs carrying over. I cancelled its queued proof run, 35508591009.

---

PR #4371, opened 2026-09-20
gam-sae manifold: normalize row loss weights over the live rows

### LINKED ISSUE #4330: gam-sae manifold: set_row_loss_weights normalizes over zero-weight rows; structure-search 0/1 estimation mask inflates dispersion and data weight by n/n_live
## Defect

`SaeManifoldTerm::set_row_loss_weights` (crates/gam-sae/src/manifold/construction.rs, the normalization at the end) divides the weights by their mean over **all** `n` rows, zero-weight rows included:

```rust
let mean = weights.iter().sum::<f64>() / weights.len() as f64;
self.row_loss_weights = Some(weights.into_iter().map(|w| w / mean).collect());
```

The rest of the term treats zero-weight rows as outside the fitted sample. `fitted_response_scalar_counts` (construction_fitted_response.rs) counts only the positive-weight rows ("A zero-weight row is excluded from estimation, so it carries no scalar in either frame"). So with `n_live < n` live rows, every live row carries weight `n / n_live` rather than 1, and `Σw = n` rather than `n_live`:

- `likelihood_dispersion = 2·data_fit / (n_live·channels)` and the non-whitened `raw_output_noise_variance` are inflated by `n / n_live`.
- The whitened raw branch of `reconstruction_residual_energies` sums **unweighted** `r²` over the live rows. Under a whitening metric the two frames are therefore on different mass scales.
- The data-fit versus penalty balance, and the ordered Beta–Bernoulli `M_k = Σ w z`, `N_eff = Σ w`, are all scaled by `n / n_live` on the live rows.

## Production path

`structure_harvest.rs` `refit_at`: the universal-inference estimation/evaluation split builds a 0/1 mask (`weights[r] = 1.0` on the estimation rows, 0 elsewhere) and calls `set_row_loss_weights`. The `RowBlockShard` doc states the contract: "a candidate is refit with the currently-held-out shards' rows at weight `0` ... and the estimation rows at weight `1`". The stored weights are in fact `n / n_est` on the estimation rows. `fit_seed` user design weights may also contain exact zeros. The doc on `set_row_loss_weights` says "Exact zeros represent rows excluded by a designed estimation split", so zeros are explicitly supported.

SPEC: "The default choices should give the best model for the user." A held-out row should not change the fit on the rows that remain, but here the dispersion and the smoothing balance of the estimation fit depend on how many rows are held out.

## Fix

Normalize by the mean over the live rows, `Σw / #{w > 0}`. A 0/1 mask is then stored as exactly 0/1, and `Σw = n_live` agrees with `fitted_response_scalar_counts`. All existing callers that pass strictly positive weights are unchanged, since for them `n_live = n`. The docs on `set_row_loss_weights`, `row_loss_weights`, the `row_loss_weights` field (term.rs) and `ard_value` also need updating.

SauersML:
Fixes #4330

## Defect
`set_row_loss_weights` divided the design weights by their mean over all rows, including zero-weight rows. `fitted_response_scalar_counts`, however, counts only the positive-weight rows as the fitted sample. Each live row therefore carried weight n/n_live instead of 1.

That inflated two quantities by n/n_live:
- the likelihood dispersion (and the non-whitened raw noise variance),
- the data-versus-penalty balance.

The structure-search estimation split hit this. Its `RowBlockShard` contract puts the estimation rows at weight 1.

## Fix
Normalize over the positive-weight rows, so a 0/1 mask is stored as exactly 0/1. Callers with strictly positive weights are unchanged.

## Test
`zero_one_estimation_mask_prices_the_live_rows_at_unit_weight` checks that a masked term's dispersion is the live-row RSS divided by the live scalar count.

Focused proof: key HN-1-4330-1.

---


SauersML:
Taken over by BF-104 (HN-1 handed off). The diff looks sound: row weights are normalized by the live-row count, which matches `fitted_response_scalar_counts`. I'm waiting on proof run 35508592114 for head c3ffd1db and will merge on green.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 33cc059a8ef8.

PR #4372, opened 2026-09-20
gam-terms: the small-χ Duchon chart must include the kernel's analytic part (#4135)

### LINKED ISSUE #4135: gam-terms basis: closed_form_anisotropic_pair_block is indefinite on the kernel-constrained space (min_eig −18% of scale at q=1 d=3 m=1 s=1 κ=0.7)
## Where

`crates/gam-terms/src/basis/matern_kernel.rs:2475` (`closed_form_anisotropic_pair_block`, via `closed_form_anisotropic_pair_block_with_origin` with `PairOrigin::Full` and `closed_form_anisotropic_pair_value_with_powers`).

It was exposed by `tests/src_modules/smooths/basis_duchon_matern_jet_derivative_tests.rs::test_pair_block_psd_in_convergent_regime` once #3677 was fixed.

## What happens

The fix for #3677 (PR #3683, head `993a28af`) gives every case a 4-dimensional constrained space: `k = d + 1 + 4` centers, `Z` = the null space of the linear side condition. Before that, `k` was fixed at 6. The focused run https://github.com/SauersML/gam/actions/runs/35499752705 (package gam-terms, target lib) fails on the **first** case, which the old fixture was already checking on only a 2-dimensional constrained space:

```
constrained pair-block not PSD: q=1 d=3 m=1 s=1 κ=0.7 min_eig=-1.967e-2 tol=1.110e-10
```

`tol = 1e-9·scale + 1e-12` puts the spectral scale at about 0.11. So `Zᵀ G Z` has an eigenvalue of about **−18% of its largest**. That is not round-off.

## Why it is wrong

With `PairOrigin::Full`, the pair block is `G_ij = ⟨D^q K(·−c_i), D^q K(·−c_j)⟩_{L²}` for the hybrid Duchon–Matérn kernel. In the convergent regime this is the Gram matrix of the functions `D^q K(·−c_i)`. The case satisfies `2(m+s) ≤ d+2q < 4(m+s)`: 4 ≤ 5 < 8. A Gram matrix is PSD on all of ℝᵏ, and so on every subspace. The kernel-constrained block `Zᵀ G Z` in particular cannot have a negative eigenvalue.

A −18% eigenvalue means the matrix does not equal that Gram matrix. Either the off-diagonal pair values (`duchon_pair_kernel_with_powers` at `r ≠ 0`) or the self-pair diagonal (`closed_form_penalty::self_pair_bundle`, the exact distributional diagonal for this strip) is wrong for q=1, d=3 (odd `d+2q = 5`), m=1, s=1, κ=0.7 with anisotropic η. An indefinite penalty block feeds REML a non-PSD `S_λ`. That breaks the `log|S|₊` and Laplace terms, and the solver would have to paper over it.

This is not the #3545 defect. #3545 concerns the pure-Duchon (κ = 0) self-pair epsilon. This case is hybrid (κ = 0.7) and takes the analytic self-pair bundle.

## Principled fix

1. Check `self_pair_bundle(q=1, m=1, s=1, κ=0.7, η)` and the `r ≠ 0` pair kernel against a direct numerical evaluation of `∫ |ξ|^{2q} · |K̂(ξ)|² · cos(ξ·r) dξ` for this case. That is the spectral form of the L² inner product of `D^q K`. Then correct whichever side disagrees.
2. The test in #3683 needs no change: it fails because the block is wrong. Do not loosen `tol` or drop the case. Once the closed form is fixed, the focused run on #3683 should go green as it stands.

## Evidence

- Run: https://github.com/SauersML/gam/actions/runs/35499752705 (log line `basis_duchon_matern_jet_derivative_tests.rs:2618:13`)
- The first failing assertion stops the test, so it is unknown whether cases 2–4 (d = 5 and 7, which previously checked a 0x0 block) also fail.

SauersML:
Fixes #4135

## Root cause
For κR ≤ 1/8, `duchon_small_chi_riesz_series_radial_derivatives` evaluated the hybrid Duchon kernel f = F⁻¹[ρ^{-2a}(κ²+ρ²)^{-b}] as its Riesz tail alone:

S = Σₙ (−1)ⁿ C(b+n−1,n) κ²ⁿ R_{a+b+n}

The kernel is actually S + E, where E = Σᵢ eᵢ(κ) R²ⁱ is the analytic part fixed by the low-frequency symbol. The partial-fraction and Schwinger charts evaluate all of f. So the kernel jumped at the chart switch, and a pair block with pairs on both sides of it was indefinite. For example, with d=3, κ=0.7 the chart gave g₁(0.1) ≈ 3.3e-6 where the true value is 0.11591.

## Fix
The chart now adds E into the same compensated sums.
- **Odd d:** Mellin–Barnes residues.
  - e₀ = Γ(a+b−d/2)Γ(d/2−a)/(Γ(b)4^{d/2}π^{d/2}Γ(d/2)) κ^{d−2a−2b}
  - The exact term ratio is e_{i+1}/e_i = −(κ²/4)(i+d/2−a)/((i+1)(i+d/2)(a+b−1−d/2−i)).
- **Even d:** the pure R²ⁱ Taylor coefficients of the partial-fraction blocks, less the log-Riesz constants that S already carries. Those coefficients are:
  - the log-Riesz finite-part constants;
  - the Matérn blocks' pure coefficients, taken from `duchon_matern_block_taylor_r2j_triplet`.
- **κ partials:** each eᵢ is κ^x times a polynomial in ψ = ln κ, so the κ partials come from its exact ψ-derivatives.
- **Truncation:**
  - uses the existing compensated band;
  - starts once R²ⁱ reaches every requested radial order;
  - skips structural zeros of the κ partial;
  - returns NaN if the term budget runs out.
- `log_riesz_coefficient` now holds the log-Riesz normalization that `riesz_block_radial_derivatives` used to inline.

## Tests
Three tests pinned the Riesz tail alone. They now pin the full kernel:
- The shared-chart test adds the odd-d E, in direct Gamma form, to its oracle.
- The d=4 small-κ test now compares against the partial-fraction Green's function. Its 3.1e3 cancellation bounds the error near 3e-12, and the test tolerance is 1e-11.
- The d=5 κ→0 test now checks the IR expansion κ⁵g/C₀ = 1 − κ²R²/10 to 1e-12. The first omitted term is 5.1e-14.

The existing `test_small_chi_chart_is_the_kernel_not_its_riesz_tail` is unchanged.

Validation: focused-rust-proof run BF-46-4135-1.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 4d05f06ba622.

PR #4373, opened 2026-09-20
gam-sae manifold/steering: lift steer_rows/steer_decode through the Tier-0 scale

### LINKED ISSUE #4332: gam-sae manifold/steering: steer_rows/steer_decode return the internal Tier-0 frame (σ not applied) while try_fitted and steering tangents are raw
## Defect

`SaeManifoldTerm::steer_rows` and `steer_decode` (`crates/gam-sae/src/manifold/steering.rs`) return `a·(Φ_k(t⊕δ) − Φ_k(t))·B_k` and `a·Φ_k(t⊕δ)·B_k` with the atom's stored decoder. On a Tier-0 standardized term, that decoder lives in the fit frame `(Z − μ)/σ`. The steering surfaces never apply σ, so they return internal-frame vectors. Every other output of the same term is in raw activation units:

- `try_fitted` lifts through `add_tier0_mean_inplace` (`x̂ = μ + σ ⊙ x̂_int`).
- `decode_at` and `decode_tangents_at` in `inference/steering.rs` multiply by `model.tier0_scale()`. Their doc states the rule: *"Every steering quantity that meets the metric (chords, tangents, doses) must therefore be mapped back to raw units, `g_raw[c] = σ_c·g_int[c]` (gh#2249 calibration confound)"*.
- The `steer_rows` doc says *"The caller adds this to the ambient activation `x`"*. That `x` is raw.

## Production path

`fit_entry.rs` (`standardizable = behavior.is_none() && crosscoder_layout.is_none()`) installs a non-uniform per-column σ, set to the column RMS, on every native (non-behavior, non-crosscoder) fit's returned term.

- **Public API.** `steer_rows` and `steer_decode` on that term return deltas that are off by σ_c per column. Adding them to a raw `x` moves it off the decoded manifold whenever σ is non-uniform.
- **`inference::steering::collateral_curve`.** It builds its target and other-atom tangent frames with `decode_tangents_at(.., model.tier0_scale())`, which is raw. It then decomposes `model.steer_rows(..)`, which is internal, against those frames. With non-uniform σ, an exactly on-tangent move projects partly off the tangent span, so the manifold arm's `collateral` and `cross_feature` (and the matched flat-control norm) are computed across two frames.

The crosscoder and behavior paths are not affected today. Crosscoder installs σ = 1 and behavior installs none. A lift through σ is therefore a no-op on them.

## Fix

Lift both outputs to raw units with `σ ⊙ ·`:

- A delta is a chord, so the mean cancels.
- A per-atom contribution does not own the shared mean, so the mean is not added.

This matches `decode_at` and `decode_tangents_at`. `steer_layer_*` inherit the lift, which is exact under the crosscoder's unit σ.

The regression test `tests_steering_e4::steer_outputs_lift_through_the_tier0_scale` installs a non-uniform σ on a fitted K=1 circle term. It checks that `try_fitted` lifts by σ, and that `steer_rows` and `steer_decode` return exactly `σ ⊙` their internal-frame values.

Out of scope, not changed here: `canonical_chart_steps` measures arc length on the internal-frame decoder. Under non-uniform σ, its "same intrinsic distance" uses the fit-frame metric rather than the raw one.

SauersML:
Fixes #4332

## Defect
A native standardized fit decodes in (Z − μ)/σ units. `steer_rows` and `steer_decode` returned the chord or contribution in that internal frame. These accessors already return raw units:
- `try_fitted`
- `decode_at`
- `decode_tangents_at`, and so the tangent frames in `collateral_curve`

## Fix
Multiply both outputs by σ. The shared mean cancels in a chord, and it is not part of a single atom's contribution. The change is a no-op for crosscoder fits (σ = 1) and behavior fits (no σ).

## Test
`steer_outputs_lift_through_the_tier0_scale` installs a non-trivial σ on a fitted circle atom. It asserts that:
- `try_fitted` lifts by σ,
- raw `steer_rows` and `steer_decode` equal the internal outputs times σ, exactly.

Focused proof: key HN-1-4332-1.

---


SauersML:
BF-104 (taking over HN-1's PRs): I closed this PR as superseded. Main already landed the same fix for #4332 in 8cd82ac7 (PR #4391), and #4332 is closed. I cancelled its queued proof run 35508593572 so it does not take a CI slot.

---

PR #4375, opened 2026-09-20
tests/quality frailty MLE: certify gam's optimum and compare the two MLEs directly

### LINKED ISSUE #4374: tests/quality frailty MLE: gam's Newton optimum is never certified stationary, and the R comparison checks distances to the truth instead of MLE agreement
## Where

`tests/quality/misc/quality_vs_synthetic_frailty_hazard_multiplier_likelihood.rs`, test `lognormal_hazard_multiplier_kernel_recovers_frailty_variance` (built into the `quality_1082` binary). Three places:

1. The gam-side Newton loop, lines 234-266.
2. The recovery bar, line 374.
3. The match-or-beat arm, lines 413-418.

## The code

```rust
for _iter in 0..100 {
    ...
    for _backtrack in 0..40 {
        next_log_sigma = log_sigma + damping * step;
        ...
        if cand_ll >= loglik { break; }
        damping *= 0.5;
    }
    let converged = (next_log_sigma - log_sigma).abs() < 1e-10;
    log_sigma = next_log_sigma;
    last_loglik = loglik;
    if converged { break; }
}
let sigma_hat_gam = log_sigma.exp();
// only check afterwards: neg_hess_at_opt > 0.0
```

```rust
let recovery_bar = (3.0 * se_sigma).max(0.05);
```

```rust
assert!(err_gam <= err_r * 1.10 + REFERENCE_AGREEMENT_TOLERANCE, ...);
// err_gam = |sigma_hat_gam - SIGMA_TRUE|, err_r = |sigma_hat_r - SIGMA_TRUE|
```

## Why it is wrong

**1. gam's optimum is never shown to be stationary.** The `converged` flag is thrown away. The loop can use up all 100 iterations and the test goes on as if it had found the MLE. Even when the flag is set, it only measures step size. If the line search uses all 40 backtracks, the step is `2^-39 * step`, which is below 1e-10. That reads as "converged" at a point that is not stationary. After the loop the only check is concavity (`neg_hess_at_opt > 0`).

The R reference is held to a stricter standard. It must pass `r_newton_displacement <= 1e-3`, the Newton step still available from its optimum. gam's own optimum never has to pass that check.

**2. The match-or-beat arm compares the wrong quantities.** Both engines maximize the same exact frailty-integrated likelihood on byte-identical data, and the module docs say so ("the two MLEs should in fact agree to quadrature precision"). What follows from that is a check on the two estimates: `|sigma_hat_gam - sigma_hat_r| <= tol`.

The test instead compares each estimate's distance to `sigma_true`. That passes a gam estimate on the other side of the truth from R's, as far as about `2.1*err_r + 1e-3` away. The `1.10` factor is not derived anywhere.

**3. The `0.05` floor on the recovery bar is not derived.** The comment says it "protects against an over-tight se from the observed info." But the observed-information SE at the certified MLE is the principled sampling-error scale. If the kernel's information were wrong, that would be a defect this test should catch, not something to widen the bar past.

## Principled fix

- Require the Newton loop to report convergence.
- Give gam's optimum the same stationarity certificate as R's. The Newton displacement `sigma_hat * |score/info|` at `sigma_hat_gam`, from gam's analytic jet, must be `<= REFERENCE_AGREEMENT_TOLERANCE`.
- Replace the match-or-beat arm with direct MLE agreement: `|sigma_hat_gam - sigma_hat_r| <= REFERENCE_AGREEMENT_TOLERANCE`. This is strictly tighter, since it implies `err_gam <= err_r + tol`.
- Drop the floor, so the bar is `recovery_bar = 3*se_sigma`.

The fix is on branch `hn23-frailty-mle-certificate` (commit fc39e3a6). It is being validated by focused-rust-proof run 35507580264, with statmod installed.

SauersML:
Fixes #4374

- **Newton convergence and stationarity.** The damped Newton loop must now report convergence. gam's optimum must also pass the same stationarity certificate as the R reference: the Newton displacement in σ, computed from gam's analytic score and information at σ̂_gam, must be ≤ `REFERENCE_AGREEMENT_TOLERANCE` (1e-3). A line search that exhausts its backtracks takes a step below 1e-10 and used to count as converged; the certificate closes that hole.
- **MLE agreement.** The match-or-beat arm (`err_gam <= 1.10*err_r + 1e-3`) is replaced by direct MLE agreement, `|σ̂_gam − σ̂_R| <= 1e-3`. Both engines maximize the same likelihood on the same data, so this is the check that follows, and it is strictly tighter.
- **Recovery bar.** The bar is `3·se_sigma`, without the underived `0.05` floor.
- **Diagnostics.** The eprintln line now also reports gam's score and Newton displacement.

Validation: focused-rust-proof run 35507580264 (key HN-23-frailty-2, `quality_1082`, with `r_packages=statmod`).

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 509e31f4a0f4.

PR #4378, opened 2026-09-20
gam-runtime: span_index_for_breakpoints assigns an interior breakpoint to its left span

### LINKED ISSUE #4376: gam-runtime: span_index_for_breakpoints' interior-breakpoint convention is never used; both anchored deviation runtimes re-implement the same left-bias correction
`gam_runtime::span::span_index_for_breakpoints` (`crates/gam-runtime/src/span.rs`) resolves a value that sits exactly on an interior breakpoint to the span on its **right** (`[left, right)` spans). No production caller wants that answer. The helper has exactly two callers, and each one immediately steps back to the left-hand span using its own copy of the same correction:

- `DeviationRuntime::left_biased_span_index_for` (`crates/gam-models/src/bms/deviation_runtime.rs:1223`):
  ```rust
  let mut span_idx = self.span_index_for(value)?;
  if span_idx > 0 && value == self.endpoint_points[span_idx] {
      span_idx -= 1;
  }
  ```
- the saved anchored deviation runtime's `left_biased_span_index_for` (`crates/gam-models/src/inference/model.rs:2787`) repeats the same lines, with the comment "LEFT-bias at interior breakpoints mirrors DeviationRuntime".

Every span lookup in both runtimes goes through these wrappers: `deviation_runtime.rs` lines 1071, 1360, 1402 and 1451, and `model.rs` lines 2756, 2905 and 2951. So the helper's documented convention is never the one the code uses. It survives only because a unit test pins it (`internal_breakpoints_use_right_hand_span`). That test's own doc says the runtimes "override" it. `DeviationRuntime::span_index_for` has no caller other than the correction.

The effect is duplicated logic: the left-hand convention the anchored cubic runtimes rely on for span-local third derivatives is defined twice, outside the helper that owns span selection. The two copies are free to drift apart. A third caller of the helper would silently get the other convention.

**Fix:**
- Make `span_index_for_breakpoints` return the left-hand span at an interior breakpoint. Span `i` becomes `(b_i, b_{i+1}]`, with the first span closed as `[b_0, b_1]`, and the doc states why only the third derivative depends on the choice.
- Reduce both `left_biased_span_index_for` wrappers to the helper call.
- Delete `DeviationRuntime::span_index_for`.
- Replace the helper test with one that pins the left-hand convention.

Both runtimes validate strictly increasing breakpoints, so the new rule gives the same span as the old helper-plus-correction for every input.

SauersML:
Fixes #4376.

The shared breakpoint lookup now assigns an interior breakpoint to its left span, matching the convention used by both live and saved cubic deviation runtimes. Their duplicated post-adjustments and the unused right-span wrapper are removed. Runtime construction checks finite, strictly increasing breakpoints; lookup remains logarithmic.

MSI validation: nine span tests pass, including an independent interval-membership oracle at adjacent representable values, extreme finite endpoints and signed zero. Complete runtime test-source compilation, all 27 production crates, and complete model unit-source metadata pass with warnings denied against a coherent dependency graph. The source-removal census records the retired wrapper and replaced convention test.


SauersML:
Merge train MT-0: merged origin/main into `claude/hn22-span-left-bias` (merge commit ea157397, no rebase or force-push). The only conflicts were `docs/source-removal-changes.json` and `docs/public-api-census-changes.json`. Both were resolved as main's entries plus this branch's new entry, and both validate with `python3 -m json.tool`. The code diff against main is identical to 796dd7db. Queued for the next local integration train, with focused filters `cargo test -p gam-runtime --lib span` and `cargo test -p gam-models --lib deviation`.

---

PR #4382, opened 2026-09-20
gam-runtime: cgroup-v2 probe admits a skewed inactive_file at its headroom

### LINKED ISSUE #4380: gam-runtime: cgroup-v2 memory probe admits 0 bytes when memory.stat inactive_file exceeds memory.current; the v1 probe clamps the same case
`crates/gam-runtime/src/cgroup_memory.rs` handles the same counter condition two opposite ways, depending on the cgroup version.

**v2 (`inspect_visible_v2_hierarchy`, around line 536 on e695ec5d).** When `inactive_file` from `memory.stat` is larger than `memory.current`, the probe returns `ProbeFailed(InconsistentCounters)`:
```rust
if inactive_file_bytes > current_bytes {
    return Err(failure(CgroupMemoryProbeFailureKind::InconsistentCounters, &stat_path, ...));
}
```
`resource.rs` reads `ProbeFailed` as zero capacity and zero available, so every governed allocation under that ceiling is refused.

**v1 (`inspect_visible_v1_hierarchy`, around line 659).** The same condition is admitted at its headroom `limit - usage` with no reclaim credit. The comment there says: "A credit larger than the charge it would reclaim has no consistent reading, so the level is admitted at its headroom with no credit. Refusing the level admitted zero bytes and refused every governed allocation on that node."

The v2 case can happen on a healthy host too. `memory.current` is the live page counter. `memory.stat` is flushed from per-CPU deltas lazily (rstat, with a batching threshold that grows with the CPU count), so a cgroup whose charge is mostly inactive page cache can read `inactive_file` a little above `memory.current` right after an uncharge. That is the exact failure the v1 comment describes: a transient counter skew turns into "admit nothing".

The same split leaves dead code behind:
- `CgroupMemoryAvailability::from_consistent_counters` returns `Option` and yields `None` only when `inactive_file > current`.
- Both probes rule that case out before calling it, v2 with the early return and v1 with a local `reclaim_credit_bytes` clamp. So both `.ok_or_else(|| failure(InconsistentCounters, ..., "memory counters became inconsistent during construction"))` branches can never run.
- `test_support::simulated_cgroup_memory_environment` has to `.expect` the same impossible `None`.
- Once the v2 early return is gone, `InconsistentCounters` has no producer at all.

**Fix:**
- Make the constructor infallible, as `from_counters`, and let it own the one rule: an `inactive_file` above the charge counter earns no reclaim credit, so the working set is the whole charge.
- Remove the v2 early return, the v1 local clamp, both unreachable `ok_or_else` branches, the `expect` in test support, and the `InconsistentCounters` variant.
- Replace `inconsistent_cache_counter_fails_closed` with a v2 test that mirrors the existing v1 one. For limit 1000, current 100 and inactive_file 101, the probe should report `V2Limited` with 900 available.

SauersML:
Fixes #4380.

When an inactive-file statistic exceeds the separately sampled charge counter, the v2 probe currently reports a failed observation and the governor admits zero capacity. Both cgroup versions now retain the observed counters and grant no reclaim credit in that case, so admission remains bounded by `max(limit - current, 0)`. Missing, malformed, and unreadable controller data still report their actual errors.

The shared constructor replaces the duplicated v1 handling and the obsolete inconsistent-counter refusal. The underlying distinction was checked against the [Linux v6.12 memory controller implementation](https://github.com/torvalds/linux/blob/v6.12/mm/memcontrol.c): `memory_current_read` reads the page counter, while statistics have a separately thresholded flush path.

MSI validation: complete runtime unit binary builds with warnings denied; eight cgroup tests and 26 resource-policy tests pass, including 216 independently calculated wide-integer counter combinations and both v1/v2 skewed-counter fixtures. All 27 production crates compile coherently with this change and #4378; complete model unit-source metadata also passes. A separate broad 110-test runtime attempt hit its 20-second suite cap and is not claimed as a full-suite pass.


SauersML:
Merge train MT-0: merged origin/main into `claude/hn22-cgroup-reclaim-clamp` (merge commit 94bdb552, no rebase or force-push). The only conflict was `docs/source-removal-changes.json`. Both sides' entries were kept, and the result validates with `python3 -m json.tool`. The code diff against main is identical to 6a48a436. Queued for the next local integration train, with focused filter `cargo test -p gam-runtime --lib cgroup resource`.

---

PR #4395, opened 2026-09-20
report: drop ReportInput::converged; a report exists only for a certified fit

### LINKED ISSUE #4385: report: ReportInput::converged is always true (fits are certified), so the "did not converge" rendering is dead code
**Area:** `crates/gam-report/src/lib.rs` `ReportInput::converged` (line 35) and its rendering (lines 559-584); producers in `crates/gam-models/src/inference/saved_summary.rs:1598, 1698, 1803` (origin/main).

SPEC: "No dead code"; "A fit object must only ever come from a converged optimization."

### What the code does

`ReportInput` carries `converged: bool`. The HTML renderer branches on it:
- `false` gives "(did not converge)" after the outer iteration count and a `conv-warn` ⚠ status;
- `true` gives a `conv-ok` status.

Every production constructor hard-codes `converged: true`:
- all three `ReportInput { .. }` literals in `saved_summary.rs`;
- `convergence_status` is read from `fit.convergence_evidence().inner_status().label()`.

That is correct as far as it goes. A report can only be built from a minted fit, and a minted fit owns a sealed `FitConvergenceEvidence`. That evidence refuses any inner status other than `Converged` (`result_types.rs:1845`, `FitConvergenceEvidence::from_serialized`) and requires a certifying analytic outer certificate. So `converged` cannot be `false` on any production path.

### Why it is wrong

- The `false` arm of both renderer branches is unreachable. The only thing that exercises it is the unit test `render_html_non_converged_shows_warning` (`lib.rs:~1822`), which sets `input.converged = false` by hand. Nothing else can produce that state.
- The field comment ("Whether the fit cleanly converged … any non-converged state is highlighted") and the renderer comment ("a reader can immediately tell a healthy fit from one that hit the iteration cap, exhausted the LM step search, or went unstable") both describe a contract the fit layer abolished. They invite a future caller to build a report from a non-certified checkpoint and pass `converged: false`. That is exactly the path SPEC forbids: a non-converged optimization rendered as a fitted-model report.

### Principled fix

1. Delete `ReportInput::converged` and both `if input.converged { .. } else { .. }` branches. Always render the iteration count and the status plainly; the status is the certified inner status label.
2. Remove the three `converged: true` initializers in `saved_summary.rs`.
3. Delete `render_html_non_converged_shows_warning`. It tests the removed feature, not a behaviour a fit can reach, so nothing is loosened. Then drop the now-unused `conv-warn` class from the report CSS if nothing else uses it.

The existence of a report then means the fit was certified. This is the same statement the type system already makes for `UnifiedFitResult`.

SauersML:
Fixes #4385

Saved reports now render the certified solve's status directly. The redundant `ReportInput.converged` field and its unreachable warning branches are removed: all three production builders previously supplied true, and sealed `FitConvergenceEvidence` rejects non-converged inner states both during fit minting and deserialization. Optimality-certificate warnings remain live.

Removed the impossible-state test and added a regression for the live status text and its HTML escaping. All43 complete gam-report tests pass. The complete eight-crate reverse dependency production closure and complete CLI unit metadata pass with warnings denied on MSI using coherent warm dependencies. The source-removal ledger records the obsolete test; no public function was removed.


SauersML:
MT-1 merge-train review: the diff looks sound to me. Dropping `ReportInput::converged` fits the SPEC line "A fit object must only ever come from a converged optimization."

The source-removal guard fails at f250ab5d9c. Checked with `python3 scripts/source_removal_guard.py --base $(git merge-base origin/main pr-4395) --head pr-4395`, it reports this removal without an acknowledgement:

- `crates/gam-report/src/lib.rs:fn:render_html_non_converged_shows_warning`

Fix: add one entry to `docs/source-removal-changes.json` and commit it on the branch: `{"base": "<merge-base sha>", "items": ["crates/gam-report/src/lib.rs:fn:render_html_non_converged_shows_warning"], "reason": "...", "evidence": "..."}`. The entry's `base` must equal the merge-base with main. Once the guard passes, I will put this PR in a train.

---

PR #4399, opened 2026-09-20
gam-sae curl_census: use the crate-wide e-BH rule and refuse fdr_alpha outside (0,1)

### LINKED ISSUE #4396: gam-sae manifold/curl_census: private e-BH copy (pair_phase::ebh_reject) ranks NaN as a tie and silently accepts any fdr_alpha
**Where:** `crates/gam-sae/src/manifold/pair_phase.rs` (`ebh_reject`) and `crates/gam-sae/src/manifold/curl_census.rs` (`ebh_ledger`, `census_shattered_circles`).

**Defect (duplicated code plus silent errors)**

1. `pair_phase.rs` now holds only a second copy of Wang–Ramdas e-BH, `ebh_reject`. Its one caller is the curl census ledger. The crate-wide rule is `gam_terms::inference::structure_evidence::e_benjamini_hochberg_in_family`, which `sparse_dict/split_lr_fdr.rs` already uses. The two copies have already drifted apart:
   - The private copy sorts with `partial_cmp(..).unwrap_or(Equal)`. A NaN e-value is therefore ranked as a tie wherever the sort happens to leave it, and it scrambles the order around it. The canonical rule refuses NaN (`EBhError::InvalidLogEvidence`).
   - The private copy returns `Vec::new()` for `alpha <= 0` or NaN, which is a silent "no discoveries". It accepts `alpha >= 1`, which gives no FDR control at all. The canonical rule refuses any `alpha` outside `(0, 1)`.
   - The private copy needs the family padded by hand with `e = 0` (`e_values.resize(family, 0.0)`). The canonical rule takes `family_size` directly, which is exactly the screened-family semantics #3658 established.
   - A family smaller than the adjudicated claims is an `assert!` panic in `ebh_ledger`. In the canonical rule it is a typed error.

2. `census_shattered_circles` never validates `cfg.fdr_alpha`, even though the derived replicate budget is `B = ceil(m/α) − 1`:
   - With `α = 0`, `m/α = inf`, which casts to `usize::MAX`. The census then asks `ring_permutation_evidence` for about 1.8e19 draws per plane that clears, which is effectively a hang.
   - With `α = NaN`, the cast gives 0, which is floored to `B = 2`, and e-BH quietly rejects nothing.
   - Both print a census that looks like an absence of structure.
   - The doc on `null_replicates` says the budget is fixed by the size of the search. That only holds for `α` in `(0, 1)`.

3. `discrete_fourier.rs` links to `crate::manifold::pair_phase` as "the phase-coupling screen". That screen no longer exists; the DFT's replicate-driven caller is `crate::null_battery`.

**Fix**
- `ebh_ledger` passes `ln e` (`e = 0` becomes `−inf`, the exact zero e-value) to `e_benjamini_hochberg_in_family(log_e, family, alpha)` and propagates its error as `Err(String)`.
- `census_shattered_circles` refuses `fdr_alpha` outside `(0, 1)` up front.
- Delete `pair_phase.rs` and `pair_phase_tests.rs`, and port the dominant-e-value unit test into `curl_census` tests.
- Point the `discrete_fourier` doc at `null_battery`.

SauersML:
Fixes #4396

**Changes**
- `curl_census::ebh_ledger` now calls `gam_terms::inference::structure_evidence::e_benjamini_hochberg_in_family(ln e, family, alpha)`, the rule `split_lr_fdr` already uses. It returns `Result<_, String>`.
  - Unadjudicated candidates enter only through `family_size`, as exact zero e-values, so the family is no longer padded by hand.
  - NaN evidence, an alpha outside (0,1), and a family smaller than its claims are now refused. Before, they were ranked as ties, returned an empty set, or panicked.
- `census_shattered_circles` refuses `fdr_alpha` outside (0,1) before deriving the replicate budget. Before, α=0 gave `B = usize::MAX` and NaN gave `B = 2` with no discoveries.
- `census_shattered_circles` refuses `null_replicates = 1`.
- A `ring_permutation_evidence` failure on a plane the screens accepted now propagates as `Err`. Before, `.unwrap_or` recorded it as the `e = 0` of a refused plane. The `.ok()?` refusals of the deterministic screens are unchanged and still bank `e = 0` by design (#3658).
- Deleted `manifold/pair_phase.rs` and `pair_phase_tests.rs`, which held only the private e-BH copy. Its unit test is ported as `ebh_ledger_rejects_only_the_dominant_e_value`.
- Fixed the stale `discrete_fourier` doc link, which now points at `null_battery`.

**Budget derivation (commit c15ec6ef)**

Moving to the log-space rule broke the linear budget. A plane that beats all `B` surrogates scores `e = B + 1`. Rank 1 rejects when `ln e ≥ ln m − ln α`. The old budget `B + 1 = ⌈m/α⌉` was computed in linear arithmetic, which rounds independently of that threshold.

At `m = 10, α = 0.05`, `⌈m/α⌉ = 200`, but `ln 200` is one ulp below `ln 10 − ln 0.05`. A lone perfect plane could therefore never be a discovery. Over α ∈ {0.05, 0.1, 0.01, 0.2, 0.025} and m ≤ 7000 this happens in 720 cases. In every one of them the linear test rejects and the log test does not.

Fix:
- `structure_evidence::e_bh_log_threshold(m, α, k) = ln m − ln α − ln k` is now the single owner of that arithmetic. Both in-family e-BH routines call it, and it is bit-identical to the old inline expressions.
- `derived_replicate_budget` takes `B + 1` as the smallest integer ≥ 3 whose log clears `e_bh_log_threshold(m, α, 1)`. It starts at `⌈m/α⌉` and steps up.
- This is derived, not a fallback. It is the exact minimum of the condition the ledger applies. Consecutive integers differ by about `1/(B+1)` in log, far more than the rounding, so the step runs at most once.

**New tests**
- `ebh_ledger_refuses_what_it_cannot_rank`
- `census_refuses_an_fdr_level_outside_the_unit_interval`, which also covers `null_replicates = 1`.
- `derived_budget_lets_a_lone_perfect_plane_clear_rank_one`, over m ∈ 1..=3000 at five α values. It checks sufficiency (`e = B + 1` alone is rejected) and minimality (`e = B` is not). It also pins `B(10, 0.05) = 200`.

Focused proof run: gam-sae lib, filter `curl_census`, key BF-104-4396-2.


SauersML:
Taken over by BF-104 (HN-1 handed off). I pushed c15ec6ef, which fixes a regression this PR introduced. Under the log-space e-BH rule, the linear budget `B + 1 = ⌈m/α⌉` could fall one ulp short of `ln m − ln α`. At m = 10, α = 0.05 this made a lone perfect plane unrejectable. The body has the derivation. Proof run key: BF-104-4396-2.

---

SauersML:
MT-1 merge-train note for BF-104, the owner: at c15ec6ef12 the source-removal guard fails. Checked with `python3 scripts/source_removal_guard.py --base $(git merge-base origin/main pr-4399) --head pr-4399`, it reports these removals without a ledger acknowledgement:

- `crates/gam-sae/.../pair_phase.rs:fn:ebh_reject`
- `.../pair_phase_tests.rs:fn:ebh_rejects_dominant_e_value`

Fix: add one `docs/source-removal-changes.json` entry that lists both items. Its `base` must be the merge-base with main, and it needs a reason and evidence. I am not touching the branch while you own it.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, e85d72607f60.

PR #4403, opened 2026-09-20
Derive the symmetry band from each matrix's assembly instead of a fixed 32 ULP (#4350)

### LINKED ISSUE #4350: gam-linalg utils: symmetry validator's SYMMETRY_ULP_ALLOWANCE = 32 is not the derived 2γ_n assembly band (split from #4038)
Split out of #4038. The solve allowance `256·n·ε` has been replaced by derived factorization bands in PR #4158. This issue tracks the second constant that #4038 named, which is still open.

## Where

`crates/gam-linalg/src/utils.rs`, `validate_finite_symmetric_matrix`:

```rust
const SYMMETRY_ULP_ALLOWANCE: f64 = 32.0;
...
let tolerance = SYMMETRY_ULP_ALLOWANCE * positive_ulp(pair_scale.max(gram_scale));
```

`gram_scale` is `√(A_ii·A_jj)`. Nothing derives the `32`.

## What the derived band is

Suppose the two triangles of `A = XᵀWX + S` come from separate `n`-term accumulations, as in a full GEMM, with `W ≥ 0`. Then

```
|fl(A_ij) − fl(A_ji)| ≤ 2γ_n Σ_k |X_ki W_k X_kj| + (penalty assembly)
                      ≤ 2γ_n √(A_ii^{XWX} · A_jj^{XWX}) + …        (Cauchy–Schwarz)
```

So the correct allowance is `2γ_n·gram_scale`, plus the penalty's own assembly band. It is not `32 ulp`. The count `n` is the number of accumulated terms (the rows of `X`), and **the validator cannot see it**. The validator takes only the assembled matrix.

At `n = 10⁴` the worst-case derived factor is about `2·10⁴·u ≈ 10⁴ ulp`. The fixed `32 ulp` can therefore *reject* legitimately assembled large-`n` Hessians that are built by a full GEMM. For a mirrored triangular update it is over-generous, since the exact defect there is 0.

## Why it was not in #4158

A derivation needs the assembly provenance: the term count, and whether the Gram is mirrored or fully computed. About 14 call sites across `gam-linalg`, `gam-solve`, `gam-models` and `gam-pyffi` pass only a matrix and a label. They include `certified_symmetric_solve`, `SymmetricMatrix` construction, `faer_ndarray` validation, `gaussian_marginal`, `gaussian_reml_multi_penalty`, `smoothing_correction` and `alo`. Fixing this means threading an `assembly_terms` argument, or a mirrored/accumulated flag, through each of them. That is an API change of its own.

SauersML:
Replaces `SYMMETRY_ULP_ALLOWANCE = 32` in `validate_finite_symmetric_matrix` with a triangle band derived from how each matrix was assembled. Every strict symmetric entry point (`validate_finite_symmetric_matrix`, `certified_spd_factorize`, `certified_spd_inverse`, `certified_symmetric_solve`, `strict_symmetric_eigh`, `SymmetricMatrix::factorize_spd`, `rank_certified_psd_pseudoinverse`, `solve_dense_block_system`) now takes a `SymmetricAssembly` argument. They all read the band from one helper, `roundoff::symmetric_assembly_band`, and refuse only when `|A_ij − A_ji|` exceeds that band.

Fixes #4350

## Derivation

Let `u = ε/2` and `γ_d = d·u / (1 − d·u)`. This is Higham's bound: a recursively summed dot product with `d` terms satisfies `|fl(Σ t_k) − Σ t_k| ≤ γ_d · Σ|t_k|`.

**Mirrored.** Here `A_ij` and `A_ji` come from the same rounded operation on the same operands. IEEE `+` and `×` are commutative and correctly rounded, so the two entries are bitwise equal and **the band is 0**. The producers that qualify:
- `(M + Mᵀ)/2`, `symmetric_part`, `mirror_lower` and `symmetrize_in_place`
- an entrywise fill of both triangles from one value
- sums and differences of mirrored matrices
- a `from_shape_fn` Gram that multiplies the same pairs in the same order
- `into_pseudoinverse` and `CertifiedSpdFactor::inverse`, which both end in `symmetrize_in_place`

**PsdAccumulation { depth: d }.** Each triangle is its own accumulation of the PSD pieces `w_k x_ki x_kj` with `w_k ≥ 0`. Each triangle lies within `γ_d Σ_k |w_k x_ki x_kj|` of the exact value. The triangle inequality therefore gives

`|A_ij − A_ji| ≤ 2 γ_d Σ_k w_k |x_ki||x_kj|`.

Cauchy–Schwarz gives `Σ_k w_k |x_ki||x_kj| ≤ sqrt(S_ii S_jj)`, where `S_ii = Σ_k w_k x_ki²` is the exact diagonal. The computed diagonal satisfies `A_ii ≥ (1 − γ_d) S_ii`. Together:

**band_ij = 2 γ_d sqrt(A_ii A_jj) / (1 − γ_d)**

- The band is `+∞` when `γ_d ≥ 1`, because the bound is then vacuous.
- It is read at the `(i, j)` block's own diagonal scale, not at `|A_ij|`, which can cancel to nothing. A well-scaled block inside a badly scaled matrix therefore keeps a tight band.
- Each diagonal is square-rooted before the product, so a large finite diagonal cannot overflow.
- A negative diagonal on a `PsdAccumulation` matrix is refused. A sum of nonnegative pieces cannot round below zero, so a negative diagonal breaks the assembly claim.

**Algebra** (`const fn` on `SymmetricAssembly`):
- `a.psd_sum(b)`: two mirrored inputs give a mirrored result. Otherwise the summand lists merge, and the one extra addition gives depth `max(d_a, d_b) + 1`.
- `a.scaled()`: a scalar multiple of a mirrored matrix stays mirrored. Otherwise `{d} → {d + 1}`, one more rounding per entry.
- `penalized_gram(rows, penalty_root_rows) = {rows + 1}.psd_sum({penalty_root_rows})`. This is XᵀWX (one extra rounding for the weight) plus SᵀS from the penalty root rows. It is used when the producer is a full GEMM or is not known.

**Depth rules at call sites:**
- An unweighted GEMM Gram over `k` inner terms has depth `k`.
- A weighted `n`-row Gram has depth `n + 1`.
- The HMC survival whitening uses `penalized_gram(3n, dim)`, because the entry, exit and derivative rows each contribute.
- The conjugate-Gaussian HMC whitening uses `{n + 1}.scaled().psd_sum(Mirrored)`.

## Contracts at call sites

- **Declared or user inputs** (penalties, prior precisions, user-supplied matrices) must be exactly symmetric: `Mirrored`. A caller that hands over an asymmetric matrix is refused. The old 32-ULP allowance let that through.
- **ρ-Hessians** are symmetrized by their consumer (`symmetrize_in_place`, then `Mirrored`), e.g. `run_rho_criterion_nuts`. `RHO_HESSIAN_ASSEMBLY` in `smoothing_correction` names this contract.
- **`gaussian_marginal`**: the declared prior uses `DECLARED_PRIOR_ASSEMBLY = Mirrored`. The posterior `Q + WᵀW` uses `POSTERIOR_PRECISION_ASSEMBLY = Mirrored.psd_sum(Mirrored)`, which is `Mirrored`: `fast_ata` mirrors its triangular Gram, and the sum adds the same two operands in both triangles.
- **GPU `fast_xt_diag_x`**: the device output is not guaranteed to be mirrored, so it is now passed through `symmetrize_in_place` on return. `design_construction::xt_diag_x_dense` now calls that same kernel instead of keeping its own copy, so there is one kernel.
- **Conditioning**: the equilibration in `design_construction` symmetrizes after the diagonal congruence, so its output is `Mirrored`.
- **Event history**: `covariance::eigenmodes` takes the assembly as a parameter.
  - `covariance_score` symmetrizes, so it passes `Mirrored`.
  - `coordinate_hessian` fills each mirrored pair from one channel, so it passes `Mirrored`.
  - The test's equilibrated covariance is rebuilt from one triangle, so it is `Mirrored`.
  - The truth covariance `L Lᵀ` is `{depth: rank}`.
- **HMC whitening**: `hessian_whitening_transform` takes the assembly as a parameter, and each caller states its assembly.

## Caveats

- **Signed-weight observed information.** The Cauchy–Schwarz step assumes `w_k ≥ 0`. For a Hessian with negative observation weights, `Σ|w_k| x_ki²` can exceed `A_ii`, so the band can be too tight. In that case the validator **refuses** a matrix it could have accepted. It never accepts one it should refuse. The `penalized_gram` HMC whitenings, where the survival and generic Hessians can carry signed observed-information weights, are the sites where this conservative side could show up.
- **GPU hooks without mirroring.** Two GPU paths do not pass through the new mirroring: `try_fast_joint_hessian_2x2` and `ResidentDesignGram::gram`. Their outputs reach the validators with the provenance of the CPU path they replace. If a device reduction splits the triangles differently, they will be refused rather than accepted silently. Follow-up work: symmetrize them on return, as done here for `fast_xt_diag_x`.

## Validation

- `focused-rust-proof` on `gam-linalg` lib, filter `certified_inverse_tests` (run key `BF-80-4350-1`). These tests cover:
  - the band follows the block scale, not a cancelled entry
  - the verdict is invariant under uniform rescaling
  - `Mirrored` admits exact symmetry only
  - the PSD band is the depth bound
  - a negative PSD diagonal is refused
- `focused-rust-proof` on the `gam` lib at the merged head, filter `init_parallelism_registers_gpu_dispatch_hook` (run key `BF-80-4350-3`). This builds every crate whose call sites now pass an assembly, with all features on.
- The branch no longer carries whole-file rustfmt changes. Each layout-only hunk was reset to main's text, and each reset was kept only when rustfmt of the result matched rustfmt of the earlier head, so the tokens did not change. The diff against main is now 41 files, +1109/−456, all semantic.
- `cross-check.yml` is disabled in this repo, so it could not be dispatched.


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, b8fcb448379a.

PR #4405, opened 2026-09-20
Score matern/periodic/sphere sweeps by across-the-function coverage of the fit's own posterior

### LINKED ISSUE #4387: tests/smooth sweeps: matern-nu, periodic-degree and sphere-penalty-order RMSE bars of 0.25 are "5 sigma" of the noise, 14-25x the function error the fixture delivers, so a fit keeping only 42-65% of the signal passes
## Where

| file | line | bar |
|---|---|---|
| `tests/basis_smooth/smooths/matern_all_nu_sweep_diagnose.rs` | 90-93 | `rmse > 0.25` ("σ=0.05, allow up to 5σ") for ν ∈ {1/2 … 9/2}, n=300 |
| `tests/basis_smooth/misc/periodic_1d_degree_sweep.rs` | 84-96 | `rmse > 0.25` ("σ=0.05 noise → 5σ = 0.25"), plus an `Err` arm that accepts any error without "panic"/"nan", for degree 1..5, k=degree+10, n=200 |
| `tests/manifolds/smooths/sphere_penalty_order_sweep.rs` | 110, 163 | `rmse > 0.25 \|\| mn < -5.0 \|\| mx > 5.0` for Wahba m=1..3 (k=30) and harmonic max_degree=4 m=1..4, n=400 |

## The code

```rust
// matern_all_nu_sweep_diagnose.rs:90
// σ=0.05, allow up to 5σ for hardest case
if rmse > 0.25 {
    failures.push(format!("`{formula}` rmse={rmse:.4} > 0.25 (5σ)"));
}
```

```rust
// sphere_penalty_order_sweep.rs:110 (and 163)
if rmse > 0.25 || mn < -5.0 || mx > 5.0 { failures.push(...) }
```

```rust
// periodic_1d_degree_sweep.rs:89
Err(e) => {
    let lower = e.to_lowercase();
    if lower.contains("panic") || lower.contains("nan") { failures.push(...) }
    else { eprintln!("[per-deg{degree}] clean rejection: {e}"); }
}
```

## Why it is wrong

1. **The bar mixes up two quantities.** Here `rmse` is the error of the fitted *function* against the known truth on a probe grid. σ = 0.05 is the SD of the *per-observation noise*. For a linear smoother, the function's estimation error is about σ·√(edf/n):
   - ≈ 0.011 for matern (edf ≈ 15, n = 300)
   - ≈ 0.013 for sphere harmonic max_degree=4 (25 columns, n = 400)

   "5σ" of the noise is 20–25× larger than that. No derivation leads from 5σ of the noise to a bound on the function error.

2. **The bar sits almost at the "fit is the mean" level.** Compare each bar with the RMSE of a flat fit at the response mean:
   - **Matern:** the truth `sin(4πt) + 0.3t` has RMS deviation from its mean of ≈ 0.71. A fit that shrinks the sine to 65% of its amplitude has RMSE ≈ 0.25 and passes.
   - **Periodic:** the truth `cos θ + 0.3 sin 2θ` has RMS ≈ 0.74. The bar passes the same 35% loss of amplitude.
   - **Sphere:** the file's own m=4 test records that a fit which collapses to the response mean scores RMSE 0.43. It also records that m=1,2,3 reach RMSE ≤ 0.018 on this data. The sweep bar of 0.25 therefore passes a fit that keeps only 42% of the signal. That is 14× the RMSE the file says these orders actually achieve. The `±5` range check on a signal whose range is ≈ 1.4 adds nothing.

   So a partial over-smoothing regression, such as REML driving λ too high for one ν, degree or order, passes all three sweeps. That regression is exactly what these sweeps exist to catch.

3. **`periodic_1d_degree_sweep` passes on a refusal.** Every case is a Gaussian periodic B-spline with k = degree + 10 ≥ 11 on 200 distinct points, which is well-posed. Yet the `Err` arm passes on any error whose text lacks "panic"/"nan", such as a non-convergence or a stationarity refusal. The "panic" substring can never fire, because a panic unwinds before the `match` runs. This is the same pattern as #4377, in a file that #4377 does not list.

## Principled fix

Replace each hard-coded RMSE bar with a bar derived from the fit's own posterior covariance.

1. **Compute the prediction SE at each probe.** For each fit, take `Σ = fit.beta_covariance()` and the probe design `X_p` that the test already builds. Set `s_i = √(x_iᵀ Σ x_i)` and `e_i = f̂(x_i) − f(x_i)`.

2. **Gate on the across-the-function coverage statistic.** Use `Q = (1/P) Σ_i e_i² / s_i²`.
   - Under the Bayesian model behind the intervals (Nychka / Wahba across-the-function coverage), `e ~ N(0, X_p Σ X_pᵀ)`. Hence `E[Q] = 1` and `Var[Q] = 2·tr(R²)/P²`, where `R` is the probe correlation matrix. Both come directly from `Σ`.
   - Gate `Q ≤ q_{1−α}` of the moment-matched scaled χ² (Satterthwaite), with a stated α, e.g. 0.01. This bound comes entirely from the fit and fixture. It fails when a fit over-smooths, because the bias adds to `e` while `s` shrinks. A mean collapse gives `Q ≫ 1`.
   - Print `Q` beside its expectation, so a very small `Q` (over-wide bands, which SPEC also counts as a defect) is visible.

3. **Periodic `Err` arm.** In `periodic_1d_degree_sweep`, require `Ok` for every degree in 1..5, since each is well-posed. A refusal must be a failure that names the degree.

4. **Remove the `±5` range checks.** They are made redundant by (2).

If any case then fails, that is a real solver or basis defect to be filed on its own. The bar must not be widened to hide it.

---

SauersML:
Fixes #4387

The three smooth sweeps used a function-RMSE bar of `0.25`, described as "5σ" of the observation noise. That is 14–25× the function error these fixtures actually deliver, so a fit keeping only 42–65% of the signal still passed. This PR replaces those bars.

**New statistic.** `gam-test-support::calibration::audit_across_function_coverage` computes Nychka's across-the-function coverage statistic:
- `Q = (1/P) Σ e_i²/s_i²`, where `s_i = √(x_iᵀ V_b x_i)`.
- Under the Bayesian model behind the intervals, `E[Q] = 1` and `Var[Q] = 2 tr(R²)/P²`, where `R` is the probe correlation matrix.
- The gate is the upper `1 − α` quantile of the Satterthwaite-matched scaled `χ²`: `(Var/2)·χ²_{2/Var}`.
- A unit test checks the gate against the exact `χ²_P/P` bound on independent probes, and checks that a fit collapsed to the mean fails.

**Sweeps.**
- `matern_all_nu_sweep_diagnose`, `periodic_1d_degree_sweep`, and the Wahba and harmonic `sphere_penalty_order_sweep` cases each gate on `Q`, with a Bonferroni share of a family-wise `α = 0.01`.
- Each test prints `Q` next to its bound.
- The `±5` range checks are removed.
- `periodic_1d_degree_sweep` now requires every well-posed degree to return `Ok`. It no longer passes on a refusal whose text lacks "panic"/"nan".

Validation: see the comments below.

https://claude.ai/code/session_01W5v9mtzjtTtPqec9DNtojF

---
_Generated by [Claude Code](https://claude.ai/code/session_01W5v9mtzjtTtPqec9DNtojF)_


SauersML:
This is a coordination note, not a blocker. #4464 (for #4411) adds a shared `gam::test_support::reference::across_function_coverage(x_probe, cov, err, alpha)` to `crates/gam-test-support`. It computes the same Nychka / Marra & Wood across-the-function statistic: Q = (1/P) Σ e²/s², bounded by the Satterthwaite g·χ²_{1−α}(1/g) with g = tr(R²)/P². The same PR adds `least_squares_projection` for an in-span target. Whichever PR lands second could call that helper instead of keeping a private copy, so the gate has a single source.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 7e4611c0b544.

PR #4410, opened 2026-09-20
gam-sae: centred discarded pair covariance with a certified band (#4351)

### LINKED ISSUE #4351: gam-sae response/compose: discarded-unit covariance is E[g(x)g(y)] − E g(x)·E g(y) by raw subtraction (same cancellation as #4124)

SauersML:
Fixes #4351

## Defect
`DiscardedLaw::covariance_diagonals` (crates/gam-sae/src/response/compose.rs) formed `C_jl = K − m_j m_l` by raw subtraction. That keeps only `u(|K| + |m_j m_l|)` of a covariance that vanishes with the discarded variances. It also cancels at small correlation and at large means.

## Fix: `response::subspace::pair_covariance`
The function carries two routes and returns whichever has the smaller certified band.

**Closed form.** `K − m_x m_y`, with this band:
- `value_rounding(K)`
- `+ γ_2(|K| + |m_x m_y|)`
- `+ |m_x| δ_y + δ_x(|m_y| + δ_y)`, where `δ = e_0 + γ_1 sup|σ'| ŝ` covers `m = a_0` evaluated at `ŝ = fl(√v)`

The whole band is then inflated by 4 roundings.

**Mehler series.** This route never forms the means. With `s_x = √v`, `s_y = √w` and `ρ = r/(s_x s_y)`:

`Cov(σ(X), σ(Y)) = Σ_{n≥1} ρⁿ a_n b_n`,  where `a_n = E[σ(b + s_x E) h_n(E)]` (from `gaussian_hermite_coefficients`, with bounds `e_n`, `f_n`).

Truncated at order `N` and evaluated by Horner, the series errs by the following terms:
- **Tail:** `|Σ_{n>N} ρⁿ a_n b_n| ≤ |ρ|^{N+1} √(E_x(N) E_y(N))`, by Cauchy–Schwarz, where `E(N) ≥ Σ_{n>N} a_n²` is the existing `unit_tail_envelope`.
- **Horner rounding:** `γ_{3N+3} Σ |ρ|ⁿ |a_n b_n|` (same operation count as `tiles::chaos_operations`).
- **Coefficient errors:** `Σ |ρ|ⁿ (e_n|b_n| + |a_n| f_n + e_n f_n)`.
- **Law rounding:** `γ_6 |ρ| ŝ_x ŝ_y sup|σ'|²`. The pieces:
  - Forming `ρ̂ = (r/ŝ_x)/ŝ_y` errs by `γ_4|ρ|`. Price's theorem, `∂_ρ Cov = s_x s_y E[σ'(X)σ'(Y)]`, turns that into `γ_4 |ρ| s_x s_y sup|σ'|²`.
  - Each `|ŝ − s| ≤ u s` moves `a_{≥1}` in ℓ² by at most `sup|σ'| |ŝ − s|`, since `∂_s σ(b + sE) = σ'(b + sE) E`.
  - The other column satisfies `‖b_{≥1}‖ ≤ s_y sup|σ'|` by the Gaussian Poincaré inequality.
  - So each scale adds `γ_1 |ρ| s_x s_y sup|σ'|²`.

Every term is proportional to `|ρ|`, so the series keeps its relative accuracy as `ρ → 0` and as `|K| ≫ |Cov|`.

**Order selection.** No grid and no knob: the order is derived.
- `N` is the first order ≥ 1 whose (inflated) tail falls within Horner's own floor `γ_{3N+3}|ρ a_1 b_1|`. That floor bounds the series' rounding from below, so a higher order cannot shrink the band by more than the tail it removes.
- The floor grows with `N` and the tail shrinks.
- The series is dropped at the first `N` whose floor plus the law rounding already reaches the closed-form band. This also ends the scan, and once `γ_{3N+3}` is infinite the floor reaches every band.

**Exact zeros.** When `v = 0`, `w = 0`, or the projected `r = 0`, the units are independent and the result is exactly `BandedEnergy::ZERO`.

**Compose step.** `covariance_row[other] = pair_covariance(activation, pair)?.value` replaces the raw subtraction. SiLU refuses exactly as it did before, because `pair_kernel` already has no SiLU closed form.

### Deviation from the dispatch sketch
The dispatch suggested a ρ-quadrature of `φ₂` in `gam-math/gaussian_activation.rs`. I used the Mehler chaos series in `gam-sae/src/response/subspace.rs` instead, for three reasons:
- It is the natural centred form: the `n = 0` term *is* `m_x m_y`, so dropping it removes the cancellation exactly.
- Its truncation bound comes from the certified `unit_tail_envelope`, which lives in `gam-sae/response/hermite.rs`. A quadrature would need a new derivative-based error bound.
- It covers ReLU and exact GELU with one code path.

`tiles::chaos_operations` becomes `pub(crate)` so the Horner count has a single source.

## Tests (`response::subspace::subspace_tests`)
References are mpmath values:
- `E σ(X)σ(Y)` by adaptive 1-D quadrature over `E₁` of the closed-form smoothing in `E₂`, minus the closed-form means;
- computed at 60 and at 90 digits;
- cross-checked against the Mehler series of the exact coefficients to 30 digits.

Every test also asserts two things about the band: `|value| > band`, and `band < u·|K|`. No subtraction `K − m m` can reach the second.

| test | law | reference | raw `K − m m` |
|---|---|---|---|
| delta, ReLU | b=1, c=2, v=w=1e-16, r=3e-17 | r (σ linear on all but Φ(−1e8)) | error > r/2 (asserted) |
| delta, exact GELU | b=0.7, c=−0.4, same law | 5.779705838062695190220885e-18 | error > half the value (asserted) |
| delta method, GELU | same | σ'(b)σ'(c)r = 5.779705838062694444815729e-18 | — |
| small ρ, ReLU | b=c=0, v=w=1, ρ=1e-6 | 2.500000795774715346413201e-7 | about 10 digits |
| small ρ, ReLU | b=0.5, c=−0.3, ρ=1e-8 | 2.641999091092812169561513e-9 | about 8 digits |
| large means | b=c=1e6, v=w=1, r=±0.3, ReLU and GELU | r | K≈1e12, error about 1e-4 |
| diagonal | ReLU b=c=0, ρ=1 | 1/2 − 1/(2π) | the closed form is kept and resolved positive |
| independent | v=0 or r=0 | exactly 0 | — |

**Delta-method remainder.** The delta-method check allows `band + inflated(remainder, 17)`. The remainder bound:
- The first-chaos term uses `a_1 = s E σ'(b + sE)` and `|Eσ'(b + sE) − σ'(b)| ≤ s² M₃/2`, with `M₃ = sup|(t³ − 4t)φ| = 0.77877…`.
- The chaos terms `n ≥ 2` are at most `ρ² ½ s⁴ M₂²`, with `M₂ = 2φ(0)`.

Measured in mpmath, the GELU gap between the exact value and the delta method is 7.45e-34, against a remainder bound of 2.92e-33.

Validation: focused-rust-proof, gam-sae lib, filters `pair_covariance response::compose::`.


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 38a7bd081e0e.

Issue #2401, opened 2026-07-23
Binomial factor-smooth fit aborts when PSD-root PIRLS reaches a sparse-native design


SauersML:
## Summary

A supported binomial GAM with a sparse high-cardinality factor design and a five-level `fs` age interaction aborts before convergence because the PSD-root P-IRLS solver rejects `WorkingCoordinateDesign::OriginalSparseNative`:

```text
Fatal outer-objective evaluation failure (outer ARC evaluation):
REML smoothing optimization failed to converge: outer eval failed:
Invalid input: dense square-root solve reached a sparse-native coordinate design
```

The same response, weights, rows, nuisance terms, and global age smooth certify when only the `fs` increment is removed. This is a solver-routing hole, not an unsupported formula or an uncertified optimum.

## Current reproduction

Observed on the current PyPI release `gamfit 0.1.258` using Python 3.12.12, NumPy 2.2.6, pandas 2.3.3, and pyarrow 25.0.0 on macOS arm64.

The deterministic 600-row fixture has:

- response `y ∈ {0, 0.5, 1}` from two binomial trials;
- constant `weights = 2`;
- 5 regions;
- 14 assay levels;
- a sparse, region-structured study factor;
- 8 standardized continuous nuisance covariates;
- region-specific age support;
- seed 42.

Passing control:

```text
y ~ whg + eef + pc1 + pc2 + pc3 + pc4 + pc5 + coverage_z
  + group(study) + group(assay) + group(region)
  + s(age_bp, k=6)
```

Result: certified model in 13.39 s with finite predictions in `[0.4321, 0.7855]`.

Failing increment:

```text
y ~ whg + eef + pc1 + pc2 + pc3 + pc4 + pc5 + coverage_z
  + group(study) + group(assay) + group(region)
  + s(age_bp, k=6)
  + s(age_bp, region, bs='fs', k=4)
```

Result: hard failure in 2.88 s with the error above. Both formulas pass `validate_formula(..., family="binomial", weights="weights")`.

A less sparse 600-row fixture with the same 174/14/5 cardinalities certifies `joint_full` in 27.66 s after one ARC cost-stall escape, so the defect is specifically triggered by the sparse-native routing boundary, not by every factor smooth.

## Root cause

Current `main` still contains the rejecting branch in `crates/gam-solve/src/pirls/gam_working_model.rs::write_fisher_design_root`:

```rust
WorkingCoordinateDesign::OriginalSparseNative => crate::bail_invalid_estim!(
[...]

PR #4418, opened 2026-09-20
pirls: stream every square-root Newton solve through blocked TSQR

### LINKED ISSUE #4414: PIRLS large-n: stiff-penalty/Firth square-root solve allocates a dense (n+r+p)×p root every inner solve
## Defect

When the augmented least-squares solve is selected (a stiff penalty via `penalty.requires_root_solve`, Firth, or Fisher curvature with the square-root route), every Newton direction of a non-sparse-native design materialises the whole augmented root `[W^{1/2}X; E; sqrt(λD²)]` densely. That is an `(n + rank(E) + p) × p` matrix, charged as **4 governed copies**, and it is rebuilt on every inner solve (`solve_fisher_direction_from_root` is called from `gam_working_model.rs:1700` and `:1778`).

Evidence on origin/main (9e4544cb):

- `crates/gam-solve/src/pirls/gam_working_model.rs:616-665`: the dense branch.
  - `try_reserve_dense_f64_copies(rows, p, 4, "PIRLS Householder QR square-root solve")` at :627, with `rows = n + penalty_rows + p`.
  - Then `Array2::zeros((rows, p).f())`, `write_fisher_design_root`, and `solve_newton_direction_from_root_with_firth_hessian`, which is a full thin QR (Q is `rows × p`) followed by `root.dot(direction)` and `root.t().dot(...)` for the certificate.
- `gam_working_model.rs:476-526`: `write_fisher_design_root`.
  - The `TransformedImplicit` arm (:500-521) also reserves an `n × p` array, densifies `x_original`, and re-forms `fast_ab(original, Qs)`. That costs O(n·p²) flops and one more `n × p` allocation **per call**, even though `DesignMatrix::row_chunk_matmul_into` already forms `X[rows,:]·Qs` chunk by chunk.
- `crates/gam-solve/src/pirls/newton_solve.rs:818-877`: an O(p²)-storage streaming `TallSkinnyQrLeastSquares` already exists. `gam_working_model.rs:537-593` uses it **only** for `OriginalSparseNative` without a Firth factor. Its fixed `p`-row blocks (`newton_solve.rs:877`) make it do `n/p` tiny `2p × p` QRs.

For scale: with n = 10⁶ and p = 100 this is about 0.8 GB per copy, so about 3.2 GB is reserved and touched on every inner iteration of every outer REML evaluation. With a larger p the governor refuses the reservation, so the fit fails outright.

SPEC.md: "The program must never run out of memory on reasonably-resourced computers." and "Fitting and inference must be fast on all scales of data, but especially large-scale data where it matters the most."

## Principled fix

Use one streaming route for every coordinate design:

- Feed the rows into `TallSkinnyQrLeastSquares`, whose block height `block_rows ≥ p` comes from the shared `gam_linalg::utils::row_chunk_for_byte_budget` (the same byte budget used by the other streamed design passes).
- Fill the design rows through a `write_working_design_rows(range, out)` helper:
  - the Firth factor's `x_dense` rows when that factor exists;
  - CSR scatter for sparse designs;
  - `row_chunk_into` for explicit designs;
  - `row_chunk_matmul_into(rows, Qs)` for implicit designs.
- Delete the dense-root solver, which leaves one source of truth.

The orthogonal block reductions give the same `R` and projected residual as a single Householder QR, so the problem's conditioning is unchanged. Peak storage becomes O((block_rows + p) × p) and no longer depends on n.

SauersML:
Fixes #4414

## Problem
The stiff-penalty/Firth augmented least-squares solve built a dense `(n + rank(E) + p) × p` root for every non-sparse-native design, and it did so on every inner solve. The governor was charged four copies. The implicit `Qs` frame also re-formed `X·Qs` (n×p, O(np²)) on each call. Only sparse-native designs used the streaming `TallSkinnyQrLeastSquares`, and those used `p`-row blocks.

## Change
- **One route for every coordinate design.** `solve_fisher_direction_from_root` fills design rows in chunks with `write_working_design_rows`:
  - Firth factor rows;
  - CSR scatter for sparse designs;
  - `row_chunk_into` for explicit designs;
  - `row_chunk_matmul_into(rows, Qs)` for the implicit frame.
  
  It scales each row by `sqrt(w)` and pushes the rows, the penalty root rows and the LM rows into the TSQR.
- **Block height.** `TallSkinnyQrLeastSquares::new(p, block_rows)` takes its block height from the shared `row_chunk_for_byte_budget(total_rows, p).max(p)`. It rejects a block shorter than `p`.
- **Memory.** Peak storage is O((block_rows+p)×p) and no longer depends on n. The reservation charges six block-sized copies: the chunk, the pending block, and the stacked factor, faer's working factor, the reflector basis and Q.
- **Dead code removed.** `solve_newton_direction_from_root_with_firth_hessian`, `write_fisher_design_root`, `write_scaled_{dense,sparse}_design` and the sparse-only branch are deleted.
- **Certificate.** Every route now uses the R-space backward-error certificate the sparse route already used: `Rᵀ(Rd + Qᵀr)`, scaled by norms accumulated from the original rows. The block reductions are orthogonal, so this is the same stationarity equation.

## Tests
- `tall_skinny_qr_is_invariant_to_its_block_height` replaces the dense-vs-blocked comparison. Blocks of `p` rows and a single block of all rows both recover the exact minimiser, and their decrements agree.
- `tall_skinny_qr_rejects_a_block_shorter_than_p` is new.
- `dense_stiff_penalty_streams_psd_root_through_blocked_qr` is new. It runs the #2401 stiff fixture through a dense design and asserts:
  - the design routes to `DenseTransformed`;
  - the coefficients are finite;
  - the fit ends in the same termination state as the sparse-native fit.
- The #2401 sparse-native regression is kept, using a shared fixture.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 987b804d3c2d.

PR #4422, opened 2026-09-20
Score the small-n / corner-case smooth suites against the truth

### LINKED ISSUE #4377: tests: six smooth "stability" suites pass on any non-"nan" error and never score a successful fit against its known truth
## Where

Six stability tests pass whenever the fit returns *any* error whose message does not contain "panic" or "nan". When the fit succeeds, the tests never compare it with the known truth.

| file | lines | cases |
|---|---|---|
| `tests/pathological/smooths/new_smooths_small_n_stability.rs` | 115-123, 145-153, 171-179, 201-209 | periodic 1-D n∈{20,50,100}; sphere Wahba k=10 n∈{20..200}; sphere harmonic max_degree=2 n∈{20..200}; cylinder `te` 32-120 rows |
| `tests/manifolds/manifolds/sphere_small_k_corner_cases.rs` | 81-98, 113-130 | Wahba k∈{2,3,4,5,6,8,12}; harmonic max_degree∈{1..6} |
| `tests/manifolds/manifolds/sphere_overresourced_small_n.rs` | 77-95, 110-128 | success only needs predictions in ±10 |
| `tests/basis_smooth/smooths/matern_extreme_length_scale.rs` | 68-84, 97-113 | success only needs predictions in ±10 |
| `tests/basis_smooth/misc/bc_minimum_k_stability.rs` | 68-77, 87-96, 109-124 | clamped / anchored k∈{4..7} |
| `tests/basis_smooth/smooths/tensor_3d_smooth_robust.rs` | 79-93, 105-118, 130-143 | 3-D `te` on noiseless data; an error only has to contain the substring `"3"`, `"dim"` or `"tensor"`; success only needs predictions in ±5 |

## The code

```rust
match try_fit_predict(&formula, &data, 2, &probes) {
    Ok(_) => eprintln!("[smallN] periodic n={n}: OK"),
    Err(e) => {
        let lower = e.to_lowercase();
        if lower.contains("panic") || lower.contains("nan") {
            failures.push(format!("n={n}: opaque: {e}"));
        } else {
            eprintln!("[smallN] periodic n={n}: clean: {e}");
        }
    }
}
```

```rust
// tensor_3d_smooth_robust.rs
Ok((mn, mx)) => { assert!(mn > -5.0 && mx < 5.0); }
Err(e) => { assert!(lower.contains("3") || lower.contains("dim") || lower.contains("tensor"), ...); }
```

## Why it is wrong

- **The "panic" check can never fire.** `fit_from_formula` returns typed errors. A real panic unwinds the test before the `match` runs, so it is never converted into a string.
- **Only one kind of error can fail these tests (tensor_3d aside).** In effect, a test fails only when an error message contains "nan", or when predictions are non-finite (which `try_fit_predict` turns into a `"non-finite: [NaN, ..]"` error).
- **A solver failure on a well-posed request counts as a pass.** For example, `RemlDidNotConverge`, a stationarity refusal, or an inner-solve stall on `y ~ s(t, periodic=true)` with n=100, or on `sphere(lat, lon, k=10)` with n=200.
- **Almost every case is well-posed.** Each is a Gaussian penalized smooth with a constant (or at most low-dimensional) null space and far more distinct points than that null space needs. The exceptions are the smallest basis sizes (Wahba k=2 and 3, harmonic max_degree=1, clamped/anchored k=4), which may be invalid options; step 2 below covers them. SPEC treats a refusal of a well-posed fit as a defect, and a test that passes on it cannot fail for the behaviour it is named for.
- **A successful fit is never scored, although the truth is known.** The periodic and sphere fixtures have `y = cos t + N(0, 0.1²)` and `0.5·sin(lat) + N(0, 0.1²)`. The cylinder and 3-D tensor fixtures are noiseless (`cos θ + 0.3h`, `cos u + 0.3 sin v + 0.2w + 0.1 cos u sin v`). Yet the only check on success is finiteness, or a range of ±5 or ±10, on signals whose range is about ±1.3. A fit that returns the mean, or the wrong sign, passes.
- **tensor_3d contradicts itself.** `tensor_3d_one_periodic_margin_works` accepts an error for the exact formula and data that `tensor_3d_seam_continuity_one_periodic_margin`, in the same file, requires to fit (`unwrap_or_else(panic!)`). The `"3"` substring also matches almost any diagnostic that contains a count.

## Principled fix

For every case:

1. **Well-posed cases.** Decide from the design alone whether the request is well-posed. For these Gaussian smooths that means distinct points ≥ the penalty null-space dimension. For each well-posed case, require `Ok`. A refusal is a failure that names the case.
2. **Invalid options.** A case that is truly an invalid option (for example, a k below the basis's documented minimum) must assert the specific typed refusal that names the option. It must not accept whatever error happens to come back.
3. **Noisy fixtures.** Score the fit against the known f at the probes, using the fit's own posterior standard error: `|f̂(x) − f(x)| ≤ z·se(x)`. Take z from a stated family-wise α over the probes (Bonferroni). This is a derived coverage bar, not a range.
4. **Noiseless fixtures (cylinder, 3-D `te`).** Bound the error by the basis's own best approximation, i.e. the unpenalized least-squares projection of y onto the same design, instead of by ±5.

Any case that then fails is a real solver defect to file separately. It must not be skipped or loosened.

Measurement runs on main that print which cases currently return `Err`: focused-rust-proof 35506877995 (pathological), 35506879370 (basis_smooth) and 35506880442 (manifolds), keys HN-23-cleanerr-1/2/3. They are still queued.

SauersML:
Fixes #4377

## What changes

The small-n and corner-case smooth suites could not fail for the behaviour they are named after. They passed on any error whose message lacked a substring, and they checked a successful fit only for finiteness or against a multiple of the noise SD. Now each one scores a successful fit against its known truth, using a derived bar (`tests/common/misc/smooth_truth_scoring.rs`):

- **Target: `f_B`.** This is the min-norm least-squares projection of the noise-free training truth onto the fit's own design, taken through the SVD. The remainder `f − f_B` is orthogonal to the columns of `X`, so a Gaussian identity-link fit is *exactly* the fit of `f_B`. The basis's approximation floor `‖f_B − f‖` is reported, but it is not charged to the fit.
- **Gate: across-the-function coverage.** The statistic is `Q = (1/P) Σ (f̂ − f_B)² / s_i²`, where `s` comes from the smoothing-corrected posterior covariance. It is bounded by the Satterthwaite scaled-χ² quantile `g·χ²_{1−α}(1/g)`, with `g = tr(R²)/P²`. `α` is a declared family-wise 0.01, Bonferroni-split over each test's cases. There are no hand-tuned tolerances.
- **Refusals are defects.** A well-posed request that gets refused now fails the test.
- **Invalid requests must be refused as formula errors.** Two cases are covered:
  - `bc=anchored, k=4` leaves the basis no free coefficient.
  - `matern(length_scale ≤ 0)` is invalid. `term_builder.rs` now routes the matern length scale through `explicit_positive_length_scale`, so a non-positive scale gets a formula error instead of being accepted.

Suites converted: `bc_minimum_k_stability`, `matern_extreme_length_scale`, `tensor_3d_smooth_robust`, `sphere_overresourced_small_n`, `sphere_small_k_corner_cases`, `new_smooths_small_n_stability`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01596npY2fJbVWMwQXkUPnfu)_


SauersML:
Verify-and-land review of head 2669f4f7: **not merging yet**. The Satterthwaite gate is derived correctly: E[Q]=1, Var[Q]=2tr(R²)/P², and g=tr(R²)/P², h=1/g match the code. The `matern` length-scale refusal in `term_builder.rs` is also correct. The problem is the centre `f_B` in the rank-deficient cases.

**1. `f_B` is not identified when rank(X) < p.** `least_squares_projection` (smooth_truth_scoring.rs:84–102) takes the minimum-norm solution. When p > n, which is the point of `sphere_overresourced_small_n` (for example 120 harmonic columns at n=50 and 80 at n=30), X β_B matches the truth only at the training rows. The null-space component of β_B is then fixed by the basis's column scaling, not by the truth. So `X_p β_B ≠ f(x_p)` at the probes, even when f lies exactly in the span, e.g. 0.5+0.3·sin(lat) in the harmonic basis. This makes the module doc's claim "When the basis expresses `f`, `f_B = f`" false in exactly these cases.

The fit's band is tightest in the heavily penalized directions where the min-norm β_B places energy. That lets the test fail on a correct fit, which contradicts SPEC.md: "A failing test should always indicate problematic behavior".

Principled fix: the orthogonality argument (`r ⟂ col(X)`, so the fit of f is the fit of f_B) fixes only X β_B, not β_B. So either:
- (a) evaluate e at the training rows, where X β_B is identified; or
- (b) when rank(X) < p, centre on the truth f itself at the probes.

**2. Noiseless fixtures** (`tensor_3d_smooth_robust` l.34, `make_cylinder` l.82). With y = f exactly, σ̂² estimates only the approximation residual. The `e ~ N(0, C)` model behind the stated α does not apply, so the Bonferroni α is not the real false-alarm rate. When f is in span, s → 0 and the gate divides by zero (l.162). These fixtures need observation noise at a declared σ before this gate applies to them.

Minor, not blocking: Duchon's hybrid `length_scale` (term_builder.rs:3724) is still a plain `option_f64` read, with no `explicit_positive_length_scale` check.

#4423 is stacked on this branch and shares `smooth_truth_scoring.rs` byte for byte, so it needs the same fix before it can land.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 4b164292c05c.

PR #4423, opened 2026-09-20
Score the fit-quality stress probes against the truth instead of degraded bars

### LINKED ISSUE #4411: tests/quality fit_quality_stress: DEGRADED (rmse < 5σ of noise, span ≥ 30%) counts as a pass, so a well-posed probe that keeps 30% of the signal passes
## Where

`tests/quality/misc/fit_quality_stress.rs`:
- lines 86-104: `categorize`
- lines 150-178: `report`
- lines 1-13: the module doc

`report` is used by every probe `#[test]` in the file (24 of its 26; the other two are `zz_measure_*` readouts), from `hifreq_cyclic_k4` to `sphere_antipodal_only`.

## The code

```rust
///   COLLAPSED if span_ratio < 0.30
///   PASS      if rmse < 2 σ AND span_ratio >= 0.70
///   DEGRADED  if rmse < 5 σ
///   COLLAPSED otherwise
fn categorize(rmse: f64, sigma: f64, span_ratio: f64) -> Category { ... }
...
assert!(!matches!(cat, Category::Collapsed), "probe {probe} collapsed");
```

## Why it is wrong

1. **A DEGRADED fit passes.** The only assertion is `cat != COLLAPSED`. Every probe therefore passes as long as the fit's function RMSE against the truth is below **5σ of the per-observation noise** and it keeps at least **30% of the truth's span**.

2. **Most probes are well-posed.** Take the `hifreq_cyclic_k{4,6,8,10}` probes:
   - The truth is `sin(kθ)`, with n = 400 and σ = 0.10.
   - The basis is `cyclic(k = 2k+6)`, which represents that frequency.
   - The function error of a correct fit is about σ·√(edf/n) ≈ 0.02.

   The DEGRADED band instead accepts RMSE up to 0.50 with a span ratio of 0.30. On a signal of RMS 0.707, that is a fit that keeps about 30% of the amplitude, i.e. an almost-complete collapse. The same holds for:
   - `hifreq_bc_*`: anchored `k = 2k+8`
   - `hifreq_sphere_*`: harmonic `max_degree ≥ l`
   - `hifreq_tensor_k4/k6`: p/n ≤ 0.44
   - `near_flat_signal`
   - `heteroscedastic_noise_mean_recovery`
   - `sparse_dense_imbalance`

   All of these can be represented and are well-posed. A partial over-smoothing regression in any of them passes.

3. **The thresholds are not derived.** 2σ, 5σ, 0.30 and 0.70 are not derived from anything. As in #4387, σ is the noise SD, not the estimation error of the function. The file itself records a previous collapse (#1866) that was only caught once the fit fell below the 0.30 span floor.

4. **The module doc is stale.** Lines 11-13 say "The only hard assertion is non-finiteness". The test also asserts non-COLLAPSED, and `report_fit_error` panics on any fit error. The doc misstates what the test guarantees.

## Principled fix

Separate the probes by whether the request is well-posed, and derive each bar.

1. **Well-posed probes.** The truth lies in, or is well resolved by, the basis. This covers the high-frequency cyclic, bc, sphere and tensor k4/k6 probes, near-flat, heteroscedastic mean and sparse-dense.
   - Gate on the fit's own across-the-function coverage statistic `Q = (1/P) Σ e_i²/s_i²`, with `s_i` taken from `Vb`.
   - Require `Q` to be at most the Satterthwaite χ² bound at a stated family-wise α. This is the construction in PR #4404.
   - DEGRADED stops being a pass.

2. **Misspecified probes.** These are the step discontinuity, the wrong period, 1% outliers under a Gaussian likelihood, antipodal-only sphere data and the multicollinear tensor. Here the best any estimator in this basis can do is the unpenalized least-squares projection of the truth onto the same design.
   - Compute that oracle `f*` on the probe grid.
   - Gate `‖f̂ − f‖ ≤ ‖f* − f‖ + z·√(mean_i s_i²)`, i.e. the approximation floor plus a derived estimation allowance.
   - Do not use a 5σ band.

3. **Fix the doc.** Correct the module doc to state what is asserted.

A probe that then fails is a real defect to file on its own. It must not be reclassified as DEGRADED.

---

SauersML:
Fixes #4411

This PR is stacked on #4422 (#4377). It reuses that PR's shared `tests/common/misc/smooth_truth_scoring.rs` helper.

`fit_quality_stress.rs` had these problems:
- It gated its probes on degraded bars, meaning multiples of the noise SD and finiteness.
- It carried a `cyclic_wrong_period` probe whose bar was arbitrary.

Changes:
- Every probe is now scored against its known truth by the across-the-function coverage statistic `Q` against the basis's best approximation `f_B`. The bound is derived as the Satterthwaite scaled-χ² quantile, at a declared family-wise α that is Bonferroni-split. See #4422 for the derivation.
- The probes covered are the high-frequency tensor at k=8/10, step, outlier, heteroscedastic and the rest.
- A truth the basis cannot express is reported as the basis's approximation floor `‖f_B − f‖`. It is not charged to the fit, so it is not papered over with a loose tolerance.

---
_Generated by [Claude Code](https://claude.ai/code/session_01596npY2fJbVWMwQXkUPnfu)_


SauersML:
Verify-and-land review of head 45a6a3c2: **not merging yet**. This branch contains #4422's two commits and shares `tests/common/misc/smooth_truth_scoring.rs` with it byte for byte, so it inherits the blocker described at https://github.com/SauersML/gam/pull/4422#issuecomment-5749884524.

The blocker: `least_squares_projection` takes the minimum-norm SVD solution, and when rank(X) < p that fixes only X β_B at the training rows, not β_B itself. The centre `f_B` at the probes then depends on the basis's column scaling rather than on the truth. That breaks SPEC.md's "A failing test should always indicate problematic behavior" for any rank-deficient probe in `fit_quality_stress.rs`.

Please rebase this onto the fixed #4422 once it's corrected. That means either scoring at the training rows, where X β_B is identified, or centring on f when rank(X) < p.

---

SauersML:
Coordination note from BF-116, which was handed #4464 (also `Fixes #4411`). I compared the two diffs for #4411:

- #4464 is standalone on main and does not depend on #4422. It keeps all 24 probes, including `cyclic_wrong_period`. That probe is well posed: π is a period of sin 2θ and the cyclic basis wraps its argument. #4464 gates every probe on the same Satterthwaite across-the-function coverage bound at α = 0.01/24, and it prints the approximation floor ‖f* − f‖ separately from the in-span error.
- The blocker raised above is that the min-norm projection is not identified when rank(X) < p. It does not bite #4464's probes. Every gated probe row lies inside the training support: the grids, the collinear band b ≈ 0.95a + 0.025, and the polar caps. Take a null vector v of X_train, a combination of compactly supported or harmonic columns that vanishes on the dense training rows. By local linear independence of the basis on an open set, v also vanishes at those interior probe rows. So f* = X_p β* there does not depend on which least-squares solution is taken.

My plan is to land #4464 for #4411 once its focused run (35510077369) is green. After that, #4423's `fit_quality_stress.rs` commit is superseded. The #4377 part of this branch can go on through #4422 on its own. If you disagree, reply here and I will hold.

---

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 2ee9d5a5374d.

PR #4437, opened 2026-09-20
reml large-n: share the cached FirthDenseOperator and dense design by Arc

### LINKED ISSUE #4431: reml large-n: τ/link-ext builders deep-copy the cached FirthDenseOperator (five n×p blocks) and the dense design each outer evaluation
**Area:** crates/gam-solve/src/reml (the τ-coordinate and link-extension hyper builders)

The dense eval bundle keeps the Firth/Jeffreys operator as `EvalShared::firth_dense_operator: Option<Arc<FirthDenseOperator>>` (reml/mod.rs:4963). That operator holds **five dense n×p matrices**: `x_dense`, `x_dense_t`, `x_reduced`, `b_base` and `p_b_base` (reml/mod.rs:4632-4690). It also holds about eight n-vectors. On origin/main (21eaaf9b7b), several live consumers deep-copy it through the `Arc` instead of sharing it:

- **hyper.rs:1523-1524** (`build_tau_hyper_coords`, which runs every outer evaluation that carries τ coordinates) does `Some(cached.as_ref().clone())`.
- **hyper.rs:1930**, inside the per-coordinate loop of the same function, does `firth_op: std::sync::Arc::new(op.clone())`. That is **one more five-block copy for every τ coordinate `j`** whose `X_τ` is non-zero.
- **hyper.rs:2180-2198** (`build_tau_fixed_drift_deriv`) does `cached.as_ref().clone()` and then wraps it in `Arc::new(op)`. The same block also runs `try_to_dense_arc` on the transformed design even when the cached operator is used and the dense design is never read.
- **hyper.rs:2919-2922** (transformed-basis τ pair callbacks) does `.map(|c| std::sync::Arc::new(c.as_ref().clone()))`, an Arc → deep copy → Arc round trip. This site also ignores `free_basis_opt`. Under an active constraint face, `beta_eval` and the `X_τ` blocks are in the free basis, but the operator stays in the full transformed basis. The shared builder `build_dense_firth_operator_for_outer_basis` (objective.rs:608) exists precisely to prevent that mismatch: "Reusing the cached full-basis operator in that situation would recreate the original value/derivative mismatch".

**Other copies of the same kind:**

- `build_dense_firth_operator_for_outer_basis` itself (objective.rs:621, 636) calls `x_transformed.to_dense()`. That deep-copies the materialised n×p design (matrix/mod.rs:1648: `Self::Materialized(matrix) => matrix.as_ref().clone()`) before the operator copies it again. It also panics instead of returning the governed error.
- `link_ext_effective_design` (hyper.rs:3460), used by the SAS/mixture/Student-t pair objects every outer Hessian evaluation, does `None => x_dense_arc.as_ref().clone()` and returns an owned n×p copy, which both callers immediately wrap in a fresh `Arc`.

`gradient_hessian.rs:2429` shows the intended pattern for the same object (`Some(cached.clone())`, an Arc bump).

**Cost:** for a Firth fit with k τ coordinates, each outer evaluation allocates and copies about (2 + k)·5·n·p doubles more than it needs. At n = 10⁶ and p = 50 that is 2 GB per copy. For comparison, the inner solve is careful to keep its footprint at O(block·p).

SPEC: "The program must never run out of memory on reasonably-resourced computers." and "Fitting and inference must be fast on all scales of data, but especially large-scale data where it matters the most."

**Fix:**
- Share the operator by `Arc` at every site.
- Route `build_tau_fixed_drift_deriv` and the transformed τ pair callbacks through `build_dense_firth_operator_for_outer_basis`, the single frame-correct builder, which also fixes the free-basis mismatch at hyper.rs:2919.
- Have that builder borrow the governed `try_to_dense_arc` instead of `to_dense()`.
- Return an `Arc<Array2<f64>>` from `link_ext_effective_design` that shares the design when no basis change applies.

The `firth_dense_operator_original` sites (hyper.rs:2270, 2484, 2650) are dead and are covered by #3730, so they are left alone here.

SauersML:
Fixes #4431

SPEC.md: "The program must never run out of memory on reasonably-resourced computers."

On every outer evaluation, the Firth/τ and link-ext hyper builders deep-copied the bundle's `FirthDenseOperator`, which holds five n×p dense blocks, and also the dense transformed design. The per-τ-coordinate loop then cloned the operator once more for each coordinate.

- `build_tau_hyper_coords`: when there is no free basis, the cached operator is shared by `Arc`, and every τ coordinate shares that one `Arc` instead of calling `Arc::new(op.clone())`.
- `build_tau_fixed_drift_deriv` and the transformed τ pair callbacks now go through `build_dense_firth_operator_for_outer_basis`. That removes the duplicated inline builders. It also fixes a frame bug in the pair callbacks: they used to build the operator in the full frame even when an active-constraint free basis was set.
- `build_dense_firth_operator_for_outer_basis` now borrows the governed dense design through `try_to_dense_arc` instead of cloning it with `to_dense()`.
- `link_ext_effective_design` now returns `Arc<Array2<f64>>` and shares the governed design when no projection is needed.

This changes no numerics apart from the free-basis frame fix.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, f21655843591.

PR #4440, opened 2026-09-20
gam-cli: joint-events keeps SavedModelError's category

### LINKED ISSUE #4439: gam-cli: joint-events flattens SavedModelError, so a refused joint model file exits with the formula code where gamfit raises DataError
`gam joint-events fit` and `gam joint-events forecast` (`crates/gam-cli/src/main/run_joint_events.rs`) turn the typed `SavedModelError` from `JointEventModel::save` and `JointEventModel::load` into a bare `String`:

```rust
model.save(&args.out).map_err(|e| e.to_string())
let model = JointEventModel::load(&args.model).map_err(|e| e.to_string())?;
```

`main.rs` then converts it back with `run_joint_events(args).map_err(CliError::from)`. `From<String>` gives `ErrorCategory::Formula` with no variant. So each of these exits with the **formula** code:
- a model file that is not a model document;
- a document of another kind (for example a GAM saved where a joint model is expected);
- a stale version;
- an inconsistent model.

`gamfit.load_joint_event_model` raises `DataError` for the same file. `saved_document_error_to_pyerr` in `gam-pyffi/src/ffi/ffi_errors.rs` hard-codes `DataError` for every non-I/O refusal.

The contract it breaks:
- `cli_errors.rs`, on `CliError`: "Typed library errors keep the category their own type declares, so the exit code names the same category as the Python exception class."
- SPEC: "It's required that the CLI, Python library, and Rust library are unified, have a single source of truth, and have behavior parity."

The root cause: `SavedModelError` declares no category, so each front end chooses its own. Python chooses Data and the CLI chooses Formula.

**Fix:**
1. Add `SavedModelError::error_category` in `gam-model-api`, as an exhaustive match with no wildcard:
   - `Kind`, `Version`, `Malformed` and `Inconsistent` are Data, because they refuse the payload.
   - `Io` is Formula, because it is the caller's path. Python keeps raising the `OSError` subclass its kind names.
2. Add `impl From<SavedModelError> for CliError`, which keeps that category.
3. Make the joint-events commands return `CliResult<()>` and use `?` on save and load. This drops the `.map_err(CliError::from)` in `main.rs`.
4. `saved_document_error_to_pyerr` raises `category_error(refused.error_category(), ..)` instead of a hard-coded `DataError`, so both front ends read one table.

**Tests:**
- `refusals_name_the_category_every_front_end_reports` (gam-model-api) builds each variant and pins its category.
- `a_cli_forecast_refuses_a_bad_model_file_with_its_category` (gam-cli suite):
  - a fitted model forecasts;
  - a non-document and a wrong-kind document exit with the data code and name their reason;
  - a missing model path exits with the formula code and names the path.

Out of scope: `fit_joint_event_model` and `condition` return `EventHistoryError`, which has no `error_category`. Both front ends flatten it (the CLI to Formula, Python through `detach_py_result` to `FormulaError`), so they agree today even though both lose the category.

SauersML:
Fixes #4439

`SavedModelError` declared no category. As a result, the CLI (through `From<String>`) reported every refused joint model file with the formula exit code, while `gamfit` hard-coded `DataError`.

- **Category.** `SavedModelError::error_category` (gam-model-api) is exhaustive:
  - `Kind`, `Version`, `Malformed` and `Inconsistent` map to Data.
  - `Io` maps to Formula; Python keeps raising the `OSError` subclass for I/O.
- **CLI.** A new `impl From<SavedModelError> for CliError` keeps that category. `run_joint_events` returns `CliResult<()>`, and its `.map_err(CliError::from)` in `main.rs` is dropped.
- **Python.** `saved_document_error_to_pyerr` raises `category_error(refused.error_category(), ..)`, so both front ends read one table. Python behaviour is unchanged: every non-I/O refusal is still `DataError`.

**Tests.**
- `refusals_name_the_category_every_front_end_reports` (gam-model-api lib).
- `a_cli_forecast_refuses_a_bad_model_file_with_its_category` (gam-cli suite):
  - a fitted model forecasts, as a positive control;
  - a non-document and a wrong-kind document exit with the data code;
  - a missing path exits with the formula code.

**Guards.** `source_removal_guard` and `public_api_census` pass. `test_census` fails only on the pre-existing `glm_outer_hessian_fd_tests.rs` entry, which main also fails.

**Validation.** focused-rust-proof runs on the gam-model-api lib and the gam-cli suite. The gam-pyffi edit is one line and cannot be focus-validated, because the workflow has no Python toolchain.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, e80be3cbae36.

PR #4459, opened 2026-09-20
survival predict: read the RP cumulative-hazard band's exp rounding from gam_linalg::roundoff

### LINKED ISSUE #3529: survival competing-risks CIF: monotonicity check uses a magic 1e-10 scaled by the GLOBAL max hazard, so one high-hazard row lets real decreases in every other row be silently clamped to zero

SauersML:
## Problem

#3668 added this RP cumulative-hazard band at `crates/gam-models/src/survival/predict.rs:4857`:

```rust
let cum_band = cum * (eta_band.exp_m1() + f64::EPSILON) / (1.0 - f64::EPSILON);
```

It hand-rolls the exp rounding factor with bare `f64::EPSILON`, and `scripts/spec_ban_ratchet.py` flags it as a new `[roundoff]` SPEC violation. Every PR branch that merges main now inherits it (`roundoff=1; 1 new`).

## Fix

The band now takes its rounding growth from the shared roundoff model:

```rust
let exp_growth = gam_linalg::roundoff::accumulation_growth(2);
let cum_band = cum * (eta_band.exp_m1() * (1.0 + exp_growth) + exp_growth);
```

`accumulation_growth(2)` is γ₂ = 2μ/(1−2μ) = ε/(1−ε), with unit roundoff μ = ε/2. It covers the two roundings in `cum · exp(η)`, so `|fl(cum·e^{η±δ}) − cum·e^{η}| ≤ cum·((e^{δ}−1)(1+γ₂) + γ₂)`.

Algebraically this is the same value as the old expression: `(x+ε)/(1−ε) = x(1+γ₂) + γ₂`. So the band's numbers don't change. What changes is that the constant now comes from the one audited rounding model instead of being written out inline, which is what the ratchet enforces.

## Validation

- [x] `cargo check --workspace --tests`: exit 0
- [x] `python3 scripts/spec_ban_ratchet.py --base origin/main`: `roundoff=0`, 0 new, 0 stale
- [x] `cargo test -p gam-models --lib` on the RP/CIF band filters: 17 passed
- [x] origin/main (f7e5ec64) merges cleanly

Fixes #3529


SauersML:
Closing as superseded. b02aafc0 (merged in #4433) landed the same `accumulation_growth(2)` band on main, and main no longer has the `exp_m1() + f64::EPSILON` line.

---

PR #4464, opened 2026-09-20
fit_quality_stress: derived across-the-function coverage gate for every probe (#4411)

### LINKED ISSUE #4411: tests/quality fit_quality_stress: DEGRADED (rmse < 5σ of noise, span ≥ 30%) counts as a pass, so a well-posed probe that keeps 30% of the signal passes

SauersML:
Fixes #4411

## What changes

`tests/quality/misc/fit_quality_stress.rs` used to classify each probe with a hand-picked category ladder: RMSE against multiples of sigma, plus a span-ratio "collapse" check. Several probes also counted a fit refusal as a pass. This PR removes all of that and puts every probe behind one derived gate. For each probe the test asserts:

1. **The fit converges.** The fit and the rebuild of its prediction design must both succeed. A refusal now panics (`verdict=FIT_FAILED`) instead of being reported as a category.
2. **The output is finite.** Every prediction and every coefficient must be finite.
3. **The fit's own band covers its error across the function** (Nychka 1988; Marra & Wood 2012). The pieces are:
   - Take the probe design X_p and the Bayesian covariance Vb. Form C = X_p Vb X_pᵀ, s_i = √C_ii, and R = D⁻¹CD⁻¹.
   - The statistic is Q = (1/P) Σ e_i²/s_i².
   - Under the band's own model, E[Q] = 1 and Var[Q] = 2 tr(R²)/P².
   - The bound is the Satterthwaite match g·χ²_{1−α}(h), with g = tr(R²)/P² and h = 1/g.
   - α = 0.01/24, a Bonferroni split over the 24 gated probes.
   - No tolerance in the gate is hand-chosen.

### What the error is measured against

The error is e = f̂ − f\*. The target is f\* = X_p β\* + o_p, with β\* = X_train⁺ (f_train − o_train). This is the unpenalized least-squares projection of the truth onto the fit's own design, which is the pseudo-true function a penalized fit on those rows estimates (White 1982).

- **Truth in the basis:** f\* = f, and the gate reduces to the plain coverage test.
- **Truth not in the basis:** the gate charges the fit only for what the basis can express. This covers the anchored endpoint pins (heteroscedastic, outlier), the step, and a frequency the basis only approximates.

Each `[fit-quality]` line splits the error into two parts: the approximation floor `floor = |f* − f|` and the in-span error `in_span_rmse = |f̂ − f*|`. The line also prints Q, the bound and the dof. The data behind each verdict is therefore in the log.

### Why not the issue's well-posed / misspecified split

The issue proposed a separate bar for "misspecified" probes. Here one gate is used for every probe, and misspecification is handled by the in-span target instead of by a second, looser threshold. A second threshold would need its own constant. The projection needs none, and it still exposes the approximation floor in the printed line.

### Decisions for individual probes

- **`cyclic_wrong_period`:** the declared period π is a period of sin 2θ, and the cyclic basis wraps its argument with `rem_euclid`. The problem is therefore well posed: the fit must recover the truth, and a refusal is a failure.
- **`tensor_multicollinear_inputs`:** the penalty identifies the directions the near-collinear data leave weak, so a refusal is a failure. The pseudo-inverse puts no weight on directions below rounding level.
- **`sphere_antipodal_only`:** the gate covers only the polar rows, where the truth is known. The equator rows are checked for finiteness, and their maximum is printed in `extra`.
- **`sparse_dense_imbalance`:** one gate covers both halves, because Q weighs each row by its own s_i.

## Shared helpers

`gam-test-support::reference` gains two helpers:
- `across_function_coverage(x_probe, cov, err, alpha)`, which returns the statistic, bound and dof.
- `least_squares_projection(design, target)`, an SVD pseudo-inverse at the SVD's own rounding tolerance max(n, p)·ε·σ_max.

The crate also gains a `gam-math` dependency for `chi_square_quantile`. #4404 and #4405 (#4387, the matern/periodic/sphere sweeps) compute the same across-the-function statistic and could call this helper instead of keeping a copy.

## Validation

- **Focused run `BF-99-4411-1`** (focused-rust-proof, package `gam`, binary `quality`) on the head commit. It covers every gated probe except `hifreq_tensor_k10`: hifreq_cyclic, hifreq_bc, hifreq_sphere, hifreq_tensor_k4/k6/k8, bimodal_sharp, near_flat_signal, heteroscedastic_noise_mean, outlier_contamination, sparse_dense_imbalance, boundary_discontinuity, tensor_multicollinear, cyclic_wrong_period and sphere_antipodal.
- **`hifreq_tensor_k10`** is longer than the 600 s budget for each filter, so it runs only in the nightly slow-timeout lane.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 9e8d982cdd51.

PR #4465, opened 2026-09-20
term_builder: honour or refuse knots=[...], knot_placement=, degree= and penalty_order= on cr and periodic smooths

### LINKED ISSUE #4447: term_builder: explicit knots=[...], knot_placement=, degree= and penalty_order= silently ignored on bs=cr and periodic ps smooths
## Summary

On `origin/main` (ce58f233), `crates/gam-terms/src/term_builder.rs` accepts an explicit `knots=[...]` list and a `knot_placement=` option on three 1-D paths and then never reads them. The fit silently uses the default basis. This is the "accepted and inert option" class (#2781/#2782/#2783). The guard test `no_whitelisted_smooth_option_is_accepted_and_inert` does not catch it, because it probes `knots` only with scalar counts and never runs the `bs=cr` or periodic-`ps` arms.

## Evidence (origin/main line numbers)

1. **`parse_ps_internal_knots` (5175-5226).** At 5187, a list is read as "no count":
   ```rust
   let knots_internal = if knots_option_is_list(options) { None } else { option_usize(options, "knots")? };
   ...
   knots_internal.is_none(),   // 5224: the `inferred` flag
   ```
   So `s(x, knots=[0.3, 0.6])` reports its basis as **inferred**. The support cap at 3082 (`if inferred && !periodic_axes[0] && unique >= 2 { support_capped_bspline_dimension(...) }`) and the factor-smooth cap at 2809 can then lower the degree of a basis whose knots the user wrote out. The factor-smooth `adaptive` flag at 2901 also reads this value.

2. **Periodic arm (3125-3148).** `s(x, periodic=true, period=1, knots=[0.2, 0.5])` builds `BSplineKnotSpec::PeriodicUniform { num_basis: n_knots + degree + 1, adaptive: inferred }` (3140) from the default count, flagged adaptive. The list is dropped without an error, and so is `knot_placement=`. A periodic basis is a uniform cyclic grid and cannot take arbitrary positions, so both must be refused.

3. **`bs=cr` arm (3149-3190).** `s(x, bs=cr, knots=[0.25, 0.5, 0.75])` goes to `capped_cr_marginal_knotspec(column, k_cr, ...)`, with `k_cr` built from the default count. `select_cr_knots` then places quantile knots, so the explicit value knots (mgcv's `knots=` for `cr`) are ignored. `k=8` combined with a list is not refused either. `knot_placement=` is whitelisted (`BSPLINE_SMOOTH_OPTION_KEYS`) but on this path it is read only by the B-spline fallback for a covariate with fewer than 3 distinct values. The cr value knots themselves always sit at quantiles.

4. **`resolve_nonperiodic_bspline_knotspec` (5400-5423).** When a list is present, it returns `Provided(knots)` before `explicit_knot_placement` is read at 5424. So `s(x, knots=[0.3, 0.6], knot_placement=quantile)` accepts a placement rule that never runs.

5. **`bs=cr` ignores `degree=` and `penalty_order=` (3151-3240).** Both are copied into `BSplineBasisSpec`, but `build_cubic_regression_basis_1d` (basis/bspline_build.rs:808) builds an interpolating cubic with the `∫f''²` penalty and never reads either field; its own comment calls `spec.degree` "inert". So `s(x, bs=cr, penalty_order=1)` fits the same model as the default, and `s(x, bs=cr, degree=2)` only changes the default size through `k_cr = n_knots + effective_degree + 1`. The tensor arm already treats these values as definitional (`CR_MARGIN_DEGREE = 3`, `CR_MARGIN_PENALTY_ORDER = 2`, 5337-5343, #2782). The 1-D arm has no matching check.

## Why it is wrong

Each input must either be honoured or refused with a clear error. Here the user asks for specific knot positions and gets a different basis without any message. SPEC: "In general, do not paper over solver issues." Dropping a user's explicit basis in silence is the input-side version of that.

## Fix

- `parse_ps_internal_knots`: an explicit list is never `inferred`, so no support cap and no pilot growth rewrites it.
- Periodic arm: refuse `knots=[...]` ("uniform cyclic grid; size the basis with k=") and refuse `knot_placement=`.
- cr arm: honour `knots=[...]` as the interior value knots. Build `[lo, sorted interior..., hi]` with `clamped_knot_vector_from_internal_positions(domain.unwrap_or(range), &positions, 0)`, which applies the same finite / strictly-inside / no-duplicate validation as the open B-spline path. Refuse `k=` together with a list, and refuse `knot_placement=`.
- `resolve_nonperiodic_bspline_knotspec`: refuse `knot_placement=` together with an explicit list.
- cr arm: refuse `degree` other than `CR_MARGIN_DEGREE` and `penalty_order` other than `CR_MARGIN_PENALTY_ORDER`, and point to `bs=ps`.
- Add a regression test covering each case, and update the `docs/formulas.md` option table.

SauersML:
Fixes #4447

## Summary
- `parse_ps_internal_knots`: an explicit `knots=[...]` list is no longer reported as an *inferred* basis, so the support cap and adaptive growth can no longer rewrite a basis whose knots the user wrote out.
- Periodic B-spline arm: refuses `knots=[...]` (a periodic basis is a uniform cyclic grid, so size it with `k=`) and refuses `knot_placement=`.
- `bs=cr` arm:
  - Honours `knots=[...]` as the interior value knots, as `[lo, sorted interior..., hi]`, built by `clamped_knot_vector_from_internal_positions(domain.unwrap_or(range), &positions, 0)`. That function applies the same finite, strictly-inside and no-duplicate validation as the open B-spline path.
  - Refuses `k=` together with a list, and refuses `knot_placement=`.
  - Refuses `degree` other than 3 and `penalty_order` other than 2 (the same `CR_MARGIN_*` constants the tensor arm uses since #2782), and points to `bs=ps`.
- `resolve_nonperiodic_bspline_knotspec`: refuses `knot_placement=` together with an explicit list, where it was previously dropped without an error.
- `docs/formulas.md`: updates the `knots` row, adds a `knot_placement` row and a short `bs=cr` paragraph.

## Test
- New test `explicit_knots_and_placement_are_honoured_or_refused` in `crates/gam-terms/src/term_builder/tests.rs`. It covers:
  - cr value knots: sorted, unsorted, with `domain=`, and a full design build;
  - each refusal: k+list, duplicate, out of range, periodic list, periodic placement, cr placement, list+placement, cr degree and cr penalty_order;
  - the default `degree=3, penalty_order=2` on cr, which still builds.

---


SauersML:
Scope addition (544d416d): the same accepted-and-inert class exists on the `smooths={}` override bridge. `apply_bspline_1d` (crates/gam-terms/src/smooth_overrides.rs:782-792 on main) wrote `degree`/`penalty_order` onto a `bs=cr` term's `NaturalCubicRegression` spec, which the cr builder never reads. It also let a `knots` vector replace the cr value knots with a `Provided` open B-spline without any message. The cr's own values, which are the Python `BSpline` defaults, still pass. Other values and a knot vector are now refused, matching the formula-side refusals and the existing `knots=<count>` refusal. `CR_MARGIN_DEGREE`/`CR_MARGIN_PENALTY_ORDER` are now `pub(crate)` so both paths share one source of truth. Test: `bspline_override_on_cr_refuses_inert_degree_penalty_and_knot_vector`. Focused proof re-dispatched as HN-48-4447-2.

---

PR #4340, opened 2026-09-20
Occupancy BIC reads a certified mixture maximum, not a Lloyd plug-in (#4237)

SauersML:
Closes #4237.

## Defect
BIC is `-2 l + p ln n`, where `l` is the **maximised** log-likelihood. `wrapped_gaussian_mixture_bic_weighted` scored each anchor rung at a hard-assignment Lloyd partition (`for _ in 0..100`) with plug-in weights and width, not at the mixture maximum. On overlapping clusters this under-scores the k=2 rung.

Fixture: 48 rows at `0.40 + 0.03 z` and 12 rows at `0.48 + 0.05 z`, on the line. On main it reads `Continuous`. The BICs at the certified maxima are:

| k | BIC |
|---|-----|
| 1 | -189.39 |
| 2 | -192.52 |
| 3 | -184.33 |

So the correct verdict is `Discrete{2}`.

## Fix
Each rung's `l` is now the certified maximum of the shared-width Gaussian mixture (wrapped on the circle) over all 2k parameters: means, softmax weights and ln sigma.
- **Solver:** `certified_newton_minimum` (opt's Newton trust region) runs on the exact jet. Its decrement bands are charged on the accumulation growth of the sums.
- **Width box, lower end:** the existing resolution floor `1/(2 ess)`.
- **Width box, upper end (circle):** the uniform limit `sigma_max = sqrt(ln(2/u)/(2 pi^2))`. Above it, the wrapped normal equals the uniform density to within one rounding (derivation in the doc comment).
- **Periodic images:** the count is now derived from the omitted tail's `z^4`-weighted mass. It is no longer a fixed ±1.
- **Uncertified rungs:** a rung that does not certify leaves the law `Indeterminate`.
- **Removed:** the Lloyd loop and the plug-in.

SPEC: "A fit object must only ever come from a converged optimization."

## Tests
- The flip fixture reads `Discrete{anchors: 2}`.
- The k=1 line maximum matches the closed-form weighted MLE to within its band.
- For 84 evenly spaced circle points, k=1 certifies at the uniform limit (`l = 0` within the band; the old plug-in gave -3.31) and the law reads `Uniform`.
- The existing collapse and recovery fixtures still hold.

Focused proof run: 35506679866 (gam-sae lib, filter `coordinate_fidelity`). It is still queued.

## Separate concern, not changed here
The weekdays fixture reads `Uniform` both before and after this change. The walk stops at the first order that does not improve on the order below it, and that verdict does not come from the plug-in.

[...]

PR #4469, opened 2026-09-20
gam-sae occupancy: end the anchor walk on a certified BIC floor, not the first non-improving order

### LINKED ISSUE #4323: gam-sae manifold/coordinate_fidelity: occupancy anchor walk stops at first non-improving k; 7-anchor weekday circle classified Uniform
**Where:** `crates/gam-sae/src/manifold/coordinate_fidelity.rs`, `classify_occupancy_weighted_impl`, in the anchor-count walk (the same walk as the unweighted classifier):

```rust
// Walk the anchor count up from two and stop at the first order that does not
// improve on the order below it
for k in 2..pairs.len() {
    ...
    if !(bic < previous_order_bic) { break; }
```

**Defect (a local stopping rule decides the model order).** BIC is not unimodal in `k` for a mixture with a shared width. While `k` is below the true anchor count, one width must cover merged clusters. Each extra anchor then buys little likelihood and costs `2·ln n`, so the BIC rises. When `k` reaches the true count, the shared width collapses to the within-cluster spread and the BIC drops by orders of magnitude. The walk stops on the first rise, so it never sees that drop.

**Reproduction.** The weekday fixture of `uniform_and_discrete_occupancy_survive_the_collapse_guard_2691`: 84 rows at `(r mod 7)/7 + 0.0005·(⌊r/7⌋ − 5.5)`, unit weights. A numpy replica of the current mixture BIC gives these values:

| k | 1 | 2 | 3 | 4 | 5 | 6 | **7** | 8 | 9 |
|---|---|---|---|---|---|---|---|---|---|
| BIC | 15.3 | 25.8 | 35.8 | 39.2 | 47.8 | 36.5 | **−310.4** | −300.9 | −291.4 |

- The walk evaluates `k = 2`, sees `k = 3` fail to improve on it, and stops.
- The winner is the uniform null (BIC `0`).
- The evidence-best law, `Discrete { anchors: 7 }`, is never evaluated.

Seven tight clusters on the circle are classified **Uniform**. The existing test only asserts "not collapsed", so it passes either way.

SPEC: "The default choices should give the best model for the user." Here the returned law is not the BIC minimiser over the orders the function itself defines.

**Direction.** Evaluate the BIC over every admissible order, `k < rows` (or up to a bound derived from the resolution floor `σ_floor = 1/(2·ess)`, which limits how many separated anchors the rows can resolve), and take the minimum. The walk is `O(k_max)` mixture fits either way. Tighten the weekday test to require `Discrete { anchors: 7 }`.

This is related to #4237, which is reworking the mixture fit in the same function, but distinct from it: the stopping rule discards the right order however well each order is maximized. The effective-row BIC scale is #4319.

SauersML:
Fixes #4323

## Defect

`classify_occupancy_weighted_impl` walked the anchor count k = 2, 3, … and stopped at the first order whose BIC did not improve on k − 1. Under a shared width the BIC is not unimodal in k. Below the true anchor count, one width has to span merged clusters, so each extra anchor costs 2 ln n and buys little likelihood. On the weekday fixture (seven clusters of twelve rows, n = 84) the BIC runs:

| k | 1 | 2 | 3 | 4 | 5 | 6 | **7** | 8 | 9 |
|---|---|---|---|---|---|---|---|---|---|
| BIC | 16.0 | 25.8 | 35.8 | 39.2 | 47.8 | 36.5 | **−310.4** | −300.9 | −291.4 |

The walk stopped at k = 2 and returned `Uniform` (BIC 0) for seven anchors.

## Fix: end the walk on a certificate, not a local rule

Every density the walk scores is f = Σⱼ πⱼ κ_σ(· − μⱼ) with a shared σ ≥ σ_floor = 1/(2·ess). On the circle κ_σ is a three-image partial sum, which is at most the wrapped normal.

1. **One width suffices.** κ_σ = κ_{σ_floor} ∗ N(0, σ² − σ_floor²), on the line and, for the wrapped normal, on the circle. So every scored f is at most f_Q = ∫ κ_{σ_floor}(· − μ) dQ(μ) for some mixing law Q, whatever its order.
2. **Lindsay's bound (Jensen).** For any positive gᵢ, let D(μ) = Σᵢ (wᵢ/M) κ_{σ_floor}(uᵢ − μ)/gᵢ. Then for every Q:
   Σ wᵢ ln f_Q(uᵢ) ≤ Σ wᵢ ln gᵢ + M ln ∫ D dQ ≤ Σ wᵢ ln gᵢ + M ln sup D =: B.
3. **Hence BIC_k ≥ −2B + 2k ln n for every k.** Once that floor reaches the best BIC so far, no later order can win: the floor rises with k and the best BIC never rises. The walk stops there.

Choices, none of which can change the winner (they only affect how early the certificate fires):
- gᵢ is the rows' kernel density at σ_floor. This is f_Q with Q the empirical law, so the bound is tight when the rows are the anchors. On the weekday fixture B = 187.8, and the seven-anchor log-likelihood is 186.2.
- sup D is bounded over cells of width σ_floor/4, using each row's distance to the cell. On the line, D falls off outside the hull, so cells cover [u_min, u_max]. On the circle they cover [0, 1).
- Kernel terms beyond `reach = σ·√(−2 ln ε)` are **charged, not dropped.** On the line that is κ(reach) per row. On the circle it is the geometric tail 2κ(reach)/(1 − e^{−reach/σ²}) per row, with reach capped at 1/2 so the −1/0/+1 images cover the window.

An order whose BIC is not computable now returns `Indeterminate`, instead of silently ending the race at the orders that did compute.

## Prototype numbers (numpy replica of the Rust semantics)

| fixture | before | after | rungs evaluated |
|---|---|---|---|
| weekday, circle (84 rows) | Uniform | **Discrete{7}** | 6 |
| lattice i/84, circle | Uniform | Uniform | 1 |
| arc 0.40–0.52 (70 rows) | — | Discrete{2} | 4 |
| wrap-straddling interval | — | Discrete{2} | — |
| 500 uniform randoms, circle | — | Uniform | 57 |

Bound dominance, measured as min over k < n of BIC_k − (−2B + 2k ln n), which must be ≥ 0:
- weekday: 3.15
- lattice: 15.8
- 120 golden-ratio points, circle: 44.0
- 120 golden-ratio points, line: 44.4
- weekday on the line: 4.0

## Tests

In `coordinate_fidelity_recovery_tests.rs`:
- `uniform_and_discrete_occupancy_survive_the_collapse_guard_2691` is tightened. Weekdays must be `Discrete { anchors: 7 }` and the lattice must be `Uniform`. It used to accept any non-collapsed law, which is how #4323 slipped through.
- `weekday_bic_is_not_unimodal_in_the_anchor_count_4323` checks the premise: BIC₃ > BIC₂, and the argmin over all k is 7.
- `order_free_bound_dominates_every_anchor_order_4323` checks that BIC_k ≥ −2B + 2k ln n for every k < n, on clustered, lattice and quasi-random rows, on both the circle and the line.

## Cost note

On continuous data the walk now runs until the certificate fires, rather than stopping at the first non-improving order. On random data this is about 0.1·n rungs (57 at n = 500). Each rung is the existing O(iter·n·k) Lloyd step, which is fine at the ess this runs at. PR #4340, which rewrites each rung as a certified Newton fit, touches the same walk hunk. The two compose: that PR's rung replaces `wrapped_gaussian_mixture_bic_weighted`, and the certificate here bounds every such rung too, since every shared-width σ ≥ σ_floor mixture is covered.

Validation: focused-rust-proof, `gam-sae` lib, filter `coordinate_fidelity`.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 2d6f3cbdf36f.

Gate: `cargo check --workspace --all-targets` on MSI at main 2d6f3cbdf3: GATE_EXIT=0, cargo check --workspace --all-targets green with the build.rs ban scanner. Conflicts with main, where any, were resolved by keeping main's newer semantics and carrying this PR's intent onto them; see the merge commit.

PR #4470, opened 2026-09-20
Derive the anisotropic Matern psi-FD bounds and fail on a missing second-diagonal block (#4444)

### LINKED ISSUE #4444: tests/basis_smooth matern_aniso_eta_design_derivative_centering_fd: second-diagonal test returns early (silent pass) and both FD bars are ~1000x the derived FD error
## Where

`tests/basis_smooth/smooths/matern_aniso_eta_design_derivative_centering_fd.rs`:
- lines 163-168: `aniso_design_raw_psi_second_diagonal_matches_single_axis_fd` returns early.
- line 187: the second-derivative tolerance.
- lines 122-127: the first-derivative tolerance in `aniso_design_raw_psi_first_derivative_matches_single_axis_fd`.

## The code

```rust
// Small-n materialized path exposes the dense diagonal seconds; if the build
// chose the operator-only path (no dense blocks), there is nothing to FD here
// (the operator path is already the native reference oracle), so skip.
if deriv.design_second_diag.len() != dim {
    return;
}
...
let h = 1e-4;
...
let scale = analytic.iter().fold(1.0_f64, |m, &v| m.max(v.abs()));
assert!(max_err < 2e-3 * scale.max(1.0), ...);
```

```rust
// first-derivative test, h = 1e-6
let scale = analytic.iter().fold(1.0_f64, |m, &v| m.max(v.abs()));
assert!(max_err < 1e-5 * scale, ...);
```

## Why it is wrong

1. **The test can pass without checking anything.** The fixture has 8 rows, so the builder takes the materialized path. The sibling test requires that path (`assert_eq!(deriv.design_first.len(), dim)`). If a regression stopped filling `design_second_diag`, this test would return early and pass, and so would any consumer that relies on it. An empty result on this fixture is a defect. It is not a case to skip.

2. **The tolerances are not derived and are far too loose.** A central second difference at `h = 1e-4` has two sources of error:
   - truncation: O(h²)·max|∂⁴X| ≈ 1e-8;
   - rounding: about 4·ε_eval·max|X|/h² ≈ 1e-6, with ε_eval a few dozen ε.

   The bar `2e-3·max(1, max|∂²X|)` is about 1000× that error. It also has an absolute floor of 1, so it stays loose even when the derivative is small.

   The first-derivative test is similar. At `h = 1e-6`, its error is about ε_eval·max|X|/h ≈ 1e-8, but the bar is `1e-5·max(1, max|∂X|)` (the fold starts at 1.0), about 1000× the derived error.

   So the analytic ∂X/∂ψ or ∂²X/∂ψ² could miss a term worth 0.1% of the derivative and both tests would still pass. #1376 and #1822 were exactly this kind of missing-term or wrong-frame defect.

## Principled fix

1. Replace the early `return` with `assert_eq!(deriv.design_second_diag.len(), dim, …)`.
2. Bound each finite difference by its own error, derived from the step and the fixture, instead of a fixed factor:
   - **Truncation:** evaluate the same stencil at `h` and at `2h`. Both are O(h²), so the Richardson estimate of the truncation error at `h` is `max|D(2h) − D(h)|/3`.
   - **Rounding:** `ε_eval·max|X|·Σ|stencil weights|`. The stencil weights sum in absolute value to `1/h` for the first derivative and `4/h²` for the second. Here `ε_eval = 64ε`, from the kernel evaluation (about 16 roundings) plus the 8-center identifiability contraction, rounded up for cancellation.
   - Assert `max|D(h) − analytic| ≤ truncation + rounding`, and print both numbers.

A test that then fails shows the analytic derivative disagrees with the realized design beyond what the finite difference can resolve. That is a real defect. The fix must not widen the bound.

---

SauersML:
Fixes #4444

## Change

`tests/basis_smooth/smooths/matern_aniso_eta_design_derivative_centering_fd.rs`

- **Second-diagonal test.** The early `return` when `design_second_diag` is empty is now `assert_eq!(deriv.design_second_diag.len(), dim, ...)`. The 8-row fixture is materialized, so an empty result is a regression.
- **Derived FD bound, both tests.** Each check now asserts `max|D(h) - analytic| <= bound`, where:
  - truncation is the Richardson estimate `max|D(2h) - D(h)| / 3`;
  - rounding is `64·eps·max|X|` times the stencil's absolute weight sum (`1/h` for the first derivative, `4/h^2` for the second).
- **Output.** Each axis prints `[aniso-fd-1]` / `[aniso-fd-2]` with the error, the bound and the magnitudes, so the margin is measured, not assumed.

No production code changes. The bound is derived from the step and the fixture, not tuned.

## Validation

Focused Rust Proof: `gam` / `basis_smooth` / filter `aniso_design_raw_psi`.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, c9aecefc2959.

PR #3601, opened 2026-09-20
Firth LAML: count the Jeffreys curvature once so the outer criterion is bounded (#3507)

SauersML:
Fixes #3507.

## Root cause

The Firth inner objective is `F(β) = F0(β) − Φ(β)`, where `Φ = ½ log|XᵀWX|`. Its mode Hessian is `H_F = H_0 − H_φ`, with `H_0 = XᵀWX + S` (SPD for every ρ) and `H_φ` the Jeffreys curvature. On a separated fit, `H_F` loses definiteness at a symmetric pitchfork, so `½ log|H_F| → −∞` there. The outer criterion had two problems:

- it **added** the Tierney–Kadane correction instead of subtracting it;
- it evaluated that correction on `H_F`.

Near the pitchfork, the correction's `H_F⁻¹` contractions blow up, and nothing bounds the criterion. In the issue's fixture the criterion is about 5.3e3 at ρ = −9.25 and about 35 at the lower edge.

## Derivation

Write `log|H_F| = log|H_0| + log|I − H_0⁻¹H_φ|`, keep the TK jets on the SPD `H_0`, and count the Jeffreys curvature once. That gives

```
C = ½log|H_0| − ½log|H_F| − ½tr(H_0⁻¹H_φ) − TK(H_0)
V = F + ½log|H_0| − ½tr(H_0⁻¹H_φ) − TK(H_0) − ½log|S|₊
```

`log|S|₊` is the pseudo-determinant over the penalty's range space. As ρ → −∞, V has slope `−r/2`, where r is the penalty rank. As ρ → +∞, V is flat.

## Implementation

- **Gradient.** The mode response is `v_k = H_F⁻¹ A_k(β − μ_k)`. The determinant part is `tr(M_H Ḣ_0) + tr(M_φ Φ̇)`, with
  - `M_H = ½(G_0 − G_F + G_0 H_φ G_0)`
  - `M_φ = ½(G_F − G_0)`

  The TK jets are taken on `H_0`. For external coordinates, `hphi_tau` is added back into the drift.
- **Hessian.** A closed-form second derivative of the determinant atom, minus the TK Hessian on `H_0`.
- **Factors.** A new `TkSpdFactor` (Cholesky, with a certified eigh arm) replaces the old `HFactor` closure. `H_F` and `H_0` are both refused, not regularized, when they are not SPD.
- **Sparse-exact TK arm.** Replaced by a refusal. Firth always routes dense, so the arm was unreachable, and its sign was wrong.
[...]

PR #4473, opened 2026-09-20
reml/firth: evaluate D(Hφ)/D²(Hφ) at V = I without identity products or P·B rebuild

### LINKED ISSUE #4461: reml/firth: every D(Hφ)/D²(Hφ) evaluation rebuilds the cached P·B Hadamard-Gram block and multiplies by the identity
## What

The Firth/Jeffreys outer derivatives use two kernels, `D H_φ[u]` and `D² H_φ[u,v]`. They are only ever needed as dense `p×p` matrices, but both are computed by passing an identity rhs to the general block-apply form. That wastes work in two ways:

1. The direction-free block `P B` (with `B = diag(w') X`) is already cached on the operator as `p_b_base` (`crates/gam-solve/src/reml/firth.rs:755`, built once in `build_from_design_factor`). Each call rebuilds it anyway.
2. Several `n×p · p×p` and `p×p · p×p` GEMMs just multiply by `I`.

Evidence on `origin/main` (08628afb):

`crates/gam-solve/src/reml/firth.rs:991-994`, `hphi_direction`:
```rust
let eye = Array2::<f64>::eye(p);
let mut out = self.hphi_direction_apply(dir, &eye);
```
Inside `hphi_direction_apply` (`firth.rs:964-972`):
```rust
let etav = fast_ab(&self.x_dense, rhs);                         // X·I
let qv = &etav * &self.w1.view().insert_axis(Axis(1));          // == b_base
let m_qv = RemlState::apply_hadamard_gram_to_matrix(             // == p_b_base, O(n·r²·p)
    &self.x_reduced, &self.k_reduced, &self.k_reduced, &qv);
```
`m_qv` is one of the kernel's three O(n·r²·p) reduced Hadamard-Gram applies, so every `D H_φ[u]` does about 1.5× the heavy work it needs.

`hphi_direction` sits on production paths:
- the Firth gradient and outer-Hessian provider: `reml_outer_engine/derivative_providers.rs:478, 503, 542`, which is `FirthAwareGlmDerivatives`
- the TK exact Firth outer Hessian: `gradient_hessian.rs` `tk_hessian_rho_canonical_logit_with_route`

`crates/gam-solve/src/reml/reml_outer_engine/derivative_providers.rs:513-516`: the full-space outer-Hessian pair loop calls this once per (k,l) ρ-pair, through `compute_ift_correction_trace`. It is the only path for Firth, because `scalar_glm_ingredients` returns `None` there.
```rust
let eye = Array2::<f64>::eye(p);
let firth_second = self.firth_op.hphisecond_direction_apply(&dir_k, &dir_l, &eye);
```
Inside (`firth.rs:1149, 1163, 1182, 1208`) there are four identity GEMMs:
```rust
let eta_rhs  = fast_ab(&self.x_dense, rhs);      // X·I          (2np²)
let p_b_rhs  = fast_ab(&self.p_b_base, rhs);     // (P B)·I      (2np²)
let p_buv_rhs = fast_ab(&p_buv_base, rhs);       // (P B_uv)·I   (2np²)
let p_uv_rhs  = fast_ab(&p_uv_base, rhs);        // (P_uv B)·I   (2np²)
```
The #1575 cached TK path (`firth.rs:1243-1400`) avoids part of this, but it still keeps an `eye` field and multiplies by it twice per pair (`firth.rs:1351, 1375`). It also stores `X·I` and `(P B)·I` copies of matrices the operator already holds.

The general-rhs forms `hphi_direction_apply` and `hphisecond_direction_apply` have no other production caller. Every production call passes `I`.

## Fix (numbers unchanged)

- `hphi_direction` evaluates the V = I form directly, as `X`, `b_base` and the cached `p_b_base`. It runs 2 heavy applies instead of 3 and skips the `X·I` GEMM.
- A new `hphisecond_direction(u, v)` returns `D²H_φ[u,v]` as a `p×p` matrix without identity products. The provider calls it.
- The #1575 cache and the direct form share one core, `hphisecond_direction_from_single_index_blocks`. The cache now holds only the per-direction blocks.
- The general-rhs forms become `#[cfg(test)]` reference oracles.

`M·I == M` bit-for-bit for finite `M`; the only exception is a signed zero. Every remaining operation runs on bit-identical inputs, so the results are unchanged. The tests check two things:
- the cached pair loop is bit-identical to the direct form
- both match the general-rhs oracle to rounding

A new synthetic test (n=4000, p=40 logit) prints before/after wall-clock timings.

Out of scope, possible follow-up: the τ-partial kernels `hphi_tau_partial_apply` and `d_beta_hphi_tau_partial_apply` (`firth.rs:1660, 3035`) rebuild the same `P·(w'⊙XV)`. They are also called with `V = I` from `hyper.rs:1987, 2107, 4178`.

SauersML:
Fixes #4461

## What

The Firth/Jeffreys p×p kernels `D H_φ[u]` and `D² H_φ[u,v]` were evaluated by feeding an identity rhs to the general block-apply forms. That rebuilt the operator's cached, direction-free Hadamard-Gram block `P·B` (`p_b_base`, one O(n·r²·p) apply) on every `D H_φ[u]`. It also ran four n×p·p×p / p×p·p×p GEMMs against `I` for every `D² H_φ` pair in the full-space outer-Hessian provider.

## Change (`crates/gam-solve/src/reml/firth.rs`, `reml_outer_engine/derivative_providers.rs`)

- `hphi_direction` evaluates the `V = I` form directly from `X`, `b_base` and the cached `p_b_base`. That is 2 heavy Hadamard-Gram applies instead of 3, with no `X·I` GEMM.
- New `hphisecond_direction(u, v)`. `FirthAwareGlmDerivatives` now calls it instead of `hphisecond_direction_apply(.., &eye)`.
- The direct form and the #1575 cached TK pair loop share one core, `hphisecond_direction_from_single_index_blocks`. `FirthSecondDirEyeCache` drops its `eye`, `X·I` and `(P B)·I` copies and holds only the per-direction blocks.
- The general-rhs `hphi_direction_apply` and `hphisecond_direction_apply` had no production caller left. They are now `#[cfg(test)]` reference oracles.

## Numbers unchanged

For finite `M`, `M·I == M` bit-for-bit; the only exception is a signed zero. Every remaining operation runs on the same inputs. The tests check three things:
- `hphisecond_eye_cached_matches_per_pair_bit_identical_1575`: the cached pair loop matches `hphisecond_direction` bit-for-bit, and both match the general-rhs oracle to 1e-13 relative.
- `firth_identity_rhs_forms_match_general_apply_and_time_hn49`: on a synthetic n=4000, p=40 logit problem, both direct forms match the general-rhs oracles, and the test prints before/after wall-clock timings (see the comment below once the focused proof finishes).
- The existing `firth_direction_*`, `firth_mixedsecond_*`, `reused_firth_direction_*` and `tk_hessian_rho_canonical_logit_firth_*` tests are unchanged.

`gradient_hessian.rs` is intentionally untouched. Its only mentions of `hphisecond_direction_apply` are comments and a `#[cfg(test)]` test, so it still compiles, and leaving it alone avoids a conflict with open PR #3601.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, a6b05215dade.

PR #4480, opened 2026-09-20
structure harvest: profile the Gaussian dispersion in the gate split-LR (#4327)

### LINKED ISSUE #4327: gam-sae structure_harvest: held-out e-value scores −½·SSE at unit dispersion; dispersion does not cancel in the split LR (anti-conservative on unstandardized behavior/crosscoder, ≈1/σ²_res underpowered on standardized fits)
## Defect

`eval_log_lik` (crates/gam-sae/src/structure_harvest.rs) is the evaluation statistic of the structure-search universal-inference gate. It is used for both the candidate plug-in and the null refit (`null_fit(..., &shard.rows)`). It scores a held-out shard as

```rust
// Gaussian log-lik up to the additive constant that cancels in every
// e-value ratio: −½·SSE (unit dispersion). The gate forms differences of
// this against the null sup, so the constant and the dispersion scale drop
// out of the certified evidence.
let reconstruction = -0.5 * sse;
```

The additive constant cancels, but the dispersion scale does not. For `y ~ N(f, σ²I)` the split log-LR is `−(SSE_alt − SSE_null) / (2σ²)`, not `−½·ΔSSE`. Fixing `σ² = 1` has two effects.

- **Anti-conservative when the true residual variance is above 1.** Take a null-true shard with `y − f₀ ~ N(0, σ²)` and a predictable alternative `f₁ = f₀ + d`. Then `E₀[exp(−½(‖y−f₁‖² − ‖y−f₀‖²))] = exp(½(σ² − 1)‖d‖²)`, which is greater than 1 for `σ² > 1`. The e-value is then not an e-value, so the alpha budget does not control false births. `fit_entry` standardizes only the plain reconstruction targets (Tier-0). Behavior and crosscoder fits run on the natural scale, where the per-scalar residual variance can be any size.
- **Underpowered by about `1/σ²_res` on standardized fits.** On a standardized target `σ²_res ≈ 1 − R²`. At an EV of 0.998, the evidence for a real structure change is shrunk about 500× in the exponent. The gate stays valid there, because it is conservative, but it rarely fires. That runs against "The default choices should give the best model for the user."
- `gate_evidence` (the Pólya–Gamma gate-block log-evidence) is added on its own natural scale. The balance between it and the reconstruction term therefore depends on the units of the target.

The comment claims the scale "drop[s] out of the certified evidence". That claim is false and is the root of the defect.

## Proposed fix

The dispersion needs to be part of the likelihood that is scored, and it needs to be handled in a way universal inference accepts:

- **Numerator.** Use `N(f_alt, σ̂²_alt)` with a predictable `σ̂²_alt`, i.e. one fixed before the shard is seen, such as the candidate's likelihood dispersion on the estimation rows: `−½·SSE_alt/σ̂²_alt − ½·count·ln σ̂²_alt`.
- **Denominator.** It must stay a sup over the null family, and that family includes σ. With the one-row shards production uses (`n_shards = n − n/2`, `fit_entry`), profiling σ per shard is degenerate. The fix therefore also needs shards large enough to profile σ, or a null sup taken over the whole evaluated stream rather than shard by shard.
- **Interim fix.** Score both sides at a shared estimation-row `σ̂²`. This at least removes the unit dependence (the `ln σ̂²` terms cancel). It is exact only when σ² is treated as known, so it should be documented that way rather than as "cancels".

A test for the fix: a null-true unstandardized fixture with `σ² ≫ 1` must not have mean e-value above 1, and rescaling the target by `c` must leave the gate's verdict unchanged.

SauersML:
Fixes #4327

## Defect
`eval_log_lik` scored the held-out reconstruction residuals as `-1/2*SSE`, which fixes the Gaussian dispersion at 1. The log-LR between two Gaussian fits is `-dSSE/(2 sigma^2)`, so the gate's e-value depended on the scale of the target:
- sigma^2 > 1: anti-conservative. The null mean is E0[e] = exp(1/2 (sigma^2-1) |d|^2).
- sigma^2 < 1: power drops by a factor of 1/sigma^2.

## Fix: exact universal inference with sigma^2 in both families
- **Alternative, per shard:** `-1/2 m ln(2 pi sigma_a^2) - SSE_s/(2 sigma_a^2)` with `sigma_a^2 = SSE_est/(n_est p)`, the candidate's own estimation-row MLE. It is fixed before any held-out row is seen, so the plug-in stays predictable.
- **Null:** `R_t = sup_{theta,sigma^2} log p(D_t) = -1/2 M_t (ln(2 pi SSE_null(D_t)/M_t) + 1)`, with the parent refit on exactly the evaluated stream prefix D_t.
  - The closure returns `R_t - R_{t-1}` (`NullStreamSup`), so `log E_t = sum_s l_alt(s) - R_t`.
  - For every null (theta0, sigma0) we have `R_t >= log p0(D_t)`. So E_t is dominated by a test martingale and Ville's inequality holds for the running sup.
- **Scale invariance:** rescaling the target by c shifts both sides by `-M ln c`, so the evidence is exactly invariant.
- **Refused cases, each an explicit error rather than a number:**
  - a zero estimation SSE (the plug-in density is undefined);
  - a prefix the null reproduces exactly (the supremum is unbounded);
  - shards streamed out of order.
- **Code changes:**
  - `eval_log_lik` is removed and replaced by `alternative_shard_log_lik`, `NullStreamSup`, `gaussian_log_lik`, `gaussian_profiled_sup_log_lik`, `stream_prefix_bounds`, `fitted_reconstruction` and `residual_sse`.
  - `EstimationEvalSplit` gains `eval_rows`.

## Reference simulation (numpy)
Setup: n=100, 8 shards of 5, intercept-only null vs OLS a+bx, alpha=0.05.

| sigma | old: certify rate under null | new: certify rate under null | new: mean final e |
|---|---|---|---|
| 0.1 | 0 | 0 | ~0.21 |
| 1 | 0 | 0 | ~0.21 |
| 30 | 0.040 | 0 | ~0.21 |
| 300 | 0.046 | 0 | ~0.21 |

Power with slope 4 sigma:

| sigma | new | old |
|---|---|---|
| 1 | 31.6% | 0% |
| 10 | 29.2% | 13.3% |

## Tests
- `structure_gate_gaussian_split_lr_is_null_valid_at_every_noise_scale_4327`: sigma in {1e-2, 1, 30}, 2000 reps. Checks certify rate <= alpha + 3SE and mean e <= 1 + 3SE.
- `structure_gate_gaussian_split_lr_is_scale_invariant_and_keeps_power_4327`: target scaled by c in {1e-3, 1e3}; the evidence is equal to 1e-9 per side, and power is >= 0.2 for a 4 sigma slope.
- `null_stream_sup_telescopes_and_refuses_broken_streams_4327`.
- `estimation_eval_split_is_disjoint` now also checks that `eval_rows` equals the shard stream.
- `production_gate_consumes_corrected_pg_normalizer` is moved to the new statistic.

## Remaining concerns (pre-existing, not introduced here)
- The latents of the held-out rows are inherited from the parent fit.
- The null is a penalized fit, not the exact MLE. The sup bound still holds for the penalized family only as an approximation.
- Each gate stream refits the null on every prefix, which costs O(N^2 p).
- The PG gate-block term is not strictly a data likelihood.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 2c8fa81bbb7d.

Issue #4056, opened 2026-09-20
gam-predict: Gaussian observation band silently reads a zero/invalid prior weight (or a mismatched weight vector) as w=1 instead of refusing like the generative path


SauersML:
## Defect

`crates/gam-predict/src/lib.rs` `gaussian_observation_variance_per_row` (origin/main ~L2244) builds the per-row Gaussian observation variance `Var(Y_i | mu_i) = sigma^2 / w_i` (#2077) for the analytic observation band, but silently substitutes `w_i = 1` whenever the weight is invalid:

```rust
match prior_weights {
    Some(weights) if weights.len() == n => Array1::from_iter(weights.iter().map(|&w| {
        if w.is_finite() && w > 0.0 { obsvar / w } else { obsvar }
    })),
    _ => Array1::from_elem(n, obsvar),
}
```

Its own doc states the policy: "`None`, a length mismatch, or a non-finite / non-positive weight falls back to `w_i = 1` for that row".

- A zero weight means `Var(Y_i) = sigma^2 / 0`, which is unbounded. The band instead reports a finite unit-weight interval, so it is anti-conservative for that row.
- A NaN or inf weight, or a negative one, is not a valid variance scale. The band silently prints a plausible interval instead of refusing.
- A weight vector of the wrong length drops ALL the weights and broadcasts the pooled `sigma^2`. That is the pre-#2077 homoscedastic band, and nothing tells the caller.

## Contrast with the generative path

The doc calls this function the analytic sibling of `scale_gaussian_sigma_by_prior_weights` (`crates/gam-models/src/inference/generative.rs:261`). That sibling refuses both cases:

- on a length mismatch it errors with "prior weights length {} does not match observation count {}";
- on `!(w.is_finite() && w > 0.0)` it errors with "Gaussian replicate prior weights must be finite and > 0; got {w}".

So on the same model and rows, `sample_replicates` refuses while the analytic band prints an interval. That is exactly the kind of analytic/generative disagreement #2077 was opened to remove.

## Production path

1. `crates/gam-pyffi/src/manifold/geometry_ffi.rs` (~L7167) forwards the output of `resolve_weight_column` as `observation_prior_weights` whenever the model's weight column is present in the prediction frame.
2. `resolve_weight_column` (`gam-models` `fit_orchestration/materialize/columns.rs:257`) rejects only negative values, so zero weights reach the band.
[...]

PR #4487, opened 2026-09-20
gam-cli: gam predict --observation-interval publishes the observation band with #2077 per-row weights

### LINKED ISSUE #4479: gam-cli: gam predict cannot produce the observation (prediction) band that Python's predict(observation_interval=True) returns, and never forwards the #2077 prior weights (SPEC parity)
## Where (origin/main)

`crates/gam-cli/src/main/run_predict.rs:1028-1042`, `run_predict_unified`:

```rust
    let request = gam_predict::interval_policy::PredictionRequest {
        interval: args.uncertainty.then_some(args.level),
        covariance_mode: resolved_covariance_mode(args, model),
        // The CLI cannot request the observation (prediction) band yet: it has
        // no `--observation-interval` switch and its CSV writers carry no
        // observation columns, while `predict(observation_interval=True)` from
        // Python can. ...
        observation_interval: false,
        // Same gap, second half: the #2077 heteroscedastic band needs the
        // per-row prior weights from `resolve_weight_column`, ...
        observation_prior_weights: None,
```

The shared policy that both front ends call (`crates/gam-predict/src/interval_policy.rs`, `resolve_prediction_request`) already builds the observation (prediction) band. For example, the Gaussian band is `m ∓ z·√(σ̂²/wᵢ + se²)` (#2077). The Python FFI requests it and publishes `observation_lower` / `observation_upper` (`crates/gam-pyffi/src/manifold/geometry_ffi.rs:7263-7282` resolves the weights and `:7442` publishes the columns). The CLI hardcodes the request off. `PredictArgs` (`crates/gam-cli/src/main/cli_args.rs`) has no switch for it, and none of the three standard CSV writers in `prediction_csv.rs` has the columns.

The policy's own docstring (`interval_policy.rs:1052-1057`) names this as one of the ways the two front ends diverged:

> the CLI hardcoded the observation-interval switch off so `gam predict --uncertainty` could never produce a prediction interval; the CLI never forwarded the #2077 per-row prior weights

## Why it is wrong

SPEC.md:

> It's required that the CLI, Python library, and Rust library are unified, have a single source of truth, and have behavior parity.

For any fitted model, `gam predict` can publish only the band on the posterior mean. It cannot publish a band for a new response. That band is the one whose coverage of fresh `y` is the question, at 90% or 95%. The posterior band on its own covers a new Gaussian response at far below nominal, because it leaves out the `σ̂²/w` term entirely. There is no CLI workaround.

A weighted Gaussian fit has a second half to this gap. Even with the switch wired, sending `None` for the weights would price every row at `w = 1`, so a weight-4 row would get twice its fitted observation SD.

## Fix

1. `PredictArgs` gets `--observation-interval`. It needs `--uncertainty`, because the band is published at `--level` beside the posterior band. It is refused together with `--conformal` (a different band), and for model classes whose dedicated paths publish no observation band: multinomial, survival, spline-scan and residual-cascade. The flag is refused rather than silently dropped.
2. When the flag is set on a weighted fit, `run_predict` loads the fit's persisted `weight_column` as a required extra column. It resolves that column with the same `resolve_weight_column` the generative path uses and forwards the result as `observation_prior_weights`. A prediction frame that lacks the column is an error, following the `run_generate_unified` precedent, instead of a silent unit weight. Python differs here: it degrades to unit weights when the column is absent (`geometry_ffi.rs:7272-7275`). That silent default belongs with #4056 and #3957, and aligning it with the refusal is left to them.
3. `run_predict_unified` forwards the flag and the weights to `resolve_prediction_request`. It publishes `observation_lower` / `observation_upper` through the three standard writers. A class that goes through the shared policy but returns no band is an error.
4. Regression test: a weighted Gaussian `y ~ x` fit, predicted at one `x` with `w = 1, 2, 4`, must meet these conditions.
   - The three rows share one mean.
   - The band is centred on that mean and is wider than the credible band.
   - The band shrinks with weight.
   - `(h₁² − h₂²)/(h₁² − h₄²) = 2/3`, which is exactly what the `σ̂²/w` term implies.

Claimed by local agent HN-55 (hn55-cli-observation-interval); draft PR incoming.

SauersML:
Fixes #4479

## Problem

Both front ends call the same shared policy, `gam_predict::interval_policy::resolve_prediction_request`. That policy already builds the observation (prediction) band that Python's `predict(observation_interval=True)` returns.

The CLI hardcoded `observation_interval: false` and `observation_prior_weights: None`. As a result, `gam predict` could not publish a band for a fresh response at all. It also could not price a weighted fit's band from each row's weight (#2077). This breaks the SPEC's CLI/Python/Rust parity requirement.

## Change

- **`cli_args.rs`**: `PredictArgs` gets `--observation-interval`, which defaults to off.
- **`run_predict.rs`**:
  - The flag needs `--uncertainty`, because the band is published at `--level` beside the posterior band.
  - The flag is refused with `--conformal`, and for multinomial, survival, spline-scan and residual-cascade models, whose dedicated paths publish no observation band. It is refused rather than silently dropped.
  - On a weighted fit, the fit's persisted `weight_column` becomes a required prediction column. It is resolved with `resolve_weight_column` and forwarded as `observation_prior_weights`.
  - A class that runs the shared policy but returns no band is an error.
- **`prediction_csv.rs`**: the three standard writers take an optional `(lower, upper)` band and append `observation_lower` / `observation_upper`. Every other call site passes `None`.
- **Test** (`crates/gam-cli/tests/suite/observation_interval_prior_weights_4479.rs`, which drives the real `gam` binary): the test fits a weighted Gaussian `y ~ x` and predicts one `x` at `w = 1, 2, 4`. It checks:
  - the three rows share one mean;
  - the band is centred on that mean and is wider than the credible band;
  - the band shrinks with weight;
  - `(h₁² − h₂²)/(h₁² − h₄²) = 2/3`, which is what `σ̂²/w` implies;
  - the flag without `--uncertainty` is refused;
  - a frame missing the weight column is refused.

## Validation

`test.yml`, `cross-check.yml` and `rust-archive.yml` are disabled on this repository, so no active workflow builds the bin-internal `cli_tests` module. For that reason the regression lives in the gam-cli integration suite. It is validated with `focused-rust-proof.yml` (package `gam-cli`, target `integration`, binary `suite`, filters `predict_observation_interval regression_predict bug_hunt_predict`, key `HN-55-4479-1`). That run also compiles the gam binary with all features under `warnings = "deny"`.

The only edits to `tests/src_modules/misc/cli_tests.rs` are the mechanical arity updates for the new `observation_interval` field and the new writer parameter (`false` / `None`). They were checked statically against the new signatures.

## Known divergence

When the weight column is absent, Python silently degrades to unit weights (`geometry_ffi.rs:7272-7275`), while this CLI refuses. The refusal follows the `run_generate_unified` precedent. The Python default belongs with #4056 and #3957.


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, f443565af607.

PR #4488, opened 2026-09-20
focused-rust-proof: derive feature flags so gam-pyffi tests can link

### LINKED ISSUE #4455: focused-rust-proof cannot link any gam-pyffi test binary: `--all-features` re-enables pyo3 `extension-module`
**Symptom.** Every `focused-rust-proof.yml` dispatch with `package=gam-pyffi` fails in "Build focused test target". The failure is a link error only: rustc accepts the crate, then `rust-lld` reports `undefined symbol: PyList_New`, `_Py_DecRef`, `PyExc_ValueError`, ... and `could not compile gam-pyffi (lib test)`. No test runs.

Example: run 35501641338 (job 106054328604), package gam-pyffi, target lib, filter `nonfinite_values_round_trip_losslessly`. The build took 1145 s and exited 101.

**Root cause.** `crates/gam-pyffi/Cargo.toml` deliberately puts pyo3's `extension-module` (no libpython link) behind the default-on feature `extension-module`, so that `cargo test -p gam-pyffi --no-default-features` links libpython and runs the unit tests. `.github/workflows/focused-rust-proof.yml` builds and runs with `cargo test -p "$PACKAGE" --all-features ...` (around lines 311 and 375), which turns `extension-module` back on. So the gam-pyffi `#[test]`s the rlib crate-type was added to exercise (#954/#964/#965/#966) can never be exercised by this workflow.

**Fix direction.** When the package is gam-pyffi, the workflow's feature selection has to exclude `extension-module`, for example `--no-default-features` plus the crate's other features. The cleaner alternative is a per-package feature input rather than a blanket `--all-features`. Either way, it needs to link libpython, which the `auto-initialize` pyo3 feature already expects. The runner also needs a Python dev library present.

SauersML:
Fixes #4455

## Problem
`focused-rust-proof.yml` built and ran every package with `cargo test --all-features`. For gam-pyffi, that turns on the default feature `extension-module`, which is pyo3's "do not link libpython" switch. So the lib-test binary always failed to link (`undefined symbol: PyList_New`, ...), and no gam-pyffi `#[test]` could ever run in a focused proof. Run 35501641338 is an example.

## Fix
"Validate inputs" now derives the feature selection from the package's own manifest, using `cargo metadata --no-deps` and jq. Rather than naming a package, it applies a rule:
- A package that declares `extension-module` is built with `--no-default-features`, plus `--features <all others>` when it has any.
- Every other package keeps `--all-features`.

For gam-pyffi this reproduces exactly what `rust-archive.yml` already does (`-p gam-pyffi --no-default-features`), which is the layout the crate's Cargo.toml documents for running its unit tests. The build step and the run step read the same `FEATURE_ARGS`, so they can never build different feature sets.

The shell logic was checked locally on four synthetic feature maps, with the results:
- `{default, extension-module}` gives `--no-default-features`;
- adding `gpu` gives `--no-default-features --features gpu`;
- `{gpu}` gives `--all-features`;
- `{}` gives `--all-features`.

The YAML parses.

## Validation
A focused proof dispatched on this branch with package gam-pyffi, target lib, filter `nonfinite_values_round_trip_losslessly`. The run uses the edited workflow from the branch ref.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 07beb92930ad.

Issue #4331, opened 2026-09-20
Survival marginal-slope K ≥ 2 fits on a finite joint law still depend on each score's units (remainder of gam#3477)


SauersML:
PR #3503 fixes gam#3477 for a single score. With K=1, a survival marginal-slope fit on an uncalibrated finite law (global or local-empirical) now solves in the score's weighted standard units.

K ≥ 2 fits still use each score in its raw units:

- In `fit_entry.rs`, `standard_units` is `None` whenever `spec.z.ncols() != 1`.
- The saved contract stores a single scalar `latent_z_normalization`.
- `build_joint_marginal_slope_predict_context` refuses any normalisation that is not the identity.

So rescaling one score (for example, cm to m) in a K ≥ 2 fit on a finite joint law changes the fitted model, not just its parameterisation.

**What the fix needs**
- A per-score normalisation stored in the saved joint latent law, used at both fit and predict time.
- Atom maps and slope-offset maps for each axis.
- ReSolve hint scaling for each score.

**Acceptance:** a K ≥ 2 version of `survival_acceptance::finite_law_score_units_3477`, where rescaling any single score leaves the predictions and fitted slopes (after back-transforming to that score's units) equal to machine precision.

Related: #2949 (the joint latent law for K ≥ 2 is a different problem) and #2938.

SauersML:
Claimed by wave8 session (claude/fix-4331-k-score-units); draft PR incoming.

---
_Generated by [Claude Code](https://claude.ai/code)_

SauersML:
Status from the gam-fb landing session (2026-09-21), what is done and what still blocks this:

An open pull request is already in flight for this issue: PR #4490, "Survival marginal-slope: solve every joint-law score in its own units (#4331)".

No files written. Per the lead's instruction, an issue with an open PR is left to that PR.
Checked with `gh pr list --state open --search "4331"` at origin/main 5496c6585eba11f540046ec838d776582874f8c8 on 2026-09-21.

SauersML:
Status from the gam-fb landing session (2026-09-21), what is done and what still blocks this:

The K >= 2 per-score unit maps landed with PR #4490; every item the issue asks for is on main

WHAT FIXED IT. `f97deaf670`, "survival marginal-slope: solve every joint-law score in its
own units (#4331)", merged from PR #4490 at 2026-09-21T11:27Z and present at this base. The
issue's three requirements, each against the code at this base:

[...]

PR #3503, opened 2026-09-20
Survival marginal-slope: solve a finite latent law in the score's standard units (#3477)

SauersML:
Fixes #3477

This builds on #3476 (merged), which fixed #3231 for BMS. That fix did not reach the survival marginal-slope family.

## What was wrong
On a finite latent law `{u_k, w_k}`, the anchor `Σ_k w_k Φ(a + b·u_k) = Φ(q)`, with `η = q·c(g) + g·z`, does not change under `z → (z−m)/s`, `u_k → (u_k−m)/s`, `g → g·s`. The row intercept absorbs the shift. So a score recorded as `C + D·z` is the same model as `z`.

The survival marginal-slope fit ignored this. It solved the global-empirical and local-empirical laws on the raw score. A polygenic-score SD of 4.51e-4 (gnomon#2402) put the slope coefficient near `b/D ≈ 2e3`, so every outer seed sat on a scale that the units set.

## Fix (`survival/marginal_slope/fit_entry.rs`, `survival/predict.rs`)
- **Units.** A K=1 fit on an uncalibrated finite law solves in the score's weighted standard units, using `finite_law_in_standard_units` from #3476, now `pub(crate)`.
  - The saved `latent_z_normalization` combines both maps: `mean' = mean + sd·m` and `sd' = sd·s`.
  - `slope_offset` is scaled by the saved sd, at fit and at predict.
- **Declared-law compression.** The certificate reads `observed_slope / s`.
- **ReSolve slope hints.** Hints are carried between score axes by `s_to/s_from`:
  - closed-form arms go from `(0,1)` to the fit units;
  - moving-law arms map through `MovingLawCandidates::score_axis`;
  - on the ζ axis the hint is dropped.
- **Test update.** The `declared_latent_law_2923` helper now reads the slope on the raw axis: `(row·β + offset)/score_sd`, with grid nodes at `mean + sd·u`.
- **Not changed:** K≥2 fits still use the scores as given. A single scalar map does not standardize a multi-score law, so this stays open under #3477.

## Test
`survival_acceptance::finite_law_score_units_3477` draws one frame (N=500, with the score mean moving on x) twice, as `z` and as `0.013 + 4.51e-4·z`, and fits both with `latent_measure = "global-empirical"`. It asserts:
- the saved sd ratio and location agree to `N·ε`;
- `max|Δβ| ≤ √ε·scale`;
- `|Δll| ≤ √ε·|ll|`;
- `max|ΔS| ≤ √ε` for the predicted survival.

[...]

Issue #3477, opened 2026-09-20
Survival marginal-slope fits on a finite latent law depend on the score's units (survival counterpart of gam#3231)


SauersML:
## What is wrong
gam#3231 (PR #3476) makes a Bernoulli marginal-slope fit on a finite latent law (global-empirical / local-empirical) independent of the score's units. It does this by solving the law in the score's weighted standard units. The survival marginal-slope family still takes the score as given.

- `row_math.rs:1956` standardizes each score column with `standardize_latent_z_with_policy(&input, weights, context, policy)`.
- The default policy is `LatentZNormalizationMode::Frozen { mean: 0.0, sd: 1.0 }` (`latent_measure.rs:632`). The module doc (`latent_measure.rs:9-11`) says this transform "is the identity: it checked, it warned, and it passed z through unchanged".
- So a survival fit whose score is recorded as `C + D·z` (gnomon#2402's raw PGS has D = 4.51e-4) has slope coefficients near `b/D`, and its outer seeds sit on a scale the units set.

On a finite law `{u_k, w_k}` the anchor is unchanged under `z → (z−m)/s`, `u_k → (u_k−m)/s`, `b → b·s`, `a → a + b·m`. So the same frame in other units is the same model, as argued in gam#3231.

## Reproducer (recipe; not run)
Use `crates/gam-models/tests/suite/bms_finite_law_score_units_3231.rs` from PR #3476, with a survival family instead of `bernoulli-marginal-slope`:
- the same draw with the score mean moving on x;
- event times from the survival marginal-slope model;
- `latent_measure = "global-empirical"`;
- fit once on `z` and once on `0.013 + 4.51e-4·z`.

Compare the log-likelihood, max |Δβ| and predicted survival.

- Expected: agreement to the fit's convergence (√ε), as the Bernoulli family now shows (Δβ 2.5e-14, Δll 5.7e-14, Δp 7.2e-16).
- On the Bernoulli family before #3476 the gap was Δβ 1.5e2, and on gnomon#2402's frame the raw-units fit refused at startup.

## Expected fix
The same fix as #3476:
- solve a finite (empirical) law on the score's weighted standard units (`weighted_location_scale`);
- compose the saved `latent_z_normalization` with those units;
- scale any slope offset by the saved sd at fit and at predict.

The standard-normal closed form keeps the score as given, because there the score's scale is part of the model.

Found while fixing gam#3231.

SauersML:
[...]

PR #4490, opened 2026-09-20
Survival marginal-slope: solve every joint-law score in its own units (#4331)

SauersML:
Fixes #4331

This is stacked on #3503 (#3477, K=1). Once #3503 merges, this PR is retargeted to `main`.

## What was wrong
A K ≥ 2 survival marginal-slope fit anchored on the joint law of its score vector read every score in the units it was recorded in. Only score 0's map was ever kept: `standardize_latent_z_matrix_with_policy` dropped the maps of columns `1..`. The save path therefore refused any K ≥ 2 model whose scores were not already standardised, and prediction required the identity normalisation.

A score recorded as `C + D·z` gave a different fit from `z`. With `D = 4.51e-4` (a polygenic-score SD), the per-score slope coefficient sat near `b/D`.

## Math
The joint-law anchor reads the drive `Σ_j g_j z_j` on the empirical joint law of the K columns. The identity is

`Σ_j g_j z_j = Σ_j (g_j s_j)·(z_j − m_j)/s_j + Σ_j g_j m_j`

so each coordinate's affine units are a coordinate choice on their own. The constant `Σ_j g_j m_j` is absorbed by the row intercept, which the anchor `Σ_k w_k Φ(a + b·u_k) = Φ(q)` solves for, and slope `j` scales by `s_j`.

The fit therefore solves every uncalibrated coordinate in its own weighted standard units `(m_j, s_j)`, from `weighted_location_scale`. This includes a coordinate whose own gate kept the Gaussian label, because the joint law replaces the per-score labels.

- **Offsets.** A raw slope offset `o` on score `j` is the standardised-axis offset `s_j·o`. Each per-score channel therefore gets its own offset column, `s_j·o + b₀`, through the new `SlopeTopology::materialize_identity_with_channel_offsets`. `materialize_with_design` now takes an `n × K` offset matrix and validates its width.
- **Persistence.** `SurvivalJointLatentLaw` now records `score_location[j]` and `score_scale[j]`. These are the composition of the policy map with the fit map: `mean' = mean + sd·m_j` and `sd' = sd·s_j`. `validate` checks them.
  - The scalar `latent_z_normalization` stays score 0's map, and the replay validator requires the two to agree.
  - The K ≥ 2 save refusal is removed.
- **Prediction.** `build_joint_marginal_slope_predict_context` reads each new score as `law.standardized_score(j, z) = (z − m_j)/s_j`. It applies `s_j·o` to channel `j`, the same factor the fit used.
- **ReSolve hints.** The closed form reads the scores raw, so `slope_hint_onto_score_units` scales each per-score coefficient range by its own `s_j`. A shared channel over K ≥ 2 scores reads one slope on the drive, which a different factor per score cannot carry, so it starts afresh.
- **Not changed.** A K ≥ 2 Gaussian closed form keeps its scores as given, because it states `z ~ N(0, I)` in those units. Moving laws and ALO replay remain K = 1 only.

## Tests
- **New: `survival_acceptance::finite_law_score_units_4331`.** It uses N = 600 and K = 2 with a global-empirical joint law, and fits the same frame twice: once with z₁ as given, and once with z₁ recorded as `0.013 + 4.51e-4·z₁`. It asserts:
  - the score-0 map is identical in both fits;
  - the score-1 scale ratio is `4.51e-4` to `N·ε`, and its location matches;
  - the scalar map equals the law's score-0 map;
  - the coefficients agree after rescaling, within `√ε·scale`;
  - the log-likelihood agrees within `√ε·|ll|`;
  - the predicted survival agrees within `√ε`.
- **Updated: `joint_latent_law_2929`.** The replay drive is now `Σ_i std_slope_i·(z_i − loc_i)/scale_i`, read from the persisted units. The reported raw-unit slopes are `std_slope/scale`.
- **Unit test: `slope_layout::distinct_channel_offsets_reach_their_own_channels`.**

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, ad57c800ac23.

PR #4499, opened 2026-09-20
Adapt the NUTS metric by the warmup's own shrinkage; drop the hand-set weights and the dim=50 switch (#3263)

### LINKED ISSUE #3263: NUTS sampler tuning is hand-set: target-accept clamp [0.90|0.92, 0.95] and hand mass-matrix shrinkage 0.10–0.18 split at dim=50 (hmc_io.rs)

SauersML:
Fixes #3263. #3370 already removed the target-accept clamp. This PR handles the other half: the hand-set metric shrinkage and the `HIGH_DIM_THRESHOLD` switch.

## What changes

- **general-mcmc pin moves adc78bb → c6e415d.** That revision's open warmup computes each window's metric from the draws:
  - It shrinks the log variances toward the identity, which is the correct target in the Laplace-whitened space, using the window's plug-in James–Stein intensity.
  - It weights each variance's error by the ESS of the squared deviations, so shrinkage is strong when warmup is short or noisy and fades as the estimate gets precise.
  - It measures the step-size window against the settled dual-averaging mean, and adapts the covariance as the inverse mass.
  - It no longer reads the caller's fixed `regularize`.
- **Deleted from `hmc_io.rs`:** `HIGH_DIM_THRESHOLD` and the four `MASS_REGULARIZE_*` / `SURVIVAL_MASS_REGULARIZE_*` weights (0.10 / 0.14 / 0.12 / 0.18).
- **One config instead of two.** `robust_mass_matrix_config` and `robust_survival_mass_matrix_config` differed only in those weights. They are now one `open_warmup_mass_matrix_config`, which every whitened NUTS run builds itself, and the `mass_cfg` plumbing is gone.
- **Stale note removed.** The `NUTS_TARGET_ACCEPT` note said targets ≥ 0.98 never end warmup; c6e415d fixed that.

## Tests

- The #3090 test `mass_matrix_configs_floor_by_regularization_not_jitter_3090` asserted that the deleted weights were a positive floor, so its premise is gone. It is replaced by `open_warmup_metric_samples_a_hundred_coordinate_unit_gaussian_3263`, which runs `run_whitened_nuts_samples` on a 100-coordinate unit Gaussian and checks two things:
  - warmup ends;
  - every coordinate's mean and variance lie within five Monte Carlo standard errors.
- Local results, after merging origin/main at 7c825599:
  - `cargo check --workspace --tests` is clean. The only warning is the `IxDyn` import in gam-pyffi, which is already on main.
  - gam-inference lib: 199 of 199 pass.
- `scripts/source_removal_guard.py --base origin/main`: verified, 8 guarded removals, acknowledged in `docs/source-removal-changes.json`.
- `scripts/public_api_census.py --base origin/main`: integrity verified, with no public function change.
- `scripts/ban_scanner.sh`: no new hit from this PR. The one new hit it reports, `gam-models/src/survival/predict.rs:4864` [roundoff], is already on main and is not in this diff.

https://claude.ai/code/session_01NhzcgxBoVfScAoxjFP1SYi

---
_Generated by [Claude Code](https://claude.ai/code/session_01NhzcgxBoVfScAoxjFP1SYi)_


SauersML:
Closing this as superseded. f978f1d0 on main already makes the same change:
- the c6e415d pin;
- `HIGH_DIM_THRESHOLD` and the four `*MASS_REGULARIZE_*` weights deleted;
- one `nuts_mass_matrix_config` at both call sites;
- the `NUTS_TARGET_ACCEPT` note corrected.

The only thing here that main lacks is `open_warmup_metric_samples_a_hundred_coordinate_unit_gaussian_3263`. It runs `run_whitened_nuts_samples` end to end on a 100-coordinate unit Gaussian and checks that warmup ends and that every coordinate's mean and variance are within 5 Monte Carlo SEs. Main's `mass_matrix_config_leaves_shrinkage_to_each_warmup_window_3263` checks only the config fields. If that end-to-end check is wanted, it can be a small test-only PR against current main.

---

PR #4500, opened 2026-09-20
tests/gam-sae #2691: ARD precision ladder asserts its anchors unconditionally

### LINKED ISSUE #4497: tests/gam-sae #2691: ARD precision ladder never asserts — exp(ln 1e-3) > 1e-3 and exp(ln 1e9) < 1e9 so anchor finds return None; refused rungs skipped
## Defect

`crates/gam-sae/src/manifold/tests_chart_collapse_2691.rs`, `zz_2691_ard_precision_ladder_collapses_the_chart` (origin/main 08628afb3), asserts nothing, for two reasons.

**1. The anchor lookup can never match.** Rungs store `log_ard.exp()` for `log_ard` in `[ln 1e-3, …, ln 1e9]`, and the only assertions sit inside

```rust
if let (Some(low), Some(high)) = (
    rungs.iter().find(|(alpha, _, _)| *alpha <= 1.0e-3),
    rungs.iter().find(|(alpha, _, _)| *alpha >= 1.0e9),
) { assert!(..); assert!(..); }
```

But in IEEE-754 double (glibc libm, and Rust's `f64::exp`/`ln` with it):

```
exp(ln(1e-3)) = 0.0010000000000000002   (> 1e-3)
exp(ln(1e9))  = 999999999.9999993       (< 1e9)
```

so both `find`s return `None`, the `if let` never matches, and the test passes with **zero assertions executed**, whatever the chart does.

**2. Refused rungs are silently dropped.** `if let Err(error) = fit { eprintln!(..); continue; }`, so even with working anchors, a refusal at either end rung would also skip the assertion. SPEC: "XFAIL pattern on tests is never allowed. A failing test should always indicate problematic behavior."

(The assertion messages also contain run-on whitespace from a lost `\` continuation.)

## Fix

* Address the anchors by ladder position (`rungs.first()` / `rungs.last()`) after `assert_eq!(rungs.len(), ladder.len())`, with a comment explaining the round-trip.
* Make an inner refusal at any rung a test failure (`unwrap_or_else(|e| panic!(..))`). The fixture is the same perfectly recoverable circle cloud at every rung, and the #2691 degenerate-chart refusal lives at the production entry (`fit_entry.rs`, `SaeFitError::DegenerateChart`), not in `run_joint_fit_arrow_schur`.
* Keep both mechanism assertions unconditional (`low` circular variance > 0.5; `high` circular variance < 1e-6 and ≤ 1 resolved chart point), and fix their messages.

SauersML:
Fixes #4497

`zz_2691_ard_precision_ladder_collapses_the_chart` executed zero assertions for two reasons:
- Its anchors were found with `alpha <= 1e-3` / `alpha >= 1e9` on `log_ard.exp()`. Because `exp(ln 1e-3) = 0.0010000000000000002` and `exp(ln 1e9) = 999999999.9999993`, both `find`s returned `None` and the guarded `if let` never ran.
- A refused rung was also skipped with `continue`.

This PR:
- Requires every rung to fit (`unwrap_or_else(|e| panic!(..))`) and asserts `rungs.len() == ladder.len()`.
- Addresses the anchors positionally (`first()` / `last()`) and documents the round-trip.
- Makes both mechanism assertions unconditional: low-precision circular variance > 0.5, and high-precision circular variance < 1e-6 with at most 1 resolved chart point.
- Fixes the assertion-message whitespace.

Validation: CI plus focused-rust-proof (gam-sae lib, filter `zz_2691_ard_precision_ladder_collapses_the_chart`, key HN-50-4497-1).

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, e379760251eb.

PR #4503, opened 2026-09-20
tests/identifiability: derived two-sided size gate for the kappa flatness test

### LINKED ISSUE #4501: tests/identifiability: kappa flatness SIZE gate allows 4/9 rejections (underived reps/2) and is one-sided
## Defect

`tests/identifiability/misc/constant_curvature_kappa_coverage_sims.rs`, `flatness_test_holds_size_across_flat_replicates` (origin/main 08628afb3, line 575):

```rust
// Allow a strict minority (≤ reps/2) to absorb the small-R binomial tail ...
assert!(rejections <= reps / 2, ...);
```

With `REPLICATE_COUNT = 9` this allows **4 of 9** false rejections at α = 0.05. The bar is not derived. The same file derives `MAX_MISSES = 2` for n = 9 from the exact binomial (its own table: "the smallest `k` holding false alarm ≤ 1%"), and applies it to every coverage gate *except* this one. Exact `Binomial(9, p)` pass probabilities:

| true size p | P(pass) at `≤ reps/2` (4) | P(pass) at `≤ MAX_MISSES` (2) |
|---|---|---|
| 0.05 (correct) | 1.0000 | 0.9916 |
| 0.10 | 0.9991 | 0.9470 |
| 0.20 | 0.9804 | 0.7382 |
| 0.30 | 0.9012 | 0.4628 |

So the current gate passes a test whose true size is 4× nominal 98% of the time, and 6× nominal 90% of the time. It cannot see size inflation.

The gate is also **one-sided**. A rejection count only catches anti-conservative behaviour. A flatness LR whose p-values pile up near 1 (a deflated statistic, a wrong reference distribution, a profile that never leaves κ̂ = 0) rejects nothing and passes, even though "conservative is a bug just like anti-conservative". The per-replicate p-values are already collected (`pvals`) but only printed.

## Fix

* `rejections <= MAX_MISSES`, with the binomial derivation in the comment (false alarm 0.84% on a correctly sized test).
* Add a two-sided calibration gate on the collected p-values. Under the null an interior χ²₁ p-value is Uniform(0,1), so `mean(p)` has mean 1/2 and sd `sqrt(1/(12·R))` (0.0962 at R = 9). Assert `|mean(p) − 1/2| ≤ 4·sqrt(1/(12·R))`, i.e. mean p ∈ [0.115, 0.885]. The exact Irwin–Hall tail makes the false alarm ≈ 7.5e-6. This fails a test whose p-values are pinned at either end.

If the tightened gates fail on main, the flatness test is mis-sized, and that is a code bug to fix, not a bar to relax.

SauersML:
Fixes #4501

`flatness_test_holds_size_across_flat_replicates` allowed `reps/2` (4 of 9) false rejections at α = 0.05. It would pass a test whose true size is 4× nominal 98% of the time. It was also one-sided, so a conservative LR (p-values pinned near 1) passed silently.

This PR:
- `rejections <= MAX_MISSES` (= 2 for R = 9). This is the file's own exact-binomial derivation, P(fail | correct size) = 0.84%.
- Adds a two-sided calibration gate on the already-collected p-values: `|mean(p) - 1/2| <= 4 * sqrt(1/(12 R))`, i.e. mean p in [0.115, 0.885] at R = 9. It follows from Uniform(0,1) null p-values; the Irwin–Hall false alarm is about 7.5e-6.

Validation: CI plus focused-rust-proof (gam integration, binary `identifiability`, filter `flatness_test_holds_size_across_flat_replicates`, key HN-50-4501-1).

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, da7ea3752982.

PR #4509, opened 2026-09-20
Carry the residual repair block's Murphy–Topel channel under the pooled joint covariance

SauersML:
Refs #2985. This does not close it: the conditional Σ(a) channel below is still open.

## What

Before this change, a calibrated Bernoulli marginal-slope fit with a residual repair block (#2924) withheld its covariance with `BmsGeneratedRegressorResidualRepairChannelUnavailable`, which 61100b27f5 added. The reason is that the generated-regressor correction had no ζ channel for the block. This PR implements that channel when the joint (z, r) covariance is the **pooled** one, under the standard-normal and global-empirical score laws. In those cases the fit now corrects and publishes its covariance.

`ResidualDriveKernel::score_zeta_sensitivity` (in `residual_repair_kernel.rs`) returns ∂s/∂ζ, summed over the three ways the block reads the calibrated score:

1. **Row channel.** `J_iᵀ ∂²ℓ/∂p∂ζ_i` from the row's Order-2 jet, where ζ is a jet axis.
2. **Pooled covariance channel.** The block uses `a = Σ̂₀₀`, `c = Σ̂_r0`, `γ = c/a` and `Σ_rr = Σ̂_rr + κ Σ̂_r0Σ̂_0r` with `κ = a⁻² − a⁻¹`, and through them the anchor's `u = γᵀβ` and `v = βᵀΣ_rrβ`. Every row's ζ enters these through `∂a/∂ζ_j = 2w_j(ζ_j − ζ̄)/W` and `∂c/∂ζ_j = w_j(r_j − r̄)/W`. The row sums of `∂ℓ/∂u` and `∂ℓ/∂v` are pulled back through them.
3. **Grid channel** (global-empirical law only). The node sensitivity is pulled back through `EmpiricalZGridBuild::node_zeta_vjp`.

`block_specs.rs` passes the sensitivity to `generated_regressor_correction`, after checking that its width matches `p_β`.

**Still withheld:** the conditional Σ(a) (the M-estimate), whose implicit derivative is not implemented yet. That case keeps the typed decline, and the reason string now names only it. So does the LocalEmpirical law, which `score_zeta_sensitivity` refuses by name.

## Tests

- `residual_score_zeta_2985_tests.rs` (unit tests):
  - Central finite-difference gates on `score_zeta_sensitivity` under the standard-normal and global-empirical laws, at derived steps.
  - A check that the probe rows are a real kernel width.
  - A check that a conditional joint covariance returns an error rather than a sensitivity.
  - A mutation check: flipping the sign of the pooled channel's `p_u·du + p_v·dv` term fails both FD gates.
- `tests/regressions/misc/bms_residual_certificate_2985.rs`:
  - The moving-law fixture, which fires the calibration, now asserts that the fit publishes its covariance and records no decline, both with and without the block.
  - The wire test uses the conditional-Σ(a) reason.
  - All three tests pass locally.
- `cargo check --workspace --tests` is clean after merging main. The focused `2985-reg` proof passed on bef2c672 (run 35502670826).

Docs: the generated-regressor paragraph in `docs/marginal-slope.md` is updated.

https://claude.ai/code/session_014q2jwvY79VUkC7eChYEGt4

---
_Generated by [Claude Code](https://claude.ai/code/session_014q2jwvY79VUkC7eChYEGt4)_


SauersML:
gam-2924 has checked this against its own #2985 derivation: the row, pooled-covariance and grid channels are the same mechanism. The grid channel reads the residual row's own anchor point, and the pooled channel differentiates `Σ̂₀₀` and `Σ̂_r0` through `κ = a⁻² − a⁻¹`.

The conditional Σ(a) channel you leave open lands as a follow-up on top of this PR once it merges. It is the tangent-linear model of `ConditionalScoreCovariance::fit` (centring, coupling ridges, log-innovations), FD-gated against refits, plus a bootstrap receipt of the corrected covariance. Please leave #2985 open for it; it closes with that commit.

SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, a1720a42eacc.

PR #4412, opened 2026-09-20
single source of truth: collapse private scalar kernels outside tests/ onto gam_math

SauersML:
Fixes #4406

This PR deletes 15 private scalar-kernel copies in 12 files outside the root `tests/` tree. Every call site now imports the `gam_math` kernel it was duplicating.

- **`logistic`:**
  - `bench/cargo_benches/joint_absolute_risk.rs`: the two-branch `sigmoid` becomes `gam_math::special::logistic`. It performs the same operations, so outputs are bit-identical. `sigmoid_bound` keeps its name, and its doc now points at `logistic`.
  - `crates/gam-cli/examples/finite_response_teacher_2946.rs`: the two-branch `logistic` copy is removed.
  - `examples/repro2748_geo_disease_matern_flexible.rs`: the naive `sigmoid` is removed.
- **`softplus`:** the bench's `softplus`, which is the same expression as `gam_math::special::softplus`, is removed.
- **`normal_cdf`:**
  - The bench's copy (statrs `erfc`) is removed.
  - The `measure_jet_range_seed` `marginal_slope_screen_response_tests` copy (statrs `erf`) is removed.
  - Six pass-through wrappers `fn normal_cdf(x) { gam_math::probability::normal_cdf(x) }` in the gam-models survival integration tests are removed.
  - `examples/repro979_margslope.rs` loses `erf_approx` (Abramowitz & Stegun 7.1.26, ~1.5e-7 absolute error), and its call `0.5*(1+erf_approx(eta/√2))` becomes `normal_cdf(eta)`.
- **`normal_pdf`:** the test-module copies in `gam-inference` `skovgaard.rs` and `measure_jet_range_seed.rs` are removed.

SPEC.md: "It's required that the CLI, Python library, and Rust library are unified, have a single source of truth".

**LoC delta (excluding the ledger):** 12 files, +22 / −91, net **−69**. The ledger `docs/source-removal-changes.json` gains one entry (+22) covering the 10 guarded removals. The bench and `examples/` copies are unguarded private items.

Guards, run locally against base 8887329660b6:
- `source_removal_guard`: 10 guarded removals, matched exactly by the ledger.
- `public_api_census`: 13908 public functions on both sides.
- `test_census`: 12870 tests on both sides.

No tests are removed, and no tolerances change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MwMvgB7i335nVFQ8vfZ44B

---
[...]

PR #4526, opened 2026-09-20
gam-models test_support: anchor on the shared Gauss-Hermite rule instead of a gauher transcription

### LINKED ISSUE #4522: single source of truth: gam-models test_support carries a Numerical Recipes gauher copy of the standard-normal Gauss-Hermite rule
`crates/gam-models/src/test_support.rs:14` (origin/main 0b9dc2c6), `gauss_hermite_probabilists(m)`, builds its own Gauss-Hermite rule. It is the Numerical Recipes `gauher` routine transcribed line for line, with the same asymptotic root seeds and the same orthonormal recurrence:

```rust
const PI_QUARTER_INV: f64 = 0.751_125_544_464_942_5;
...
z = match i {
    0 => (2.0 * n + 1.0).sqrt() - 1.85575 * (2.0 * n + 1.0).powf(-1.0 / 6.0),
    1 => z - 1.14 * n.powf(0.426) / z,
    2 => 1.86 * z - 0.86 * x[0],
    3 => 1.91 * z - 0.91 * x[1],
    _ => 2.0 * z - x[i - 2],
};
... for _ in 0..200 { /* Newton */ if (z - z1).abs() <= 3e-16 * (1.0 + z.abs()) { break; } }
```

It then maps the result to the standard-normal scale (`u = √2·x`, `w/√π`, ascending).

The workspace already has the one owner of exactly this rule: `gam_math::quadrature::standard_normal_gauss_hermite_rule(n)` (crates/gam-math/src/quadrature.rs:261). It uses Golub-Welsch, returns ascending nodes and returns weights with unit mass. gam-models already depends on gam-math.

The copy is used by the lib tests of `latent_anchor`, `latent_law_compression`, `survival::marginal_slope::{joint_latent_law, slope_geometry}`. Those tests exercise the anchor and the laws, not the rule. So they should anchor on the same Gauss-Hermite law production uses, not on a 200-iteration hand Newton with its own hard-coded stop band.

SPEC.md:
- "The goal of this project is never to copy existing reference implementations."
- "It's required that the CLI, Python library, and Rust library are unified, have a single source of truth"

Not covered by #3919 / PR #4022, which route the five physicists-to-probabilists conversions but leave this file alone. Not covered by #4406 / PR #4412 either, which handle scalar kernels. The bench `bench/cargo_benches/joint_absolute_risk.rs` has a second `gauher` copy (line 533). It is left out here because PR #4412 is actively editing that file.

**Fix.** Make `gauss_hermite_probabilists` a split of `standard_normal_gauss_hermite_rule(m)` into its node and weight columns, and delete the transcription (-55 lines). The tests' own assertions stay as they are: unit mass within 1e-12, ascending nodes, moments to 1e-10 and 1e-9, the closed-form anchor to 1e-9, and the node-count tolerance ladder. They then check the shared rule.

SauersML:
Fixes #4522

`crates/gam-models/src/test_support.rs` `gauss_hermite_probabilists` was a line-for-line transcription of Numerical Recipes' `gauher`, including its seeds, its recurrence and a 200-step Newton loop. It now splits `gam_math::quadrature::standard_normal_gauss_hermite_rule(m)` into its node and weight columns. That function is the workspace's one standard-normal Gauss-Hermite rule, and gam-models already depends on gam-math. The change removes 55 lines.

Callers are the lib tests of `latent_anchor`, `latent_law_compression`, `survival::marginal_slope::{joint_latent_law, slope_geometry}`. Their assertions do not change (unit mass, ascending nodes, moments, the closed-form anchor, the node-count tolerance ladder at m = 16/64/128). They now exercise the rule that production uses.

The local constant `PI_QUARTER_INV` is recorded in `docs/source-removal-changes.json`. Local guard results:
- source_removal_guard: 8 -> 7, 1 guarded removal
- public_api_census: pass
- test_census: pass

The bench's second `gauher` copy (`bench/cargo_benches/joint_absolute_risk.rs`) is left out, because PR #4412 is editing that file.

Validation: focused-rust-proof on the gam-models lib filters `latent_anchor`, `latent_law_compression`, `joint_latent_law` and `slope_geometry`.

---


SauersML:
Added 1dc102c5: a second copy of the same rule, in production this time. `crates/gam-models/src/multinomial_posterior.rs:1755` `gauss_hermite_rule` rebuilt the standard-normal rule by hand from the physicists' rule. It scaled the nodes by `SQRT_2`, renormalized the weights to sum to one, and carried its own weight-mass refusal. That is exactly `gam_math::quadrature::standard_normal_gauss_hermite_rule`, which already refuses a non-finite or non-positive mass.

It now reads the shared rule and unzips the pairs into the node and weight columns the Smolyak axes use. The net change is -12 lines.

The focused proof has been redispatched as HN-58-4522-2. It adds `multinomial_posterior` to the filters and cancels the -1 run.

The integration fixture `tests/survival_acceptance/declared_latent_law_2923.rs::Law::gauss_hermite` stays as it is. It is an intentionally independent oracle built by bisection, "so the fixture does not share code with the crate".

---

PR #4534, opened 2026-09-20
arrow_schur: closed-form spectral per-row ridge replaces the escalation ladder

### LINKED ISSUE #4483: arrow_schur: per-row H_tt ridge is a 10×-per-rung reactive ladder with magic seed/cap, and its max(1.0) diag clamp breaks the claimed unit invariance
**Where** (origin/main ce58f23374)

`crates/gam-solve/src/arrow_schur/factorization.rs`, in `factor_one_row_result` (line 926). On the non-evidence path this is the per-row `H_tt` factorisation behind every arrow-Schur Newton step (SAE and latent-coordinate fits).

```rust
const RIDGE_GROWTH_FACTOR: f64 = 10.0;          // 1030
const RIDGE_SEED_DIAG_FRACTION: f64 = 1.0e-10;  // 1031
const RIDGE_CAP_DIAG_FRACTION: f64 = 1.0e-12;   // 1032
const RIDGE_CAP_SCALE: f64 = 1.0e12;            // 1033
let diag_scale = row_block_diag_scale(row, d);
let ridge_cap = ridge_t.max(RIDGE_CAP_DIAG_FRACTION * diag_scale) * RIDGE_CAP_SCALE;
```

When the κ-proxy / pivot check fails (`1138-1149`) or the Cholesky errors (`1232-1246`), the loop refactors at `ridge_eff·10`. That is up to 10-12 full re-factorisations per row, until the block passes or `ridge_cap` is exceeded.

Helpers (lines 136-221):
- `row_block_diag_scale` = `max_a |H_tt[a,a]|` **`.max(1.0)`**
- `safe_spd_pivot_min(s)` = `√ε·s.max(1.0)`
- `safe_spd_kappa_max(d)` = `1/(√ε·d)`

**Why it is wrong**

SPEC: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code." The dispatch rules also say that "a reactive retry, fallback, jitter or ridge bump that exists because a solve failed is a violation; a derived construction ... is not".

1. **A reactive ladder with arbitrary constants.** The growth factor 10, seed 1e-10 and cap 1e12 are not derived from anything. The accepted ridge is whichever decade happened to pass. It overshoots the ridge actually needed by up to 10x, and each rung costs a whole `d×d` Cholesky.
2. **The unit-invariance claim in the comment is false.** The comment (lines 1018-1029) says "Every quantity is a FRACTION of `diag_scale`, so the policy is invariant to the units of the latent block", and that "the ridge [never exceeds] the curvature it is damping". But `row_block_diag_scale` clamps to `max(1.0)`, and `safe_spd_pivot_min` clamps again. Take a latent block whose curvature is O(1e-6); that is exactly what rescaling the latent coordinates by 1e3 produces.
   - `diag_scale = 1`, so the pivot floor is the absolute `√ε ≈ 1.5e-8` instead of `√ε·1e-6 ≈ 1.5e-14`.
   - A perfectly SPD block with a condition number of only about 70 (`min pivot ≈ 1.4e-8`) is therefore rejected and ridged. In units where its curvature is O(1), the same block passes untouched.
   - The seed ridge is an absolute `1e-10`, and the cap is `1e-12·1·1e12 = 1`. The ladder is therefore allowed to add a ridge up to **10^6 times** the block's own curvature and still report success.
   - The same fit with the latent coordinates in other units takes a different Newton step and a different `ridge_escalated` verdict.

**Principled fix (one shot, derived)**

On failure at `ridge_t`, compute the symmetric eigen-decomposition of the `d×d` block (d is the small latent dimension). Then set the ridge in closed form to the smallest `r` that passes both acceptance tests exactly:

```
r* = max( ridge_t,
          pivot_min − λ_min,
          (λ_max − κ_max·λ_min) / (κ_max − 1) )
```

with `pivot_min = √ε·max_a|H_aa|` (no `max(1.0)`) and `κ_max = safe_spd_kappa_max(d)`. Then factor once at `r*`. The construction is rigorous for the existing acceptance test: every Cholesky pivot `L_ii²` of an SPD matrix lies in `[λ_min, λ_max]`, so `(maxL/minL)² ≤ κ₂` and `min L_ii² ≥ λ_min`, which means a block with `κ₂(H+r*I) ≤ κ_max` and `λ_min(H+r*I) ≥ pivot_min` passes. `PerRowFactorIllConditioned`/`Failed` remains only for non-finite blocks, where `r*` is non-finite.

This also:
- deletes the four constants and the loop;
- makes the ridge exactly unit-covariant, since `r*` scales with `H`;
- makes `ridge_escalated` report the ridge that was actually needed.

Tests to keep green: the #578 ill-conditioned-row tests and #1273. Add a test that rescaling the latent block by `c` rescales the applied ridge by exactly `c²`.

SauersML:
Fixes #4483

## Problem
`factor_one_row_result` conditioned a row block that failed the safe-inversion gate by walking a fixed ridge ladder: `RIDGE_ESCALATION_FACTOR`, `RIDGE_FLOOR`, `RIDGE_CAP_FLOOR`/`RIDGE_CAP_SCALE` and a retry loop. That is a reactive ridge bump driven by arbitrary constants. SPEC: "In general, do not paper over solver issues." and "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code." The gate floor `safe_spd_pivot_min` also clamped the diagonal scale at `max(·, 1)`. As a result, a perfectly conditioned block in small units (for example `[[1e-20]]`) was treated as singular, so the gate was not unit-covariant.

## Derivation
Let `M = H + rI` with symmetric spectrum `[λmin, λmax]`.

**Interlacing.** Each Cholesky pivot `L_ii²` is the Schur complement of a leading principal block. By interlacing it lies in `[λmin + r, λmax + r]`. So:
- the gate's pivot proxy is `≥ λmin + r`;
- its κ proxy `(max pivot)/(min pivot)` is `≤ (λmax + r)/(λmin + r)`.

**Sufficient conditions.** The gate is therefore guaranteed to pass once both of these hold:
- `r ≥ p − λmin`, with `p = √ε · max_a |H_aa|` (no clamp);
- `r ≥ (λmax − κ λmin)/(κ − 1)`, with `κ = safe_spd_kappa_max(d) = 1/(√ε d)`.

**Rounding margin.** Floating-point Cholesky (Higham Thm 10.3, with `‖|L||Lᵀ|‖ ≤ d(1 − dγ)⁻¹‖M‖`) and the symmetric eigensolver each perturb the spectrum by at most `c_ch (s + r)`, where:
- `c_ch = dγ_{d+1}/(1 − dγ_{d+1})`;
- `γ_k = k u/(1 − k u)`;
- `s = max|λ̂|`.

We charge `c = 2 c_ch` and solve both inequalities with that margin:

```
pivot_ridge = (p − λmin + c s) / (1 − c)
kappa_ridge = (λmax − κ λmin + c s (1 + κ)) / (κ (1 − c) − 1 − c)
r*          = max(ridge_t, pivot_ridge, kappa_ridge)
```

**Control flow.** The row is first factored at `ridge_t`, exactly as before; evidence paths are unchanged. If that factor fails the gate, the row is factored **once** at `r*` and the gate is checked again. A remaining failure returns the typed `PerRowFactorIllConditioned` / `PerRowFactorFailed` error to the outer LM ridge loop. There is no retry loop and none of the four constants remain; `ridge_escalated = r* > ridge_t`.

## Behaviour changes
| block | old ladder ridge | new ridge |
|---|---|---|
| `[[1,2],[2,1]]` (λ = −1, 3) | 10 | 1.00000012 |
| `diag(1, 1e-9)` | 1e-7 | 2.88e-8 (the single shot still fails; LM recovers, as the existing test asserts) |
| `[[1e-20]]` | escalated against a unit floor | accepted as is, `ridge_escalated = false` |

- `row_block_diag_scale` and `safe_spd_pivot_min` no longer clamp at 1. Every consumer of these helpers stays mutually consistent because they share the helpers:
  - the GPU batch check;
  - gauge qualification;
  - the certified-shift pivot floor;
  - evidence spectral deflation.
- A zero block at ridge 0 now errors, and the outer LM ridge recovers from it.

## Tests
- `factor_one_row_accepts_tiny_scalar_block_in_its_own_units_4483`
- `factor_one_row_ridge_is_closed_form_minimal_and_unit_covariant_4483`: blocks at scales 2^-40, 1 and 2^40. It checks that:
  - the factor is the one computed at `closed_form_row_ridge`;
  - the gate passes;
  - the excess over the exact-arithmetic minimal ridge lies in `[0, 8c]`;
  - `ridge/scale` is scale-invariant to within `4c·ρ`.
- The existing `factor_one_row_conditions_barely_pd_block_via_ridge`, `lm_escalation_recovers_from_ill_conditioned_row` and `ill_conditioning_tolerated_returns_cache_with_exact_logdet` tests are kept.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 24abc9b28102.

Issue #4103, opened 2026-09-20
GLM full conformal under-covers at small n: frozen smoothing parameter (refused:glm_frozen_penalty) needs the honest ρ-re-selecting map


SauersML:
`model.predict(interval="conformal", training_data=...)` for binomial / Poisson / NB / Gamma (`crates/gam-models/src/inference/full_conformal_glm.rs`) is exact for the fitting map with the penalty **frozen at the training fit**. When the fit selected a smoothing parameter, that map is not symmetric in the n + 1 points. The route already labels those rows `conformal_certificate = -7` (`refused:glm_frozen_penalty`), but the sets under-cover measurably at small n.

From the 1000-rep benchmark (`bench/pygam_audit/conformal_coverage.md`, `y ~ s(x, k=10)`, alpha = 0.1, 10 test rows per rep):

| DGP | n | full-conformal coverage | MCSE | verdict |
|---|---|---|---|---|
| binomial | 30 | 0.8805 | 0.0037 | under |
| binomial | 100 | 0.8937 | 0.0031 | under |
| binomial | 1000 | 0.8998 | 0.0030 | nominal |
| poisson | 30 | 0.8912 | 0.0034 | under |
| poisson | 100 | 0.9028 | 0.0030 | nominal |

**Control:** the same DGPs, seeds and routes with `y ~ x` (a single ridge penalty) cover at 0.9009 ± 0.0035 (binomial) and 0.8997 ± 0.0034 (Poisson) at n = 30. So the enumeration, the randomized tie-break and the Poisson count walk are exact, and the gap comes from ρ being frozen.

**What's needed:** the honest map for the GLM families, the GLM analogue of `honest_full_conformal` (Layer 3, Gaussian only today). It would re-select ρ by REML on each augmented multiset `(training ∪ {(x_*, z)})` and certify it, like the Gaussian arm. For Bernoulli that is only two refits per test row, so it is cheap even before any homotopy. Per the conformal-families spec, this certificate belongs to conformal-honest-rho; the route in `crates/gam-predict/src/conformal_routes.rs` would consume it in place of the `-7` refusal.

Part of the pyGAM audit (bench/pygam_audit).

SauersML:
Claimed by wave8 session (claude/fix-4103-glm-honest-rho); draft PR incoming.

---
_Generated by [Claude Code](https://claude.ai/code)_

SauersML:
[...]

PR #4538, opened 2026-09-20
full-conformal GLM: honest re-selecting map for single-penalty Bernoulli fits (#4103)

SauersML:
Part of #4103. This lands the honest map for single-penalty Bernoulli fits. Multi-penalty fits and the count/Gamma families still need work; see "What remains".

## Problem
`GlmFullConformalSubstrate` fitted every augmented data set at the training fit's **frozen** `Sλ̂`. The n training responses selected `λ̂` without the test row, so the fitting map is not symmetric in the n+1 augmented rows. The conformal exchangeability argument therefore fails, and every penalized row reported `refused:glm_frozen_penalty` (-7).

## Fix: re-select the strength on every augmented data set (Bernoulli, one penalty)
For level `z ∈ {0,1}`, `ρ̂(z)` is the **certified** `OuterProblem` optimum (the same seed-path engine every smoothing selection uses, the pattern of `honest.rs::refit_at`) of the Laplace marginal likelihood of the n+1 augmented rows:

```
V(ρ) = Σ_j ℓ_j(β̂) + ½ e^ρ β̂ᵀSβ̂ + ½ ln|H| − ½ r ρ,   H = X_augᵀ W X_aug + e^ρ S,  r = rank(S)
V′  = ½λβ̂ᵀSβ̂ + ½ tr(H⁻¹Ḣ) − ½r,                    Ḣ = X_augᵀ diag(w′η̇) X_aug + λS
V″  = ½λβ̂ᵀSβ̂ + λβ̂ᵀSβ̇ + ½ tr(H⁻¹Ḧ) − ½ tr(H⁻¹ḢH⁻¹Ḣ), Ḧ = X_augᵀ diag(w″η̇² + w′η̈) X_aug + λS
β̇ = −H⁻¹λSβ̂,   β̈ = −H⁻¹(Ḣβ̇ + λSβ̂ + λSβ̇),   w′ = w(1−2μ),  w″ = w − 6w²
```

- **Domain.** `coordinate_domain(resolvability_interval(γ))`, where γ are the penalty-range generalized eigenvalues of the augmented start-weight Gram `X_augᵀW₀X_aug` against `S`, and W₀ is taken at the augmented prevalence `(Σy_aug + ½)/(n+2)`. The bound is derived from the problem, not hand-supplied.
- **Seed.** The domain midpoint, not the stored strength, which only the training rows saw.
- **Symmetry.** The domain, the seed and every criterion value are symmetric functions of the augmented multiset. So `ρ̂(z)` is symmetric too, and the rank argument gives finite-sample coverage. Rescaling the stored `Sλ` translates V and the domain together, so the set does not depend on `λ̂`.
- **Membership.** Each level is decided by the existing certified count solve at `e^{ρ̂(z)}S`.
- **Convergence.** Only a run with `result.converged()` counts as `ρ̂(z)`. SPEC: "A fit object must only ever come from a converged optimization."
- **Per-row certificate.** The certificate is now per row (`GlmFullConformalSet.certificate`):
  - An augmented Gram with no penalty-range curvature refuses with `augmented_gram_singular` (-3).
  - An uncertified engine run refuses with `refit_failed` (-6).
  - Both return the frozen set with that refusal. There is no retry and no fallback search.

Certificate map (`ConformalGlmFamily::certificate`):

| Fit | Certificate |
|---|---|
| unknown penalty count | -2 |
| NB | -7 |
| 0 penalties | exact_frozen (0) |
| Bernoulli, 1 penalty | honest_refit (1) |
| Bernoulli, ≥2 penalties | multi_penalty (-1) |
| other families, ≥1 penalty | -7 |

The route (`conformal_routes.rs`), the FFI `interval_method` text, the `_model.py` docstring and the Python route test are updated. `s(x)` is double-penalty by default, so the binomial Python test now expects -1.

## Tests (crates/gam-models/src/inference/full_conformal_glm.rs)
- **`laml_derivatives_match_finite_differences`:** analytic V′ and V″ against central differences, h=1e-4, 1e-6 relative, at ρ∈{−3,−0.5,1,3.5} and z∈{0,1}. The tolerance derivation is in the test doc.
- **`honest_bernoulli_levels_match_brute_force_at_their_reselected_strength`:** each level against a direct solve at its own ρ̂.
- **`honest_bernoulli_set_does_not_depend_on_the_stored_strength`:** scale 1 and scale 7 give identical sets.
- **`monte_carlo_coverage_is_nominal_for_the_honest_bernoulli_map`:** 1000 reps, n=99, α=0.1, coverage within 2 MCSE two-sided.
- **`certificate_is_exact_only_when_nothing_was_selected`:** the new map.

Validation: focused proof `BF-59-4103-1` (run 35508562007, gam-models lib `full_conformal_glm`), queued.

## What remains for #4103
- **Multi-penalty Bernoulli, which covers the benchmark's default double-penalty `s(x)`.** The per-penalty `S_j` are not persisted in the conformal payload, only `Σ λ̂_j S_j`. Re-selecting the ratios needs the payload to carry each `S_j`.
- **Poisson, NB and Gamma honest maps.** Their tails are proved at a fixed penalty. NB θ and the Gamma dispersion would also have to be re-selected on each augmented set.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 2b734f38ffb9.

PR #4545, opened 2026-09-20
duchon scale tests: gate recovery on across-the-function coverage

### LINKED ISSUE #4505: tests/basis_smooth duchon_scale_and_memory: recovery bars (0.20 fixed across n, 0.5x trivial) are 50-200x the estimation error and pass a fit keeping half the signal
## Where

`tests/basis_smooth/smooths/duchon_scale_and_memory.rs`:
- line 150 (`duchon_1d_recovers_truth_across_increasing_n`): `recovery_rmse < 0.20`, the same bar for n = 2 000, 10 000 and 40 000.
- line 239 (`duchon_2d_recovers_smooth_surface`): `recovery_rmse < 0.5 * trivial`.

## The code

```rust
// 1-D: y = sin(2πx) + N(0, 0.05²), duchon(x, k=40), n ∈ {2k, 10k, 40k}
// "0.20 is a principled non-degeneracy bar that still catches a blown-up or
//  collapsed fit without encoding the noise floor."
assert!(recovery_rmse < 0.20, ...);

// 2-D: y = sin(2πx)cos(2πz) + N(0, 0.05²), duchon(x, z, k=60), n = 4 000
// "Half of it is a principled non-degeneracy bar"
assert!(recovery_rmse < 0.5 * trivial, ...);
```

## Why it is wrong

Both tests are named for truth recovery, but they only assert that the fit is not degenerate.

**1-D test**
- The truth is a single-period sine that the k=40 basis resolves, as the file itself says: "the achievable error is the noise floor, not under-resolution bias".
- For a linear smoother, the RMSE of the function estimate is about σ·√(edf/n):
  - about 0.004 at n = 2 000;
  - about 0.0009 at n = 40 000.
- The bar of 0.20 is 50× to 200× larger, and it does not tighten with n. So the "across increasing n" sweep never checks that the error falls with n.
- A fit that keeps only 72% of the sine's amplitude (RMSE ≈ 0.2) passes at every n.

**2-D test**
- `0.5·trivial` is about 0.25 on a surface with RMS 0.5, so a fit that keeps half the surface passes.
- The estimation error at n = 4 000 with edf ≈ 40 is about 0.005, which is 50× below the bar.

**Consequence.** A partial over-smoothing regression in the Duchon REML path passes both tests, for example λ driven too high at large n. That is the regression a scale test exists to catch. The pattern is the same as #4387 and #3760, in a file those issues do not cover.

## Principled fix

1. **Gate on across-the-function coverage.** Gate each fit on its own statistic `Q = (1/P) Σ (f̂ − f)²/s²`, with `s² = x_pᵀ Vb x_p` at the test grid.
2. **Use a derived bound.** Require `Q` to be at most the Satterthwaite bound `g·χ²_{1−α}(1/g)`, where `g = tr(R²)/P²` and `R` is the probe correlation matrix of `Vb`. Use α = 0.01. This is the same construction as PR #4404.
3. **Resulting behaviour.** The bar tightens with n automatically, because `s` shrinks. A fit that keeps a fraction of the signal fails, because the bias adds to `f̂ − f` while `s` does not grow.
4. **Keep the finiteness and memory checks** as they are.

A case that then fails is a real Duchon or REML defect to be filed on its own. The bar must not be widened.

---

SauersML:
Fixes #4505

`duchon_1d_recovers_truth_across_increasing_n` and `duchon_2d_recovers_smooth_surface` passed or failed on hand-picked RMSE bars (`< 0.20` and `< 0.5 * trivial`). Nothing derives those numbers. This PR replaces both bars with the Nychka across-the-function coverage gate from PR #4404, which is built entirely from the fit's own posterior:

- **Posterior covariance at the probes:** `C = X_p V_b X_pᵀ`, with `s_i = √C_ii` and `e_i = f̂ − f`.
- **Statistic:** `Q = (1/P) Σ e_i²/s_i²`. If the band is calibrated, `Q` has mean 1 and variance `2 tr(R²)/P²`, where `R = D⁻¹CD⁻¹`.
- **Bound:** a Satterthwaite two-moment match gives `g · χ²_{1−α}(1/g)`, with `g = tr(R²)/P²`.
- **Rate:** α = 0.01 family-wise. The 1-D sweep splits it Bonferroni-style over its three n values.

The RMSE and the trivial-predictor RMS are still printed, but only as diagnostics. The finiteness checks are unchanged.

https://claude.ai/code/session_0187xuFaLvee3fS86ygeJZsN

---
_Generated by [Claude Code](https://claude.ai/code/session_0187xuFaLvee3fS86ygeJZsN)_


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, a3723cc7debc.

PR #4547, opened 2026-09-20
Gate the Duchon scale recovery tests on the fit's own coverage (#4505)

### LINKED ISSUE #4505: tests/basis_smooth duchon_scale_and_memory: recovery bars (0.20 fixed across n, 0.5x trivial) are 50-200x the estimation error and pass a fit keeping half the signal

SauersML:
Fixes #4505

## Problem
`tests/basis_smooth/smooths/duchon_scale_and_memory.rs` used bars that only catch a degenerate fit:
- the 1-D test used `recovery_rmse < 0.20` at n = 2 000, 10 000 and 40 000;
- the 2-D test used `recovery_rmse < 0.5·trivial`.

For a resolved truth the estimation error is σ·√(edf/n):
- about 0.004 at n = 2 000 and 0.001 at n = 40 000 in 1-D;
- about 0.005 in 2-D.

So both bars are 50 to 200 times too loose. A fit that keeps only 72% of the sine, or half the surface, passes. The 1-D bar also never tightens with n.

## Fix
Each fit is now gated on its own across-the-function coverage statistic. This is the same construction as #4404.

- **Statistic.** With `C = X_p Vb X_pᵀ` on the probe grid, `s_i = √C_ii` and `R = D⁻¹CD⁻¹`:
  - `Q = (1/P) Σ (f̂_i − f_i)²/s_i²`;
  - under the band's model, `e ~ N(0, C)`, so `E[Q] = 1` and `Var[Q] = 2·tr(R²)/P²`.
- **Bound.** The moment-matched (Satterthwaite) scaled χ² gives `Q ≤ g·χ²_{1−α}(1/g)`, where `g = tr(R²)/P²`.
- **Significance level.** Family-wise α = 0.01. The 1-D sweep splits it Bonferroni-style over its three independent fits (α = 0.01/3 each). The 2-D test has one fit and uses α = 0.01.
- **Why this bar is right.**
  - It tightens with n on its own, because `s` shrinks.
  - A fit that loses signal fails: the lost signal enters `e`, but `s` does not grow to cover it.
- **Unchanged.** The finiteness checks stay as they were.
- **Removed.** The trivial-predictor RMS helper is gone because nothing uses it any more.

If a case fails, that is a real Duchon/REML defect to file separately. The bar is not to be widened.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 0a97413edd3e.

PR #4551, opened 2026-09-20
estimate/SAS link: fit (ε, δ) at the LAML optimum, not a MAP under a fixed prior

### LINKED ISSUE #4482: estimate/SAS link: fitted (ε, δ) is a MAP under a hard-coded 1e-4 ridge + 1e-2 edge barrier on log δ, inside fixed |ε|≤8, |logδ|≤12 boxes
**Where** (origin/main ce58f23374)

- `crates/gam-solve/src/estimate/evaluation.rs:3-7`:
  ```rust
  pub(crate) fn sas_log_deltaridgeweight() -> f64 {
      // Weak fixed stabilization for the SAS tail parameter to avoid
      // boundary/flat-region pathologies in outer optimization.
      1e-4
  }
  ```
- `evaluation.rs:10-14`: `sas_log_delta_edge_barrierweight()` returns `1e-2`. The barrier is `2w·ln cosh(raw_log_δ / B)`, with `B = SAS_LOG_DELTA_BOUND`.
- `evaluation.rs:49-52`: `sas_epsilon_bound()` returns `8.0`, with the comment "Fixed smooth bound on raw SAS epsilon". The effective ε is `8·tanh(raw/8)`.
- `crates/gam-solve/src/mixture_link.rs:30`: `SAS_LOG_DELTA_BOUND: f64 = 12.0`. The effective `log δ = smooth_bound_jet(raw, 12)`, so `δ ∈ [e^{-12}, e^{12}]`.
- `crates/gam-solve/src/estimate/optimizer.rs`:
  - `1508-1512`: `sasridgeweight = sas_log_deltaridgeweight()` whenever a SAS link is estimated.
  - `1987-1995`: `sas_ridge_cost = 0.5·w·logδ² + barrier`.
  - `2011` (value-only), `2027` (value, gradient and Hessian) and `2114` (EFS): the ridge and barrier are added to every outer cost evaluation.
  - `2055-2063`: the matching gradient and Hessian terms.

**Why it is wrong**

SPEC: "Hard-coded knobs and magic constants, especially if arbitrary, should be avoided when possible in production code." "Hand-supplied search boxes or bounds are never allowed." "In general, do not paper over solver issues."

1. **The ridge and barrier change the estimand.** The outer criterion for an estimated SAS link minimises `LAML(ρ, ε, logδ) + 0.5·1e-4·logδ² + 2·1e-2·ln cosh(logδ/12)`. The reported `(ε, δ)` is therefore a penalised MAP estimate under an arbitrary Gaussian-plus-log-cosh prior on `log δ`, not the REML/LAML optimum the fit claims. The comments give the reason: "avoid boundary/flat-region pathologies" and "keep raw log-delta away from tanh-saturation edges where ... outer gradients become uninformative". That is the textbook case of a counter-term added to hide a solver problem, a flat criterion or a boundary optimum, rather than reporting it. The weights `1e-4` and `1e-2` have no units and no derivation. Their pull relative to the LAML curvature in `log δ` changes with n, so the bias does not vanish in a predictable way.
2. **The beta-logistic link already removed the same ridge.** The comment at `optimizer.rs:1982-1986` (#2902 row 34) records that the block used to carry "this weak ridge on both coordinates" because of a gauge, and that the ridge was dropped once the gauge was removed. SAS keeps the ridge, but it has no such gauge. `sinh(δ·asinh η − ε)` is not invariant under any rescaling of β that δ can absorb: the tails go as `η^δ`, and only the linearisation at η = 0 trades δ against the scale of η. So neither ridge nor barrier is needed for identifiability.
3. **The effective parameter bounds are hand-supplied.** `|ε| ≤ 8` and `|log δ| ≤ 12` are fixed boxes on the fitted link shape, not bounds derived from the data's resolution. The raw-chart domain itself (`sas_epsilon_domain`, `smooth_bound_support`) is derived from the gradient resolution, but only *inside* these fixed effective boxes.

**Fix**

- Delete `sas_log_deltaridgeweight`, `sas_log_delta_edge_barrierweight`, `sas_log_delta_edge_barriercost*`, and every cost, gradient and Hessian addition of them in `optimizer.rs` (including EFS).
- Take the effective box from the problem's resolution, as the beta-logistic block does with `rho_domain::precision_box()`: `log δ` and `ε` are log-shape coordinates with no penalty spectrum. This avoids a fixed 8/12 tanh chart.
- Where the LAML is flat in `log δ` (a tail shape the data do not resolve), the certificate should report a boundary or flat-direction optimum honestly. It should not be steered to an interior point by a prior.
- Update the barrier finite-difference test (`estimate_policy_tests.rs:~1165`) accordingly, and add a test that the fitted `(ε, δ)` is a stationary point of the unpenalised LAML.

SauersML:
Fixes #4482

## What was wrong
When the SAS link was estimated, the outer criterion was `LAML + 0.5·1e-4·logδ² + 2·1e-2·ln cosh(logδ/12)`. The reported `(ε, δ)` was therefore a MAP estimate under a hand-set prior, not the LAML optimum that the certificate claims. The prior's weights have no units or derivation. SAS `sinh(δ·asinh η − ε)` has no gauge that δ could absorb, so neither term is needed for identifiability. The beta-logistic block already dropped the same ridge (#2902 row 34).

The optimizer also searched a raw ε that was mapped through a fixed `ε = 8·tanh(raw/8)` chart. That chart is a hand-supplied `|ε| ≤ 8` box.

## Change
- Deleted `sas_log_deltaridgeweight`, `sas_log_delta_edge_barrierweight` and the barrier cost/grad/Hessian helpers. Also removed every place they were added in `optimizer.rs`: value, value+gradient+Hessian, and EFS.
- Deleted the ε tanh chart (`sas_epsilon_bound`, `sas_effective_epsilon*`, `sas_epsilon_domain`) and its chain rules. Raw ε is now the link's ε. It is searched over `precision_box()`, the same box the other link-shape coordinates use (mixture logit, beta-logistic `[ε, log δ]`, Student-t).
- Updated the comments that described the removed terms (EFS universal form, unified link evaluator). Removed the two tests that only checked the deleted chart and barrier derivatives. The fixed-η ε-sensitivity test now checks `du/dε` directly.

## Not in this PR
The log-δ domain still comes from `smooth_bound_jet(·, SAS_LOG_DELTA_BOUND = 12)` inside the link itself. That chart is also used by user-fixed SAS links. The outer box is its derived support (`smooth_bound_support`). Replacing the link's internal log-δ map is a change to the link definition, and I have left it as follow-up.

https://claude.ai/code/session_01RfYoyQkWBazTS5JGU8mjMT

---
_Generated by [Claude Code](https://claude.ai/code/session_01RfYoyQkWBazTS5JGU8mjMT)_


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 143971019a49.

Issue #4466, opened 2026-09-20
saved model: write-only payload fields (sas_param_covariance, mixture_link_param_covariance, slope_formulas, baseline_slopes, latent_score_contract, beta_noise) and a dead working_response accessor


SauersML:
Found by the save/load round-trip audit on origin/main `b4b0ef1761` (unchanged since `ce58f23374`). Line numbers below are on that commit.

## Defect

`FittedModelPayload` (crates/gam-models/src/inference/model.rs) has six fields that production code never reads, or reads only through a dead fallback. Each is serialized into every affected model file, checked for finiteness at load, and then ignored. Two of them duplicate state the payload already stores canonically. One of them replicates a scalar in a way no reader could interpret.

| field | written at | read by |
|---|---|---|
| `sas_param_covariance` (model.rs:696) | model_payload_builders.rs:530; `synchronize_stateful_link_metadata` model.rs:3792, 3803 | nothing (only `validate_numeric_finiteness` model.rs:6431 and a test) |
| `mixture_link_param_covariance` (model.rs:694) | model_payload_builders.rs:526; model.rs:3814 | nothing (only model.rs:6427) |
| `slope_formulas` (model.rs:702) | model_payload_builders.rs:867, 1272 | nothing (a CLI/FFI parity test compares it) |
| `baseline_slopes` (model.rs:825) | model_payload_builders.rs:876, 1289, **2423** | nothing (the same parity test) |
| `latent_score_contract: Option<SavedLatentScoreContract>` (model.rs:775, struct at model.rs:1129) | never; only `None` in `new()` | nothing |
| `beta_noise` (model.rs:714) | model_payload_builders.rs:1115, fed by `fit.block_by_role(BlockRole::Scale)` at 2565/2673/2738 | only as `location_scale_noise_beta(fit).or_else(\|\| self.payload().beta_noise ...)` in gam-predict/src/lib.rs:1001, 1075, 1088 |

There are also two accessors with no caller anywhere in the workspace (crates/gam-solve/src/model_types/result_types.rs:6680-6689):

```rust
pub(crate) fn working_geometry(&self) -> Option<&WorkingGeometry> { ... }
pub fn working_response(&self) -> Option<&Array1<f64>> {
    self.working_geometry().map(|working| &working.response)
}
```

## Why it is wrong

* **The link covariance is stored twice.**
[...]

PR #4554, opened 2026-09-20
saved model: the fit's blocks are the only coefficient store (#4507)

### LINKED ISSUE #4507: saved model: coefficient blocks stored twice (beta_link_wiggle, survival_beta_*); Gaussian/binomial location-scale read the copy unchecked
Found by the save/load round-trip audit on origin/main `b4b0ef1761`. Line numbers below are on that commit.

## Defect

A saved model stores four coefficient blocks twice. Each lives once in the canonical `fit_result.blocks` and again as a flat copy on the payload (crates/gam-models/src/inference/model.rs):

| payload copy | canonical block | written at (model_payload_builders.rs) |
|---|---|---|
| `beta_link_wiggle` (model.rs:746) | `fit_result` `BlockRole::LinkWiggle` | 522 (standard), 1123 (Gaussian/binomial location-scale), 1474 (survival location-scale) |
| `survival_beta_time` (model.rs:903) | `BlockRole::Time` | 1414, 1500, 1556 |
| `survival_beta_threshold` (model.rs:905) | `BlockRole::Threshold` | 1501 |
| `survival_beta_log_sigma` (model.rs:907) | `BlockRole::Scale` | 1502 |

The model classes treat these copies inconsistently.

* **Standard link-wiggle.** `saved_link_wiggle` (model.rs:4177-4236) requires the copy to bit-match the fit's `LinkWiggle` block, then uses the block. Its own comment gives the reason: *"accepting either source independently would let point prediction and uncertainty describe different models."*
* **Survival location-scale.** `validate_survival_location_scale_saved_fit` (model.rs:1871-1983) also requires every copy to equal its block, and then uses only the block widths. To keep that validator passing, the posterior-draw path `saved_model_with_survival_coefficients` (crates/gam-models/src/survival/predict.rs:662-709) writes each sampled block back into all four copies.
* **Gaussian and binomial location-scale.** These reach the `_` arm at model.rs:4237:
  ```rust
  _ => payload
      .beta_link_wiggle
      .clone()
      .ok_or_else(|| FittedModelError::MissingField { ... })?,
  ```
  Nothing checks this copy against the fit's `LinkWiggle` block.
  * `GaussianLocationScalePredictor` and `BinomialLocationScalePredictor` (gam-predict/src/lib.rs:1014, 1081) take the point wiggle from `runtime.link_wiggle.beta`, which is this copy.
  * They take the covariance from `fit.beta_covariance()`, which is indexed by the fit's blocks, including `LinkWiggle`.
  * The link-wiggle SE gradient `link_wiggle_eta_gradient_rows` (gam-predict/src/lib.rs:866) builds `dq/dq0` from `runtime.beta`, the copy, and pairs it with that covariance.
* **`survival_beta_time` on other survival models.** Apart from the survival location-scale validator, nothing reads `survival_beta_time`. Plain survival and survival marginal-slope models write it (builders 1414, 1556) and never read it.

The in-memory Gaussian location-scale result has the same duplication. `GaussianLocationScaleResult::beta_link_wiggle` is `fit.block_states.get(2)` (fit_orchestration/fit.rs:821). The response-scale remap therefore has to rescale this copy by hand, separately from the `LinkWiggle` block (fit.rs:1216 and 1255).

## Why it is wrong

The saved state holds a copy of each quantity instead of relying on its single source. The class that trusts the copy without a check is exactly the class where the two can drift apart. Several kinds of change touch the `fit_result` blocks without touching the copy:

* a posterior draw;
* a response-scale remap;
* a hand-edited or partially migrated file;
* a future refactor.

After any of these the file loads silently, with the point wiggle from one model and the uncertainty from another. The survival draw path shows the cost is real: it works only because it copies every block back by hand. The standard path refuses a mismatch; the location-scale path hides it.

Per policy: "Breaking changes are welcome. Keep no compat shims and no dead code."

## Round-trip reproducer (synthetic)

1. Fit and save a Gaussian location-scale link-wiggle model. Use the one from `crates/gam-predict/src/gaussian_location_scale_wiggle_replay_3001_tests.rs`, or any Gaussian or binomial location-scale model with `linkwiggle()`.
2. In the JSON, multiply `payload.beta_link_wiggle` by 2 and leave `fit_result` alone.
3. Reload. On main the file loads without error and predicts a different μ. The link-wiggle SEs apply the original block's covariance around the doubled coefficients.
4. Make the same edit on a standard binomial link-wiggle model. Load fails with "standard link-wiggle payload coefficients disagree with the fitted LinkWiggle block".

## Principled fix

Make the `fit_result` blocks the only place coefficients are stored.

* Delete `beta_link_wiggle`, `survival_beta_time`, `survival_beta_threshold` and `survival_beta_log_sigma` from `FittedModelPayload`, along with their builder inputs, writes and finiteness checks.
* In `saved_link_wiggle`, take `beta` from `fit_result.block_by_role(BlockRole::LinkWiggle)` for every model class. Link-wiggle metadata without that block is a hard error. The Standard arm keeps its `[Mean, LinkWiggle]` layout and index-shift checks.
* `validate_survival_location_scale_saved_fit` reads block widths from the fit alone.
* `saved_model_with_survival_coefficients` drops its four-way copy.
* Optional follow-on: remove `GaussianLocationScaleResult::beta_link_wiggle`, so the response-scale remap has only one block to rescale.

Old model files still load, because the payload does not `deny_unknown_fields`.

* For a consistent file, the removed keys were exact copies of the fit blocks, so ignoring them changes nothing.
* For an inconsistent file, the silent mix of two models becomes the canonical one.

SauersML:
Fixes #4507

**This PR stacks on #4494 (#4466).** Its branch is built on `hn46-4466-dead-payload-fields`, so until #4494 merges the diff also shows that PR's commit.

## What changed
The saved-model payload stored four coefficient blocks twice: once in `fit_result.blocks` and again as flat copies (`beta_link_wiggle`, `survival_beta_time`, `survival_beta_threshold`, `survival_beta_log_sigma`).

The Gaussian/binomial location-scale arm of `saved_link_wiggle` read `payload.beta_link_wiggle` without checking it against the fit's `LinkWiggle` block. The covariance, however, was indexed by the fit's blocks. So a copy that drifted gave predictions whose point wiggle came from one model and whose uncertainty came from another.

- The four payload fields are deleted, together with their builder inputs (`LocationScaleWiggle::beta_link_wiggle`, `SurvivalTransformationInputs::survival_beta_time`, `SurvivalLocationScaleInputs::{beta_link_wiggle, survival_beta_*}`, `LatentWindowInputs::beta_time`), their writes and their finiteness checks.
- The plumbing that only fed the standard copy is also deleted: `BinomialMeanWiggle{TermFitResult,FrozenFit}::saved_warp_beta` and `StandardFitResult::wiggle_saved_warp_beta`.
  - That vector was the fit's own wiggle block. `finalize_binomial_mean_wiggle_saved_frame` does not touch that block, and the old loader already required the two to match bit for bit.
  - It is still validated for non-negativity at fit time.
- `saved_link_wiggle` now takes the coefficients from the fit's `LinkWiggle` block for every model class.
  - Link-wiggle metadata without a `fit_result`, or without that block, is a hard error.
  - The standard arm keeps its `[Mean, LinkWiggle]` layout check and its frozen-index-shift checks.
- `validate_survival_location_scale_saved_fit` reads block widths from the fit alone, via `survival_saved_block_width`.
- `saved_model_with_survival_coefficients` no longer mirrors each sampled block into four payload copies.

## Round trip
A file whose `beta_link_wiggle` disagrees with its `fit_result` no longer loads as a silent mix of two models. The copy is not read. Old files still load, because the payload has no `deny_unknown_fields`; for a consistent file the removed keys were exact copies of the blocks.

## Tests
- `saved_prediction_runtime_rejects_location_scale_survival_payload_drift` tested the comparison with the copy. With no copy there is nothing to drift, so it is replaced by `..._fit_missing_threshold_block`, which pins the refusal by name.
- The linkwiggle-derivative test now builds its model with the fit block via `intercept_only_binomial_location_scale_model`.
- The CLI survival location-scale test counts covariance width from the fit's blocks.
- `stateful_saved_model_sync` pins 92 payload keys.
- Locally, `source_removal_guard.py` passes (2 acknowledged removals in `docs/source-removal-changes.json`), as do `public_api_census.py` and `test_census.py`. The ledger entry will be re-keyed to the main tip just before merge.

## Follow-on (not in this PR)
The in-memory `GaussianLocationScaleFitResult` / `BinomialLocationScaleFitResult::beta_link_wiggle` still duplicates the fit's `LinkWiggle` block. `fit_orchestration/fit.rs` rescales it by hand during the response-scale remap, separately from the block. This is the optional follow-on named in #4507.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 0954975341ff.

PR #4555, opened 2026-09-20
tests: two-sided MCSE coverage gates for smoothing-corrected bands

### LINKED ISSUE #4511: tests/python test_smoothing_correction_first_order: coverage gate for the smoothing-corrected bands is one-sided (>= 0.90), so an over-estimated correction (the only way J V_rho J^T can err) passes
## Where

`tests/test_smoothing_correction_first_order.py:148-149` on origin/main (a11e685398), in `test_corrected_intervals_cover_the_mean_and_partial_dependence`:

```python
    mean_cover, pd_cover = float(np.mean(mean_hits)), float(np.mean(pd_hits))
    # Across-the-function coverage of Bayesian GAM intervals is nominal on
    # average (Nychka 1988; Marra & Wood 2012). With 40 replicates the
    # replicate-level spread of a 95% average is ~0.01-0.02, so 0.90 fails
    # only on a real loss of coverage.
    assert mean_cover >= 0.90, (family, mean_cover)
    assert pd_cover >= 0.90, (family, pd_cover)
```

A sibling with the same one-sided shape and a single seed is `tests/generic_variance_link_cells_test.py:180`:

```python
    covered = np.mean((lower <= mu_grid) & (mu_grid <= upper))
    assert covered >= 0.85, f"{family}/{link}: band covers {covered:.3f} of the truth"
```

## Why it is wrong

1. **The gate only checks one direction, and it is the direction the code under test cannot err in.** The first test in the same file checks `se_corr >= se_cond`, because `J V_rho J^T` is a Gram matrix and the smoothing correction can only *add* variance. So the one way the first-order correction can be miscalibrated is by adding too much. That happens if `V_rho` is overestimated, the Jacobian `J` is too large, or the active subspace is too big. Each of these gives intervals that over-cover. The coverage test exists to validate that correction, but `mean_cover >= 0.90` passes every over-covering outcome up to 1.0. That includes a correction that inflates every SE by 50%, which gives about 0.997 nominal coverage. SPEC.md makes conservative intervals a bug, and this test cannot detect one.
2. **The test's own comment gives the resolution for a two-sided check, then leaves it unused.** It says the replicate-level spread of the 40-replicate average is "~0.01-0.02". At that spread, 0.95 ± 3 SE is roughly [0.89, 1.0] for the noisiest cell and [0.92, 0.98] for the tightest. So an upper bar would bind in the cells where the SE is small. The test hard-codes a single lower number instead of using the spread it has already measured.
3. **`pd_hits` mixes a null term with non-null terms.** `x2` has truth identically zero. When its smooth is shrunk to the null space, the band covers zero at every grid point, which is coverage 1.0. That is exactly the over-coverage regime. Averaging it with `x1` and `x3` pushes `pd_cover` up and hides any shortfall or excess in the two terms that carry signal.
4. **`generic_variance_link_cells_test.py:180` has the same one-sided shape on one seed.** Across-the-function coverage of a single fit is not nominal; only its average over replicates is, and that is what Nychka and Marra & Wood cite. So `>= 0.85` on one draw is at once:
   - too weak to catch over-coverage, since any band wider than needed passes up to 1.0;
   - not derived from any stated spread.

## Principled fix

In `test_corrected_intervals_cover_the_mean_and_partial_dependence`:

- **Mean coverage.** Keep the per-replicate hits. Compute the Monte-Carlo SE from the replicates themselves, `se = np.std(hits, ddof=1) / sqrt(N_REP)`. Then assert `abs(cover - 0.95) <= z * se` with a stated z, for example 4 (a two-sided family-wise rate of about 6e-5 per assertion). Print `cover` and `se` in the message. This fails on under-coverage and on over-coverage alike, with a bar derived from the replicates rather than chosen by hand.
- **Partial dependence.** Gate `x1` and `x3` two-sided in the same way.
- **The null term `x2`.** Gate it separately: first on the band containing zero, and second on its width shrinking toward zero as the smoothing parameter goes to infinity. This keeps the expected 1.0 coverage of a null-space-shrunk term out of the calibration statistic.

In `generic_variance_link_cells_test.py`, replace the single-seed `>= 0.85` with a replicated across-the-function coverage average, gated two-sided against its own Monte-Carlo SE in the same way. If the fixture cost rules out replicates, drop the coverage claim from that test. A number that cannot fail on a conservative band should not stand in for a calibration check.

Do not loosen anything to land the fix. If a cell then over-covers, that is the finding: the first-order correction is inflating the SEs, which is a defect in the correction.

(Also checked: `tests/test_inference_exposures.py:101` (`>= 0.7`, split conformal with n_cal = 60 and n_test = 60) and `tests/test_saved_model_carries_no_training_rows.py:76` (`>= 0.8`, full conformal with n = 300 and n_test = 200). There the lower bars are about 3.7 SD below the exact conformal coverage law. At those test-set sizes an upper bar at the same z sits at 1.0 and cannot bind, so they are not part of this issue. The related `crates/gam-test-support` verdict is #3534, and the marginal-slope Python gate is #3893. This issue covers neither file.)

---

SauersML:
Fixes #4511

## Problem
The coverage gates on the smoothing-corrected bands only checked one side.
- `test_smoothing_correction_first_order.py` used `>= 0.90`.
- `generic_variance_link_cells_test.py` used `>= 0.85` on a single seed.

`J V_rho J^T` is a Gram, so the first-order correction can only add variance. Its only failure mode is over-coverage, and a lower bar passes that at any size.

## Fix
**`test_corrected_intervals_cover_the_mean_and_partial_dependence`**
- Mean coverage, `s(x1)` and `s(x3)` are each gated two-sided: `|cover - 0.95| <= 4 * MCSE`.
  - MCSE is `std(per_rep, ddof=1)/sqrt(N_REP)`, taken from the replicates. This is the pattern in `tests/test_full_conformal_families_route.py`.
  - At nominal coverage the gate fails with probability `2(1-Phi(4)) = 6.3e-5` per assertion.
  - The failure message prints the coverage and the MCSE.
- The null term `s(x2)` (truth identically 0) is no longer averaged into the signal terms.
  - Its truth is the double penalty's infinite-lambda limit, where the band contains 0 by construction.
  - It is therefore gated one-sided, on containing the zero truth: `cover >= 0.95 - 4*MCSE`.

**`test_generic_cell_gam_recovers_truth_and_bands_cover_it`**
- The single-seed `>= 0.85` is replaced by 40 replicates (seeds 101..140) with the same two-sided MCSE gate.
- The per-replicate `explained > 0.95` recovery check is kept.
- Across-the-function coverage is nominal only on average over replicates (Nychka 1988; Marra & Wood 2012).

No tolerance is loosened. If a cell over-covers, that is the defect the issue predicts (an inflated first-order correction) and gets fixed in the correction.

## Not included
The issue also suggests a width-shrinkage check on `s(x2)`. It is left out because I could not derive a bound for it: REML stops at a finite lambda. Only the containment half of that suggestion is gated.

---


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, aefc4beed7f9.

PR #4557, opened 2026-09-20
Clear the ten ban-scanner hits that stop a source build at HEAD

nathanneuro:
`maturin develop` / `cargo build` on a checkout of main aborts in build.rs: the ban-scanner reports 10 violations across 5 rules, so the tree cannot be built from source as it stands. Each is fixed by using or dropping the value rather than silencing it:

- stratum_boundary_2939_tests.rs: two `let _ = guard.observe_off_stratum(..)` become bare calls (the verdict is not must_use; the file's other calls are already bare and the streak assertions that follow are what the test checks).
- penalty_coordinate.rs: the should_panic test calls `project_out_null_directions` for its panic, without binding the result.
- spatial_optimization.rs: `debug_assert_eq!` on the slot/hyper lengths is a plain `assert_eq!`, so the invariant holds in release builds too.
- tests_interval_warm_surrogate_3711.rs: the all-interval fixture's empty edit closure now asserts the premise its test name states (no exact-failure rows) instead of editing nothing.
- bms/custom_family_impl.rs: `persistent_warm_start_fingerprint` uses its `specs` / `options` the way the survival marginal-slope family does — refuse to key a record when the block specs do not match the response rows or the inner tolerance is not a finite positive number — instead of ignoring them.

No behaviour change for a well-formed fit; the fingerprint now returns None for mismatched specs where it previously keyed a record anyway.

Claude-Session: https://claude.ai/code/session_01VK8shv74ukAk2a7CrbDDNT


SauersML:
Landing note (gam-fb, 2026-09-21): merged into main by a two-parent merge, 41b3bd77a351.

